PR #7, opened 2020-07-01
Adding CI

eddynaka:
## Changes
- Adding GitHub Actions
- Adding nuget.config with MyGet V3 and dotnet5
- Adding nuget.config and other files to sln (same as the other OT for dotnet solution)


cijothomas:
GH actions are now triggered.

PR #9, opened 2020-07-02
adding dependabot

eddynaka:
## Changes:
- added dependabot to update only OT dependency. Later we can change that to enable all dependencies if needed.

## Tests:
- GitHub Repo to exemplify: https://github.com/eddynaka/github-dependabot/pulls


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  # Enable version updates for nuget
+  - package-ecosystem: "nuget"
```
SergeyKanzhelev, line 4: will it update from MyGet or from NuGet? Do we want to keep up with the MyGet versions at all?

eddynaka: i really don't know if that would update for myget. i will try in my repo and see if nuget is accepting myget/nuget

eddynaka: @SergeyKanzhelev , with our latest release, i removed the myget feed from this repository, so it won't update from it. Let me know what do you think. Thank you.

MikeGoldsmith: I think having both is good, especially during beta. Nuget is for formally released packages, myget is for bleeding edge daily builds.

eddynaka: we were discussing if we will keep publishing every merge to myget. Probably that will in the next .net SIG

PR #13, opened 2020-07-27
Updating projects to beta nuget package

eddynaka:
## Changes
- updated projects to use OT from nuget (latest beta)
- removed feed from public dotnet / MyGet


src/OpenTelemetry.Exporter.Stackdriver/Implementation/ActivityExtensions.cs
```diff
@@ -15,15 +15,15 @@
...
+using System.Diagnostics;
 using System.Linq;
 using Google.Cloud.Trace.V2;
 using Google.Protobuf.WellKnownTypes;
 using OpenTelemetry.Trace;
-using OpenTelemetry.Trace.Export;
 
 namespace OpenTelemetry.Exporter.Stackdriver.Implementation
 {
-    internal static class SpanExtensions
```
MikeGoldsmith, line 26: Looks like some name refactoring has been included here - please move to separate PR :+1:

src/OpenTelemetry.Exporter.Stackdriver/StackdriverTraceExporter.cs
```diff
@@ -23,14 +24,14 @@
...
 using OpenTelemetry.Exporter.Stackdriver.Implementation;
-using OpenTelemetry.Trace.Export;
+using OpenTelemetry.Trace;
 
 namespace OpenTelemetry.Exporter.Stackdriver
 {
     /// <summary>
     /// Exports a group of spans to Stackdriver.
     /// </summary>
-    public class StackdriverTraceExporter : SpanExporter
```
MikeGoldsmith, line 33: Should be in separate PR.

src/OpenTelemetry.Exporter.Stackdriver/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+using OpenTelemetry.Exporter.Stackdriver;
+
+namespace OpenTelemetry.Trace
```
MikeGoldsmith, line 20: Should be in separate PR.

src/OpenTelemetry.Instrumentation.Azure/OpenTelemetryBuilderExtensions.cs
```diff
@@ -30,8 +30,8 @@ public static class OpenTelemetryBuilderExtensions
         /// </summary>
         /// <param name="builder"><see cref="OpenTelemetryBuilder"/> being configured.</param>
         /// <returns>The instance of <see cref="OpenTelemetryBuilder"/> to chain the calls.</returns>
-        public static OpenTelemetryBuilder AddAzureClientsDependencyInstrumentation(
-            this OpenTelemetryBuilder builder)
+        public static TracerProviderBuilder AddAzureClientsInstrumentation(
```
MikeGoldsmith, line 33: Separate PR.

src/OpenTelemetry.Exporter.Stackdriver/Implementation/ActivityExtensions.cs
eddynaka: I renamed because when we updated, the project didn't compile because we don't expose SpanExporter anymore.

src/OpenTelemetry.Exporter.Stackdriver/StackdriverTraceExporter.cs
eddynaka: I renamed because when we updated, the project didn't compile because we don't expose SpanExporter anymore.

src/OpenTelemetry.Exporter.Stackdriver/TracerProviderBuilderExtensions.cs
eddynaka: ok. i will remove.

src/OpenTelemetry.Instrumentation.Azure/OpenTelemetryBuilderExtensions.cs
eddynaka: ok. i will remove

src/OpenTelemetry.Exporter.Stackdriver/TracerProviderBuilderExtensions.cs
eddynaka: removed

src/OpenTelemetry.Instrumentation.Azure/OpenTelemetryBuilderExtensions.cs
eddynaka: removed

src/OpenTelemetry.Exporter.Stackdriver/Implementation/ActivityExtensions.cs
MikeGoldsmith: Shouldn't the build be failing on master if `SpanData` is no longer exposed?

eddynaka: well, in master, we was using an alpha version, which still had that version.

eddynaka:
> Considering the "refactor" is required in order to bump the package and change the source, LGTM.

I will create a new PR adding the missing classes/extensions to be similar to our main repo.

src/OpenTelemetry.Exporter.Stackdriver/Implementation/ActivityExtensions.cs
MikeGoldsmith: I'm not really following - the changes to update the package feed to use the new beta should be isolated from the refactoring. The current master branch, currently pointing at myget, should have failed because if `SpanData` was missing.

Unless the version that was pointed to in the myget feed was behind master? In that case, I'd still prefer to see separate PRs to do the refactoring and update the package feed. Separate PRs help to reduce PR complexity and make smaller changes that can be reverted if required.

eddynaka: @MikeGoldsmith , i updated not only the feed, but the version as well. We were using 0.2.0-alpha.404 and now, with this pr, we are using 0.4.0-beta.2. With this change, SpanData doesn't exist, so, I had to update to Activity style. What i can do is:
1. update to a beta inside myget, To the latest one (that will include the refactor to activity style)
2. update the feed from myget to nuget

cijothomas:
@MikeGoldsmith Please re-review. The refactoring is not really refactoring, as they are necessary changes to update main sdk version.

PR #23, opened 2020-09-01
Renaming to Contrib namespace

eddynaka:
## Changes:
- Renaming to Contrib namespace

## What to do next:
- In a separate PR I will update the build workflows, props, etc to be similar to main opentelemetry-dotnet repo.


eddynaka:
> LGTM! @SergeyKanzhelev I was thinking about your suggestion to use `OpenTelemetryContrib.` I can't decide which I like better. Leaning slightly more towards `OpenTelemetry.Contrib` as it is done here. We could also add text to the package title and/or description to add clarity? Something along the lines of "OpenTelemetry .NET vendor contributed XYZ instrumentation/exporter"?

I was thinking about that. But, those i would do in a separate PR, because I still have to change some props, build files, workflows, etc.

PR #30, opened 2020-09-15
Add .NET Remoting instrumentation.

mbakalov:
Hi,

Wanted to start a PR for contributing Otel instrumentation support for .NET Remoting.

[Remoting](https://docs.microsoft.com/en-us/previous-versions/dotnet/articles/ms973857(v=msdn.10)?redirectedfrom=MSDN) is a pretty old and not-anymore-recommended technology, but some existing apps (including ours) are still using it, so it would be good to have support if possible.

Remoting is (was :)) pretty extensible and (among other things) has a concept of [sinks](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/tdzwhfy3(v=vs.100)) which are layers through which request and response pass through during the remote invocation.

To instrument the remote calls we use [IDynamicMessageSink](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.remoting.contexts.idynamicmessagesink?view=netframework-4.8) that is hooked into the pipeline to intercept all calls leaving or entering current AppDomain (see also [here](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty?view=netframework-4.8)):
```csharp
RemotingContext.RegisterDynamicProperty(
  new TelemetryDynamicSinkProvider(),
  null,
  RemotingContext.DefaultContext);
```

Since Remoting is a .NET Framework-only thing, the new project `OpenTelemetry.Contrib.Instrumentation.Remoting` is targeting "net452;net46" only. I wasn't sure about "net452", but I saw that the "main" OpenTelemetry [SDK](https://github.com/open-telemetry/opentelemetry-dotnet/blob/91ee24f4175c254f924390b6f356f92ed4899d23/src/OpenTelemetry/OpenTelemetry.csproj#L3) was doing this, so I also added it.

The code already works OK in a few test apps that I have built and in a test version of our "main" app, but is still incomplete. I have some TODOs in there for things I am unsure about:
* Configurable `IPropagator`.
* It is probably incorrect to stop _any_ current Activity on call return.
* If the transport layer (http) is instrumented, the outgoing hook is called before the http request happens, but the incoming hook is called _after_ the http response is processed. If http is also instrumented, then we are trying to become a child of that http activity, but I am not sure if it is done 100% correctly.
* Missing tests and README.md. I will add some tests that set up 2 AppDomains and try to do different Remoting calls between them.
* I probably missed many more Otel conventions / semantics in there, please let me know if anything else need to be looked into!

Cheers!


cijothomas:
I wasn't sure about "net452" - .NET452 is still officially supported by Microsoft, hence we decided to support it in the main SDK/API. It'd be good to support it here as well, as chances are high that customers of .NET Remoting might be on old .net frameworks.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,127 @@
...
+        public void ProcessMessageStart(IMessage reqMsg, bool bCliSide, bool bAsync)
+        {
+            // Are we executing on client?
+            if (bCliSide)
+            {
+                // Start new outgoing activity
+                var act = RemotingActivitySource.StartActivity(ActivityOutName, ActivityKind.Client);
+                if (act != null)
+                {
+                    act.SetTag("uri", reqMsg.Properties["__Uri"]);
```
cijothomas, line 59: its recommended to protect these `SetTag` inside If (act.IsAllDataRequested) {}

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            // This will be called:
+            // 1) On the server after we are done processing the call and are returning back to client
+            //    => Current activity should be the "ActivityIn", we need to stop it
+            //
+            // 2) On the client, when the server call returns
+            //    => Current activity should be the "ActivityOut", we need to stop it as well
+
+            // TODO: is this wrong? Don't just stop _any_ current activity, check if it is either of the above explicitly?
+            var act = Activity.Current;
```
cijothomas, line 107: In general, I think the idea is: "Stop what you started, and nothing else."

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Fixed. I've changed tags (and `DisplayName`) to now be set only if `IsAllDataRequested`. I've also added some exception-specific tags which are always set in case of an exception regardless of `IsAllDataRequested`.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Thanks, makes sense! I've added checks based on operation type and `activity.DisplayName` similar to how it was done in ASP.NET instrumentation I think.

It is still possible to have something like that:
* Instrumentation starts `Remoting-In` activity
* Server code starts child `aServerActivity` and doesn't stop it
* Remoting call completes, `ProcessMessageFinish` fires, current activity is still `aServerActivity`...

It is probably possible to capture the `Remoting-In` activity in the call-context and then stop it on `ProcessMessageFinish`, but the still-active `aServerActivity` is likely an error (missing Stop/Dispose) in the server code, so perhaps it is ok to log it and move on?

mbakalov:
Thank you @cijothomas for your comments!!

I've done some cleanup, added better exception support (exception details are now assigned to the proper semantic tags) and added tests.

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Unit test project for OpenTelemetry .NET Remoting instrumentation</Description>
+    <TargetFrameworks>net452;net46</TargetFrameworks>
```
eddynaka, line 5: since we don't have any specific code for net46, we could remove it.

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
```diff
@@ -0,0 +1,20 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
+    <PackageReference Include="Moq" Version="4.14.5" />
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="coverlet.collector" Version="1.2.0" />
```
eddynaka, line 13: could you change to something like this:
```xml
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPkgVer)" />
    <PackageReference Include="Moq" Version="$(MoqPkgVer)" />
    <PackageReference Include="xunit" Version="$(XUnitPkgVer)" />
    <PackageReference Condition="$([MSBuild]::IsOsPlatform('Windows'))" Include="xunit.runner.visualstudio" Version="$(XUnitRunnerVisualStudioPkgVer)">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
    <DotNetCliToolReference Include="dotnet-xunit" Version="$(DotNetXUnitCliVer)" />
```

And you can remove coverlet.collector

eddynaka:
@mbakalov , can you add a README.md file showing how to use, what to expect and some references for .net remoting as well?

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
mbakalov: Done. Removed from both the tests and the instrumentation projects. No net46-specific code in either.

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
mbakalov: Done!

mbakalov:
> 
> 
> @mbakalov , can you add a README.md file showing how to use, what to expect and some references for .net remoting as well?

Thanks, added in the last commit.

mbakalov:
Hmm, noticed that my new tests are failing with `dotnet test` but work in Visual Studio. Investigating...

eddynaka:
> Hmm, noticed that my new tests are failing with `dotnet test` but work in Visual Studio. Investigating...

for dotnet-windows workflow, you will have to add net452 to test it And you will have to add a OS codition in the workflow, so it won't be tested on linux.

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
eddynaka: thank you!

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj
eddynaka: thank you!

mbakalov:
> 
> 
> > Hmm, noticed that my new tests are failing with `dotnet test` but work in Visual Studio. Investigating...
> 
> for dotnet-windows workflow, you will have to add net452 to test it And you will have to add a OS codition in the workflow, so it won't be tested on linux.

Hmm interesting! Looking at the current workflow, the "Windows / build-test" only runs the "net461" targets:
```shell
dotnet test **\bin\**\net461\*Tests.dll
```

Should I add "net461" back to my test project so that it at least gets tested without modifying the workflows? Looks like that is the reason I haven't seen my tests fail in CI: I either had "net46" or "net452" as TargetFrameworks, but not "net461" which is what the workflow is looking for.

I also found that the tests fail or pass due to how the runners handle AppDomains. I had it working under the ReSharper's runner, but the xUnit/VS built-in runners do something different and `ShadowCopyFiles` doesn't work for them. I'll make a commit now to make it work for xUnit/VS built-in at least. Don't like this flaky behavior. I _think_ it can be fixed properly if I put my test class into a separate assembly, but I wanted to avoid creating a separate project just for that purpose.

eddynaka:
Just change the ci and add net452 to the list

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

mbakalov:
Oh. I _thought_ I fixed this by adding a `.runsettings` file on my test project to do `RunConfiguration.DisableAppDomain=true` which works across all runners. BUT the CI just does `dotnet test .dll` and so doesn't know about .csproj/.runsettings. I am trying to find a nice way to fix this now.

eddynaka:
> Oh. I _thought_ I fixed this by adding a `.runsettings` file on my test project to do `RunConfiguration.DisableAppDomain=true` which works across all runners. BUT the CI just does `dotnet test .dll` and so doesn't know about .csproj/.runsettings. I am trying to find a nice way to fix this now.

@mbakalov , can you change to this:
`dotnet test --no-build --framework ${{ matrix.version }}`

And see if that works for you?

mbakalov:
> 
> 
> > Oh. I _thought_ I fixed this by adding a `.runsettings` file on my test project to do `RunConfiguration.DisableAppDomain=true` which works across all runners. BUT the CI just does `dotnet test .dll` and so doesn't know about .csproj/.runsettings. I am trying to find a nice way to fix this now.
> 
> @mbakalov , can you change to this:
> `dotnet test --no-build --framework ${{ matrix.version }}`
> 
> And see if that works for you?

Looks like that will try to run given framework tests for all test projects:

```shell
> dotnet test --no-build --framework net452

Test run for opentelemetry-dotnet-contrib\test\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests.dll(.NETFramework,Version=v4.5.2)
The test source file "opentelemetry-dotnet-contrib\test\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests.dll" provided was not found.
C:\Program Files\dotnet\sdk\3.1.401\Microsoft.TestPlatform.targets(32,5): error MSB4181: The "Microsoft.TestPlatform.Build.Tasks.VSTestTask" task returned false but did not log an error.

A total of 1 test files matched the specified pattern.

Test Run Successful.
Total tests: 2
     Passed: 2
 Total time: 1.9279 Seconds

> $LASTEXITCODE
1
```

I am reading some more about AppDomains meanwhile to try to understand how to do this reliably.

eddynaka:
> > > Oh. I _thought_ I fixed this by adding a `.runsettings` file on my test project to do `RunConfiguration.DisableAppDomain=true` which works across all runners. BUT the CI just does `dotnet test .dll` and so doesn't know about .csproj/.runsettings. I am trying to find a nice way to fix this now.
> > 
> > 
> > @mbakalov , can you change to this:
> > `dotnet test --no-build --framework ${{ matrix.version }}`
> > And see if that works for you?
> 
> Looks like that will try to run given framework tests for all test projects:
> 
> ```shell
> > dotnet test --no-build --framework net452
> 
> Test run for opentelemetry-dotnet-contrib\test\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests.dll(.NETFramework,Version=v4.5.2)
> The test source file "opentelemetry-dotnet-contrib\test\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Tests.dll" provided was not found.
> C:\Program Files\dotnet\sdk\3.1.401\Microsoft.TestPlatform.targets(32,5): error MSB4181: The "Microsoft.TestPlatform.Build.Tasks.VSTestTask" task returned false but did not log an error.
> 
> A total of 1 test files matched the specified pattern.
> 
> Test Run Successful.
> Total tests: 2
>      Passed: 2
>  Total time: 1.9279 Seconds
> 
> > $LASTEXITCODE
> 1
> ```
> 
> I am reading some more about AppDomains meanwhile to try to understand how to do this reliably.

@mbakalov , I think i found the answer. In the pipeline, add this:

`dotnet test **\bin\**\${{ matrix.version }}\*Tests.dll -- RunConfiguration.DisableAppDomain=true` (with the space between `--` and the command. 

With that, the tests worked fine. And you can delete the extra file.

mbakalov:
> @mbakalov , I think i found the answer. In the pipeline, add this:
> 
> `dotnet test **\bin\**\${{ matrix.version }}\*Tests.dll -- RunConfiguration.DisableAppDomain=true` (with the space between `--` and the command.
> 
> With that, the tests worked fine. And you can delete the extra file.

@eddynaka - thanks!! I was hoping that it would work, but I cannot get it to run this way locally. Strangely, I am getting this error:
```shell
PS > dotnet test **\bin\**\net452\*Tests.dll -- RunConfiguration.DisableAppDomain=true
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Unhandled exception. System.NotImplementedException: The method or operation is not implemented.
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.CLIRunSettingsArgumentExecutor.Initialize(String argument)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.ArgumentProcessorFactory.<>c__DisplayClass20_0.<WrapLazyProcessorToInitializeOnInstantiation>b__0()
   at System.Lazy`1.PublicationOnlyViaFactory(LazyHelper initializer)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.GetArgumentProcessors(String[] args, List`1& processors)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.Execute(String[] args)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Program.Main(String[] args)
```

The runsettings do get parsed correctly when I just do `dotnet test` or pass in a .csproj:
```shell
PS > dotnet test .\test\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.csproj -- RunConfiguration.DisableAppDomain=true
Test run for C:\Users\mikha\Projects\opentelemetry-dotnet-contrib\test\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.dll(.NETFramework,Version=v4.5.2)
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.

Test Run Successful.
Total tests: 2
     Passed: 2
 Total time: 2.0141 Seconds
```

But not when I pass in a .dll:
```shell
PS > dotnet test .\test\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests\bin\Debug\net452\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.dll -- RunConfiguration.DisableAppDomain=true
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Unhandled exception. System.NotImplementedException: The method or operation is not implemented.
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.CLIRunSettingsArgumentExecutor.Initialize(String argument)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Processors.ArgumentProcessorFactory.<>c__DisplayClass20_0.<WrapLazyProcessorToInitializeOnInstantiation>b__0()
   at System.Lazy`1.PublicationOnlyViaFactory(LazyHelper initializer)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.GetArgumentProcessors(String[] args, List`1& processors)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Executor.Execute(String[] args)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.Program.Main(String[] args)
```

I do see this in the [vstest repo](https://github.com/microsoft/vstest/blob/63817611638bd221e0bd8f0f953fb4619c642877/src/vstest.console/Processors/CLIRunSettingsArgumentProcessor.cs#L100):
```csharp
public void Initialize(string argument)
{
    throw new NotImplementedException();
}

public void Initialize(string[] arguments)
{
    // if argument is null or doesn't contain any element, don't do anything.
    if (arguments == null || arguments.Length == 0)
    {
        return;
    }

    Contract.EndContractBlock();

    // Load up the run settings and set it as the active run settings.
    try
    {
        // Append / Override run settings supplied in CLI
        CreateOrOverwriteRunSettings(this.runSettingsManager, arguments);
    }

    // ...
}
```

I am now wondering if I am doing something wrong, or if it is indeed by design, or perhaps an issue in vstest.

It does work correctly with a .dll when I pass in runsettings via a file:
```shell
PS > dotnet test **\bin\**\net452\*Tests.dll --settings .\test\OpenTelemetry.Contrib.Instrumentation.Remoting.Tests\test.runsettings
Microsoft (R) Test Execution Command Line Tool Version 16.7.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.

Test Run Successful.
Total tests: 2
     Passed: 2
 Total time: 2.0025 Seconds
```

I am going to investigate some more but any objections if I add test/test.runsettings file (basically move my file outside of my project one folder up) and reference that in CI if I can't find a better solution?

eddynaka:
U wouldn't need the full + settings with that config. Can you give me write permissions so I can try in ur branch? Ping me on gitter tomorrow

mbakalov:
Looks like that is (was!) an issue with vstest. I just updated to .NET 5.0.100-rc.1.20454.5 and the same command works:

```shell
PS > dotnet test **\bin\**\net452\*Tests.dll -- RunConfiguration.DisableAppDomain=true
Microsoft (R) Test Execution Command Line Tool Version 16.8.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     2, Skipped:     0, Total:     2, Duration: 290 ms - OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.dll (net452)
```

In the new SDK, the vstest got updated to "16.8.0" and doesn't have that issue. But it fails on "16.7.0" which is in .NET Core 3.1 that is in CI.

eddynaka:
> Looks like that is (was!) an issue with vstest. I just updated to .NET 5.0.100-rc.1.20454.5 and the same command works:
> 
> ```shell
> PS > dotnet test **\bin\**\net452\*Tests.dll -- RunConfiguration.DisableAppDomain=true
> Microsoft (R) Test Execution Command Line Tool Version 16.8.0
> Copyright (c) Microsoft Corporation.  All rights reserved.
> 
> Starting test execution, please wait...
> A total of 1 test files matched the specified pattern.
> 
> Passed!  - Failed:     0, Passed:     2, Skipped:     0, Total:     2, Duration: 290 ms - OpenTelemetry.Contrib.Instrumentation.Remoting.Tests.dll (net452)
> ```
> 
> In the new SDK, the vstest got updated to "16.8.0" and doesn't have that issue. But it fails on "16.7.0" which is in .NET Core 3.1 that is in CI.

Probably I wasn't facing this because i'm using vspreview + dotnet preview.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,124 @@
+# .NET Remoting Instrumentation for OpenTelemetry.Contrib .NET
+
+This is an instrumentation library, which instruments [.NET Remoting](https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-3.0/72x4h507(v=vs.85))
```
cijothomas, line 3: Can you remove "en-us/" part from docs link, so it'll always redirect based on user's settings.

mbakalov: Done!

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,237 @@
...
+                    nameof(TelemetryDynamicSink),
+                    nameof(this.ProcessMessageFinish),
+                    $"{ActivityInName} or {ActivityOutName}",
+                    act.OperationName);
+            }
+        }
+
+        private static void SetStartingActivityAttributes(Activity activity, IMessage msg)
+        {
+            if (activity != null && activity.IsAllDataRequested)
```
CodeBlanch, line 192: It looks like for client case null check happens before coming into the method, so for client it is a double-check here. Maybe move the null check out of here and check before entering for the server call? That way you don't pay the method invocation for server for null, and client doesn't get hit twice when not null.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,237 @@
...
+                    if (returnMsg.Exception == null)
+                    {
+                        act.SetStatus(Status.Ok);
+                    }
+                    else
+                    {
+                        act.SetStatus(Status.Unknown);
+
+                        var e = returnMsg.Exception;
+                        act.SetTag(AttributeExceptionType, e.GetType().FullName);
```
CodeBlanch, line 172: FYI you can use the [RecordException extension](https://github.com/open-telemetry/opentelemetry-dotnet/blob/f1ebf05ecb3e96d2069f105aa9a585ac81008552/src/OpenTelemetry.Api/Trace/ActivityExtensions.cs#L120) here.

Main benefit to that is it will add the details as an event which is what is called for [in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md).

src/OpenTelemetry.Contrib.Instrumentation.Remoting/RemotingInstrumentation.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace OpenTelemetry.Trace
+{
+    internal class RemotingInstrumentation : IDisposable
+    {
+        public RemotingInstrumentation(RemotingInstrumentationOptions options)
+        {
+            // See https://docs.microsoft.com/en-us/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty?view=netframework-4.8
+            // This will register our dynamic sink to listen to all calls leaving or entering
+            // current AppDomain.
+            RemotingContext.RegisterDynamicProperty(
```
CodeBlanch, line 30: Safe to be called multiple times? Users can register multiple TracerProviders so might need to check static variable here or something to make it safe?

CodeBlanch:
Thanks for the contribution @mbakalov! Sent a few comments.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Makes sense, fixed. Changed the signature to use `IMethodMessage` as well for clarity so that can use better API instead of looking things up in `Properties`.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Thanks! I _did_ read that page before and completely missed the point about the whole point about the Event, thought it was about span attributes... Fixed now.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/RemotingInstrumentation.cs
mbakalov: Nope, that would have been a bug - thanks!

I've added checks on both Register and Unregister. I am not sure if anyone will ever call that from different thread, but added a `lock` just in case. I imagine it will be called so rarely that a big lock won't be an issue?

mbakalov:
> 
> 
> Thanks for the contribution @mbakalov! Sent a few comments.

@CodeBlanch thanks for the comments - very useful!

CodeBlanch (approved):
LGTM

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,232 @@
...
+        private static readonly ConcurrentDictionary<string, string> ServiceNameCache = new ConcurrentDictionary<string, string>();
+
+        private readonly RemotingInstrumentationOptions options;
+
+        public TelemetryDynamicSink(RemotingInstrumentationOptions options)
+        {
+            this.options = options;
+        }
+
+        public void ProcessMessageStart(IMessage reqMsg, bool bCliSide, bool bAsync)
```
cijothomas, line 65: what is the impact if this method throws an exception? Does it make sense to wrap the entire method inside try..catch..log blocks?

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
cijothomas: Can you check this for this entire PR . there are a lot of links which has en-us in it!
(we may be able to automatically flag this in CI)

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,232 @@
...
+            // This will be called:
+            // 1) On the server after we are done processing the call and are returning back to client
+            //    => Current activity should be the "ActivityIn", we need to stop it
+            //
+            // 2) On the client, when the server call returns
+            //    => Current activity should be the "ActivityOut", we need to stop it as well
+
+            try
+            {
+                if (this.options.Filter?.Invoke(replyMsg) == false)
```
cijothomas, line 141: do we have to run the Filter in `ProcessMessageFinish` as well? If this was filtered in `ProcessMessageStart`, then we wouldn't have started a new activity. So activity will be null, or someone else's activity, and this case seems handled already in the below codes..

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
cijothomas: > Server code starts child aServerActivity and doesn't stop it

Do you means Asp.Net starting its activity, and not stopping it?

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
mbakalov: Thanks, missed the ones in the code itself, updated now. I see there are two more in the Common msbuild props, won't touch them as part of this PR but can do a tiny separate one to clean those up as well.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Thanks - that would have been an appdomain unhandled exception! Added try-catch-log for both start and finish.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Hmm, not sure about this one.

If we _do_ run the filter then we know that we except our activity back at this point. Is it worth logging that we got someone else's (or null) instead?

This is related to the other thread about `aServerActivity`. In scenarios like below, when someone doesn't stop their activity or stops ours - should we log such things in general? Or is that too noisy and better to just continue silently?
```csharp
// Scenario 1:
{
    ProcessMessageStart
    {
        RemotingActivitySource.StartActivity("RequestIn");
    }

    // If user code starts activity and doesn't stop it:
    SomeOtherActivitySource.StartActivity("user-operation")

    ProcessMessageFinish
    {
        if (Activity.Current?.OperationName != "RequestIn")
        {
            // Log Warning
        }
    }
}

// Scenario 2:

{
    ProcessMessageStart
    {
        RemotingActivitySource.StartActivity("RequestIn");
    }

    // Or if user code stops "our" activity
    Activity.Current?.Stop();

    ProcessMessageFinish
    {
        if (Activity.Current?.OperationName != "RequestIn")
        {
            // LogWarning
        }
    }
}
```

cijothomas: Logging is done to EventSource. I think we should log warning level log message. If somethings doesn't work as expected, then those logs are the only option for us to investigate. (assuming we cannot attach debugger)

mbakalov: I looked at how this was done in other instrumentations ([ASP.NET Core](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6633dc846e20c72460ae478f86e7225e32110717/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L143), [ASP.NET](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6633dc846e20c72460ae478f86e7225e32110717/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs#L201)) and they do simply silently continue when they encounter an activity that they didn't start.

I changed the Remoting one to be the same. It'll only log an error now if something went wrong processing "our" activity.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,224 @@
...
+
+                    // Do we already have an incoming activity?
+                    // Existing activity might be started by an instrumentation layer higher up the
+                    // stack. E.g. if we are using http as a transport for remoting, and we have
+                    // instrumented incoming http request, then that instrumentation happens before
+                    // our "ProcessMessageStart" and we just need to attach to that existing activity.
+                    var parentActivity = Activity.Current;
+                    if (parentActivity != null)
+                    {
+                        ourActivity = RemotingActivitySource.StartActivity(ActivityInName, ActivityKind.Server);
```
cijothomas, line 116: I am a bit unsure about this part. If incoming http request is instrumented and creates activity, would it be extracing the same context as the one this instrumentation would have extracted from the callcontext?

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,124 @@
...
+using OpenTelemetry;
+using OpenTelemetry.Trace;
+
+namespace ExampleClient
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            Sdk.CreateTracerProviderBuilder()
```
cijothomas, line 36: its good to show examples disposing the provider when done. Consider doing 'using' statement.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,124 @@
...
+                .AddRemotingInstrumentation()
+                .AddConsoleExporter()
```
cijothomas, line 38: its best to mention that AddConsoleExporter() requires additional nuget.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,124 @@
...
+{
+    public class Global : System.Web.HttpApplication
+    {
+        private IDisposable _tracerProvider;
+
+        protected void Application_Start(object sender, EventArgs e)
+        {
+            _tracerProvider = Sdk.CreateTracerProviderBuilder()
+                .AddRemotingInstrumentation()
+                .AddJaegerExporter(o =>
```
cijothomas, line 63: please mention that this requires installing the jagerexporter package.

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                else
+                {
+                    Assert.Throws<Exception>(() => obj.DoStuff(exceptionMessage));
+                }
+
+                AppDomain.Unload(ad);
+            }
+
+            Assert.Equal(4, activityProcessor.Invocations.Count); // OnStart/OnEnd/OnShutdown/Dispose called.
+            var activity = (Activity)activityProcessor.Invocations[1].Arguments[0];
```
cijothomas, line 75: can you validate the ActivityKind - client/server as well.

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
mbakalov: Done!

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
mbakalov: Done

src/OpenTelemetry.Contrib.Instrumentation.Remoting/README.md
mbakalov: Done

test/OpenTelemetry.Contrib.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
mbakalov: Done as well

src/OpenTelemetry.Contrib.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
mbakalov: Thank you! It did sometimes extract incorrect context, in a case when server has aspnet instrumentation enabled but the client didn't enable http instrumentation.

I ended up changing this part quite a bit to cover various cases. I am a bit unsure if `parentActivity.TraceId == callContext.TraceId` check is robust enough if I just want to inherit whatever context the aspnet instrumentation establishes. The "call stack" I am trying to support is:
* Client RemotingOut
  * Client HttpOut
    * Server AspNet
      * Server RemotingIn

I didn't want to just discard the AspNet Activity in this case as it is nice to be able to see all 4 elements of this trace.

JanKnipp:
Is this still being followed? .Net Remoting instrumentation would be extremly helpful

cijothomas:
@mbakalov Would you be interested in continue working on this?

mbakalov:
> @mbakalov Would you be interested in continue working on this?

Ohhh sorry - been away from this again for a VERY long time.

Yes, I will try to revive this! I'll make a new PR though: I think a lot in the API has changed since I wrote this originally and it would be easier to start fresh.

PR #43, opened 2020-10-16
Update OpenTelemetry to 0.7.0-beta1

alexvaluyskiy:


src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
```diff
@@ -118,31 +118,15 @@ public override void OnStopActivity(Activity activity, object payload)
...
-                else if (statusCode == 401)
-                {
-                    activity.SetStatus(Status.Unauthenticated);
-                }
-                else if (statusCode == 400)
-                {
-                    activity.SetStatus(Status.AlreadyExists);
-                }
-                else
+                var status = Status.Error;
```
alexvaluyskiy, line 121: I copied this approach from `HttpClient` instrumentation. But I think, I should set `Unset` status, and change it only if an error was occured

src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
```diff
@@ -102,9 +102,7 @@ public override void OnStopActivity(Activity current, object valueValue)
 
         public override void OnException(Activity activity, object valueValue)
         {
-            Status status = Status.Unknown;
-            activity.AddTag(SpanAttributeConstants.StatusCodeKey, SpanHelper.GetCachedCanonicalCodeString(status.CanonicalCode));
-            activity.AddTag(SpanAttributeConstants.StatusDescriptionKey, valueValue?.ToString());
```
eddynaka, line 107: I think we can maintain this line

alexvaluyskiy: I've changed it to this code in order to save a description
```csharp
activity.SetStatus(Status.Error.WithDescription(valueValue?.ToString()))
```

eddynaka (approved):

src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
cijothomas: @eddynaka Can you suggest here?

eddynaka: We did the same thing:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/378ba54b77690f670c1ab526281d0349e1001348/src/OpenTelemetry.Api/Trace/SpanHelper.cs#L30-L40

PR #45, opened 2020-10-26
Add ActivitySourceAdapter to AddAzureInstrumentation

rajkumar-rangaraj:
Current implementation does not collect data. Modified the code to use `ActivitySourceAdapter`.

Now, we could collect traces using below instrumentation code.

```csharp

.AddAzureInstrumentation()
.AddConsoleExporter(opt => opt.DisplayAsJson = true)
.Build();

```

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
@@ -17,7 +17,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using OpenTelemetry.Instrumentation;
 using OpenTelemetry.Trace;
 
@@ -28,14 +27,15 @@ namespace OpenTelemetry.Contrib.Instrumentation.Azure.Implementation
         internal const string ActivitySourceName = "AzureSDK";
         internal const string ActivityName = ActivitySourceName + ".HttpRequestOut";
         private static readonly Version Version = typeof(AzureSdkDiagnosticListener).Assembly.GetName().Version;
-        private static readonly ActivitySource AzureSDKActivitySource = new ActivitySource(ActivitySourceName, Version.ToString());
+        private readonly ActivitySourceAdapter activitySource;
 
         // all fetchers must not be reused between DiagnosticSources.
         private readonly PropertyFetcher<IEnumerable<Activity>> linksPropertyFetcher = new PropertyFetcher<IEnumerable<Activity>>("Links");
 
-        public AzureSdkDiagnosticListener(string sourceName)
+        public AzureSdkDiagnosticListener(string sourceName, ActivitySourceAdapter activitySource)
             : base(sourceName)
         {
+            this.activitySource = activitySource;
         }
 
         public void OnCompleted()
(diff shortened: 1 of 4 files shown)


eddynaka (approved):

rajkumar-rangaraj:
Adding @pakrym

src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
```diff
@@ -74,35 +74,18 @@ public override void OnStartActivity(Activity activity, object valueValue)
...
         {
-            // nothing to be done.
+            this.activitySource.Stop(activity);
         }
 
-        public override void OnException(Activity activity, object valueValue)
+        public override void OnException(Activity activity, object payload)
         {
-            activity.SetStatus(Status.Error.WithDescription(valueValue?.ToString()));
+            activity.SetStatus(Status.Error.WithDescription(payload?.ToString()));
```
eddynaka, line 88: shouldn't we get only the exception message here?

For aspnetcore I did:
```csharp
if (!this.stopExceptionFetcher.TryFetch(payload, out Exception exc) || exc == null)
{
     AspNetCoreInstrumentationEventSource.Log.NullPayload(nameof(HttpInListener), nameof(this.OnException));
     return;
}
activity.RecordException(exc);
activity.SetStatus(Status.Error.WithDescription(exc.Message));
```

reyang (approved):
LGTM.

src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
```diff
@@ -28,14 +27,15 @@ internal class AzureSdkDiagnosticListener : ListenerHandler
...
         internal const string ActivityName = ActivitySourceName + ".HttpRequestOut";
         private static readonly Version Version = typeof(AzureSdkDiagnosticListener).Assembly.GetName().Version;
-        private static readonly ActivitySource AzureSDKActivitySource = new ActivitySource(ActivitySourceName, Version.ToString());
+        private readonly ActivitySourceAdapter activitySource;
 
         // all fetchers must not be reused between DiagnosticSources.
         private readonly PropertyFetcher<IEnumerable<Activity>> linksPropertyFetcher = new PropertyFetcher<IEnumerable<Activity>>("Links");
 
-        public AzureSdkDiagnosticListener(string sourceName)
+        public AzureSdkDiagnosticListener(string sourceName, ActivitySourceAdapter activitySource)
```
cijothomas, line 35: these instrumentations were left in main repo *without* any tests, and we just migrated it as is from main repo. Without unit tests, we cannot validate if this is doing the right thing.

src/OpenTelemetry.Contrib.Instrumentation.Azure/Implementation/AzureSdkDiagnosticListener.cs
rajkumar-rangaraj: The changes to this specific line of code is not a part of this PR. Just renamed the variable name. Yes we should change it as per ASP.NET Core, but wanted to keep this PR very simple and will get this addressed later.

cijothomas:
Closing this as the current plan from AzureSDK does not require this instrumentation.

PR #48, opened 2020-10-27
Adjust MassTransit to the semantic conventions

### LINKED ISSUE #47: Adjust MassTransit to the semantic conventions
Currently, MassTransit instrumentation does not follow the semantic conventions for messaging traces: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md

By default, MassTransit creates an activity for publishing and receiving messages, and populate default tags to be compliant with OpenTracing. But these tags do not conform to the semantic convention spec, for example
- Span names do not match the spec
- Default tag names do not match the spec
- A lot of redundant tags
- Incorrect `span.kind`
- `net.peer.*` tags are not populated correctly
- The instrumentation produces an incorrect `peer.service` tag
- Detect a sync communication (Request-Reply) and set `span.kind=client` or `span.kind=server` 
- Detect temporary destinations (in request-reply) and set `messaging.temp_destination` tag. Also, span name should be changed to `(temporary) send`
- Detect `messaging.destination_kind` (topic|queue). Is it possible to detect it based on `Activity` or `SendContext`?

alexvaluyskiy:
Partially resolves #47

Fixed
- Span names do not match the spec (partially)
- Default tag names do not match the spec
- A lot of redundant tags
- Incorrect `span.kind`
- `net.peer.*` tags are not populated correctly
- The instrumentation produces an incorrect `peer.service` tag

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
@@ -90,24 +90,15 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
         {
             if (activity.OperationName == OperationName.Transport.Send)
             {
-                activity.DisplayName = DisplayNameHelper.GetSendOperationDisplayName(this.GetTag(activity.Tags, TagName.PeerAddress));
-
                 this.ProcessHostInfo(activity);
 
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
                 this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
 
-                activity.SetTag(TagName.SpanKind, null);
-                activity.SetTag(TagName.PeerAddress, null);
-                activity.SetTag(TagName.PeerHost, null);
-                activity.SetTag(TagName.PeerService, null);
-
                 activity.SetTag(TagName.SourceAddress, null);
             }
             else if (activity.OperationName == OperationName.Transport.Receive)
             {
-                activity.DisplayName = DisplayNameHelper.GetReceiveOperationDisplayName(this.GetTag(activity.Tags, TagName.PeerAddress));
-
                 this.ProcessHostInfo(activity);
 
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
@@ -115,15 +106,22 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
 
                 activity.SetTag(TagName.MessageId, null);
 
-                activity.SetTag(TagName.SpanKind, null);
-                activity.SetTag(TagName.PeerAddress, null);
-                activity.SetTag(TagName.PeerHost, null);
-                activity.SetTag(TagName.PeerService, null);
-
                 activity.SetTag(TagName.MessageTypes, null);
                 activity.SetTag(TagName.SourceAddress, null);
                 activity.SetTag(TagName.SourceHostMachine, null);
             }
+            else if (activity.OperationName == OperationName.Consumer.Consume)
+            {
+                this.RenameTag(activity, TagName.ConsumerType, SemanticConventions.AttributeMessagingMassTransitConsumerType);
+            }
+            else if (activity.OperationName == OperationName.Consumer.Handle)
+            {
+            }
+
+            activity.SetTag(TagName.SpanKind, null);
+            activity.SetTag(TagName.PeerService, null);
+            activity.SetTag(TagName.PeerAddress, null);
+            activity.SetTag(TagName.PeerHost, null);
         }
 
         private void ProcessHostInfo(Activity activity)
--- a/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
@@ -95,6 +95,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
                 this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
                 this.RenameTag(activity, TagName.InitiatorId, SemanticConventions.AttributeMessagingMassTransitInitiatorId);
+                this.RenameTag(activity, TagName.CorrelationId, SemanticConventions.AttributeMessagingMassTransitCorrelationId);
 
                 activity.SetTag(TagName.SourceAddress, null);
             }
@@ -105,6 +106,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
                 this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
                 this.RenameTag(activity, TagName.InitiatorId, SemanticConventions.AttributeMessagingMassTransitInitiatorId);
+                this.RenameTag(activity, TagName.CorrelationId, SemanticConventions.AttributeMessagingMassTransitCorrelationId);
 
                 activity.SetTag(TagName.MessageId, null);
 
--- a/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
@@ -94,6 +94,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
 
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
                 this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
+                this.RenameTag(activity, TagName.InitiatorId, SemanticConventions.AttributeMessagingMassTransitInitiatorId);
 
                 activity.SetTag(TagName.SourceAddress, null);
             }
@@ -103,6 +104,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.MassTransit.Implementation
 
                 this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
                 this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
+                this.RenameTag(activity, TagName.InitiatorId, SemanticConventions.AttributeMessagingMassTransitInitiatorId);
 
                 activity.SetTag(TagName.MessageId, null);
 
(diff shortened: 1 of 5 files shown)


alexvaluyskiy:
Could you please help to review this PR @phatboyg @dbarbashov ?

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
```diff
@@ -55,33 +53,173 @@ public async Task MassTransitInstrumentationConsumerAndHandlerTest()
...
-            var handles = this.GetActivitiesFromInvocationsByOperationName(activityProcessor.Invocations, OperationName.Consumer.Handle);
+                Assert.NotNull(actualActivity);
+                Assert.NotNull(expectedMessageContext);
 
-            foreach (var activity in sends)
-            {
-                Assert.Equal("SEND /input_queue", activity.DisplayName);
+                Assert.Equal("/input_queue send", actualActivity.DisplayName);
+                Assert.Equal(ActivityKind.Producer, actualActivity.Kind);
+                Assert.Equal("loopback", actualActivity.GetTagValue(SemanticConventions.AttributeMessagingSystem)?.ToString());
```
alexvaluyskiy, line 65: This tag could be `rabbitmq`, `activemq`, etc for all real providers.
But for the in-memory provider, we have `loopback`. Is it OK?

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
```diff
@@ -72,18 +78,88 @@ private ActivityKind GetActivityKind(Activity activity)
...
+            else if (activity.OperationName == OperationName.Transport.Receive)
+            {
+                this.ProcessHostInfo(activity);
+
+                this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
+                this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
+
+                activity.SetTag(TagName.MessageId, null);
+
+                activity.SetTag(TagName.MessageTypes, null);
```
alexvaluyskiy, line 109: remove `input-address` tag as well

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
```diff
@@ -55,33 +53,173 @@ public async Task MassTransitInstrumentationConsumerAndHandlerTest()
...
+                Assert.Null(actualActivity.GetTagValue(TagName.DestinationAddress));
+
+                Assert.Null(actualActivity.GetTagValue(TagName.SpanKind));
+                Assert.Null(actualActivity.GetTagValue(TagName.PeerService));
+
+                Assert.Null(actualActivity.GetTagValue(TagName.PeerAddress));
+                Assert.Null(actualActivity.GetTagValue(TagName.PeerHost));
+                Assert.Null(actualActivity.GetTagValue(TagName.MessageTypes));
+                Assert.Null(actualActivity.GetTagValue(TagName.SourceAddress));
+                Assert.Null(actualActivity.GetTagValue(TagName.SourceHostMachine));
```
alexvaluyskiy, line 137: remove `input-address` tag as well

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
```diff
@@ -55,33 +53,173 @@ public async Task MassTransitInstrumentationConsumerAndHandlerTest()
...
+                    await harness.Stop();
+                }
+
+                var expectedMessageContext = harness.Sent.Select<TestMessage>().FirstOrDefault()?.Context;
+                var actualActivity = this.GetActivitiesFromInvocationsByOperationName(activityProcessor.Invocations, OperationName.Transport.Receive).LastOrDefault();
+
+                Assert.NotNull(actualActivity);
+                Assert.NotNull(expectedMessageContext);
+
+                Assert.Equal("/input_queue consume", actualActivity.DisplayName);
```
alexvaluyskiy, line 117: Should we keep a slash, for the span name?

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
```diff
@@ -55,33 +53,173 @@ public async Task MassTransitInstrumentationConsumerAndHandlerTest()
...
+                {
+                    await harness.Stop();
+                }
+
+                var expectedMessageContext = harness.Sent.Select<TestMessage>().FirstOrDefault()?.Context;
+                var actualActivity = this.GetActivitiesFromInvocationsByOperationName(activityProcessor.Invocations, OperationName.Consumer.Handle).LastOrDefault();
+
+                Assert.NotNull(actualActivity);
+                Assert.NotNull(expectedMessageContext);
+                Assert.Equal("TestMessage/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests consume", actualActivity.DisplayName);
```
alexvaluyskiy, line 215: We should use `process` operation name here, instead of `consume`

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
```diff
@@ -55,33 +53,173 @@ public async Task MassTransitInstrumentationConsumerAndHandlerTest()
...
+                var actualActivity = this.GetActivitiesFromInvocationsByOperationName(activityProcessor.Invocations, OperationName.Transport.Receive).LastOrDefault();
+
+                Assert.NotNull(actualActivity);
+                Assert.NotNull(expectedMessageContext);
+
+                Assert.Equal("/input_queue consume", actualActivity.DisplayName);
+                Assert.Equal(ActivityKind.Consumer, actualActivity.Kind);
+                Assert.Equal("loopback", actualActivity.GetTagValue(SemanticConventions.AttributeMessagingSystem)?.ToString());
+
+                Assert.Equal(expectedMessageContext.MessageId.ToString(), actualActivity.GetTagValue(SemanticConventions.AttributeMessagingMessageId)?.ToString());
```
alexvaluyskiy, line 121: We could detect, that a destination is `temporary` based on `input-address` tag, for example
 `rabbitmq://localhost/TempQueuName_bus_yyyoyyyxieyyud4xbdc8w34sy1?temporary=true`
--

test/OpenTelemetry.Contrib.Instrumentation.MassTransit.Tests/MassTransitInstrumentationTests.cs
phatboyg: I would lose the slash, tbh.

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
```diff
@@ -72,18 +78,92 @@ private ActivityKind GetActivityKind(Activity activity)
...
+            else if (activity.OperationName == OperationName.Transport.Receive)
+            {
+                this.ProcessHostInfo(activity);
+
+                this.RenameTag(activity, TagName.MessageId, SemanticConventions.AttributeMessagingMessageId);
+                this.RenameTag(activity, TagName.ConversationId, SemanticConventions.AttributeMessagingConversationId);
+                this.RenameTag(activity, TagName.InitiatorId, SemanticConventions.AttributeMessagingMassTransitInitiatorId);
+                this.RenameTag(activity, TagName.CorrelationId, SemanticConventions.AttributeMessagingMassTransitCorrelationId);
+
+                activity.SetTag(TagName.MessageId, null);
```
alexvaluyskiy, line 111: MassTransit has a bug. It sets `message-id` two times in `receive` activity

alexvaluyskiy:
I have a couple of open questions related to semantic conventions for a messaging system
- Detect a sync communication (Request-Reply) and set span.kind=client or span.kind=server
- Detect temporary destinations (in request-reply) and set messaging.temp_destination tag. Also, the span name should be changed to (temporary) send
- Detect messaging.destination_kind (topic|queue). Is it possible to detect it based on Activity or SendContext?

@phatboyg Is it possible to detect these parameters based on activities' tags or on `SendContext`/`ConsumeContext`?

phatboyg:
I don't think so, definitely not with any consistency. Transports ultimately own the implementation of temporary queues, exchanges, topics, etc. and the message producers do not "own" that behavior. Anything you did to detect temporary would be a "guess" and likely expensive (pattern matching) at runtime.

alexvaluyskiy:
This pull request is waiting for the review for one month.
@eddynaka @cijothomas could you please help with the review?

eddynaka (approved):

julioct:
@alexvaluyskiy @cijothomas  Could we get this merged? Also, can this be used with OTel 1.0.0 RC1?

alexvaluyskiy:
It should be merged
But it will compatible only with Otel 0.7.0, like all other Contrib packages

phatboyg:
Rather than push this, would it make more sense to get MassTransit converted to ActivitySource, and spend the effort there to align with the new approach? As long as it continues to work with Application Insights out of the box, I'm okay with making the change to ActivitySource.

eddynaka:
Yesterday I created a draft pr showing how we can use the new package (1.0RCx), without needing the internal classes. https://github.com/open-telemetry/opentelemetry-dotnet/pull/1654. We could try to do the same here.

alexvaluyskiy:
@phatboyg it would be great if you implement `ActivitySource` based tracing directly in MassTransit.
But in this case it should be compatible with both OpenTracing (AppInsights) and OpenTelemetry semantic conventions.

julioct:
> Rather than push this, would it make more sense to get MassTransit converted to ActivitySource, and spend the effort there to align with the new approach? As long as it continues to work with Application Insights out of the box, I'm okay with making the change to ActivitySource.

That would be awesome!

cijothomas:
Converting MassTransit from DiagnosticSource to ActivitySource *alone* would break existing users of MassTransit. We can try to do a small amount of work in MassTransit, and keep backward compatibility, and still work well with Opentelemetry. I'll provide a draft PR for the approach (no ETA), but the following would be roughly the steps involved:

1. Update MassTransit (MT) to 5.0.0 to DiagnosticSource.
2. Create an ActivitySource in MT.
3. Assuming MT is currently using new Activity("operationname").Start() to start activity, replace it with ActivitySource.StartActivity(). If MT was extracting context and doing activity.SetParentId, then that context must be used as parentContext in the StartActivity call.
4. Stop the activity as before.
5. Create a dummy ActivityListener and add it to ActivitySource. This ActivityListener should do the following:
```
var activityListener = new ActivityListener();
            activityListener.ShouldListenTo = (activitySource) => activitySource.Name.Equals("name of ActivitySource from step2");
            activityListener.Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.PropagationData;
            ActivitySource.AddActivityListener(activityListener);
```
This step ensures backward compatibility in the sense that, Activity will always be created, irrespective of Otel sampling decisions.
6. Continue to emit DS events as before.

7. In OpenTelemetry Instrumentation, do AddSource("ActivitySourcename from step2")
8. Use DiagnosticSource callbacks to enrich the activity with more tags.

phatboyg:
That would be great. There is a discussion on the MassTransit repository for this effort, I can add your comments to the action items.

cijothomas (approved):

PR #50, opened 2020-11-12
Update root activity's sampling decision when using ActivitySource.StartActivity()

lupengamzn:
When using `ActivitySource.StartActivity()` to create root activity, the sampling decision is decided [before](https://github.com/dotnet/runtime/blob/ae9da6432360728c77ab10b1ef0c414d1a9ee287/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/ActivitySource.cs#L207) invoking `ActivityListener`s. If user is using traceid-dependent sampler, root activity's sampling decision needs to be updated as well while replacing trace id. Here, we provide a new API called `AddXRayTraceIdWithSampler` to override the trace id and update sampling decision for this case.

### DIFF
--- a/src/OpenTelemetry.Contrib.Extensions.AWSXRay/AWSXRayIdGenerator.cs
+++ b/src/OpenTelemetry.Contrib.Extensions.AWSXRay/AWSXRayIdGenerator.cs
@@ -19,6 +19,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


lupengamzn:
This PR has been there for one month, @cijothomas @eddynaka could you take a look at it? Thanks :smile:

lupengamzn:
@SergeyKanzhelev Can you take a look at this PR, thanks a lot!

SergeyKanzhelev (approved):

PR #51, opened 2020-11-16
Add stackdriver metrics exporter

humphrieslk:


humphrieslk:
The failures in the checks on this PR seem to be indicative of breaking changes from the `Update to 0.8.0-beta.1`. I suppose I may need to separate out that update in a separate PR and fix what has been broken, assuming it is straight forward.. Otherwise, someone else with more context may need to make those fixes.

billybraga:
An progress being made here? Is there still a lot to be done? Can we help?

PR #54, opened 2020-12-26
Migrate masstransit without ActivitySourceAdapter

eddynaka:
Hi,

this a PR to upgrade the OpenTelemetry to latest and remove the dependency to ActivitySourceAdapter.

Testing locally, I have only one issue for a loopback check.

Can you help me?

@alexvaluyskiy
@phatboyg


src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
```diff
@@ -42,9 +46,33 @@ public override void OnStartActivity(Activity activity, object payload)
...
+            string displayName = this.GetDisplayName(activity);
+            ActivityContext activityContext = default;
+            Activity originalActivityCurrent = Activity.Current;
+            if (activity.ParentSpanId == default)
+            {
+                Activity.Current = null;
+            }
+            else
+            {
+                activityContext = new ActivityContext(activity.TraceId, activity.ParentSpanId, activity.ActivityTraceFlags, activity.TraceStateString, true);
```
alexvaluyskiy, line 58: Could we use `activityContext = activity.Context` here? Or is it mandatory to create a new context here?

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
```diff
@@ -42,9 +46,33 @@ public override void OnStartActivity(Activity activity, object payload)
                 return;
             }
 
-            activity.DisplayName = this.GetDisplayName(activity);
+            string displayName = this.GetDisplayName(activity);
+            ActivityContext activityContext = default;
+            Activity originalActivityCurrent = Activity.Current;
```
alexvaluyskiy, line 51: Should we use an activity instance from the method's parameters, instead of the global Activity?

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
```diff
@@ -42,9 +46,33 @@ public override void OnStartActivity(Activity activity, object payload)
...
+
+            activity = ActivitySource.StartActivity(activity.OperationName, this.GetActivityKind(originalActivityCurrent), activityContext);
+
+            if (activity == null)
+            {
+                Activity.Current = originalActivityCurrent;
+            }
+            else
+            {
+                foreach (var tag in originalActivityCurrent.TagObjects)
```
alexvaluyskiy, line 69: `originalActivityCurrent` could be null here

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
eddynaka: @cijothomas , do u know what would be the difference between using activity.Context and creating our own? I remember that you suggested this in the AspNetCore migration as well.

eddynaka: @alexvaluyskiy , i checked the values:
image

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
eddynaka: fixed

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
eddynaka: fixed

alexvaluyskiy:
@eddynaka I've tested these changes, but unfortunately it does not work for me. `Receive` span cannot find a related `Send` parent span.
MassTransit doing a context propagation on it's own side. If we create other activity on send - this activity will have a different `spanId`. But MT will use an original activity Id and propagate it via headers.

eddynaka:
> @eddynaka I've tested these changes, but unfortunately it does not work for me. `Receive` span cannot find a related `Send` parent span.
> MassTransit doing a context propagation on it's own side. If we create other activity on send - this activity will have a different `spanId`. But MT will use an original activity Id and propagate it via headers.

Thanks for taking a look. Based on this, I think we won't be able to accomplish the complete migration...

src/OpenTelemetry.Contrib.Instrumentation.MassTransit/Implementation/MassTransitDiagnosticListener.cs
cijothomas: The new activity being created must be a "sibling" of the original activity. This is achieved by setting the parent of the original activity as the parent of the to-be-created activity. Using activity.Context will make the new activity as a child of the original activity, instead of sibling.

cijothomas:
> > @eddynaka I've tested these changes, but unfortunately it does not work for me. `Receive` span cannot find a related `Send` parent span.
> > MassTransit doing a context propagation on it's own side. If we create other activity on send - this activity will have a different `spanId`. But MT will use an original activity Id and propagate it via headers.
> 
> Thanks for taking a look. Based on this, I think we won't be able to accomplish the complete migration...

I'll also take a look and see if we can fix this.
(HttpClient/AspNetCore also does propagation itself, so it maybe facing same issue.)

phatboyg:
I'm happy to get the changes put into MassTransit to make it work, I just don't know what's needed from looking at it myself.

cijothomas:
> I'm happy to get the changes put into MassTransit to make it work, I just don't know what's needed from looking at it myself.

Do you mean changing the MassTransit library itself?

phatboyg:
Yes, so that both the existing DiagnosticSource listeners continue to work, while also making it support the data needed to support open telemetry.

cijothomas:
> > Yes, so that both the existing DiagnosticSource listeners continue to work, while also making it support the data needed to support open telemetry.
> 
> Would love to help with this. (will get back by next week, as we have some urgent issues to resolve in the main repo this week).

@phatboyg Sorry for not getting back earlier. I will get back by next week after wrapping up 1.0 release. The same problem is being discussed in .NET runtime, Asp.Net Core, Azure SDKs as well. I hope to get a more official recommendation soon, which we can use in MassTransit as well.

srumyantsev:
@cijothomas there are any news about recommendations which you mentioned above?

cijothomas:
> @cijothomas there are any news about recommendations which you mentioned above?

1. We are investigating an approach in the main repo to totally eliminate the need of `ActivitySourceAdapter` and have the `TracerProvider` natively support the legacy activities. If this attempt is successful, then we can make MassTransit (and other instrumentations which produce activity in legacy way), work without any hacks. PR where this is being investigated: https://github.com/open-telemetry/opentelemetry-dotnet/pull/1836/files
This is a good short term approach and doesn't have to wait for any change in the library itself.

2. To offer a path forward from legacy activity creation, to ActivitySource based approach, there is active PR in Asp.NetCore repo where an effort is underway to move AspNetCore to ActivitySource. There is some work done in AzureSdks in the same direction. These PRs are in active discussions at this stage. Once that approach is validated, we can use that to MassTransit (and other libraries as well) - This is more of a long term approach, and only work for the newer versions of the library.

I know its a long wait, but we want to make sure we explore all approaches before closing :) Hoping to solve this (#1 above) by next week and (#2 above) in next few weeks.

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/pull/1836/files is merged. ElasticSearch instrumentation is being modified to leverage the new feature https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/83

Once ES pr is merged, we can do similar for Masstransit as well.

PR #63, opened 2021-02-05
Individual component versioning through tag prefixes

srprash:
MinVer provides a feature to [version multiple projects in a single repo independently](https://github.com/adamralph/minver#can-i-version-multiple-projects-in-a-single-repo-independently) by defining a tag prefix for each project. I believe we can leverage this feature by defining specific tag prefix for each project and then creating a tag with `<projectPrefix>-<version>` when we want to do a release for the project.

For example, if the project name is `OpenTelemetry.Contrib.Extensions.AWSXRay`, the tag prefix will be `Extensions.AWSXRay-` and the tag name will be `Extensions.AWSXRay-1.0.0-rc2`. 

Any new project added to the repo must define the `<MinVerTagPrefix>` in its `csproj` file.

I checked the [naming requirements](https://git-scm.com/docs/git-tag#Documentation/git-tag.txt-lttagnamegt) for a git tag name but didn't find a restriction on the length. Please call out here if you know of any such limit.

### DIFF
--- a/src/OpenTelemetry.Contrib.Exporter.Stackdriver/OpenTelemetry.Contrib.Exporter.Stackdriver.csproj
+++ b/src/OpenTelemetry.Contrib.Exporter.Stackdriver/OpenTelemetry.Contrib.Exporter.Stackdriver.csproj
@@ -10,5 +10,8 @@
     <PackageReference Include="Google.Cloud.Trace.V2" Version="2.0.0" />
     <PackageReference Include="OpenTelemetry" Version="0.7.0-beta.1" />
   </ItemGroup>
-
+    
+  <PropertyGroup>
+    <MinVerTagPrefix>Exporter.Stackdriver-</MinVerTagPrefix>
+  </PropertyGroup>
 </Project>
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Contrib.Exporter.Stackdriver/OpenTelemetry.Contrib.Exporter.Stackdriver.csproj
```diff
@@ -3,12 +3,11 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <Description>Stackdriver .NET Exporter for OpenTelemetry.</Description>
     <PackageTags>$(PackageTags);Stackdriver;Google;GCP;distributed-tracing</PackageTags>
+    <MinVerTagPrefix>Exporter.Stackdriver-</MinVerTagPrefix>
```
cijothomas, line 6: could you validate locally that "." in the prefix is okay.

srprash: Yup. `.` is okay

cijothomas (approved):
Thanks! this is the same approach used in the sdk repo as well. So far it works well. :)

PR #64, opened 2021-02-07
GH workflows for PR build and Releases

srprash:
### Changes:
- Combining separate PR build workflows for windows and linux into one.
- Adding workflows for releasing projects on tag creation event.

Probably it's not the best approach to have a separate release workflow for each project because there's a lot of code duplication. But so far I couldn't find a way to reuse or include one common workflow into specific ones. Happy to discuss a better way to do this.


eddynaka (approved):

.github/workflows/package-Exporter.Stackdriver.yml
```diff
@@ -0,0 +1,50 @@
+name: Pack OpenTelemetry.Contrib.Exporter.Stackdriver
```
cijothomas, line 1: todo: add a section in contributing.md about new projects, that it require minver prefix and a dedicated workflow.

srprash: Can we get this one merged? And I'll do a separate PR for contribution guidance.

srprash:
We should also remove the common pack and upload to myget workflow: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/.github/workflows/dotnet-core.yml
since it runs on any tag creation and nightly.

cijothomas (approved):

PR #66, opened 2021-02-09
Removing common build workflow

srprash:
This workflow currently runs on any tag creation and also nightly to build the solution and upload to myget. Since now all the projects have their own build workflow triggered on respective tags, we don't need this common workflow.


eddynaka (approved):

cijothomas:
@srprash Do you want to add a daily trigger to the individual workflows as well, as people would get daily builds to myget?

srprash:
I think that will be up to individual component owners if they want to make their daily builds available on myget. But for starters, we can have them in the current individual projects.

PR #68, opened 2021-02-10
Updating contribution guidance

srprash:
## Changes:
* Added guidance for contributing a new project based on the recent changes.
* Added list of approvers and maintainers
* Added the badge for contributors to the repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -151,22 +151,48 @@ These files are used to configure the _StyleCop.Analyzers_ which runs during
 build. Breaking the rules will result in a build failure.
 
 ## Contributing a new project
-This repo is a great place to contribute a new instrumentation, exporter or any kind of extension. Although the projects within this repo share some properties and configurations, they are built and released independently. So if you are creating a new project within `/src` and corresponding test project within `/test`, here are a few things you should do to ensure that your project is automatically built and shipped through CI.
 
-- Include [`OpenTelemetry`](https://www.nuget.org/packages/OpenTelemetry) package in your project. You can choose the version that you want to depend on. Usually it is a good idea to use the latest version. Example:
+This repo is a great place to contribute a new instrumentation, exporter or
+any kind of extension. Although the projects within this repo share some
+properties and configurations, they are built and released independently.
+So if you are creating a new project within `/src` and corresponding test
+project within `/test`, here are a few things you should do to ensure that
+your project is automatically built and shipped through CI.
+
+* Include [`OpenTelemetry`](https://www.nuget.org/packages/OpenTelemetry)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -149,3 +140,50 @@ This repository also includes [stylecop ruleset
...
+## Contributing a new project
+
+This repo is a great place to contribute a new instrumentation, exporter or
+any kind of extension. Although the projects within this repo share some
+properties and configurations, they are built and released independently.
+So if you are creating a new project within `/src` and corresponding test
+project within `/test`, here are a few things you should do to ensure that
+your project is automatically built and shipped through CI.
+
+* Include [`OpenTelemetry`](https://www.nuget.org/packages/OpenTelemetry)
```
cijothomas, line 153: We cannot assume every contrib project must depend on the SDK. Instrumentations are only required to depend on Otel.API. (The DiagnosticSource instrumentations are an exception)

srprash: Ah. I see what you mean. I notice that the writing a new instrumentation has been updated to use the ActivitySource API instead of DiagnosticSource. https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/extending-the-sdk/README.md#writing-own-instrumentation-library

cijothomas: Yes. But most instrumentations need to depend on Otel.API, not just System.Diagnostics.DiagnosticSource alone.
Otel API provides:
Baggage 
Propagators
Builder extensions

CONTRIBUTING.md
```diff
@@ -149,3 +140,56 @@ This repository also includes [stylecop ruleset
...
+You can choose the version that you want to depend on.
+Usually it is a good idea to use the latest version. Example:
+
+  ```xml
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry" Version="1.0.0-rc4" />
+  </ItemGroup>
+  ```
+
+* The assembly and nuget versioning is managed through
```
cijothomas, line 170: probably too much info here, as the actual version is based on git tag, which is good candidate to be put in RELEASING.md file (once we have one).
For contributing.md, we can simply state "your project has minvertagprefix which matches project name", and an example.

cijothomas (approved):
Okay to merge. We can iterate this and make it more better.
Especially around what components can be hosted here, what would we do if the original contributor is unresponsive to issues etc.

CONTRIBUTING.md
srprash: +1 for having a RELEASING.md file and having most of the info about versioning there.

cijothomas (approved):

PR #73, opened 2021-02-21
Fix project name for test step

srprash:
I missed to notice that the tests for the projects have `.Tests` suffixed to them and isn't the exact name as their source projects. So this PR aims to fix the test step in the workflows by explicitly calling the test projects rather than using the `PROJECT` variable.
I could have generalized this too but for some reason the EntityFrameworkCore project doesn't exactly follow the `.Tests` convention rather has the `Tests`  suffix. Maybe we can fix it later for the sake of consistency in the repo.

### DIFF
--- a/.github/workflows/package-Exporter.Stackdriver.yml
+++ b/.github/workflows/package-Exporter.Stackdriver.yml
@@ -33,7 +33,7 @@ jobs:
       run: dotnet build src/${{env.PROJECT}} --configuration Release --no-restore -p:Deterministic=true
 
     - name: dotnet test ${{env.PROJECT}}
-      run: dotnet test test/OpenTelemetry.Contrib.Exporter.Stackdriver.Tests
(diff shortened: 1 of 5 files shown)


.github/workflows/package-Exporter.Stackdriver.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       run: dotnet build src/${{env.PROJECT}} --configuration Release --no-restore -p:Deterministic=true
 
     - name: dotnet test ${{env.PROJECT}}
-      run: dotnet test test/${{env.PROJECT}}
+      run: dotnet test test/OpenTelemetry.Contrib.Exporter.Stackdriver.Tests
```
eddynaka, line 36: You could just concat? Something like ${{env.PROJECT}}.Tests

eddynaka (approved):

cijothomas (approved):

PR #77, opened 2021-02-23
Fix test project name in AWS instrumentation

srprash:
Sorry I missed this one before merging in the whole PR. Doing the same fix as done for the rest of the workflows in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/73


cijothomas (approved):

srprash:
We should also look into getting the workflows trigger and the branch protection check conflict to work. Otherwise some checks will always be hanging.

PR #79, opened 2021-02-24
Remove X-Ray specific documentation from README

anuraaga:
Setting the trace ID is only for X-Ray users, but this instrumentation is unrelated to X-Ray.

And it's almost never a good idea to set the global propagator to just the xray propagator, at least it should be both w3c and xray. But as the instrumentation is directly using xray propagation (nice), it's not necessary anyways.


srprash:
True. Good call out. Usage of xray trace id and xray propagator is covered in the OpenTelemetry.Contrib.Extensions.AWSXRay [readme](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Contrib.Extensions.AWSXRay/README.md) so its not required here. We should update the examples there to use the CompositeTextMapPropagator then.

srprash (approved):

PR #81, opened 2021-02-24
Instrumentation.AWS - Changed project dependency to package reference

lupengamzn:
When referencing the source code of `Extensions.AWSXRay`, the [MinVer](https://github.com/adamralph/minver#how-it-works) will try to infer the version number for `Extensions.AWSXRay`, and causes stable version of `Instrumentation.AWS` to depend on a pre-release version of `1.0.2-alpha` for `Extensions.AWSXRay`, which is in conflict with what we wanted as `1.0.1`. Changing the project reference to package reference will resolve this.


srprash (approved):
Thanks

srprash:
We should discuss this with other folks on a long term solution for auto versioning

PR #91, opened 2021-03-12
Update stackdriver README.md

aabmass:
Great to see this exporter from the community! Just want to leave a note for customers so they are not confused about Google's support of this exporter


cijothomas:
@aabmass does Google have an officially supported exporter?

@SergeyKanzhelev - This exporter was 'lift and shift ed' from the main repo, and not sure who is the original author. We can remove this exporter completely, if we cannot find an owner. Any thoughts on this?

aabmass:
@cijothomas not right now, we are just relying on OTLP + Collector.

PR #95, opened 2021-03-17
Modify ElasticSearch Instrumentation to take advantage of sampling decision

utpilla:
Extends the optimization done in [PR #1894](https://github.com/open-telemetry/opentelemetry-dotnet/pull/1894) to optimize ElasticSearch Instrumentation

## Changes
- Added a check to return from the OnStartActivity method of the listener if the instrumentation is suppressed
- Wrap the logic in OnStartActivity to perform operations such as fetch uri, set the activity DisplayName, etc. only when - activity.IsAllDataRequested is set to true
- Added unit tests to check if the instrumentation respects `Sdk.SuppressInstrumentation` and `activity.IsAllDataRequested`

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
@@ -60,68 +60,74 @@ namespace OpenTelemetry.Contrib.Instrumentation.ElasticsearchClient.Implementati
 
         public override void OnStartActivity(Activity activity, object payload)
         {
-            var uri = this.uriFetcher.Fetch(payload);
+            // By this time, samplers have already run and
+            // activity.IsAllDataRequested populated accordingly.
 
-            if (uri == null)
-            {
-                ElasticsearchInstrumentationEventSource.Log.NullPayload(nameof(ElasticsearchRequestPipelineDiagnosticListener), nameof(this.OnStartActivity));
-                return;
-            }
-
-            ActivityInstrumentationHelper.SetActivitySourceProperty(activity, ActivitySource);
-            ActivityInstrumentationHelper.SetKindProperty(activity, ActivityKind.Client);
-
-            var method = this.methodFetcher.Fetch(payload);
-            activity.DisplayName = this.GetDisplayName(activity, method);
-
-            if (this.options.SuppressDownstreamInstrumentation)
-            {
-                SuppressInstrumentationScope.Enter();
-            }
-
-            if (!activity.IsAllDataRequested)
+            if (Sdk.SuppressInstrumentation)
             {
                 return;
             }
 
-            var elasticIndex = this.GetElasticIndex(uri);
-            activity.DisplayName = this.GetDisplayName(activity, method, elasticIndex);
-            activity.SetTag(SemanticConventions.AttributeDbSystem, DatabaseSystemName);
+            if (activity.IsAllDataRequested)
+            {
+                var uri = this.uriFetcher.Fetch(payload);
 
-            if (elasticIndex != null)
-            {
-                activity.SetTag(SemanticConventions.AttributeDbName, elasticIndex);
-            }
+                if (uri == null)
+                {
+                    ElasticsearchInstrumentationEventSource.Log.NullPayload(nameof(ElasticsearchRequestPipelineDiagnosticListener), nameof(this.OnStartActivity));
+                    return;
+                }
 
-            var uriHostNameType = Uri.CheckHostName(uri.Host);
-            if (uriHostNameType == UriHostNameType.IPv4 || uriHostNameType == UriHostNameType.IPv6)
-            {
-                activity.SetTag(SemanticConventions.AttributeNetPeerIp, uri.Host);
-            }
-            else
-            {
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
```diff
@@ -60,68 +60,74 @@ public ElasticsearchRequestPipelineDiagnosticListener(ElasticsearchClientInstrum
...
+                {
+                    SuppressInstrumentationScope.Enter();
+                }
 
-            if (elasticIndex != null)
-            {
-                activity.SetTag(SemanticConventions.AttributeDbName, elasticIndex);
-            }
+                var elasticIndex = this.GetElasticIndex(uri);
+                activity.DisplayName = this.GetDisplayName(activity, method, elasticIndex);
```
cijothomas, line 93: DisplayName was already set few lines above. One of them could be removed

cijothomas (approved):

cijothomas:
@ejsmith Would you be willing to put yourself as a codeowner for this ES instrumentation, and review/approve the PRs?

ejsmith:
> @ejsmith Would you be willing to put yourself as a codeowner for this ES instrumentation, and review/approve the PRs?

Yeah, that would be good with me.

src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
```diff
@@ -60,68 +60,74 @@ public ElasticsearchRequestPipelineDiagnosticListener(ElasticsearchClientInstrum
...
-            if (uri == null)
+            if (Sdk.SuppressInstrumentation)
             {
-                ElasticsearchInstrumentationEventSource.Log.NullPayload(nameof(ElasticsearchRequestPipelineDiagnosticListener), nameof(this.OnStartActivity));
                 return;
             }
 
-            ActivityInstrumentationHelper.SetActivitySourceProperty(activity, ActivitySource);
-            ActivityInstrumentationHelper.SetKindProperty(activity, ActivityKind.Client);
+            if (activity.IsAllDataRequested)
```
ejsmith, line 71: Previously, we were doing some things even when `IsAllDataRequested` was false. Is it best practice now to not do anything?

ejsmith: I see that other instrumentations like the HTTP one are still gathering some info even when `IsAllDataRequested` is false. But this PR is changing it so that we don't do anything. I'm concerned that by removing the basic level of information that the activity spans will be so generic you won't know what they are for.

src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
ejsmith: The reason it was set above was because that used to be outside of the check for `IsAllDataRequested` and gave the activity a basic display name and if `IsAllDataRequested` was true it set it to a more details display name.

src/OpenTelemetry.Contrib.Instrumentation.Elasticsearch/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs
utpilla: If `IsAllDataRequested` is set to `false`, we don't export the activity as `Processor.OnEnd` is not called. So it wouldn't really be an issue if the activity doesn't have that basic level of information as it would never be sent to `Processors`.

ejsmith: Hmm... so why does the HTTP instrumentation still do some things then? https://github.com/open-telemetry/opentelemetry-dotnet/blob/caab6d63f47052d9f8fadd40485be313ce171f6e/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L85-L110

utpilla: HTTP instrumentation does not enrich the activity if `IsAllDataRequested` is set to `false`. It only propagates context as we don't want broken traces. If HTTP instrumentation does not propagate context, then the external system which receives the HTTP call will create new TraceId for the same transaction, and this will lead to broken traces. That's why, we ensure that the context is propagated regardless of the sampling decision as the other system that is receiving the HTTP call might upsample the activity and export it. In that case, we want to retain the same TraceId for that operation.

In case of Elasticsearch, there is no context propagation happening. We only enrich the activity, which wouldn't be necessary if we are not going to export that activity.

ejsmith: Ok, that makes sense. Guess we would just need to de-duplicate the DisplayName setting then.

ejsmith (approved):
LGTM

Issue #103, opened 2021-04-09
Support for tracing on AWS Lambda


lupengamzn:
This issue is mainly about instrumenting the SDK wrapper for tracing AWS Lambda handler.

What the wrapper needs to implement for an incoming Lambda handler span:
1. Inherit trace id `_X_AMAZON_TRACE_ID` from AWS Lambda runtime or upstream
2. Record AWS Lambda resource [attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/86aeab1e0a7e6c67be09c7f15ff25063ee6d2b5c/specification/trace/semantic_conventions/instrumentation/aws-lambda.md)

Will keep updating and submit a PR

Kielek:
Closing per @Oberon00 comment.

PR #112, opened 2021-05-12
Added Tracing Wrapper for AWS Lambda Handler

lupengamzn:
Targeting issue: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/103

Add manual instrumentation support for tracing AWS Lambda function. The tracing wrapper will create incoming activity for Lambda function, which inherits trace id from AWS Lambda and contains resource/span attributes indicated in [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/86aeab1e0a7e6c67be09c7f15ff25063ee6d2b5c/specification/trace/semantic_conventions/instrumentation/aws-lambda.md)


src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/README.md
```diff
@@ -0,0 +1,65 @@
...
+        // add other instrumentations
+        .AddLambdaConfigurations()
+        .Build();
+```
+
+## Instrumentation
+
+### Lambda Function
+
+1. Create a function with the same signature as the original Lambda function. Call
```
srprash, line 26: I think it's better to use "wrapper" and "actual" keywords for functions. Something like "Create a wrapper function with the same signature as your actual lambda function...".

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/README.md
```diff
@@ -0,0 +1,65 @@
...
+`AWSLambdaWrapper.Trace()` API and pass `TracerProvider`, original Lambda function
+and its inputs as parameters.
+
+2. Set the Lambda handler field point to the new function.
```
srprash, line 30: Is the example below for point 1? If so, we should move it to that point.
Also, rephrase point 2 to something like "In your lambda function configuration, set this wrapper function as the lambda handler." Add an example or screenshot if possible.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        public static TracerProviderBuilder AddLambdaConfigurations(this TracerProviderBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder.AddSource(AWSLambdaUtils.ActivitySourceName);
+            builder.SetResourceBuilder(ResourceBuilder
+                .CreateEmpty()
```
srprash, line 42: Instead of an empty resource, can we use telemetry sdk resource?  It comes with attributes about the OTel SDK.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Resources/ResourceBuilderExtensions.cs#L97

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// Extension class for TracerProviderBuilder.
+    /// </summary>
+    public static class TracerProviderBuilderExtensions
+    {
+        /// <summary>
+        /// Add AWS Lambda configurations.
+        /// </summary>
+        /// <param name="builder"><see cref="TracerProviderBuilder"/> being configured.</param>
+        /// <returns>The instance of <see cref="TracerProviderBuilder"/> to chain the calls.</returns>
+        public static TracerProviderBuilder AddLambdaConfigurations(this TracerProviderBuilder builder)
```
srprash, line 33: Better to call it `AddAWSLambdaInstrumentation()`? We are actually adding an instrumentation so let's just call it that. Using "configurations" might confuse users into thinking that there's actually a way to configure the lambda instrumentation in some way here.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    /// <summary>
+    /// Class for getting AWS Lambda related attributes.
+    /// </summary>
+    internal static class AWSLambdaUtils
+    {
+        internal const string ActivitySourceName = "Amazon.AWS.AWSLambdaInstrumentation";
+        private const string CloudProvider = "aws";
+        private const string AWSRegion = "AWS_REGION";
+        private const string AWSXRayLambdaTraceHeaderKey = "_X_AMZN_TRACE_ID";
```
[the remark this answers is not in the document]
lupengamzn: I think it will not cause confusion as the only xray trace key we can get in Lambda is from environment variable, so probably not need to specify it's from ev, unless we have other way to fetch the key, say from Lambda context, instead of ev. Could you provide more context on this?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -0,0 +1,111 @@
...
+            return new string[0];
+        };
+
+        internal static ActivityContext GetParentContext()
+        {
+            // Currently get trace header from Lambda runtime environment variable
+            // https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+            // TODO: Add steps to extract trace header from http header
+
+            var traceHeaderKey = Environment.GetEnvironmentVariable(AWSXRayLambdaTraceHeaderKey);
```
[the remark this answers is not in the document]
lupengamzn: Right, it should be a value. And the `traceHeaderKey` could be either the environment variable we retrieve from Lambda or trace header value from upstream http header(see the todo), so I think it's good to name it `traceHeaderValue`

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -0,0 +1,111 @@
...
+            return Environment.GetEnvironmentVariable(FunctionVersion);
+        }
+
+        internal static string GetAccountId(string functionArn)
+        {
+            // The fifth item of function arn: https://github.com/open-telemetry/opentelemetry-specification/blob/86aeab1e0a7e6c67be09c7f15ff25063ee6d2b5c/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#all-triggers
+            // Function arn format - arn:aws:lambda:<region>:<account-id>:function:<function-name>
+
+            var items = functionArn.Split(':');
+            if (items.Length >= 5)
```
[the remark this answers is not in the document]
lupengamzn: The Lambda `arn` should be formatted beforehand before we're able to retrieve it. The fifth value should always be the account id for Lambda: https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#versioning-versions-using

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/README.md
lupengamzn: Will add

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/README.md
lupengamzn: For each component, it indicates steps one by one to do the configuration or instrumentation for using Lambda handler or Lambda handler hosting Asp.Net Core serverless application. And the sample below simply shows a complete steps to include configurations and instrumentation for a simple Lambda function. If we put it in step 1, seems a little bit duplicated and we may have to provide an example for the AspNet Core as well which I think makes the doc even more verbose. Will rephrase the point 2.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
lupengamzn: We don't actually add instrumentation in the `TracerProvider`, the instrumentation is manual in this PR and cx has to call API in this library to get the Lambda handler instrumented. What we add to `TracerProvider` is a resource detector and source.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/README.md
```diff
@@ -0,0 +1,103 @@
...
+    var httpClient = new HttpClient();
+    _ = S3Client.ListBucketsAsync().Result;
+    _ = httpClient.GetAsync("https://aws.amazon.com").Result;
+    return input?.ToString();
+}
+```
+
+### Lambda Function - Asp.Net Core
+
+Override the `FunctionHandlerAsync` function in `LambdaEntryPoint.cs` file. Call
```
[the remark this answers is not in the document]
lupengamzn: Good callout, will add the link

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
lupengamzn: I'm guessing you're asking to add the telemetry sdk **after** we create the `ResourceBuilder` as there are only two options: `CreateEmpty()` and `CreateDefault()`, which makes no difference in Lambda as what `CreateDefault()` added will also be added by our resource detector using `CreateEmpty()`. And I didn't see we need to include the telemetry sdk resource attribute for a Lambda span in the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/86aeab1e0a7e6c67be09c7f15ff25063ee6d2b5c/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#all-triggers) so I need to confirm, but IMO, I agree it's good to include.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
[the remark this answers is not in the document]
lupengamzn: > The env var may not be present if Active tracing is not enabled on the function. We can validate the `traceHeaderKey` and return preemptively.

The EV value would always be present, no matter the active tracing is on and off, the only difference is when active tracing is off, the `Sampled` entry would be 0 in the header.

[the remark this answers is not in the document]
srprash: > > The env var may not be present if Active tracing is not enabled on the function. We can validate the `traceHeaderKey` and return preemptively.
> 
> The EV value would always be present, no matter the active tracing is on and off, the only difference is when active tracing is off, the `Sampled` entry would be 0 in the header.

You're right. Tested this myself. I was confused by [this section in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/86aeab1e0a7e6c67be09c7f15ff25063ee6d2b5c/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#determining-the-parent-of-a-span) where it says "_The environment variable will be set and the Context will be sampled only if AWS X-Ray has been enabled for the Lambda function._". @anuraaga Maybe we should correct the wordings? Also, in such a case, when to extract trace context from http headers?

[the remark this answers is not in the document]
anuraaga: Yeah the key part of that sentence is `will be sampled only` but if it's not clear it's worth updating. To confirm, I remember hearing about cases where the variable isn't set, I think usually if X-Ray has never been enabled on an account. I don't know if this is still the case but Shengxin mentioned confusing behavior regards to the variable a while back.

That being said, either way we shouldn't write this logic with the assumption the variable is set, that's a feature of Lambda and we wouldn't want to break if it changed without our noticing.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/AssemblyInfo.cs
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Runtime.CompilerServices;
+
+#if SIGNED
```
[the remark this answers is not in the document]
lupengamzn: Just removed

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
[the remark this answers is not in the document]
srprash: That's fine.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
srprash: Yeah. Can be done as a follow up.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
srprash: Hmm. I don't really know what we can better call this function. Typically calling something "configuration" means that customers are able to configure something here but that's not the case right now. Do we see some actual configuration options going into this method in the future? If not, we should rename it as `AddAWSLambdaProperties` or something. Also be sure to call it "AWS"Lambda than just Lambda to be more specific.

lupengamzn: Yes, I would foresee that we're going to add more config options in future, so we can keep the name. Also, will add the `AWS` to the name to specify.

srprash (approved):
LGTM!

PR #119, opened 2021-05-25
Cleanup diagnosticsource listeners

cijothomas:
Avoids taking SDK dependency.
These checks are redundant since 1.1, as the SDK does this check (Supress) and all Activity start/stop goes through SDK only now.


alanwest (approved):
Cool beans!

utpilla (approved):

PR #171, opened 2022-01-11
Persistent File storage

vishweshbankwar:
Moving this here from OpenTelemetry-dotnet repo [1845] (https://github.com/open-telemetry/opentelemetry-dotnet/pull/1845)

Fixes [1278](https://github.com/open-telemetry/opentelemetry-dotnet/issues/1278)

Sample code:

```c#
var testDirectory = new DirectoryInfo(Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), "test"));
using var storage = new LocalFileStorage(testDirectory.FullName);

var data = Encoding.UTF8.GetBytes("Hello, World!");

// Create blob.
IPersistentBlob blob1 = storage.CreateBlob(data);

// List blob.
foreach (var blobItem in storage.GetBlobs())
{
	Console.WriteLine(((LocalFileBlob)blobItem).FullPath);
}

// Get blob.
IPersistentBlob blob2 = storage.GetBlob();

if(blob2 != null)
{
	blob2.Lease(10);
        System.Text.Encoding.UTF8.GetString(blob2.Read());
	blob2.Delete();
}
```

TODO:
Add additional test cases covering all the methods.
Add GH workflow for release


src/OpenTelemetry.Contrib.PersistentStorage/LocalFileBlob.cs
```diff
@@ -0,0 +1,120 @@
...
+// limitations under the License.
+// </copyright>
+
+using System;
+using System.IO;
+
+namespace OpenTelemetry.Contrib.PersistentStorage
+{
+    /// <summary>
+    /// The <see cref="LocalFileBlob"/> allows to save a blob
```
reyang, line 23: Minor, maybe consider `FileStorageBlob` or `FileBlob`. The fact that the implementation is using file API doesn't necessarily require the file to be local (e.g. it could be UNC path, or a local path backed by a remotely shared folder).

src/OpenTelemetry.Contrib.PersistentStorage/LocalFileStorage.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// <param name="retentionPeriodInMilliseconds">
+        /// Retention period in milliseconds for the blob.
+        /// Default is 2 days.
+        /// </param>
+        /// <param name="writeTimeoutInMilliseconds">
+        /// Controls the timeout when writing a buffer to blob.
+        /// Default is 1 minute.
+        /// </param>
+        public LocalFileStorage(
+                                string path,
```
reyang, line 63: Curious, is this the style recommendation from C#?

src/OpenTelemetry.Contrib.PersistentStorage/LocalFileStorage.cs
```diff
@@ -0,0 +1,177 @@
...
+                                int maintenancePeriodInMilliseconds = 6000,
+                                long retentionPeriodInMilliseconds = 172800,
+                                int writeTimeoutInMilliseconds = 6000)
+        {
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+
+            this.directoryPath = PersistentStorageHelper.CreateSubdirectory(path);
```
reyang, line 74: We need to validate the input range (e.g. if `maintenancePeriodInMilliseconds` is a crazy number like every 1ms or negative).

src/OpenTelemetry.Contrib.PersistentStorage/LocalFileStorage.cs
```diff
@@ -0,0 +1,177 @@
...
+            PersistentStorageHelper.RemoveExpiredBlobs(this.directoryPath, this.retentionPeriodInMilliseconds, this.writeTimeoutInMilliseconds);
+        }
+
+        private bool CheckStorageSize()
+        {
+            var size = PersistentStorageHelper.GetDirectorySize();
+            if (size >= this.maxSizeInBytes)
+            {
+                StorageEventSource.Log.Warning($"Persistent storage max capacity has been reached. Currently at {size / 1024} KB. " +
+                                                "Telemetry will be lost. Please consider increasing the value of storage max size in exporter config.");
```
reyang, line 170: I suggest that we remove "Telemetry will be lost" here. Whether we drop telemetry or not is depending on multiple factors.

src/OpenTelemetry.Contrib.PersistentStorage/LocalFileStorage.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            PersistentStorageHelper.RemoveExpiredBlobs(this.directoryPath, this.retentionPeriodInMilliseconds, this.writeTimeoutInMilliseconds);
+        }
+
+        private bool CheckStorageSize()
+        {
+            var size = PersistentStorageHelper.GetDirectorySize();
+            if (size >= this.maxSizeInBytes)
+            {
+                StorageEventSource.Log.Warning($"Persistent storage max capacity has been reached. Currently at {size / 1024} KB. " +
```
reyang, line 169: Is `{size / 1024}` accurate? (e.g. is there a chance that directory size is something like `512` and we ended up reporting 0KB? - I guess it is depending on the file system and OS implemetations)

src/OpenTelemetry.Contrib.PersistentStorage/OpenTelemetry.Contrib.PersistentStorage.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <PackageTags>$(PackageTags);Shared</PackageTags>
```
reyang, line 5: Hmm, why do we need "Shared" nuget tag? Is this a common practice?

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        internal static void RemoveExpiredBlobs(string directoryPath, long retentionPeriodInMilliseconds, long writeTimeoutInMilliseconds)
+        {
+            var currentUtcDateTime = DateTime.UtcNow;
+
+            var leaseDeadline = currentUtcDateTime;
+            var retentionDeadline = currentUtcDateTime - TimeSpan.FromMilliseconds(retentionPeriodInMilliseconds);
+            var timeoutDeadline = currentUtcDateTime - TimeSpan.FromMilliseconds(writeTimeoutInMilliseconds);
+
+            foreach (var file in Directory.EnumerateFiles(directoryPath).OrderByDescending(f => f))
```
reyang, line 111: Consider rename the bound variable `f` to `filename` for better readability.

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
```diff
@@ -0,0 +1,260 @@
...
+        internal static long GetDirectorySize()
+        {
+            return Interlocked.Read(ref directorySize);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static void WriteAllBytes(string path, byte[] buffer)
+        {
+            File.WriteAllBytes(path, buffer);
+            UpdateDirectorySize(buffer.Length);
```
reyang, line 143: What if the file got removed by other processes, do we update/fix the directory size periodically in the maintenance job?

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
```diff
@@ -0,0 +1,260 @@
...
+        {
+            return Interlocked.Add(ref directorySize, fileContentLength);
+        }
+
+        internal static DateTime GetDateTimeFromBlobName(string filePath)
+        {
+            var fileName = Path.GetFileNameWithoutExtension(filePath);
+            var time = fileName.Substring(0, fileName.LastIndexOf('-'));
+            DateTime.TryParseExact(time, "yyyy-MM-ddTHHmmss.fffffffZ", CultureInfo.InvariantCulture, DateTimeStyles.None, out var dateTime);
+            return dateTime.ToUniversalTime();
```
reyang, line 199: What if `TryParseExact` failed (e.g. the operating system automatically created some file such as `Thumbs.db` or `.fseventsd`)?

reyang (approved):
LGTM in general, I've left many suggestions/comments but I think we can address them in a separate PR given the size of the change.

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
vishweshbankwar: We do not check the directorySize periodically. We only update the directory size when a new file is added or a file is removed. The folder where data is stored is unique to one process so chances of other process removing the file is less assuming each process has unique identity.
```c#
string appIdentity = Environment.UserName + "@" + Path.Combine(baseDirectory, Process.GetCurrentProcess().ProcessName);
string subdirectoryName = GetSHA256Hash(appIdentity);
subdirectoryPath = Path.Combine(path, subdirectoryName);
Directory.CreateDirectory(subdirectoryPath);
```

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
vishweshbankwar: This needs to be handled accordingly - will fix it.

src/OpenTelemetry.Contrib.PersistentStorage/OpenTelemetry.Contrib.PersistentStorage.csproj
vishweshbankwar: We do not - will fix it.

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
reyang: Thanks! Consider addressing this in a separate PR to keep each PR scoped.

reyang (approved):

src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
reyang: > The folder where data is stored is unique to one process so chances of other process removing the file is less assuming each process has unique identity (not guaranteed).

I'm bit confused - what is the definition of "one process"? Is it a unique instance of process (identified by PID)? If that's the case, a process restart (which would have a different PID) would cause the new process not able to pick up the file left by the previous process.

reyang: My current understanding is that we affinitize the storage folder to an "app", and by default the app is identified by user name + proc name. Let's say we have `dotnet.exe` CLI being invoked thousands of times during a project build time, each of them would have different PID, but we treat them as the same app, so they could run concurrently and access the same folder.

vishweshbankwar: This is my bad
1) We do recalculate the directorySize on maintenance job.
2) And Yes storage folder is affinitize to an "app"

rajkumar-rangaraj (approved):
LGTM

test/OpenTelemetry.Contrib.PersistentStorage.Tests/OpenTelemetry.Contrib.PersistentStorage.Tests.csproj
```diff
@@ -0,0 +1,23 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net461</TargetFrameworks>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNETTestSdkPkgVer)" />
+    <PackageReference Include="Moq" Version="$(MoqPkgVer)" />
```
cijothomas, line 11: do we use this anywhere?

src/OpenTelemetry.Contrib.PersistentStorage/StorageEventSource.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Diagnostics.Tracing;
+using System.Runtime.CompilerServices;
+
+namespace OpenTelemetry.Contrib.PersistentStorage
+{
+    [EventSource(Name = EventSourceName)]
+    internal sealed class StorageEventSource : EventSource
+    {
+        public static StorageEventSource Log = new StorageEventSource();
+        private const string EventSourceName = "OpenTelemetry-Shared";
```
cijothomas, line 27: Use a name matching the project.

src/OpenTelemetry.Contrib.PersistentStorage/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog - OpenTelemetry.Contrib.PersistentStorage
+
+## 1.0.0 [To Be Released]
+
+This is the first release for the `OpenTelemetry.Contrib.PersistentStorage`
+project.
+
+For more details, please refer to the
```
cijothomas, line 8: use relative links

src/OpenTelemetry.Contrib.PersistentStorage/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog - OpenTelemetry.Contrib.PersistentStorage
+
+## 1.0.0 [To Be Released]
```
cijothomas, line 3: simply say Unreleased.

cijothomas (approved):

PR #176, opened 2022-01-19
Fix todo [OpenTelemetry.Contrib.PersistentStorage]

vishweshbankwar:


src/OpenTelemetry.Contrib.PersistentStorage/PersistentStorageHelper.cs
```diff
@@ -35,7 +35,7 @@ internal static void RemoveExpiredBlob(DateTime retentionDeadline, string filePa
             if (filePath.EndsWith(".blob", StringComparison.OrdinalIgnoreCase))
             {
                 DateTime fileDateTime = GetDateTimeFromBlobName(filePath);
-                if (fileDateTime < retentionDeadline)
+                if (fileDateTime != DateTime.MinValue && fileDateTime < retentionDeadline)
```
rajkumar-rangaraj, line 38: Is there a use case where this could happen?

PR #177, opened 2022-01-20
Event counter listener

hananiel:
Resolves open-telemetry/opentelemetry-dotnet-contrib#215

This PR provides the option to use any EventCounter as metric instrumentation, using an [EventListener](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventlistener?view=net-6.0) to "subscribe" to all Event sources the developer has specified.

The listener reads the event counter payload and passes it to the [MeterTelemetryPublisher](https://github.com/hananiel/opentelemetry-dotnet-contrib/blob/b3a7d27320aaccf3a979b65f2fb445a60a29cd58/src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/MeterTelemetryPublisher.cs#L25).

The publisher creates the instrumentation based on user provided definition or if none is provided,  by using the information given by the event counter (see [Inferring the Instrument Type](https://github.com/hananiel/opentelemetry-dotnet-contrib/blob/b3a7d27320aaccf3a979b65f2fb445a60a29cd58/src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/MeterTelemetryPublisher.cs#L72)).

The configuration is very easy. You can add all event counters of an event source just by
```csharp
 options.AddEventSource("Microsoft-AspNetCore-Server-Kestrel")
```

or define each event counter fine grained by

```csharp
options.AddEventSource("Microsoft-AspNetCore-Server-Kestrel")
 .With("connections-per-second", "The number of connections per update interval", InstrumentationType.Gauge)
 .With("total-connections", "The number of total connections", InstrumentationType.Gauge)
```


src/OpenTelemetry.Contrib.EventCounterListener/OpenTelemetry.Contrib.EventCounterListener.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <Description>OpenTelemetry Metrics for Dotnet EventCounters</Description>
+        <PackageTags>$(PackageTags);</PackageTags>
+        <MinVerTagPrefix>EventCounterListener-</MinVerTagPrefix>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Bcl.HashCode" Version="1.1.1" />
+        <PackageReference Include="Microsoft.Diagnostics.Monitoring.EventPipe" Version="5.0.257301" />
```
cijothomas, line 12: we shouldn't need this dependency

src/OpenTelemetry.Contrib.EventCounterListener/EventCounterListener.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Reflection;
+using Microsoft.Diagnostics.Monitoring.EventPipe;
+using OpenTelemetry.Contrib.EventCounterListener.EventPipe;
+
+namespace OpenTelemetry.Contrib.Instrumentation.EventCounterListener
+{
+    /// <summary>
+    /// EventCounterListener that subscribes to EventSource Events.
+    /// </summary>
+    public class EventCounterListener : EventListener
```
cijothomas, line 15: does this need to be public?

src/OpenTelemetry.Contrib.EventCounterListener/EventCounterListener.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        /// <summary>
+        /// Processes a new EventSource event.
+        /// </summary>
+        /// <param name="eventData">Event to process.</param>
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
+        {
+            if (eventData == null)
+            {
+                throw new ArgumentNullException(nameof(eventData));
```
cijothomas, line 46: does anyone catch this?

src/OpenTelemetry.Contrib.EventCounterListener/README.md
```diff
@@ -0,0 +1,58 @@
...
+                    .AddConsoleExporter()
+                    .Build();
+
+            System.Threading.Thread.Sleep(15000); // Give it some time to record metrics
+
+        }
+    }
+}
+```
+
```
cijothomas, line 55: the usage is not clear... could you describe with an example?

hananiel: If you are looking for how configuration is done, I havent coded that up yet, just making sure what I have so far is in the ballpark. Using the extension method provided (`AddEventCounterListener`) adds the EventListener that then republishes counters via metricsAPI. For example running the program listing in the `Readme.md` above produces the following output:

- I assumed all counters will be instrumented as guages 
- Since this requires aggregation in the `collect` interval, the appropriate one is last value

`Export cpu-usage, Unit: CPU Usage, Meter: OpenTelemetry.Contrib.EventCounterListener/0.0.0.0                                                                                                                                                 (2022-01-20T03:48:23.3248145Z, 2022-01-20T03:48:37.2994387Z] DoubleGauge                                                                                                                                                                     Value: 0                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  Export working-set, Unit: Working Set, Meter: OpenTelemetry.Contrib.EventCounterListener/0.0.0.0                                                                                                                                             (2022-01-20T03:48:23.3275819Z, 2022-01-20T03:48:37.2994398Z] DoubleGauge                                                                                                                                                                     Value: 31                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 Export gc-heap-size, Unit: GC Heap Size, Meter: OpenTelemetry.Contrib.EventCounterListener/0.0.0.0                                                                                                                                           (2022-01-20T03:48:23.3276834Z, 2022-01-20T03:48:37.2994400Z] DoubleGauge                                                                                                                                                                     Value: 5                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  Export gen-0-gc-count, Unit: Gen 0 GC Count, Meter: OpenTelemetry.Contrib.EventCounterListener/0.0.0.0                                                                                                                                       (2022-01-20T03:48:23.3280650Z, 2022-01-20T03:48:37.2994402Z] DoubleGauge                                                                                                                                                                     Value: 0   `

cijothomas: without seeing how a user would configure, its hard to comment on it.
Some qns:
What are the inputs provided by the user? Do they pick each and every eventcounter?
Given there are different types of EventCounters, do we let user let us know if they want it treated as counter/gauge/histogram?

(Ofcourse we don't want to solve it in 1st iteration, but want to get a feel of how would the final version look like )

NuGet.config
```diff
@@ -4,6 +4,8 @@
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
     <add key="MyGet" value="https://www.myget.org/F/opentelemetry/api/v3/index.json" />
+    <add key="Dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
```
cijothomas, line 7: we can remove this, as we won't need any packages from non nuget.org/myget.org

src/OpenTelemetry.Contrib.EventCounterListener/README.md
hananiel: > What are the inputs provided by the user? Do they pick each and every eventcounter?

I was thinking they can filter by EventSource, and then get more granular at the particular counter. 
>Given there are different types of EventCounters, do we let user let us know if they want it treated as counter/gauge/histogram?

I have made a list of [well known eventcounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters) and ran [Reiley's "algorithm" for picking an instrument](https://github.com/open-telemetry/opentelemetry-specification/pull/1617#issue-858179203). In each case they made no sense to add them, since the eventCounters themselves are preaggregated, it makes no sense to further aggregate them - let me know if I am wrong about this or a different guide to pick instruments. Dependending on the answer to this,  we could also make this the default option while giving users customizability to change this. 

If you need to get on a zoom call for 15 mins, I am available today or next week is fine as well.

twenzel:
Can we rename the projects `OpenTelemetry.Contrib.EventCounterListener` (+`.Tests`) to `OpenTelemetry.Contrib.Instrumentation.EventCounterListener` to be the same as the used namespace and as the other projects?

src/OpenTelemetry.Contrib.EventCounterListener/README.md
hananiel: I did take a crack at what usage might look like if it needs to be able to support multiple instrument types. However if we can ascertain that we dont need to use other instrument types because of the nature of eventcounters, then everything will be very simple.  
```csharp
   this.meterProvider = Sdk.CreateMeterProviderBuilder()
                 .AddEventCounterListener(options =>
                 {
                    options.Sources = new[]
                    {
                        new EventSourceOption
                        {
                            EventSourceName = "System.Runtime",
                            RefreshInterval = "1",
                            EventCounters = new[] { new EventCounter{ Name = "cpu-usage", Type = "ObservableGuage" } },
                        },
                    };
                 })
                .AddReader(metricReader)
                .Build();
```

hananiel: I did realize something I had misunderstood about different eventcounter types. I'll rework this.

twenzel:
Hi @cijothomas   🙋🏼‍♂️,
we think the code is now fine for a review. 
Please have look and tell us what you think.

NuGet.config
hananiel: removed ..

examples/eventcounters/Examples.EventCounters.AspNetCore/Examples.EventCounters.AspNetCore.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
```
cijothomas, line 1: i'd suggest to not have examples here. The readme can explain the usage without a full blown example application.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCounter.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the event counter name.
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Gets or sets the metric description.
+        /// </summary>
+        public string Description { get; set; }
```
cijothomas, line 34: is this the description of the outputted Metric?

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCounter.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Gets or sets the metric type.
+        /// </summary>
+        public MetricType? Type { get; set; } = MetricType.LongSum;
```
cijothomas, line 39: It'd be good to ask user which instrument to use for this eventcounter (like ObservableCounter or ObservableGauge)

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCountersOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// Options for <see cref="EventCounterListener"/>.
+    /// </summary>
+    public class EventCountersOptions
+    {
+        /// <summary>
+        /// Gets or sets the interval in seconds.
+        /// </summary>
+        public int RefreshIntervalSecs { get; set; } = 1;
```
cijothomas, line 30: its not clear why we have RefreshInterval in this class and also in `EventSourceOtion`

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/SystemRuntime/SystemRuntimeOptionsExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using OpenTelemetry.Contrib.Instrumentation.EventCounters;
+using OpenTelemetry.Contrib.Instrumentation.EventCounters.SystemRuntime;
+
+namespace OpenTelemetry.Metrics
+{
+    /// <summary>
+    /// Extension methods for the event counters options.
+    /// </summary>
+    public static class SystemRuntimeOptionsExtensions
```
cijothomas, line 25: maybe good to keep the initial scope smaller. All this (along with grpc/aspnetcore) can be added by users without these as well. If there is a need, we can always add new things.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCounter.cs
twenzel: Yes.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCountersOptions.cs
twenzel: The property in `EventSourceOtion`  will be removed.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/SystemRuntime/SystemRuntimeOptionsExtensions.cs
twenzel: So should we remove these predefined counters or should we move those to a separate project each?

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/EventCounter.cs
twenzel: The user can define whether a ObservableCounter (MetricType.LongSum, MetricType.DoubleSum) or a ObservableGauge (MetricType.LongGauge, MetricType.DoubleGauge) or a Histogram (MetricType.Histogram) should be used.

```csharp
 options.AddEventSource("Microsoft-AspNetCore-Server-Kestrel")
             .With( "connections-per-second", "The number of connections per update interval to the web server", MetricType.LongGauge)
);
```

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/README.md
```diff
@@ -0,0 +1,128 @@
...
+                    .AddConsoleExporter()
+                    .Build();
+
+            System.Threading.Thread.Sleep(15000); // Give it some time to record metrics
+        }
+    }
+}
+```
+
+Configuration can be done completely using the IConfiguration.
```
cijothomas, line 59: dont think we need to document the usage of IConfiguration or anything else here. Its not related to this package.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/README.md
```diff
@@ -0,0 +1,128 @@
...
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Contrib.Instrumentation.EventCounters;
+
+namespace DotnetMetrics
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            using var meterprovider = Sdk.CreateMeterProviderBuilder()
```
cijothomas, line 40: could you describe the overall approach/design used in the PR description? Its a bit hard to figure out the design in a big PR without some writeup.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/SystemRuntime/SystemRuntimeOptionsExtensions.cs
cijothomas: remove now. The readme.md can show how users can enable them. (all they need is the name of eventsource,counter, whether to treat it as gauge/counter).
Once this PR is merged, we can discuss the best options on how to support the well known counters - either as part of this project itself, or a separate project (nugets) etc.

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/EventCounterListener.cs
```diff
@@ -0,0 +1,243 @@
...
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            // Keeping track of all EventSources here, as this call may happen before initialization.
+            this.allEventSourcesCreated.Enqueue(eventSource);
+
+            // options might be null when this method gets called
+            // before our constructor was completed
+            if (this.isInitialized)
+            {
+                this.EnableEventSource(eventSource);
```
Im5tu, line 125: There's a small chance here that if we're not initialized that we loose the reference to the event source

Im5tu (commented):
Sorry it took me a few days to see the message @cijothomas - looks good, just a minor comment. Think we need to be aware that some EventSource's are available on startup as they are only captured on first use. So this could be before the EventListener has been created depending on the configuration of the app. Code as is will miss this if it's not fully initialized from what I can see

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/EventCounterListener.cs
twenzel: This scenario is already handled. The base EventListener constructor will call `OnEventSourceCreated` on every existing EventSource. Therefor our OnEventSourceCreated will [queue](https://github.com/hananiel/opentelemetry-dotnet-contrib/blob/b3a7d27320aaccf3a979b65f2fb445a60a29cd58/src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/EventCounterListener.cs#L119) all eventsources and "[enable](https://github.com/hananiel/opentelemetry-dotnet-contrib/blob/b3a7d27320aaccf3a979b65f2fb445a60a29cd58/src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/EventCounterListener.cs#L67)" then afterwards.

cijothomas: Yup! This is well known and is handled by OpenTelemetry SDK as well for its internal logging (https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/SelfDiagnosticsEventListener.cs)

src/OpenTelemetry.Contrib.Instrumentation.EventCounters/InstrumentationType.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    /// Defines the type of the created instrumentation.
+    /// </summary>
+    public enum InstrumentationType
+    {
+        /// <summary>
+        /// Creates a counter instrumentation of long values which is an instrument that reports monotonically
+        /// increasing values.
+        /// </summary>
+        Counter = 0,
```
cijothomas, line 28: Counter is an sync instrument, so we should not use it from EventCounter. 
Only supported instruments should be the ObservableCounter, ObservableGauge, and ObservableUpDownCounter (when that instrument is added)

twenzel: It's only named "Counter". The [publisher creates](https://github.com/hananiel/opentelemetry-dotnet-contrib/blob/b3a7d27320aaccf3a979b65f2fb445a60a29cd58/src/OpenTelemetry.Contrib.Instrumentation.EventCounters/Implementation/MeterTelemetryPublisher.cs#L86) an `ObservableCounter`.
Should we rename the `Counter` enum value to `ObservableCounter`?

twenzel: The names were renamed.

twenzel:
@cijothomas Anything else missing?

ZEXSM:
@hananiel have you considered receiving diagnostic data via EventPipeSession, as is done by e.g. dotnet counters, instead of subscribing to an EventListener?

https://docs.microsoft.com/en-us/dotnet/core/diagnostics/eventpipe
https://docs.microsoft.com/en-us/dotnet/core/diagnostics/diagnostics-client-library

twenzel:
@ZEXSM EventPipes should be used to retrieve the data for out-proc scenarios. EventListeners can be used for in-proc. Since we're running in-proc the EventListeners are the prefered way.

See https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters

cijothomas:
@hananiel @twenzel To unblock and make progress, I'd suggest to break this down into smaller PRs, similar to suggestion for runtime metrics PR (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/207#issuecomment-1057606111).

For this one, consider following
1. Add project/workflow
2. Extension method to enable, with just a basic support of specifying list of EventSource/EventCounter name. Default to ObservableGauge for all.
3. Add usage and example
4. Add capability to add allow user to specify which Instrument to use for each eventcounter.
5. Release 1st alpha and iterate based on initial feedback.

(its generally very hard to get people to review big PRs, so I'd strongly suggest small PRs, each addressing a specific concern, to  constantly make progress, and hopefully attract more reviewers.)

hananiel:
Closing in lieu of #365

PR #179, opened 2022-01-25
FIX #178 incompatibility between EFCore + SqlClient libs

### LINKED ISSUE #178: Incompatibility between EFCore and SqlClient instrumentation libs
There are incompatibilities between:

- OpenTelemetry.Instrumentation.SqlClient v1.0.0-rc8
- OpenTelemetry.Contrib.Instrumentation.EntityFrameworkCore v1.0.0-beta2

# Repro steps:

1. Create an empty project
2. Add EFCore + OpenTelemetry instrumentation libs (EFCore + SqlClient)
3. Do any request using EFCore DbContext

# Expected behavior
Both EFCore + SqlClient spans should be correctly reported with the SqlClient one created as a child of EFCore span.

# Actual behavior
SqlClient spans are created as children of EFCore ones, yet EFCore spans are never completed and therefore never reported. Then, resulting in SqlClient spans pointing to a non-existing parent span ID.

# Explanation
Both SqlClient and EFCore intrumentation libs rely on AsyncLocal mechanism to correctly start/enrich/complete their spans. However when the database operation is completed both diagnostic events are triggered and there is no order guarantee between both listeners.

Most of the times EFCore code runs before SqlClient has the opportunity to stop the current span. When EFCore lib reaches this point:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/9373eb54d52ef43b9e0cb2284f86f668cad1dbc2/src/OpenTelemetry.Contrib.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L197 

since the `activity` is still the SqlClient span, the `EntityFrameworkCoreCommandExecuted` is ignored and the span is never completed.

There is a similar issue reported [I44](https://github.com/opentracing-contrib/csharp-netcore/issues/44) in OpenTracing lib before they released a fix [PR82](https://github.com/opentracing-contrib/csharp-netcore/pull/82)

# Possible solution

The most obvious solution is to stop relying on AsyncLocal and start using some kind of local data struct to store those references, similar to what was done in OpenTracing lib. I'll try to create a PR with this approach.

Thank you in advance for your time!

andrembpontes:
Use internal ConcurrentDictionary to store managed activities. Using internal state instead of relying on AsyncLocal solves race conditions detected between EFCore and SqlClient instrumentation libs.

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System;
+using System.Collections.Concurrent;
 using System.Data;
 using System.Diagnostics;
 using OpenTelemetry.Instrumentation;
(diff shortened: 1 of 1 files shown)


prochnowc:
@andrembpontes Can you please re-open the PR ?

apontes-pagerduty:
Hey! I would like to get it merged but I received no feedback at all. The PR was auto closed due to inactivity

PR #190, opened 2022-01-28
Add MassTransit activity enrich possibility

glucaci:
It will be helpful to have enrich possibility on the MassTransit activity.


glucaci:
Just saw that it's another PR open for this

PR #203, opened 2022-02-04
Include an option to hide dbstatement tag in ElasticSearch

simaoribeiro:
Hi there.

Similar to what already exists for SQL instrumentation https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/README.md#advanced-configuration, this pull request includes an option to exclude the `db.statement` tag preventing the capture of sensitive data.

The option is named `SetDbStatementForRequest` and is enabled by default.


simaoribeiro:
Another question to discuss in this pull request is that the specification mentions that, when `db.statement` is not included, `db.operation` is required. 
ElasticSearch instrumentation doesn't have a `db.operation` but has a `db.method` tag that is not included in the database semantic convention. Is it a good solution to replace the tags keeping the same value?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

examples/elasticsearch/Examples.ElasticSearch/Examples.ElasticSearch.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
cijothomas, line 1: I'd suggest to remove this example from this PR. We haven't formulated a plan for how this repo would maintain examples in the long term. (too many examples, one per project, is going to be very difficult to maintain.).
Until the plan is sorted out, I'd recommend to not add examples.

simaoribeiro: Hi @cijothomas ,

I removed the sample project as suggested. Can you re-trigger the build please?

simaoribeiro:
> Another question to discuss in this pull request is that the specification mentions that, when `db.statement` is not included, `db.operation` is required. ElasticSearch instrumentation doesn't have a `db.operation` but has a `db.method` tag that is not included in the database semantic convention. Is it a good solution to replace the tags keeping the same value?
> 
> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

Also, regarding this question, is it something that should be addressed?

cijothomas:
@ejsmith - Could you review this PR?

ejsmith (approved):
Looks good to me.

ejsmith:
> > Another question to discuss in this pull request is that the specification mentions that, when `db.statement` is not included, `db.operation` is required. ElasticSearch instrumentation doesn't have a `db.operation` but has a `db.method` tag that is not included in the database semantic convention. Is it a good solution to replace the tags keeping the same value?
> > https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes
> 
> Also, regarding this question, is it something that should be addressed?

I don't know about this. I think I copied this from the http client library instrumentation.

simaoribeiro:
@ejsmith I will create an issue to discuss the tag nomenclature so we don't block this PR

cijothomas (approved):

Issue #201, opened 2022-02-02
AWSEKSResourceDetector - Validate ClusterName/ContainerID independently before adding it to the resource


swetharavichandrancisco:
EKS Resources in Java are decoupled for ClusterName and ContainerID

Reference: https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource/EksResource.java
```
  String clusterName = getClusterName(httpClient);
    if (clusterName != null && !clusterName.isEmpty()) {
      attrBuilders.put(ResourceAttributes.K8S_CLUSTER_NAME, clusterName);
    }

    String containerId = dockerHelper.getContainerId();
    if (containerId != null && !containerId.isEmpty()) {
      attrBuilders.put(ResourceAttributes.CONTAINER_ID, containerId);
    }
```

For .Net, both cluster name and container id are required to add the EKS resources. It makes sense to have them separated like Java because it makes kubernetes api calls from amazon-cloudwatch to extract cluster name and that won't be needed to extract container id.

```
if (clusterName == null && containerId == null)
{
         return resourceAttributes;
}

```

swetharavichandrancisco:
Will create a PR for the same

srprash:
Hi @swetharavichandrancisco 

> For .Net, both cluster name and container id are required to add the EKS resources.

I think that's not the case here. If both `clusterName` and `containerId` are null, nothing is added as resource attribute. However, if either or both of these values are not-null, they will be captured as resource attribute.
The only downside could be that we may capture null value for one of the attribute from `clusterName` and `containerId`.
Java implementation solves this by independently validating each attribute and then adding it to the resource, which can be done as an improvement here as well.
Please feel free to open a PR and I will be happy to review it. Thanks!

swetharavichandrancisco:
Thank you @srprash for the clarification. 

In order to get the `clusterName`, we need to have CloudWatch agent/Fluent Bit installed. 
With the current code, `ClusterName` is extracted from the configmaps of amazon-cloudwatch namespace -`https://kubernetes.default.svc/api/v1/namespaces/amazon-cloudwatch/configmaps/cluster-info`.

As per my understanding, having CloudWatch agent/FluentD  is an optional step in EKS which is mainly used for sending metrics/logs to CloudWatch.  Please correct me if I'm wrong. 

[...]

PR #205, opened 2022-02-08
 #201 AWSEKSResourceDetector  - Validate ClusterName/ContainerID independently before adding it to the resource

swetharavichandrancisco:
#201 AWSEKSResourceDetector - Validate ClusterName/ContainerID independently before adding it to the resource

- Validates ClusterName/ContainerID independently before adding it to the resource
- Includes bug fix - IsEKSProcess should return non-null auth value to confirm if its running in EKS environment.
- Handled exceptions separately with more error information
- Added Unit Tests


srprash (approved):
Thanks for making this change. LGTM.

Issue #204, opened 2022-02-08
Provide .NET Runtime instrumentation libraries


reyang:
# Feature Request

.NET has several runtime metrics exposed via EventCounters https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs.

In the near future it is impossible for .NET runtime to completely switch over to the Meter API since these APIs are not backported to old versions of runtime.

**Describe the solution you'd like:**

The ask here is to provide an instrumentation library which exposes runtime metrics via Meter API, so users who need these could leverage them without having to use EventCounter wrappers.

**Describe alternatives you've considered.**

Leveraging something like https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/177 or having the users implementing the Meter/Instrument by themselves.

Leveraging the EventCounter listener/adapter has lots of problems with the scheduling - EventCounters would require the user to configure the polling schedule, which makes it less efficient for pull exporters (e.g. Prometheus), and hard to manage for push exporters (e.g. if OTLP exporter is exporting every 10 seconds, what's the proper polling period for EventCounters?)

twenzel:
I would like to contribute to this issue. I'll try to create a PR in the next days.

steinarox:
Nice to see i am not the only one that looks for this 😎 
For UX i like the [prometheus-net instrumentation](https://github.com/djluck/prometheus-net.DotNetRuntime#start-collecting-metrics) where you can add all or some groups of "default" metrics

omamoo:
@reyang Thanks for the sharing with this feature request, It will be useful so much to  add such kind of instrumentation.

twenzel:
@reyang Can this be closed?

reyang:
@twenzel it's not 100% done but I think YES we should close it and iterate on your contributions. Thank you @twenzel for leading this effort!

PR #207, opened 2022-02-16
Runtime instrumentation

twenzel:
This PR implements Runtime instrumentation (#204 ).

It create instrumentation for all runtime metrics unless configured.

All information is retrieved directly (similar to [RuntimeEventSource](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs)).
There are only metrics for values which can be retrieved via Runtime API.


src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal class AssembliesInstrumentation : IRuntimeInstrumentation
+    {
+        private readonly Meter meter;
+        private readonly ObservableCounter<int> assemblyCounter;
+
+        public AssembliesInstrumentation(Meter meter)
+        {
+            this.meter = meter;
+            this.assemblyCounter = meter.CreateObservableCounter("assembly-count", () => AppDomain.CurrentDomain.GetAssemblies().Length, description: "Number of Assemblies Loaded");
```
alanwest, line 30: I think the names for the metrics in this PR should be prefixed with the `process.runtime.dotnet.` namespace. In this case something like `process.runtime.dotnet.assembly_count` (using an `_` rather than `-` since I think that's more commonly what I've seen in other semantic conventions).

The spec has some general guidance here for runtime metrics: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---processruntimeenvironment

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/PerformanceInstrumentation.cs
```diff
@@ -0,0 +1,42 @@
...
+        private readonly IEventCounterStore eventCounterStore;
+        private readonly ObservableGauge<double> cpuTimeCounter;
+        private readonly ObservableGauge<double> workingSetCounter;
+
+        public PerformanceInstrumentation(Meter meter, IEventCounterStore eventCounterStore)
+        {
+            this.meter = meter;
+            this.eventCounterStore = eventCounterStore;
+
+            this.eventCounterStore.Subscribe(CpuTimeCounterName, EventCounterType.Mean);
```
alanwest, line 36: The instrumentation classes should probably implement `IDisposable` and clean up subscriptions in the event the `MeterProvider` is shutdown.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
twenzel: Thank you. Names are now prefixed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/PerformanceInstrumentation.cs
twenzel: Great hint. Subscription clean up is now done on Dispose.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// </summary>
+        /// <param name="eventCounterStore">Instance to store the data to.</param>
+        public EventCounterListener(IEventCounterStore eventCounterStore)
+        {
+            this.eventCounterStore = eventCounterStore ?? throw new ArgumentNullException(nameof(eventCounterStore));
+
+            try
+            {
+                this.refreshIntervalDictionary = new Dictionary<string, string>();
+                this.refreshIntervalDictionary.Add("EventCounterIntervalSec", "1");
```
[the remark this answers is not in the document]
twenzel: This PR is dedicated to the runtime metrics and is using the API directly to retrieve the values. 
Unfortunately sometimes the API is internal, so we've to fallback to the EventCounter. 
The advantage of a this instrumentation is that it can use the direct API as soon as the runtime make it public. It will perform better as retrieving the values from event counters with less overhead. 
The EventCounter instrumentation from #177 is a general purpose implementation to retrieve any existing EventCounter.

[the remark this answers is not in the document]
reyang: I think ideally we want this "Runtime Instrumentation" to use the OpenTelemetry Metrics API rather than relying on the EventCounter. The main reason is efficiency and accuracy - especially considering Prometheus scenario that [using an asynchronous instrument](https://github.com/open-telemetry/opentelemetry-dotnet/blob/f2fcb53214d737b089bad523218b1ef37dda30c2/docs/metrics/learning-more-instruments/Program.cs#L33) will make sure we only invoke the callback when Prometheus is scaping metrics, rather than having to poll the EventCounter based on some predetermined schedule.

If there are cases where we cannot use OpenTelemetry Metrics API (e.g. there is no proper API - such as ObservableUpDownCounter) or we cannot access the internal functionality of .NET runtime (e.g. if a certain EventCounter is implemented by calling some non-public API), we should discuss and see if we can find better approaches.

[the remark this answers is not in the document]
reyang: > Unfortunately sometimes the API is internal, so we've to fallback to the EventCounter.

I suggest that we call out the list of APIs, and we should work with the .NET Runtime team to see if it makes sense to make these public.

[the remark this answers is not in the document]
reyang: > @hananiel the goal for #207 is to have a pure OpenTelemetry Metrics API based instrumentation library for .NET runtime - that does not rely on EventCounter at all (as a consequence, provides better efficiency, sampling accuracy and probably better mode - e.g. different gen of GC could be modeled as metric dimensions/attributes rather than each modeled as a separate metric).

omamoo:
@twenzel Can you make the `MetricPrefix` configurable? `process_runtime_dotnet` is prefix matching Prometheus but differ than Datadog metrics for example.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/GcInstrumentation.cs
```diff
@@ -0,0 +1,102 @@
...
+            this.eventCounterStore = eventCounterStore;
+
+            this.eventCounterStore.Subscribe(GcTimeCounterName, EventCounterType.Mean);
+            this.eventCounterStore.Subscribe(Gen0SizeCounterName, EventCounterType.Mean);
+            this.eventCounterStore.Subscribe(Gen1SizeCounterName, EventCounterType.Mean);
+            this.eventCounterStore.Subscribe(Gen2SizeCounterName, EventCounterType.Mean);
+            this.eventCounterStore.Subscribe(LohSizeCounterName, EventCounterType.Mean);
+            this.eventCounterStore.Subscribe(PohSizeCounterName, EventCounterType.Mean);
+
+            this.gcHeapSizeCounter = meter.CreateObservableCounter($"{RuntimeMetrics.MetricPrefix}gc_heap_size", () => (double)(GC.GetTotalMemory(false) / 1_000_000), "MB", "GC Heap Size");
```
omamoo, line 60: Can I ask a question?
Why `ObservableCounter` and not `ObservableGauge`? 

From the docs: 
ObservableCounter - ObservableCounter is an Instrument which reports monotonically increasing value(s)
Observable Gauge - ObservableGauge is an asynchronous Instrument which reports non-additive value(s)

The heap size is non-additive value and can decrease.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/GcInstrumentation.cs
```diff
@@ -0,0 +1,102 @@
...
+            this.gen0GCCounter = meter.CreateObservableGauge($"{RuntimeMetrics.MetricPrefix}gen_0-gc_count", () => GC.CollectionCount(0), description: "Gen 0 GC Count");
+            this.gen1GCCounter = meter.CreateObservableGauge($"{RuntimeMetrics.MetricPrefix}gen_1-gc_count", () => GC.CollectionCount(1), description: "Gen 1 GC Count");
+            this.gen2GCCounter = meter.CreateObservableGauge($"{RuntimeMetrics.MetricPrefix}gen_2-gc_count", () => GC.CollectionCount(2), description: "Gen 2 GC Count");
+            this.allocRateCounter = meter.CreateObservableGauge($"{RuntimeMetrics.MetricPrefix}alloc_rate", () => GC.GetTotalAllocatedBytes(), "B", "Allocation Rate");
```
omamoo, line 64: Questions again here, `GC.GetTotalAllocatedBytes` is `IncrementingPollingCounter`
[RuntimeEventSource ](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs)

From the docs - Provides a variant of `EventCounter` for variables that are ever-increasing, such as the number of exceptions in the runtime

Why not to use `ObservableCounter`?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/GcInstrumentation.cs
twenzel: Fixed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/GcInstrumentation.cs
twenzel: Fixed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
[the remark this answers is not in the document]
twenzel: @reyang These are the APIs that are internal and should be made public:
- GC.GetLastGCPercentTimeInGC()
- GC.GetGenerationSize() // for generation 0-4
- System.Reflection.Assembly.GetAssemblyCount()
- Exception.GetExceptionCount()
- RuntimeEventSourceHelper.GetCpuUsage()

Should I make PRs or are you discuss these requests internally?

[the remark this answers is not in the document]
reyang: @twenzel Thank you! Let me discuss these with the .NET runtime team today, I will get back with the meeting notes.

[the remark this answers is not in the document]
cijothomas: I'd suggest to remove any EventCounter specific code from this PR. 
in the 1st version, we can support only those metrics which are exposed by runtime. 
and as runtime exposes more, we can modify this to support them as well.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptionsExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+// </copyright>
+
+using OpenTelemetry.Contrib.Instrumentation.Runtime;
+
+namespace OpenTelemetry.Metrics
+{
+    /// <summary>
+    /// Extensions to configure the runtime metrics options.
+    /// </summary>
+    public static class RuntimeMetricsOptionsExtensions
```
cijothomas, line 24: I think we don't need to have this (atleast in an initial release).As `RuntimeMetricsOptions` already has public properties to enable each category of metrics.

cijothomas (commented):
Overall approach seems good for me , except that we need to remove EventCounter specifics from this PR. It'll limit the number of metrics available, but we can work with runtime team to expose more, without going through EventCounters.

Will do another review once EventCounter stuff is removed (that should reduce the PR size by half)

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptionsExtensions.cs
reyang: > the general idea - start with the absolute minimum public API. We can always add more public APIs based on feedbacks from initial release.

+1 👍

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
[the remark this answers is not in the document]
twenzel: Ok, EventCounter specific code was removed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace OpenTelemetry.Contrib.Instrumentation.Runtime.Implementation
+{
+    internal class AssembliesInstrumentation : IRuntimeInstrumentation
+    {
+        private readonly Meter meter;
+        private readonly ObservableCounter<int> assemblyCounter;
+
+        public AssembliesInstrumentation(RuntimeMetricsOptions options, Meter meter)
+        {
+            this.meter = meter;
```
cijothomas, line 29: no need of storing meter here.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
```diff
@@ -0,0 +1,33 @@
...
+            this.assemblyCounter = meter.CreateObservableCounter($"{options.MetricPrefix}assembly_count", () => AppDomain.CurrentDomain.GetAssemblies().Length, description: "Number of Assemblies Loaded");
```
cijothomas, line 30: and don't think this counter is needed to be stored either. In short, all this class needs to do is create the observableInstrument with the callback.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
twenzel: Removed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/AssembliesInstrumentation.cs
twenzel: Removed.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Initializes a new instance of the <see cref="RuntimeMetrics"/> class.
+        /// </summary>
+        /// <param name="options">The options to define the metrics.</param>
+        public RuntimeMetrics(RuntimeMetricsOptions options)
+        {
+            this.meter = new Meter(InstrumentationName, InstrumentationVersion);
+
+            if (options.IsGcEnabled)
+            {
+                this.instrumentations.Add(new GcInstrumentation(options, this.meter));
```
cijothomas, line 47: maybe its easier to create the Observable*** right here, without additional IRunTimeInstrumentation abstractions? It doesn't seem to be serving much purpose since the EventCounter part is totally removed.

twenzel: Yep, you're right. For the 1st version this might be "oversized" 😉.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+    <PropertyGroup>
+        <TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
+        <Description>dotnet runtime instrumentation for OpenTelemetry .NET</Description>
+        <PackageTags>$(PackageTags);OpenTelemetry;runtime</PackageTags>
+        <MinVerTagPrefix>Instrumentation.Runtime-</MinVerTagPrefix>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Extensions.Options" Version="2.1.1" />
```
cijothomas, line 11: This dependency seems not required.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFrameworks>netcoreapp3.1;net6.0</TargetFrameworks>
```
cijothomas, line 4: could we do netstandard2.0 and net461?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
twenzel: Do we need to support `IDeferredMeterProviderBuilder`? This package is required (at the moment) to retrieve RuntimeMetricsOptions from the service provider.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
twenzel: In addition to `netcoreapp3.1;net6.0`?  Sure.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
cijothomas: `IDeferredMeterProviderBuilder` is really nice to have, but not *required*. Some suggestions:
1. Maybe remove it for now and we can add it back later.
2. If keeping, consider checking the pattern followed in https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis 

(My personal preference would be option1, so that we can quickly unblock this PR and do a 1st release.)

twenzel: Removed in order to unblock this PR.

cijothomas: thanks. we can revisit this after an initial release.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/OpenTelemetry.Contrib.Instrumentation.Runtime.csproj
cijothomas: thanks.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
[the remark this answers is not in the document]
reyang: > @reyang These are the APIs that are internal and should be made public:
> 
> * GC.GetLastGCPercentTimeInGC()
> * GC.GetGenerationSize() // for generation 0-4
> * System.Reflection.Assembly.GetAssemblyCount()
> * Exception.GetExceptionCount()
> * RuntimeEventSourceHelper.GetCpuUsage()
> 
> Should I make PRs or are you discuss these requests internally?

@cijothomas and I had a discussion with the .NET runtime team (@maryamariyan, @noahfalk and @tarekgh), here goes a quick summary:

1. Some of the methods (e.g. `RuntimeEventSourceHelper.GetCpuUsage()`) were introduced into the runtime to reduce dependencies, there is likely going to be public APIs in other .NET runtime assemblies that meet the requirements here.
2. Some methods might have no good alternatives that are public. From .NET perspective, @noahfalk can help to see what's the direction - e.g. we might make certain API public in the future version of runtime, or design a public APIs that provide the same or even more feature.
3. Given this OpenTelemetry package will have to work with existing versions of the runtime, we need use reflection (probably in combination with some conditional compilation or runtime detection), for the functions that we took reflection approach here, .NET team could add comment in the source code to avoid breaking OpenTelemetry accidentally - before we can fully move to public APIs.
4. All of the above doesn't need to block this PR, we can address these 5 missing instruments in separate PRs.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Advanced configuration
+
+By default all available runtime metrics will be added. It's also possible to
+specify only the required metrics:
+
+```csharp
+using var meterProvider = Sdk.CreateMeterProviderBuilder()
+    .AddRuntimeMetrics(options => options
+        .WithGc()
```
cijothomas, line 67: modify the examples to match the updated version.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    /// <summary>
+    /// Options to define the runtime metrics.
+    /// </summary>
+    public class RuntimeMetricsOptions
+    {
+        /// <summary>
+        /// Gets or sets a prefiix used for the metric names.
+        /// </summary>
+        public string MetricPrefix { get; set; } = "process_runtime_dotnet_";
```
cijothomas, line 27:
```suggestion
        public string MetricPrefix { get; set; } = "process.runtime.dotnet.";
```

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---processruntimeenvironment

cijothomas: Also, this need not be public.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether assembly metrics should be collected.
+        /// </summary>
+        public bool? AssembliesEnabled { get; set; }
+
+        /// <summary>
+        /// Gets a value indicating whether all metrics are enabled.
+        /// </summary>
+        internal bool IsAllEnabled => this.GcEnabled == null
```
cijothomas, line 61: have you considered making everything opt-in? i.e by default, user gets nothing by simply adding AddRunTimeMetrics(). One needs to explicitly enable individually to get them.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -0,0 +1,104 @@
...
+        /// </summary>
+        /// <param name="options">The options to define the metrics.</param>
+        public RuntimeMetrics(RuntimeMetricsOptions options)
+        {
+            this.meter = new Meter(InstrumentationName, InstrumentationVersion);
+
+            if (options.IsGcEnabled)
+            {
+                this.meter.CreateObservableGauge($"{options.MetricPrefix}gc_heap_size", () => (double)(GC.GetTotalMemory(false) / 1_000_000), "MB", "GC Heap Size");
+                this.meter.CreateObservableGauge($"{options.MetricPrefix}gen_0-gc_count", () => GC.CollectionCount(0), description: "Gen 0 GC Count");
```
noahfalk, line 46: I would suggest we use a single separator rather than a mixture of both "-" and "_".

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                this.meter.CreateObservableGauge($"{options.MetricPrefix}monitor_lock_contention_count", () => Monitor.LockContentionCount, description: "Monitor Lock Contention Count");
+                this.meter.CreateObservableCounter($"{options.MetricPrefix}threadpool_thread_count", () => ThreadPool.ThreadCount, description: "ThreadPool Thread Count");
+                this.meter.CreateObservableGauge($"{options.MetricPrefix}threadpool_completed_items_count", () => ThreadPool.CompletedWorkItemCount, description: "ThreadPool Completed Work Item Count");
+                this.meter.CreateObservableCounter($"{options.MetricPrefix}threadpool_queue_length", () => ThreadPool.PendingWorkItemCount, description: "ThreadPool Queue Length");
+                this.meter.CreateObservableCounter($"{options.MetricPrefix}active_timer_count", () => Timer.ActiveCount, description: "Number of Active Timers");
+            }
+#endif
+
+            if (options.IsPerformanceEnabled)
```
noahfalk, line 79: Perhaps this should be named "Memory", or "Process Memory"? "Performance" seems very broad.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
noahfalk: @cijothomas - I think whatever separator we use here should be the same one used in the instrument names for consistency? Also working set and cpu usage feel like outliers here. According to the naming specifications they would belong under process.cpu and process.memory

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
reyang: I would propose int64 `process.runtime.dotnet.gc.heap`, and the unit is `Bytes`.

For GC count, use `process.runtime.dotnet.gc.count`, and put `gen` as dimension/attribute.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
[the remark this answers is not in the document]
noahfalk: @reyang @cijothomas - Talking with Maoni about the percent-time-in-gc counter a similar issue arises that I see with the cpu % counter. It seems like what we'd want to do is represent these things as ObservableCounters which measure the number of seconds where something was true:
CPU = seconds/interval where CPU was in use
GC = seconds/interval where GC had all threads paused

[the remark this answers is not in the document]
reyang: Take an example to make sure I understand the proposal:

| Time since program started | Total time when GC had all threads paused |
| -------------------------- | ----------------------------------------- |
| 1.5 seconds                | 0.1 second                                |
| 6.5 seconds                | 1.1 second                                |
| 16.5 seconds               | 2.1 second                                |

* T0: 0s
* T1: 1.5s
* T2: 6.5s
* T3: 16.5s

Option 1:

* ObservableCounter("process.runtime.dotnet.gc.time")
* at T1, record 0.1 // (T0, T1)
* at T2, record 1.1 // (T0, T2)
* at T3, record 2.1 // (T0, T3)

The consumer can configure the exporter to use delta temporality, so proper rate can be calculated downstream.

Option 2:

* at T1, record 0.067 // 0.1 / 1.5
* at T2, record 0.2 // (1.1 - 0.1) / (6.5 - 1.5)
* at T3, record 0.1 // (2.1 - 1.1) / (16.5 - 6.5)

It seems Option 2 won't work, as the data being recorded won't fit any existing instrument.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
twenzel: So then the `configure` parameter in `MeterProviderBuilderExtensions.AddRuntimeMetrics` should be mandatory, right?. Otherwise it makes no sense to add AddRuntimeMetrics().

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
twenzel: This "category" is used to provide the cpu-usage as well, which was removed for the 1st version.
What about "ProcessPerformance"?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/Implementation/EventCounterListener.cs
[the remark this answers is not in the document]
noahfalk: Yep exactly, option 1 is what I was suggesting we should start doing and option 2 is the problematic path that the cpu-usage EventCounter is doing right now.

Option (2) works ok with an ObservableGauge only if there is guaranteed exactly one observer/exporter, allowing the callback to infer the interval by timestamping when it was last called. As soon as there is a 2nd observer then the callback can't distinguish if the call at T2 is a new observer trying to measure the interval [0, T2] or the same observer that called at T1 now trying to measure [T1, T2].

At risk of getting a little into the weeds, the current runtime implementation of GetLastGCPercentTimeInGC() doesn't capture a timestamp of when it was last called, but the way it avoids doing that produces a measurement people are likely to find confusing. In your example above, imagine that the various GCs occured at:
GC 1: 1.0-1.2  (0.1 sec paused)
GC 2: 2.5-3.9  (0.9 sec paused)
GC 3: 4.0-4.1  (0.1 sec paused)
GC 4: 15.0-16.2 (1 sec paused)
You can see GCs can also do some work while the threads aren't paused. Adding up the pause times gives the same 0.1, 1.1 and 2.1 measurements as above when observed at T=1.5, 6.5, and 16.5.
If you measure at some time T, GetLastGCPercentTimeInGC() finds the most recent GC X that finished before time T and calculates GC_x_pause/ ( GC_x_end - GC_x-1_end). For example if you query at T=6.5 it determines GC 3 was the most recent completed GC before that point which ended at 4.1. Before that GC 2 ended at 3.9. That gives 0.1 pause time/(4.1-3.9) = 50% time in GC. Users were probably hoping to see (1.1-0.1)/(6.5-1.5)=20% time in GC and will be surprised not to see that. So the current GC stats don't suffer non-determinism when subjected to multiple observers, but only by using a surprising calculation that isn't necessarily a good approximation of what users really wanted.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
noahfalk: I noticed later that we were planning to prefix all these metrics with "process.runtime.dotnet." and the OT semantic conventions suggest this value should be named something like "process.memory.xyz". Maybe we should leave this instrument out of this Meter and put it with a different collection of non-runtime process information?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
cijothomas: we could make some of them as enabled by default.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
omamoo: @noahfalk / @cijothomas 
I almot sure that `process.runtime.dotnet` won't work for Prometheus, since the separator must be '_'
should  we do it configurable somehow?

cijothomas: @omamoo I am not sure I follow fully.. The names would be as per the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---processruntimeenvironment - are you saying these names won't work for prometheus?

reyang: @omamoo PrometheusExporter will do the conversion from `.` to `_`. It's not a concern of the instrumentation library nor the SDK, individual exporter will have to do the escape/mapping.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/ff20a89a37d53b518acb62f73637cae84460ffaa/src/OpenTelemetry.Exporter.Prometheus/Implementation/PrometheusSerializer.cs#L177-L186

reyang (approved):
There are many nice discussions and I think all of them can be addressed in follow up PRs.

I would recommend that in this PR we focus on making the CI clean (100% pass), leave the other nice things for follow up PRs (and I'm happy to send a follow up PR if that helps).

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
omamoo: @reyang 
Thanks for sharing. so  i thinks that `.` is the better choice too

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetricsOptions.cs
twenzel: @cijothomas ok, which one should be enabled by default? "Performance"?

cijothomas: pick anything random for now or make all Enabled by default. We can merge and get a nuget out, and start collecting some feedback, and adjust accordingly.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
twenzel: @noahfalk Does the [semantics ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---processruntimeenvironment) didn't say "process.runtime.xyz.memory.usage" ?

noahfalk: Sorry I'm not sure I understood the question? 
I believe the semantics are distinguishing memory allocated by a runtime's GC (process.runtime.dotnet.something) vs. all memory allocated by the process for any purpose (process.memory.something). The original EventCounters for the runtime were created prior to these semantic conventions existing and they weren't attempting to separate process-wide information from runtime-specific information.

cijothomas:
I will proceed and merge the PR to make progress.

1. Create the framing (i.e project, workflow, extension method to configure etc). -- this PR. (We can revisit the actual instrument name, logic in follow ups)
2. Support one metric, and add usage example
3. Add more metrics which are straightforward to implement.
4. Release a alpha nuget with couple of metrics.
5. Add more metrics... 
6. Based on feedback, we can adjust subsequent releases.

cijothomas (approved):
Looks good for an initial merge. Lets follow up with more fixes.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
reyang: Most of the OpenTelemetry semantic conventions are still at very early stages.
Take [process.memory.usage and process.memory.virtual](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process) for example, the names are bit confusing, and memory measurements are much broader if we consider things like private working set, shared working set, commit size, reserved size, managed vs. native/unmanaged memory.

If I understand @noahfalk correctly, I think the intention is to put .NET runtime specific things under "process.runtime.dotnet.*", and put operating system specific things under a different place that doesn't have "runtime.dotnet" in the name. In other words, we need to ask ourselves - would this metric also apply to C++, if yes, it should not have "dotnet" in the name.

PR #214, opened 2022-03-04
Add cpu time

twenzel:
Add metric for `process.cpu.time`.

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -32,7 +32,6 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
         internal static readonly string InstrumentationVersion = AssemblyName.Version.ToString();
         private static string metricPrefix = "process.runtime.dotnet.";
         private readonly Meter meter;
-        private double previousProcessorTime;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RuntimeMetrics"/> class.
@@ -83,9 +82,9 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
                 this.meter.CreateObservableGauge($"{metricPrefix}memory.usage", () => (double)(Environment.WorkingSet / 1_000_000), "MB", "Working Set");
             }
 
-            if (options.IsCpuEnabled)
+            if (options.IsProcessEnabled)
             {
-                this.meter.CreateObservableGauge("process.cpu.usage", this.CalculateCpuUsage, "CPU usage");
+                this.meter.CreateObservableGauge("process.cpu.time", () => Process.GetCurrentProcess().TotalProcessorTime.TotalSeconds, "s", "Total processor time of this process");
             }
 
             if (options.IsAssembliesEnabled)
@@ -107,21 +106,5 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
             return gcInfo.HeapSizeBytes != 0 ? gcInfo.FragmentedBytes * 100d / gcInfo.HeapSizeBytes : 0;
         }
 #endif
-
-        private double CalculateCpuUsage()
-        {
-            var cpuUsage = 0.0;
-
-            var endTime = Process.GetCurrentProcess().TotalProcessorTime.TotalMilliseconds;
-
-            if (this.previousProcessorTime != 0)
-            {
-                cpuUsage = (endTime - this.previousProcessorTime) / (100 * Environment.ProcessorCount);
-            }
-
-            this.previousProcessorTime = endTime;
-
-            return cpuUsage;
-        }
     }
 }
--- a/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -15,6 +15,7 @@
 // </copyright>
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Reflection;
 using System.Threading;
@@ -31,6 +32,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
         internal static readonly string InstrumentationVersion = AssemblyName.Version.ToString();
         private static string metricPrefix = "process.runtime.dotnet.";
         private readonly Meter meter;
+        private double previousProcessorTime;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RuntimeMetrics"/> class.
@@ -81,6 +83,11 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
                 this.meter.CreateObservableGauge($"{metricPrefix}memory.usage", () => (double)(Environment.WorkingSet / 1_000_000), "MB", "Working Set");
             }
 
+            if (options.IsCpuEnabled)
+            {
+                this.meter.CreateObservableGauge($"{metricPrefix}cpu.usage", this.CalculateCpuUsage, "CPU usage");
+            }
+
             if (options.IsAssembliesEnabled)
             {
                 this.meter.CreateObservableCounter($"{metricPrefix}assembly.count", () => AppDomain.CurrentDomain.GetAssemblies().Length, description: "Number of Assemblies Loaded");
@@ -100,5 +107,21 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
             return gcInfo.HeapSizeBytes != 0 ? gcInfo.FragmentedBytes * 100d / gcInfo.HeapSizeBytes : 0;
         }
 #endif
+
+        private double CalculateCpuUsage()
+        {
+            var cpuUsage = 0.0;
+
+            var endTime = Process.GetCurrentProcess().TotalProcessorTime.TotalMilliseconds;
+
+            if (this.previousProcessorTime != 0)
+            {
+                cpuUsage = (endTime - this.previousProcessorTime) / (100 * Environment.ProcessorCount);
+            }
+
+            this.previousProcessorTime = endTime;
+
+            return cpuUsage;
+        }
     }
 }
--- a/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -85,7 +85,7 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
 
             if (options.IsCpuEnabled)
             {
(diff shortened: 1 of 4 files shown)


test/OpenTelemetry.Contrib.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -54,5 +54,27 @@ public void RuntimeMetricsAreCaptured()
...
+
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                 .AddRuntimeMetrics(options =>
+                 {
+                     options.CpuEnabled = true;
+                 })
+                 .AddReader(metricReader)
+                .Build();
+
+            await Task.Delay(1000);
```
cijothomas, line 72: this is not required, right?

cijothomas:
```
var afterTenMsec = DateTime.UtcNow.AddMilliseconds(10);
while (DateTime.UtcNow < afterTenMsec) { }
```

Can we do something like above, and forceflush, and validate that the value is >=10 ?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -100,5 +107,21 @@ private static double GetFragmentation()
             return gcInfo.HeapSizeBytes != 0 ? gcInfo.FragmentedBytes * 100d / gcInfo.HeapSizeBytes : 0;
         }
 #endif
+
+        private double CalculateCpuUsage()
+        {
+            var cpuUsage = 0.0;
+
+            var endTime = Process.GetCurrentProcess().TotalProcessorTime.TotalMilliseconds;
```
cijothomas, line 115: won't ObservableCounter work better? We can simply pass `Process.GetCurrentProcess().TotalProcessorTime.TotalMilliseconds` value to it. (and let SDK/downstream handle calculation?

cijothomas: @reyang @noahfalk to share thoughts here.

test/OpenTelemetry.Contrib.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
twenzel: Something like this
``` csharp
var afterTenMsec = DateTime.UtcNow.AddMilliseconds(10);
            while (DateTime.UtcNow < afterTenMsec)
            {
                meterProvider.ForceFlush(MaxTimeToAllowForFlush);
            }

            Assert.True(exportedItems.Count > 10);
```

cijothomas: Not really..`while (DateTime.UtcNow < afterTenMsec) { }` is intentionally spinning CPU.
Flush after that.
There'll be single exported item.
But its value should be atleast 10.

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
reyang: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/207#discussion_r816413966

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -100,5 +107,21 @@ private static double GetFragmentation()
...
+
+            if (this.previousProcessorTime != 0)
+            {
+                cpuUsage = (endTime - this.previousProcessorTime) / (100 * Environment.ProcessorCount);
```
cijothomas, line 119: what is the relevance of 100 here?

test/OpenTelemetry.Contrib.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -54,5 +53,55 @@ public void RuntimeMetricsAreCaptured()
...
+
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                 .AddRuntimeMetrics(options =>
+                 {
+                     options.CpuEnabled = true;
+                 })
+                 .AddInMemoryExporter(exportedItems)
+                .Build();
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
```
cijothomas, line 70: right after this, we can validate that the value is 0. (1st time we return 0 as cpu usage, right)

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -81,6 +83,11 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
                 this.meter.CreateObservableGauge($"{metricPrefix}memory.usage", () => (double)(Environment.WorkingSet / 1_000_000), "MB", "Working Set");
             }
 
+            if (options.IsCpuEnabled)
+            {
+                this.meter.CreateObservableGauge($"{metricPrefix}cpu.usage", this.CalculateCpuUsage, "CPU usage");
```
noahfalk, line 88: Perhaps this should be a [process metric](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process) rather than a runtime metric? This measurement has no information that is specific to the .NET runtime, it is solely extracting information that the OS tracks for every process. 
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/207#discussion_r818272869

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
twenzel: Percent?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
twenzel: As mentioned in the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process) we should (also) provide a "process.cpu.time" with the implementation of `Process.GetCurrentProcess().TotalProcessorTime.TotalMilliseconds`

noahfalk: I'd suggest we have process.cpu.time be a replacement for this rather than in addition to this. However if we do decide have a runtime metric ObservableCounter still seems like the prefered way to do it. 

I could be messing up my math, but the current code doesn't appear to calculate the intended result. For example assume we have a 4 core machine, measurements are taken once a minute and the process is using all available CPU. At T=1 min TotalMilliseconds returns 4*60*1000=240_000. At T=2 min TotalMilliseconds returns 2*4*60*1000=480_000. The metric computes (480K-240K)/(100*4) = 600%

twenzel: I've changed to `process.cpu.time`.

twenzel:
@cijothomas @noahfalk Any further changes required?

src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -81,6 +82,11 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
                 this.meter.CreateObservableGauge($"{metricPrefix}memory.usage", () => (double)(Environment.WorkingSet / 1_000_000), "MB", "Working Set");
             }
 
+            if (options.IsProcessEnabled)
+            {
+                this.meter.CreateObservableCounter("process.cpu.time", () => Process.GetCurrentProcess().TotalProcessorTime.TotalSeconds, "s", "Total processor time of this process");
```
cijothomas, line 87: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process

mention this link in the comment for future readers reference.

cijothomas: wondering if we can do UserProcessorTime and PrivilegedProcessorTime with state tag?

test/OpenTelemetry.Contrib.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -54,5 +53,53 @@ public void RuntimeMetricsAreCaptured()
...
+
+            // simple CPU spinning
+            var spinDuration = DateTime.UtcNow.AddMilliseconds(10);
+            while (DateTime.UtcNow < spinDuration)
+            {
+            }
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            Assert.True(exportedItems.Count > 0);
```
cijothomas, line 78: this would be exactly 1?

noahfalk:
@cijothomas - do you want process metrics to be included in an assembly that is named as being Runtime instrumentation? This is more an question for how OpenTelemetry expects instrumentation packages to map to particular pieces of instrumentation. Ideally I want .NET to follow the same conventions that other language ecosystems are following.

noahfalk:
> Any further changes required?

Other than the question above of what assembly this metric belongs in I think it looks good. Thanks @twenzel!

test/OpenTelemetry.Contrib.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
cijothomas, line 78: also validate the name of the metric

reyang:
> > @cijothomas - do you want process metrics to be included in an assembly that is named as being Runtime instrumentation? This is more an question for how OpenTelemetry expects instrumentation packages to map to particular pieces of instrumentation. Ideally I want .NET to follow the same conventions that other language ecosystems are following.
> 
> Agree its not really a good fit in the runtime instrumentation. I am okay to have this here now for the initial versions. We'll need to add each metric one by one (to iron out collection issues), and once all are ready, we can split to more specific packages.

Maybe something like "os" (operating system). I think ultimately we have to decide whether we wait for the OpenTelemetry Semantic Convention to be stable, or ship something that follows .NET style or some other established styles, and such decision is unlikely to be made right away. So this PR looks good to me for now 👍

reyang (approved):

cijothomas (approved):
LGTM.
There are some minor comments we can follow up.

PR #218, opened 2022-03-09
Add process.cpu.count

twenzel:
Adds the "process.cpu.count" metric to the runtime metrics.


src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -85,6 +85,7 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
             if (options.IsProcessEnabled)
             {
                 this.meter.CreateObservableCounter("process.cpu.time", () => Process.GetCurrentProcess().TotalProcessorTime.TotalSeconds, "s", "Total processor time of this process");
+                this.meter.CreateObservableCounter("process.cpu.count", () => Environment.ProcessorCount, description: "The number of available logical CPUs");
```
cijothomas, line 88: can you add links to semantic conventions, if they exist.?
Also, this doesn't look like a Counter

twenzel: Specification is not yet approved https://github.com/open-telemetry/opentelemetry-specification/pull/2392

cijothomas: You can put a link to this PR as a code comment, which will help future readers. (Before a stable release, we can cleanup any references, and point to official spec)

PR #219, opened 2022-03-09
Move memory metric to process metrics

twenzel:
According to  https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process moving the "memory.usage" to the process metrics.

Change metric to expose value in bytes instead of MB.
Add "process.memory.virtual" also

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -78,14 +78,11 @@ namespace OpenTelemetry.Contrib.Instrumentation.Runtime
             }
 #endif
 
-            if (options.IsMemoryEnabled)
-            {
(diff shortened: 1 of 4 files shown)


src/OpenTelemetry.Contrib.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -78,14 +78,11 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
...
 
-            if (options.IsMemoryEnabled)
-            {
-                this.meter.CreateObservableGauge($"{metricPrefix}memory.usage", () => (double)(Environment.WorkingSet / 1_000_000), "MB", "Working Set");
-            }
-
             if (options.IsProcessEnabled)
             {
                 this.meter.CreateObservableCounter("process.cpu.time", this.GetProcessorTimes, "s", "Processor time of this process");
+                this.meter.CreateObservableGauge("process.memory.usage", () => Process.GetCurrentProcess().WorkingSet64, "By", "The amount of physical memory in use");
```
cijothomas, line 84: no unit tests added to cover this, but okay to address later.

cijothomas (approved):

PR #228, opened 2022-03-11
WCF server example use TelemetryServiceBehavior instead of  TelemetryEndpointBehavior

vonckm-kadaster:
The WCF Server Example is using the `client endpoint` extensions instead of the `server service` extension. This pull request sets the example to telemetryServiceBehavior


vonckm-kadaster:
i don't want to sign EasyCLA, because i'm on my organization user account. Maybey somebody else can change the example

PR #295, opened 2022-04-14
GenevaExporter - LogSerialization Fixes

cijothomas:
1. Fixes an issue which dropped the LogRecord.State, when it is a KVP of size1

2. Refactored  LogExporter code a bit:
LogExporter obtains the IReadOnlyList<KeyValuePair<string, object>> by either casting the State, or from the StateValues.
Serializes each KVP to form Part C fields.
Body is only populated if IncludeFormatMessage is true *OR*  special casing {OriginalFormat} from Log extension methods.

As a result:
The below log statement, will result in empty body, as there is no formatter provided by the user, and Exporter will not attempt "ToString() on the state" to populate body.
`logger.Log(LogLevel.Information, default, state: "somestringasdata", exception: null, formatter: null);`

The below log statement will also result in empty body, if `IncludeFormattedMessage = false`, else uses the formatted message from user. Again, Exporter does not attempt any "ToString()" of the state.

`logger.Log(LogLevel.Information, default, state: "somestringasdata", exception: null, formatter: (state, ex) => "Example formatted message.");`

In short, Exporter no longer attempts to do "ToString() on state to populate body.".

3. Tests are refactored. (could be further improved, coming in next PR), categorized into "Raw Log method", vs "using extension methods on Log (with message templating)."

~4. Expecting slight perf boost (less if...else ladder), will run benchmarks and update numbers, before merging.~
// * Summary *

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19044.1586 (21H2)
Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.201
  [Host]     : .NET 6.0.3 (6.0.322.12309), X64 RyuJIT
  DefaultJob : .NET 6.0.3 (6.0.322.12309), X64 RyuJIT

Main branch

|                   Method |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------------- |-----------:|---------:|---------:|-------:|----------:|
| LoggerWithGenevaExporter | 1,100.1 ns | 14.49 ns | 12.85 ns | 0.0610 |     256 B |
|       SerializeLogRecord |   797.9 ns |  7.26 ns |  6.43 ns | 0.0057 |      24 B |

with this PR

|                   Method |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------------------------- |-----------:|---------:|---------:|-------:|----------:|
| LoggerWithGenevaExporter | 1,085.7 ns | 14.71 ns | 13.04 ns | 0.0610 |     256 B |
|       SerializeLogRecord |   794.0 ns |  7.82 ns |  6.93 ns | 0.0057 |      24 B |

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
@@ -185,6 +185,7 @@ namespace OpenTelemetry.Exporter.Geneva
             if (logRecord.State == null)
             {
                 // When State is null, OTel SDK guarantees StateValues is populated
+                // TODO: Debug.Assert?
                 listKvp = logRecord.StateValues;
             }
             else
@@ -327,8 +328,8 @@ namespace OpenTelemetry.Exporter.Geneva
             {
                 var entry = listKvp[i];
 
-                // Iteration #1 - Get those fields which become dedicated column
-                // i.e all PartB fields and opt-in part c fields.
+                // Iteration #1 - Get those fields which become dedicated columns
+                // i.e all Part B fields and opt-in Part C fields.
                 if (entry.Key == "{OriginalFormat}")
                 {
                     cursor = MessagePackSerializer.SerializeAsciiString(buffer, cursor, "body");
@@ -342,7 +343,7 @@ namespace OpenTelemetry.Exporter.Geneva
                     // TODO: the above null check can be optimized and avoided inside foreach.
                     if (entry.Value != null)
                     {
-                        // Geneva doesn't support null.
+                        // null is not supported.
                         cursor = MessagePackSerializer.SerializeUnicodeString(buffer, cursor, entry.Key);
                         cursor = MessagePackSerializer.Serialize(buffer, cursor, entry.Value);
                         cntFields += 1;
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
@@ -181,20 +181,16 @@ namespace OpenTelemetry.Exporter.Geneva
 
         internal int SerializeLogRecord(LogRecord logRecord)
         {
-            bool isUnstructuredLog = true;
             IReadOnlyList<KeyValuePair<string, object>> listKvp;
             if (logRecord.State == null)
             {
+                // When State is null, OTel SDK guarantees StateValues is populated
                 listKvp = logRecord.StateValues;
             }
             else
             {
+                // Attempt to see if State could be ROL_KVP.
                 listKvp = logRecord.State as IReadOnlyList<KeyValuePair<string, object>>;
-            }
-
-            if (listKvp != null)
-            {
-                isUnstructuredLog = listKvp.Count == 1;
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -181,22 +181,18 @@ internal bool IsUsingUnixDomainSocket
 
         internal int SerializeLogRecord(LogRecord logRecord)
         {
-            bool isUnstructuredLog = true;
             IReadOnlyList<KeyValuePair<string, object>> listKvp;
             if (logRecord.State == null)
             {
+                // When State is null, OTel SDK guarantees StateValues is populated
```
reyang, line 187: Maybe consider a Debug.Assert?

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -325,81 +321,72 @@ internal int SerializeLogRecord(LogRecord logRecord)
...
 
-            if (isUnstructuredLog)
+            bool hasEnvProperties = false;
+            bool bodyPopulated = false;
+            for (int i = 0; i < listKvp?.Count; i++)
+            {
+                var entry = listKvp[i];
+
+                // Iteration #1 - Get those fields which become dedicated column
+                // i.e all PartB fields and opt-in part c fields.
```
reyang, line 331:
```suggestion
                // i.e. all Part B fields and opt-in Part c fields.
```

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -325,81 +321,72 @@ internal int SerializeLogRecord(LogRecord logRecord)
...
             cntFields += 1;
 
-            if (isUnstructuredLog)
+            bool hasEnvProperties = false;
+            bool bodyPopulated = false;
+            for (int i = 0; i < listKvp?.Count; i++)
+            {
+                var entry = listKvp[i];
+
+                // Iteration #1 - Get those fields which become dedicated column
```
reyang, line 330:
```suggestion
                // Iteration #1 - Get those fields which become dedicated columns
```

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -325,81 +321,72 @@ internal int SerializeLogRecord(LogRecord logRecord)
...
+                    cntFields += 1;
+                    bodyPopulated = true;
+                    continue;
+                }
+                else if (this.m_customFields == null || this.m_customFields.ContainsKey(entry.Key))
+                {
+                    // TODO: the above null check can be optimized and avoided inside foreach.
+                    if (entry.Value != null)
+                    {
+                        // Geneva doesn't support null.
```
reyang, line 345: // null is not supported

reyang (approved):
LGTM.

utpilla:
I think the CHANGELOG needs to be updated.

utpilla (approved):

PR #297, opened 2022-04-14
Auto flush activity processor implementation and tests (#296).

rypdal:
Implemented auto flush activity processor as described in the issue: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/296


src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AutoFlushActivityProcessor"/> class.
+        /// </summary>
+        /// <param name="predicate">Predicate that should return <c>true</c> to initiate a flush.</param>
+        /// <param name="timeoutMilliseconds">Timeout (in milliseconds) to use for flushing.</param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown when the <c>timeoutMilliseconds</c> is smaller than -1.
+        /// </exception>
+        internal AutoFlushActivityProcessor(Predicate<Activity> predicate, int timeoutMilliseconds)
```
cijothomas, line 40: if we keep this as public, then users can create their own logic, as opposed to providing `FlushAfterLocalServerSideRootSpans` etc from this package itself? Is there a reason why you prefer to hide this and instead offer more specific statics to create the Processor itself?

rypdal: I thought about a set of static methods with predefined setup of predicates for convenient/frequent usages. In our project we need FlushAfterLocalServerSideRootSpans for example. But may be other users may have their own logic. I'll remove those static methods and open the constructor.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Diagnostics;
+using System.Threading;
+
+namespace OpenTelemetry.Trace
+{
+    /// <summary>
+    /// Activity processor that flushes its containing <see cref="TracerProvider"/> if an ended
+    /// activity matches a predicate.
+    /// </summary>
+    public sealed class AutoFlushActivityProcessor : BaseProcessor<Activity>
```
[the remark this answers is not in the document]
rypdal: Sure, I'll provide an example.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            this.timeoutMilliseconds = timeoutMilliseconds;
+        }
+
+        /// <inheritdoc/>
+        public override void OnEnd(Activity data)
+        {
+            if (this.predicate(data))
+            {
+                (this.ParentProvider as TracerProvider)?.ForceFlush(this.timeoutMilliseconds);
```
reyang, line 56: Could this be improved so the check and type conversion are performed once instead of in every `OnEnd`?

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,60 @@
...
+                throw new ArgumentOutOfRangeException(nameof(timeoutMilliseconds));
+            }
+
+            this.timeoutMilliseconds = timeoutMilliseconds;
+        }
+
+        /// <inheritdoc/>
+        public override void OnEnd(Activity data)
+        {
+            if (this.predicate(data))
```
reyang, line 54: What if the predicate throws?

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace OpenTelemetry.Trace
+{
+    /// <summary>
+    /// Activity processor that flushes its containing <see cref="TracerProvider"/> if an ended
+    /// activity matches a predicate.
```
reyang, line 25: What if the predicate triggered new Activity creation?

reyang (changes requested):
Need to understand the overall design:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/297/files#r850791969

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/297/files#r850790567

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: The `this.ParentProvider` is set in the internal method of `BaseProcessor` class: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/BaseProcessor.cs#L143. So, in this code we don't know if `this.ParentProvider` is already set or not. What we can optimise here is to check if `this.ParentProvider` is already defined (not null) and store result of check/conversion for subsequent use. So, we do check/conversion once and assume that the `SetParentProvider(...)` will not be called again although theoretically it's still possible.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: It's assumed that predicate is defined as a logical expression and doesn't contain more complex code. Predicate triggering new Activity creation seems more exotic scenario.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
Oberon00: Would be nice if there was an OnAttachToProvider/OnSetParentProvider hook.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: IMHO predicate's exception should be bubbled to the OnEnd caller as this class is not defining the predicate code. Otherwise it can be swallowed and logged with debug level.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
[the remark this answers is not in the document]
rypdal: Added an example to the Readme.md

src/OpenTelemetry.Extensions/README.md
```diff
@@ -13,3 +13,25 @@ future.
...
+```cs
+public static TracerProviderBuilder AddMyExporter(this TracerProviderBuilder builder, MyExporterOptions options)
+{
+    var exporter = new MyExporter(options);
+
+    var autoFlushProcessor = new AutoFlushActivityProcessor(
+        a => a.Parent == null && (a.Kind == ActivityKind.Server || a.Kind == ActivityKind.Consumer), 5000);
+
+    return builder
+        .AddProcessor(new SimpleActivityExportProcessor(exporter))
```
alanwest, line 34: Would using the BatchActivityExporterProcessor in the example be more interesting? Using the Simple one, more than likely the flush afterwards would not have anything to flush.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
reyang: What should the user expect (e.g. if a user provides a buggy predicate that throws, should he expect the processor to fail and the data to be dropped, or the predicate itself to be ignore - so no Flush is called but the data still flows, or some other behavior?)

src/OpenTelemetry.Extensions/README.md
```diff
@@ -13,3 +13,25 @@ future.
...
+        .AddProcessor(autoFlushProcessor);
```
reyang, line 35: Might be more user friendly with:

```suggestion
        .AddAutoFlushProcessorProcessor(a => a.Parent == null && (a.Kind == ActivityKind.Server || a.Kind == ActivityKind.Consumer), 5000);
```

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
reyang: Consider something like https://github.com/open-telemetry/opentelemetry-dotnet/blob/cab5b26a24f666cca6089e380adb83805a02eee4/src/OpenTelemetry/Metrics/MetricReader.cs#L116-L124

src/OpenTelemetry.Extensions/README.md
reyang: +1

reyang (approved):
LGTM once https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/297/files#r851260803 is addressed (e.g. calling out the responsibility and expectation in the xmldoc)

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: Do you mean only comment from the reference or also the code preventing parallel calls to the `OnCollect(...)` method ?

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: I agree. Data should still flow. Wrapped the predicate with `try...catch` block and logged exception.

src/OpenTelemetry.Extensions/README.md
rypdal: Updated example.

src/OpenTelemetry.Extensions/README.md
rypdal: Although there is no such extension method, but updated example with this method as it looks more self explanatory and compact.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
Oberon00: What does it mean for the data (not) to "flow"? Are expectations if a Span Processor's OnEnd throws an exception documented? I think we should not make any additional guarantees / set any additional expectations here.

reyang: The way I imagine: "if the predicate throws exception, the `AutoFlushActivityProcessor` will swallow the exception and treat the predicate as False (which means `ForceFlush` will not be applied)".

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,108 @@
...
+        private readonly int timeoutMilliseconds;
+
+        private TracerProvider tracerProvider;
+        private bool canForceFlush = true;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AutoFlushActivityProcessor"/> class.
+        /// </summary>
+        /// <param name="predicate">Predicate that should return <c>true</c> to initiate a flush.</param>
+        /// <param name="timeoutMilliseconds">Timeout (in milliseconds) to use for flushing.</param>
```
CodeBlanch, line 42: Maybe mention `Timeout.Infinite` in the xml like:

`Timeout (in milliseconds) to use for flushing. Specify <see cref="Timeout.Infinite"> to wait indefinitely."

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,108 @@
...
+                this.canForceFlush = this.tracerProvider != null;
+                if (!this.canForceFlush)
+                {
+                    return;
+                }
+            }
+            else if (this.ParentProvider == null)
+            {
+                return;
+            }
```
CodeBlanch, line 82: Could we simplify this a bit?

```csharp
            if (!this.canForceFlush)
            {
                return;
            }

            if (this.tracerProvider == null)
            {
                this.tracerProvider = this.ParentProvider as TracerProvider;
                this.canForceFlush = this.tracerProvider != null;
                if (!this.canForceFlush)
                {
                    return;
                }
            }
```

`ParentProvider` will be set before `OnEnd` is called. Mostly certain about that 😄

rypdal: If we are sure that "`ParentProvider `will be set before `OnEnd` is called" and this behaviour will not be changed then we can simplify the condition. I can mention in the comment that it's assumed `ParentProvider` isn't a null when the `OnEnd` is called. Otherwise for safety we may keep the code as is, because anyway next method execution there will be either `this.canForceFlush = fasle` or `this.tracerProvider == null`, i.e. the condition `this.ParentProvider != null` will not be even checked.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="timeoutMilliseconds">Timeout (in milliseconds) to use for flushing.</param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown when the <c>timeoutMilliseconds</c> is smaller than -1.
+        /// </exception>
+        /// <remarks>
+        /// It's assumed that the predicate is defined as a lambda expression which is executed quite fast and
+        /// doesn't contain more complex code. The predicate must not create new Activity instances,
+        /// otherwise the behavior is undefined. Any exception thrown by the predicate will be swallowed and logged.
+        /// </remarks>
+        public AutoFlushActivityProcessor(Predicate<Activity> predicate, int timeoutMilliseconds)
```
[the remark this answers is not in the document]
rypdal: `10000` milliseconds sounds as reasonable default value. Added default.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: Thanks. Added it to the doc.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: I stick with "swallow and log" approach and mentioned this behaviour in the doc.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+                {
+                    return;
+                }
+            }
+            else if (this.ParentProvider == null)
+            {
+                return;
+            }
+
+            var shouldFlash = this.RunPredicate(data);
```
utpilla, line 87: I'm assuming you meant `shouldFlush`? 😂

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
utpilla: Does this need to be checked on every `OnEnd` call? Could this done in the ctor? This check doesn't use the input `Activity` at all.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            if (!this.canForceFlush)
+            {
+                return;
+            }
+
+            if (this.tracerProvider == null && this.ParentProvider != null)
+            {
+                this.tracerProvider = this.ParentProvider as TracerProvider;
+                this.canForceFlush = this.tracerProvider != null;
```
utpilla, line 76: How about using `this.canForceFlush = this.ParentProvider is TracerProvider;` instead?

src/OpenTelemetry.Extensions/README.md
```diff
@@ -13,3 +13,25 @@ future.
...
+public static TracerProviderBuilder AddMyExporter(this TracerProviderBuilder builder, MyExporterOptions options)
+{
+    return builder
+        .AddProcessor(new BatchActivityExportProcessor(
+                    new MyExporter(options),
+                    options.BatchExportProcessorOptions.MaxQueueSize,
+                    options.BatchExportProcessorOptions.ScheduledDelayMilliseconds,
+                    options.BatchExportProcessorOptions.ExporterTimeoutMilliseconds,
+                    options.BatchExportProcessorOptions.MaxExportBatchSize))
+        .AddAutoFlushProcessorProcessor(a => a.Parent == null && (a.Kind == ActivityKind.Server || a.Kind == ActivityKind.Consumer), 5000);
```
utpilla, line 35: This extension method has to be added to the code.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown when the <c>timeoutMilliseconds</c> is smaller than -1.
+        /// </exception>
+        /// <remarks>
+        /// It's assumed that the predicate is defined as a lambda expression which is executed quite fast and
+        /// doesn't contain more complex code. The predicate must not create new Activity instances,
+        /// otherwise the behavior is undefined. Any exception thrown by the predicate will be swallowed and logged.
+        /// In case of an exception the predicated is treated as false which means flush will not be applied.
+        /// </remarks>
+        public AutoFlushActivityProcessor(Predicate<Activity> predicate, int timeoutMilliseconds = 10000)
```
utpilla, line 54: I think we should be using `Func<Activity, bool>` instead of `Predicate<Activity>` for the following reasons:

1. It's more descriptive in its syntax
2. .NET framework libraries also tend to use `Func<T, bool>` instead of `Predicate<T>`. For e.g.: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitylistener.shouldlistento?view=net-6.0

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: In ctor it's not guaranteed that the `this.ParentProvider` is already set (see this [comment,](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/297#discussion_r851254737) please).

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: Sure :-) Typo is fixed.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// to wait indefinitely.
+        /// </param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// Thrown when the <c>timeoutMilliseconds</c> is smaller than -1.
+        /// </exception>
+        /// <remarks>
+        /// It's assumed that the predicate is defined as a lambda expression which is executed quite fast and
+        /// doesn't contain more complex code. The predicate must not create new Activity instances,
+        /// otherwise the behavior is undefined. Any exception thrown by the predicate will be swallowed and logged.
+        /// In case of an exception the predicated is treated as false which means flush will not be applied.
```
cijothomas, line 52:
```suggestion
        /// In case of an exception the predicate is treated as false which means flush will not be applied.
```

src/OpenTelemetry.Extensions/README.md
```diff
@@ -13,3 +13,25 @@ future.
 
 Adds a log processor which will convert log messages into events and attach them
 to the currently running Activity.
+
+## Traces
+
+### AutoFlushActivityProcessor
+
+Processor that flushes its containing TracerProvider if an ended activity matches a predicate. 
+Note that this processor must be added *after* exporter related span processors.
```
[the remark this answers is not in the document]
rypdal: Applied.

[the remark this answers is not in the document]
rypdal: Applied.

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
rypdal: Applied.

src/OpenTelemetry.Extensions/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace OpenTelemetry.Trace
+{
+    /// <summary>
+    /// Extension methods to simplify registering of auto flush Activity processor.
+    /// </summary>
+    public static class TracerProviderBuilderExtensions
```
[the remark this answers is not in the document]
rypdal: For now it's sufficient to keep only the `AddAutoFlushActivityProcessor(...)` extension method as a part of public API. The `AutoFlushActivityProcessor` is made internal.

cijothomas (approved):
Please add entry in changelog.md as well
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Extensions/CHANGELOG.md

cijothomas:
@rypdal Thanks for your very 1st contribution to the repo!

rypdal:
Thanks for review and discussions: @cijothomas, @Oberon00, @alanwest, @reyang, @utpilla, @CodeBlanch !
Thanks to @Oberon00 for the idea of auto flush activity processor !

PR #313, opened 2022-04-18
Remove unnecessary using statements + add rule

### LINKED ISSUE #275: Enforce unnecessary `using` statements to fail CI
There are unnecessary using statements cluttering files. 

Related issue https://github.com/open-telemetry/opentelemetry-dotnet/pull/2411
Note: The PR fixing the above only did it for the source code and not the tests I believe.

mic-max:
Fixes #275


reyang (approved):

SergeyKanzhelev (approved):

src/OpenTelemetry.Instrumentation.Wcf/TelemetryServiceBehavior.cs
```diff
@@ -14,9 +14,11 @@
 // limitations under the License.
 // </copyright>
 
+#pragma warning disable IDE0005 // Using directive is unnecessary.
```
utpilla, line 17: Why are we disabling the rule for this specific instance?

mic-max: I tried all combinations of removing these "unnecessary" using statements, but in each scenario VS would then complain and tell me they were needed. I wasn't sure what to do so disabled via source for this one spot.

PR #322, opened 2022-04-20
Geneva Exporter - Throw on TableNameMappings `null` value

mic-max:
Throw exception when null value is used in `TableNameMappings` for Geneva Exporter Options class.
Includes unit test.


reyang (approved):

cijothomas:
@mic-max please address conflicts.

cijothomas (approved):

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -64,6 +65,8 @@ public GenevaLogExporter(GenevaExporterOptions options)
             var tempTableMappings = new Dictionary<string, string>(options.TableNameMappings.Count, StringComparer.Ordinal);
             foreach (var kv in options.TableNameMappings)
             {
+                Guard.ThrowIfNull(kv.Value);
```
TimothyMothra, line 68: I would recommend including the Key name in the exception message here.

utpilla: Would it be better to do this validation in the `GenevaExporterOptions` class, like how it's done for `PrepopulatedFields`: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L36-L73

mic-max:
What should be done when a null key exists in the `TableNameMappings`, if anything?

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -27,11 +27,36 @@ public class GenevaExporterOptions
...
+    public IReadOnlyDictionary<string, string> TableNameMappings
+    {
+        get => this._tableNameMappings;
+        set
+        {
+            Guard.ThrowIfNull(value);
+
+            foreach (var entry in value)
+            {
+                if (entry.Value is null)
```
cijothomas, line 45: we can check if key is null/empty as well here.

mic-max: Actually, C# already doesn't allow `null` to be a key in a Dictionary :)

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -27,11 +27,36 @@ public class GenevaExporterOptions
...
+                {
+                    throw new ArgumentNullException(entry.Key, $"{nameof(this.TableNameMappings)} must not contain null values.");
+                }
+            }
+
+            var copy = new Dictionary<string, string>(value.Count);
+            foreach (var entry in value)
```
cijothomas, line 52: we can validate and copy with a single foreach. There is no need of dual iteration here.

cijothomas (approved):

PR #352, opened 2022-05-11
GenevaTraceExporter bug fix to not export activities not sampled in

cijothomas:
Fixed by creating Reentrant Exporter specific to Activity.

(Will explore actually making ReEntrantExporter a public thing in the main SDK separately)

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/ReentrantActivityExportProcessor.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/ReentrantActivityExportProcessor.cs
@@ -1,0 +1,56 @@
+﻿// <copyright file="ReentrantActivityExportProcessor.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


src/OpenTelemetry.Exporter.Geneva/ReentrantActivityExportProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    protected override void OnExport(Activity data)
+    {
+        if (!data.Recorded)
+        {
+            return;
+        }
+
+        this.exporter.Export(CreateBatch(data));
```
reyang, line 52: Maybe reverse the flow? (no strong opinion though)

```suggestion
        if (data.Recorded)
        {
            this.exporter.Export(CreateBatch(data));
        }
```

reyang (approved):

src/OpenTelemetry.Exporter.Geneva/ReentrantActivityExportProcessor.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace OpenTelemetry.Exporter.Geneva;
+
+// This export processor exports without synchronization.
+// Once OpenTelemetry .NET officially support this,
+// we can get rid of this class.
+// This is currently only used in ETW export, where we know
+// that the underlying system is safe under concurrent calls.
+internal class ReentrantActivityExportProcessor : BaseExportProcessor<Activity>
+{
+    static ReentrantActivityExportProcessor()
```
utpilla, line 31: Can we just have this class inherit `ReentrantExportProcessor<Activity>` instead? That way we could reuse the `ctor` logic of the `ReentrantExportProcessor<T>` class and we would only need to override the `OnExport` method.

cijothomas: Done!

utpilla (approved):

mic-max (approved):

PR #354, opened 2022-05-11
Add eventsource - PersistentStorage.Abstractions

vishweshbankwar:


src/OpenTelemetry.Extensions.PersistentStorage.Abstractions/PersistentStorageAbstractionsEventSource.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace OpenTelemetry.Extensions.PersistentStorage
+{
+    [EventSource(Name = EventSourceName)]
+    internal sealed class PersistentStorageAbstractionsEventSource : EventSource
+    {
+        public static PersistentStorageAbstractionsEventSource Log = new PersistentStorageAbstractionsEventSource();
+        private const string EventSourceName = "OpenTelemetry-Extensions-PersistentStorage-Abstractions";
+
+        [NonEvent]
```
cijothomas, line 29: why do we need this nonevent which calls the actual event?

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OpenTelemetryProtocolExporterEventSource.cs follow this example. use noneevents for avoiding allocation cost for exception scenarios. for everything else, use the event direclty

vishweshbankwar: @cijothomas - Please review if it looks good now.

src/OpenTelemetry.Extensions.PersistentStorage.Abstractions/PersistentStorageAbstractionsEventSource.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void LogException(string message, Exception ex)
+        {
+            if (this.IsEnabled(EventLevel.Error, EventKeywords.All))
+            {
+                this.LogException(message, ToInvariantString(ex));
+            }
+        }
+
+        [Event(1, Message = "{0} : {1}", Level = EventLevel.Error)]
+        public void LogException(string message, string ex)
```
utpilla, line 40: I'm curious to know if having dedicated events for `PersistentBlob` and `PersistentBlobProvider` make it easier when troubleshooting?

vishweshbankwar: The event names will be different so, it will be more clear where the exception occurred. I will update it

PR #360, opened 2022-05-16
Add release workflow - PersistentStorage.Abstractions

vishweshbankwar:


.github/workflows/package-Extensions.PersistentStorage.Abstractions.yml
```diff
@@ -0,0 +1,49 @@
+name: Pack OpenTelemetry.Extensions.PersistentStorage.Abstractions
```
cijothomas, line 1: don't know if this was discussed earlier - did we consider naming it as .API instead of .Abstrations? 

Not a blocker.

cijothomas (approved):

.github/workflows/package-Extensions.PersistentStorage.Abstractions.yml
vishweshbankwar: Yes - abstractions was the recommendation.
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/221#discussion_r852360001

cijothomas: okay, that matches general .NET namings.
(FYI - the current otel.API was originally named abstrations, but later changed to API, following OTel spec!)

vishweshbankwar: I see -did not know that. should we keep it API then? @cijothomas @CodeBlanch

cijothomas: Given this is not a spec-ed thing, lets stick with Abstractions as Mikel suggested.

utpilla (approved):

utpilla:
@vishweshbankwar  Update the .sln file as well.

PR #363, opened 2022-05-17
[Opentelemetry.Extensions.PersistentStorage] Update implementation to use Abstractions

vishweshbankwar:
Fixes #.

## Changes


src/OpenTelemetry.Extensions.PersistentStorage/FileBlobProvider.cs
```diff
@@ -115,31 +116,37 @@ public IEnumerable<IPersistentBlob> GetBlobs()
...
+        protected override bool OnTryCreateBlob(byte[] buffer, int leasePeriodMilliseconds, [NotNullWhen(true)] out PersistentBlob blob)
         {
-            return this.GetBlobs().FirstOrDefault();
+            blob = this.CreateFileBlob(buffer, leasePeriodMilliseconds);
+            if (blob == null)
+            {
+                return false;
+            }
+
+            return true;
```
TimothyMothra, line 127: nit: consider simplifying this:

```suggestion
            return blob != null;
```

test/OpenTelemetry.Extensions.PersistentStorage.Tests/FileBlobProviderTests.cs
```diff
@@ -16,143 +16,146 @@
...
             // Get blob.
-            IPersistentBlob blob2 = storage.GetBlob();
+            Assert.True(blobProvider.TryGetBlob(out var singleBlob));
 
-            Assert.Single(storage.GetBlobs());
+            Assert.Single(blobProvider.GetBlobs());
 
             // Verify file name from both create blob and get blob are same.
-            Assert.Equal(((FileBlob)blob1).FullPath, ((FileBlob)blob2).FullPath);
+            Assert.Equal(((FileBlob)blob).FullPath, ((FileBlob)singleBlob).FullPath);
```
TimothyMothra, line 44: nit: imo "blob" and "singleBlob" isn't descriptive of what's being tested here.

I think the original "blob1" and "blob2" are fine here.
Could also use "createBlob" and "getBlob".

test/OpenTelemetry.Extensions.PersistentStorage.Tests/FileBlobProviderTests.cs
```diff
@@ -16,143 +16,146 @@
...
                 maintenancePeriodInMilliseconds,
                 retentionPeriodInMilliseconds,
                 writeTimeOutInMilliseconds);
 
             var data = Encoding.UTF8.GetBytes("Hello, World!");
-            var blob1 = (FileBlob)storage.CreateBlob(data);
+            Assert.True(blobProvider.TryCreateBlob(data, out var blob));
 
             // Wait for maintenance job to run
             Thread.Sleep(4000);
```
TimothyMothra, line 94: can the maintenance job be manually triggered to avoid the Sleep?

test/OpenTelemetry.Extensions.PersistentStorage.Tests/FileBlobProviderTests.cs
```diff
@@ -16,143 +16,146 @@
...
 
             // Blob will be deleted as retention period is 1 sec
-            Assert.False(File.Exists(blob1.FullPath));
+            Assert.False(File.Exists(((FileBlob)blob).FullPath));
 
             testDirectory.Delete(true);
```
TimothyMothra, line 99: Consider putting the `Delete()` in a `finally`. If any of the above asserts fail, I think this won't be executed.

vishweshbankwar: I think, we dont need to do that. I will do a follow up PR and remove it.

test/OpenTelemetry.Extensions.PersistentStorage.Tests/FileBlobProviderTests.cs
vishweshbankwar: No, it is a timer event. Will follow up to see if there is another way to avoid this. Added TODO item.

src/OpenTelemetry.Extensions.PersistentStorage/FileBlobProvider.cs
TimothyMothra: This class has two more `if` blocks that can be simplified in the same way. GitHub wouldn't let me give suggestions for the others.

TimothyMothra (approved):
LGTM

cijothomas (approved):
Consider changelog entry as this has a lot of user facing deletions

src/OpenTelemetry.Extensions.PersistentStorage/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 1.0.0-alpha.3
 
+* [Update implementation to use
+  Opentelemetry.Extensions.PersistentStorage.Abstractions](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/363)
```
utpilla, line 6: This should be under "Unreleased" section.

vishweshbankwar: Done

Issue #215, opened 2021-10-26
EventCounter -> Metric API instrumentation


cijothomas:

mic-max:
I am working on making this instrumentation more stable. I will be

- Add / enable testing
- Cleaning up implementation
- Looking into different possible bad scenarios, like enabling the extension twice on the same provider
- Filtering out metrics covered by other instrumentation (`Runtime`, `Process`)
  - Ensure it works in conjunction with other instrumentation so none of the events will be lost
- Checking for memory leaks and minimize memory footprint as it will be enabled for the full lifetime of most applications that use it
  - Don't allocate memory after startup

PR #365, opened 2022-05-18
Create a EventCounter Listener that subscribes to all event counters

hananiel:
Resolves Issue #215 

Created a focussed PR that subscribes to all Event Counter Events. 
The configuration to selectively listen to particular EventCounter Sources and/or Counters will be added as a separate PR. 

This will replace PR #177


src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,245 @@
...
+using System.Globalization;
+using System.Reflection;
+
+namespace OpenTelemetry.Instrumentation.EventCounters
+{
+    internal class EventCounterListener : EventListener
+    {
+        // EventSourceCreated can run before constructor, so constructor injection of the options dependency will not work.
+        // Using static to set options at host build time.
+        public static EventCounterMetricsOptions EventCounterMetricsOptions = new EventCounterMetricsOptions();
```
cijothomas, line 31: its a bit odd. But I think we can fix this later.

hananiel: Ill fix this, we can wait until initialization to read options.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,245 @@
...
+                    {
+                        instrumentType = InstrumentType.Counter;
+                    }
+                    else if (key.Equals("Increment", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Increment indicates we have to calculate rate.
+                        instrumentType = InstrumentType.Gauge;
+                        calculateRate = true;
+                    }
+                    else if (key.Equals("IntervalSec", StringComparison.OrdinalIgnoreCase))
```
cijothomas, line 133: I'd suggest to ignore this whole thing. lets just trust that EventCounter will fire based on our configuration, and we always assume it is correct.

cijothomas: it should simple this whole class a lot!

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,245 @@
...
+                        counterName = payload.Value.ToString();
+                    }
+                    else
+                    if (key.Equals("DisplayName", StringComparison.OrdinalIgnoreCase))
+                    {
+                        counterDisplayName = payload.Value.ToString();
+                    }
+                    else if (key.Equals("Mean", StringComparison.OrdinalIgnoreCase))
+                    {
+                        instrumentType = InstrumentType.Counter;
```
cijothomas, line 125: not reading the value anywhere?

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                    else if (key.Equals("Increment", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Increment indicates we have to calculate rate.
+                        instrumentType = InstrumentType.Gauge;
+                        calculateRate = true;
```
cijothomas, line 131: not reading the value?

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: instrument type is passed to RecordMetric in line 162.

cijothomas: but where do we read the actual value from the payload?

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: You mean just the section that reads and verifies actual interval or just the whole calculation of recordedValue based on rate?

cijothomas: the part which reads "actual" interval. If we simply assume its always what we configured, it'll simplify a lot.
(I dont recollect if there is a reason for us to *not trust* eventcounters to not respect the configured interval.)

(Will ask the eventcounter owners to chime in here, if needed)

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: We are inferring it. If countertype field is mean, then the mean field is available. If it is Sum, we look at the increment field. So it is effectively the same , but I can clarify the intent by reading the field. Here is some diagnostics code I am basing my thinking on: https://github.com/dotnet/diagnostics/blob/9ccbd1e94c1571c4af35bda26e580474aed3a37f/src/Microsoft.Diagnostics.Monitoring.EventPipe/Counters/TraceEventExtensions.cs#L41

cijothomas: I am not sure I follow. If we are not reading the value from the payload here, we wont get the value from anywhere else. Even the linked code is reading the value from payload.

hananiel: Ok yes thats a problem - fixing it.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: I made this change to add list of sources to a concurrent bag,  then enable events after initialization with options from constructor.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: I made this change to remove the logic to read actual interval.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: Ok fixed this and my bad for not actually validating this. But now seems to match up nicely with the dotnet-counters tool.  
CompToDotnetCounter

cijothomas: Thanks! could you add a unit test to validate the flow? you can use inmemoryexporter, and inspect the actual value. Can create our own eventcounter like an example from this : https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/EventCounterCollector/EventCounterCollector.Tests/TestEventCounter.cs

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,229 @@
...
+            }
+            catch (Exception ex)
+            {
+                EventCountersInstrumentationEventSource.Log.ErrorEventCounter(source.Name, ex.Message);
+            }
+        }
+
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
+        {
+            if (!this.isInitialized || !eventData.EventName.Equals("EventCounters"))
```
cijothomas, line 79: Where does the "EventCounters" name come from?

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,229 @@
...
+                        counterName = payload.Value.ToString();
+                    }
+                    else
+                    if (key.Equals("DisplayName", StringComparison.OrdinalIgnoreCase))
+                    {
+                        counterDisplayName = payload.Value.ToString();
+                    }
+                    else if (key.Equals("Mean", StringComparison.OrdinalIgnoreCase))
+                    {
+                        instrumentType = InstrumentType.Counter;
```
cijothomas, line 126: this feels a bit odd.. "Mean" means the user has user EventCounter. It has nothing to do with the Otel concept of "Counter". This should be a gauge.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,229 @@
...
+                        actualValue = payload.Value.ToString();
+                    }
+                    else if (key.Equals("Increment", StringComparison.OrdinalIgnoreCase))
```
cijothomas, line 129: This would be the incrementingeventcounter, which would be more like OTel's Counter. so this would map to ObservableCounter.
(need to write a unit test to fully see how this whole things works end2end. The EventCounter terminology can be very confusing!)

birojnayak:
> Resolves Issue #215
> 
> Created a focussed PR that subscribes to all Event Counter Events. The configuration to selectively listen to particular EventCounter Sources and/or Counters will be added as a separate PR.
> 
> This will replace PR #177

this is definitely going to be useful.. waiting for this CR to be merged

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: @cijothomas  I added a test for each EventCounterType.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: It is hardhoded in the implementation class:  https://github.com/dotnet/runtime/blob/e026f392b2bb93bc9509ac5403d3e8cd7d3eee72/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventCounter.cs#L129

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: You are correct. I changed it.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
hananiel: It would appear so and I followed your suggestion but I am not sure this is the right instrument. I feel like this should be an  observable guage as well, because the eventcounterListener always sees a sum aggregation. But if we use the ObservableCounter and set the metric reader options to Delta, these values dont make sense. As it is if we use cumulative, it does make sense.

stebet:
Would absolutely love to see this make it in! Let me know if I can help @cijothomas @hananiel

cijothomas (approved):
Looks good for an initial PR. We can iterate on this.

src/OpenTelemetry.Instrumentation.EventCounters/README.md
```diff
@@ -0,0 +1,52 @@
...
+                .AddEventCounters(options =>
+                {
+                   options.RefreshIntervalSecs = 55;
+                })
+                .AddConsoleExporter()
+                .Build();
+    }
+}
+```
+
```
utpilla, line 49: It would be nice to add a sample output on the console to the README.

src/OpenTelemetry.Instrumentation.EventCounters/OpenTelemetry.Instrumentation.EventCounters.csproj
```diff
@@ -0,0 +1,17 @@
...
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <Description>OpenTelemetry Metrics instrumentation for Dotnet EventCounters</Description>
+        <PackageTags>$(PackageTags);metrics;eventcounters</PackageTags>
+        <MinVerTagPrefix>Instrumentation.EventCounters-</MinVerTagPrefix>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="OpenTelemetry.Api" Version="1.2.0-rc2" />
```
utpilla, line 11: You could use `1.3.0`.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            var metricKey = new MetricKey(eventSourceName, counterName);
+            var description = string.IsNullOrEmpty(displayName) ? counterName : displayName;
+            this.lastValue[metricKey] = calculateRate ? value / this.options.RefreshIntervalSecs : value;
+            switch (instrumentType)
+            {
+                case InstrumentType.ObservableCounter:
+                    if (!this.metricInstruments.ContainsKey(metricKey))
+                    {
+                        this.metricInstruments[metricKey] = this.meter.CreateObservableCounter(counterName, () => this.ObserveDouble(metricKey), description: description);
```
utpilla, line 155: Use the `TryAdd` method of the `ConcurrentDictionary` to add instruments for thread-safe implementation.

utpilla:
@hananiel Thanks for creating this PR! Could you please follow the [contributing doc](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#contributing-a-new-project) to add additional files such as package workflow, issue template etc. You could also refer to any of the existing projects.

The other comments could be addressed in a follow-up PR.

utpilla (approved):
Approved with suggestions.

src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs
```diff
@@ -0,0 +1,199 @@
...
+                        this.metricInstruments.TryAdd(metricKey, this.meter.CreateObservableCounter(counterName, () => this.ObserveDouble(metricKey), description: description));
+                    }
+
+                    break;
+
+                case InstrumentType.ObservableGauge:
+
+                    if (!this.metricInstruments.TryGetValue(metricKey, out Instrument instrument))
+                    {
+                        this.metricInstruments[metricKey] = this.meter.CreateObservableGauge(counterName, () => this.ObserveDouble(metricKey), description: description);
```
utpilla, line 164: You could use `TryAdd` here as well.

test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCounterListenerTests.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            await Task.Delay(2000);
+            this.meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            this.meterProvider.Dispose();
+
+            Assert.True(metricItems.Count > 1);
+        }
+
+        [Fact(Skip = "Unstable")]
```
utpilla, line 50: Consider using `InMemoryExporter` that takes in a collection of `MetricSnapshot` as we persist the metric values in that case.

test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCounterListenerTests.cs
```diff
@@ -0,0 +1,200 @@
...
+        public async Task TestEventCounterMetricsAreCaptured()
```
utpilla, line 51: These need not be `async Task`. You could use `Task.Delay(...).Wait()` instead.

utpilla (approved):
LGTM. Left some suggestions to improve the unit tests which could be done in a follow-up PR.

magoogli:
This feature is great @hananiel, just what I was looking for. Can I help with the additional work to select which event counters to listen to, or is this in progress already?

mic-max:
> This feature is great @hananiel, just what I was looking for. Can I help with the additional work to select which event counters to listen to, or is this in progress already?

I did some work on that here https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/620
There could be additional need to listen to an eventName from a specific eventSource though. The options I configured only listen to the listed event sources and listed event names, not a particular event from a specific event source if that makes sense.

PR #375, opened 2022-05-25
Allow serialization of Unicode characters for Exception type

utpilla:
Addressing the issue mentioned here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/345#discussion_r879913448

## Changes
- Use `SerializeUnicodeString` method to serialize `LogRecord.Exception.GetType().FullName`


test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
```diff
@@ -924,5 +928,10 @@ private void AssertFluentdForwardModeForLogRecord(GenevaExporterOptions exporter
             // Epilouge
             Assert.Equal("DateTime", timeFormat["TimeFormat"]);
         }
+
+        // A custom exception class with non-ASCII character in the type name
+        private class CustomExceptionû : Exception
```
reyang, line 933: Consider a different approach (e.g. reflection, dynamic types) to avoid non-ASCII characters in the source code.

reyang (changes requested):
I just realized that many source files in this repo use UTF-8 BOM, probably worth considering some improvements (many tools don't work well with non-ASCII characters in the source code - e.g. low level debuggers, old fashioned terminal). This doesn't have to block the PR.

The blocker is https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/375/files#r881894745.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
utpilla: I replaced the character by its code. This should be good enough?

reyang: Thanks!

reyang (approved):

PR #385, opened 2022-06-02
[Exporter.Instana] Check if env vars are defined

zivaninstana:
If environment variables are not defined but Instana exporter is added that can cause the application to crash. 
We have to check if env vars are defined before we try to start reporting to Instana backend.


cijothomas (approved):

zivaninstana:
@cijothomas could we release new version of Exporter.Instana on NuGet?

Issue #335, opened 2022-04-29
Bringing .NET Runtime metrics instrumentation to Stable


xiang17:
Opening an issue for gaining feedbacks and making improvements on .NET Runtime metric instrumentation, regarding which metrics to expose, what each metrics means exactly, and how to calculate/fetch each metrics along with clear documentation for each metrics.

The goal is to have well defined and documented metrics for .NET Runtime users to have better knowledge of a system in a user-friendly way, in order to get the Runtime metrics instrumentation library to a stable stage.

Currently the Runtime metrics already works upon finish of the issue #204, but it's still in Alpha stage. The metrics instrumentations are initiated in [`RuntimeMetrics.cs`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs). The metrics are mostly from EventCounters: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs

# UI

The primary goal is to provide enable users to utilize user-friendly UI to gain insights in .NET Runtime behavior. Here is a simple demo for the concept. The Grafana dashboard graphs I set up in my laptop look this (see appendix on how to do it):
![Dotnet-Grafana](https://user-images.githubusercontent.com/9310587/166079695-6617614e-69ab-43d4-844f-3414d40c3556.png)

Or this:
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/9310587/166079722-2ad30203-164e-45a1-9139-dccaaa60469c.png)

# Metrics

*TODO*: List of metrics grouped in CPU, Memory, Garbage Collection, Threadpool, Process, and etc.

# Documentation

Clear description on what each metric means. Given that developers and support engineers might rely on this to debug or diagnose issues, the description should reveal reasonably enough insight into internal details.

Help from Runtime team is appreciated especially for this part.

# Appendix

Instruction on how to set up the graphs:
1. Set up Prometheus and Grafana using [this guide](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/getting-started-prometheus-grafana).
[...]

Issue #383, opened 2022-05-23
NewRelic metric errors (monotonic values must be positive)


ejsmith:
# Bug Report

I'm sending my data to New Relic and getting a lot of errors from the `AddRuntimeMetrics` integration reporting .net metrics.

_One or more OTLP metric data point(s) were dropped because monotonic values must be positive._

`process.runtime.dotnet.gc.fragmentation, process.runtime.dotnet.threadpool.queue.length, process.runtime.dotnet.threadpool.thread.count`

Anyone have any ideas how to fix this issue?

PR #404, opened 2022-06-07
[Do Not Merge] Add description doc for DotNet Runtime metrics

xiang17:
For issue #335.

Creating a doc for a description for each runtime metrics. This doc will both serve as a design doc and as a formal descriptive doc for users in future.

I'd like to ask for clarification on a few metrics that I'm not sure whether they are monotonically increasing or not.

Can I assume the metrics in [the EventCounter implementation](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs) using IncrementingPollingCounter are all monotonically increasing? More specifically on this one:
- GC.GetTotalAllocatedBytes()

I think the following metrics are monotonically increasing but the EventCounter implementation used PollingCounter rather than IncrementingPollingCounter:
- il.bytes.jitted: System.Runtime.JitInfo.GetCompiledILBytes()
- methods.jitted.count: System.Runtime.JitInfo.GetCompiledMethodCount()
- assembly.count: AppDomain.GetAssemblies().Length

I think these are monotonically increasing, but would like to confirm:
- Process.UserProcessorTime and Process.PrivilegedProcessorTime (for current process as retrieved by: Process.GetCurrentProcess())

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/description.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/description.md
@@ -127,7 +127,7 @@ The metrics in this section can be enabled by setting the
 | Name                    | Description                            | Units          | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
 |-------------------------|----------------------------------------|----------------|-------------------|------------|------------------|------------------|
 | process.cpu.utilization | CPU utilization of this process        | `1`            | ObservableGauge   | `Double`   |                  |                  |
-| process.cpu.time        | Processor time of this process         | `s`            | ObservableCounter | `Int64`    | type             | user, system     |
+| process.cpu.time        | Processor time of this process         | `s`            | ObservableCounter | `Int64`    | state            | user, system     |
 | process.cpu.count       | The number of available logical CPUs   | `{processors}` | ObservableGauge   | `Int64`    |                  |                  |
 | process.memory.usage    | The amount of physical memory in use   | `By`           | ObservableGauge   | `Int64`    |                  |                  |
 | process.memory.virtual  | The amount of committed virtual memory | `By`           | ObservableGauge   | `Int64`    |                  |                  |
--- a/src/OpenTelemetry.Instrumentation.Runtime/description.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/description.md
@@ -77,7 +77,7 @@ These metrics are only available for NET6_0_OR_GREATER.
 |-------------------------------------------------|--------------------------|-------------|-------------------|------------|------------------|------------------|
 | process.runtime.dotnet.**il.bytes.jitted**      | IL Bytes Jitted          | `By`        | ObservableCounter | `Int64`    |                  |                  |
 | process.runtime.dotnet.**methods.jitted.count** | Number of Methods Jitted | `{methods}` | ObservableCounter | `Int64`    |                  |                  |
-| process.runtime.dotnet.**time.in.jit**          | Time spent in JIT        | `ns`        | ObservableCounter | `Double`   |                  |                  |
+| process.runtime.dotnet.**time.in.jit**          | Time spent in JIT        | `ns`        | ObservableCounter | `Int64`   |                  |                  |
 
 [JitInfo.GetCompiledILBytes](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledilbytes?view=net-6.0#system-runtime-jitinfo-getcompiledilbytes(system-boolean)):
 Gets the number of bytes of intermediate language that have been compiled.
--- a/src/OpenTelemetry.Instrumentation.Runtime/description.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/description.md
@@ -45,7 +45,6 @@ Gets the total committed bytes of the managed heap.
 
 - [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
 Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
-(The number of generations `i` is limited by [GC.MaxGeneration](https://docs.microsoft.com/dotnet/api/system.gc.maxgeneration))
 
 ## JIT Compiler related metrics
 
--- a/src/OpenTelemetry.Instrumentation.Runtime/description.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/description.md
@@ -13,18 +13,7 @@ The metrics in this section can be enabled by setting the
 
 | Name                                          | Description              | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
 |-----------------------------------------------|--------------------------|-----------|-------------------|------------|------------------|------------------|
-| process.runtime.dotnet.**gc.totalmemorysize** | Total allocated bytes    | `By`      | ObservableGauge   | `Int64`    |                  |                  |
 | process.runtime.dotnet.**gc.count**           | Garbage Collection count | `{times}` | ObservableCounter | `Int64`    | gen              | gen0, gen1, gen2 |
-
-Question for .NET team: is GC.GetTotalMemory(false) always equal to the sum of
-GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes (i from 0 to 4)?
-
-I need to decide whether it makes sense to include both of them in the memory/GC
-size metrics.
-
-- [GC.GetTotalMemory](https://docs.microsoft.com/dotnet/api/system.gc.gettotalmemory):
-The number of bytes currently thought to be allocated.
-It does not wait for garbage collection to occur before returning.
 
 - [GC.CollectionCount](https://docs.microsoft.com/dotnet/api/system.gc.collectioncount):
 The number of times garbage collection has occurred for the specified generation
--- a/src/OpenTelemetry.Instrumentation.Runtime/description.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/description.md
@@ -1,0 +1,193 @@
+# Runtime metrics description
+
+Metrics name are prefixed with the `process.runtime.dotnet.` namespace, following
+the general guidance for runtime metrics in the
+[specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#runtime-environment-specific-metrics---processruntimeenvironment).
+Instrument Units [should follow](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-units)
+the Unified Code for Units of Measure.
+
+## GC related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsGcEnabled` switch.
+
+| Name                                          | Description              | Units     | Instrument Type ([*](README.md#instrument-types)) | Value Type | Attribute Key(s) | Attribute Values |
+|-----------------------------------------------|--------------------------|-----------|---------------------------------------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.totalmemorysize** | Total allocated bytes    | `By`      | ObservableGauge                                   | `Int64`    |                  |                  |
+| process.runtime.dotnet.**gc.count**           | Garbage Collection count | `{times}` | ObservableCounter                                 | `Int64`    | gen              | gen0, gen1, gen2 |
+
+Question for .NET team: is GC.GetTotalMemory(false) always equal to the sum of GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes (i from 0 to 4)?
+
+I need to decide whether it makes sense to include both of them in the memory/GC size metrics.
+
+- [GC.GetTotalMemory](https://docs.microsoft.com/dotnet/api/system.gc.gettotalmemory):
+The number of bytes currently thought to be allocated.
+It does not wait for garbage collection to occur before returning.
+
+- [GC.CollectionCount](https://docs.microsoft.com/dotnet/api/system.gc.collectioncount):
+The number of times garbage collection has occurred for the specified generation
+of objects.
+
+### Additional GC metrics only available for NETCOREAPP3_1_OR_GREATER
+
+| Name                                               | Description                                      | Units | Instrument Type ([*](README.md#instrument-types)) | Value Type          | Attribute Key(s) | Attribute Values |
+|----------------------------------------------------|--------------------------------------------------|-------|---------------------------------------------------|---------------------|------------------|------------------|
+| process.runtime.dotnet.**gc.allocated.bytes**      | Bytes allocated over the lifetime of the process | `By`  | ObservableCounter                                 | `Int64`             |                  |                  |
+| process.runtime.dotnet.**gc.fragmentation.ratio**  | GC fragmentation ratio                           | `1`   | ObservableGauge                                   | `Double`            |                  |                  |
+
+- [GC.GetTotalAllocatedBytes](https://docs.microsoft.com/dotnet/api/system.gc.gettotalallocatedbytes):
+  Gets a count of the bytes allocated over the lifetime of the process. The returned
+value does not include any native allocations. The value is an approximate count.
+
+- GC fragmentation ratio is calculated as:
+If `GC.GetGCMemoryInfo().HeapSizeBytes != 0`,
+the value is
+`GC.GetGCMemoryInfo().FragmentedBytes * 1.0d / GC.GetGCMemoryInfo().HeapSizeBytes`,
+otherwise the value is `0`.
+
+  - [GCMemoryInfo.FragmentedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.fragmentedbytes?view=netcore-3.1):
+  Gets the total fragmentation when the last garbage collection occurred.
+  - [GCMemoryInfo.HeapSizeBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.heapsizebytes?view=netcore-3.1#system-gcmemoryinfo-heapsizebytes):
+  Gets the total heap size when the last garbage collection occurred.
+
+### Additional GC metrics only available for NET6_0_OR_GREATER
+
+| Name                                    | Description        | Units | Instrument Type ([*](README.md#instrument-types)) | Value Type | Attribute Key(s) | Attribute Values           |
+|-----------------------------------------|--------------------|-------|---------------------------------------------------|------------|------------------|----------------------------|
+| process.runtime.dotnet.**gc.committed** | GC Committed Bytes | `By`  | ObservableGauge                                   | `Int64`    |                  |                            |
+| process.runtime.dotnet.**gc.heapsize**  |                    | `By`  | ObservableGauge                                   | `Int64`    | gen              | gen0, gen1, gen2, loh, poh |
+
+- [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes?view=net-6.0#system-gcmemoryinfo-totalcommittedbytes):
+Gets the total committed bytes of the managed heap.
+
+- [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
+Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+(The number of generations `i` is limited by [GC.MaxGeneration](https://docs.microsoft.com/dotnet/api/system.gc.maxgeneration))
+
+## JIT Compiler related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsJitEnabled` switch.
+
+These metrics are only available for NET6_0_OR_GREATER.
+
+| Name                                            | Description              | Units       | Instrument Type ([*](README.md#instrument-types)) | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------|--------------------------|-------------|---------------------------------------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**il.bytes.jitted**      | IL Bytes Jitted          | `By`        | ObservableCounter                                 | `Int64`    |                  |                  |
+| process.runtime.dotnet.**methods.jitted.count** | Number of Methods Jitted | `{methods}` | ObservableCounter                                 | `Int64`    |                  |                  |
+| process.runtime.dotnet.**time.in.jit**          | Time spent in JIT        | `ms`        | ObservableCounter                                 | `Double`   |                  |                  |
+
+[JitInfo.GetCompiledILBytes](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledilbytes?view=net-6.0#system-runtime-jitinfo-getcompiledilbytes(system-boolean)):
+Gets the number of bytes of intermediate language that have been compiled.
+The scope of this value is global.
+
+[JitInfo.GetCompiledMethodCount](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledmethodcount?view=net-6.0#system-runtime-jitinfo-getcompiledmethodcount(system-boolean)):
+Gets the number of methods that have been compiled.
+The scope of this value is global.
+
+[JitInfo.GetCompilationTime](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompilationtime?view=net-6.0#system-runtime-jitinfo-getcompilationtime(system-boolean)):
+Gets the amount of time the JIT Compiler has spent compiling methods.
+The scope of this value is global.
+
+## Threading related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsThreadingEnabled` switch.
+
+These metrics are only available for NETCOREAPP3_1_OR_GREATER.
+
+| Name                                                        | Description                          | Units       | Instrument Type ([*](README.md#instrument-types)) | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------------------|--------------------------------------|-------------|---------------------------------------------------|------------|------------------|------------------|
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+| process.runtime.dotnet.**gc.fragmentation.ratio** | GC fragmentation ratio                           | `1`   | ObservableGauge   | `Double`   |                  |                  |
+
+- [GC.GetTotalAllocatedBytes](https://docs.microsoft.com/dotnet/api/system.gc.gettotalallocatedbytes):
+  Gets a count of the bytes allocated over the lifetime of the process. The returned
+value does not include any native allocations. The value is an approximate count.
+
+- GC fragmentation ratio is calculated as:
+If `GC.GetGCMemoryInfo().HeapSizeBytes != 0`,
+the value is
+`GC.GetGCMemoryInfo().FragmentedBytes * 1.0d / GC.GetGCMemoryInfo().HeapSizeBytes`,
```
reyang, line 47: Might need to call out that the result from `GetGCMemoryInfo` is used to get both `FragmentedBytes` and `HeapsizeBytes` so the behavior is transactional.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+
+- [GC.CollectionCount](https://docs.microsoft.com/dotnet/api/system.gc.collectioncount):
+The number of times garbage collection has occurred for the specified generation
+of objects.
+
+### Additional GC metrics only available for NETCOREAPP3_1_OR_GREATER
+
+| Name                                              | Description                                      | Units | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|---------------------------------------------------|--------------------------------------------------|-------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.allocated.bytes**     | Bytes allocated over the lifetime of the process | `By`  | ObservableCounter | `Int64`    |                  |                  |
```
reyang, line 37: Is this monotonically increasing?

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+the Unified Code for Units of Measure.
+
+## GC related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsGcEnabled` switch.
+
+| Name                                          | Description              | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-----------------------------------------------|--------------------------|-----------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.totalmemorysize** | Total allocated bytes    | `By`      | ObservableGauge   | `Int64`    |                  |                  |
```
reyang, line 16: I think many of the `ObservableGauge` here should be `ObservableUpDownCounter` once .NET has the support for up-down counters.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+  Gets the number of thread pool threads that currently exist.
+- [ThreadPool.CompletedWorkItemCount](https://docs.microsoft.com/dotnet/api/system.threading.threadpool.completedworkitemcount?view=netcore-3.1):
+  Gets the number of work items that have been processed so far.
+- [ThreadPool.PendingWorkItemCount](https://docs.microsoft.com/dotnet/api/system.threading.threadpool.pendingworkitemcount?view=netcore-3.1):
+  Gets the number of work items that are currently queued to be processed.
+- [Timer.ActiveCount](https://docs.microsoft.com/dotnet/api/system.threading.timer.activecount?view=netcore-3.1):
+  Gets the number of timers that are currently active. An active timer is registered
+  to tick at some point in the future, and has not yet been canceled.
+
+## Process related metrics
```
reyang, line 122: I have a general question - do we want these to be in the `OpenTelemetry.Instrumentation.Runtime` package, or we want them to be in a different package (and if yes, what should be the name of that package)?

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsGcEnabled` switch.
+
+| Name                                          | Description              | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-----------------------------------------------|--------------------------|-----------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.totalmemorysize** | Total allocated bytes    | `By`      | ObservableGauge   | `Int64`    |                  |                  |
+| process.runtime.dotnet.**gc.count**           | Garbage Collection count | `{times}` | ObservableCounter | `Int64`    | gen              | gen0, gen1, gen2 |
+
+Question for .NET team: is GC.GetTotalMemory(false) always equal to the sum of
```
xiang17, line 19: @tarekgh can you give comment on this?

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+
+These metrics are only available for NET6_0_OR_GREATER.
+
+| Name                                            | Description              | Units       | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------|--------------------------|-------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**il.bytes.jitted**      | IL Bytes Jitted          | `By`        | ObservableCounter | `Int64`    |                  |                  |
+| process.runtime.dotnet.**methods.jitted.count** | Number of Methods Jitted | `{methods}` | ObservableCounter | `Int64`    |                  |                  |
+| process.runtime.dotnet.**time.in.jit**          | Time spent in JIT        | `ns`        | ObservableCounter | `Double`   |                  |                  |
+
+[JitInfo.GetCompiledILBytes](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledilbytes?view=net-6.0#system-runtime-jitinfo-getcompiledilbytes(system-boolean)):
```
xiang17, line 82: If it's running on AOT mode, would this return zero or throw exception?

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: I've reworded it to get the properties from calling `GetGCMemoryInfo` once.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+|-----------------------------------------|--------------------|-------|-----------------|------------|------------------|----------------------------|
+| process.runtime.dotnet.**gc.committed** | GC Committed Bytes | `By`  | ObservableGauge | `Int64`    |                  |                            |
+| process.runtime.dotnet.**gc.heapsize**  |                    | `By`  | ObservableGauge | `Int64`    | gen              | gen0, gen1, gen2, loh, poh |
+
+- [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes?view=net-6.0#system-gcmemoryinfo-totalcommittedbytes):
+Gets the total committed bytes of the managed heap.
+
+- [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
+Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+(The number of generations `i` is limited by [GC.MaxGeneration](https://docs.microsoft.com/dotnet/api/system.gc.maxgeneration))
```
xiang17, line 67: `GC.MaxGeneration` returned `2` when GC Count has values for gen0, gen1, gen2 and GC heap size has values for all five generations. I will remove mention of this property and use 3 for GC Count and `GC.GetGCMemoryInfo().GenerationInfo.Length` for heap sizes.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+  to the process context.
+
+## Assemblies related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsAssembliesEnabled` switch.
+
+| Name                                      | Description                 | Units          | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------|-----------------------------|----------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**assembly.count** | Number of Assemblies Loaded | `{assemblies}` | ObservableCounter | `Int64`    |                  |                  |
```
xiang17, line 172: .NET runtime has the ability to [unload assemblies](https://docs.microsoft.com/en-us/dotnet/standard/assembly/load-unload), so `AppDomain.GetAssemblies().Length` could decrease in some occasions. The instrument type should be `ObservableUpDownCounter` (before `ObservableUpDownCounter` is ready, I'll use `ObservableGauge` in implementation). I'll update the doc later.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+## Process related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsProcessEnabled` switch.
+
+| Name                    | Description                            | Units          | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------|----------------------------------------|----------------|-------------------|------------|------------------|------------------|
+| process.cpu.utilization | CPU utilization of this process        | `1`            | ObservableGauge   | `Double`   |                  |                  |
+| process.cpu.time        | Processor time of this process         | `s`            | ObservableCounter | `Int64`    | state            | user, system     |
+| process.cpu.count       | The number of available logical CPUs   | `{processors}` | ObservableGauge   | `Int64`    |                  |                  |
```
xiang17, line 131: `process.cpu.count` metric was never officially added (see https://github.com/open-telemetry/opentelemetry-specification/pull/2392). I'll remove it later.

src/OpenTelemetry.Instrumentation.Runtime/description.md
tarekgh: Your comment is for .NET Framework and not for .NET Core as .NET Core doesn't allow more than one appomain in the process. 

> AppDomain.GetAssemblies().Length could decrease in some occasions.

I am not sure if this is accurate either. You may unload the whole appdomain and not individual assembly. So, technically the `AppDomain.GetAssemblies().Length` doesn't decrease. but the total assemblies in the process can decrease. 

In .NET Core, uses AssemblyLoadContext for unloading all assemblies in that context by unloading the context. 

CC @noahfalk

src/OpenTelemetry.Instrumentation.Runtime/description.md
noahfalk: My understanding is that they aren't the same. GC.GetGCMemoryInfo() is giving a snapshot of historical data that was computed when the most recent GC started. GC.GetTotalMemory(false) is giving the live value. Aside from timing difference there may also be a difference in the accuracy. The GC has a central pool of memory that it gives out to threads in chunks (8 or 16KB each I think?). Each chunk is called an allocation context inside the runtime. Each thread then sub-allocates individual .NET objects from its private allocation context. I wouldn't be surprised if GetTotalMemory approximates that the entire allocation_context as being allocated memory whereas GetGCMemoryInfo() might be accurately measuring how much of each allocation_context has been given out by the thread which owns it.

src/OpenTelemetry.Instrumentation.Runtime/description.md
noahfalk: We'll probably need to to get more precise on this one. Some different possible interpretations:
- Total amount of memory ever allocated from the GC since process start
- Total amount of memory commited by the GC currently
- Total amount of memory for live .NET objects + memory for free objects (objects that existed in the past and were collected but the memory hasn't yet been reused)
- Total amount of memory for live .NET objects only
- Total amount of memory for rooted .NET objects (objects that would still be live after a full GC)

src/OpenTelemetry.Instrumentation.Runtime/description.md
noahfalk: I'd vote separate package and OpenTelemetry.Instrumentation.Process as a potential name?

src/OpenTelemetry.Instrumentation.Runtime/description.md
noahfalk: I haven't tested it, but I would assume that assemblies in any AssemblyLoadContext, including an unloadable one, appear in the AppDomain.GetAssemblies() list. In that case AppDomain.GetAssemblies().Length could decrease because a load context unloaded.

tarekgh: Will such a counter have really a useful meaning in .NET core apps? And how can users handle such a counter in .NET Framework apps? or we don't care much about .NET Framework cases?

reyang: My answer is YES. I can think of a valid use case - certain online service / background job run out of memory after hours and the engineers can tell from the "assembly loaded" metrics that the process kept loading plugins without unloading them.

This happened in my previous team, where a backend processing job detects new XML schema and compiles it to assembly (for sake of perf) and load it dynamically, the backend job ended up loading tens of thousands of assemblies.

tarekgh: Sounds good then!

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+| process.memory.usage    | The amount of physical memory in use   | `By`           | ObservableGauge   | `Int64`    |                  |                  |
+| process.memory.virtual  | The amount of committed virtual memory | `By`           | ObservableGauge   | `Int64`    |                  |                  |
```
xiang17, line 133: Note: `VirtualMemorySize64` is giving me numbers in the range of 2TB (numbers like 2204055367680, 2204056940544), when `WorkingSet64` is only 20~40 MB. I need to figure out how/why the `VirtualMemorySize64` is so big. (Probably something to do with how my runner is written.)

image

reyang: https://docs.microsoft.com/en-us/windows/win32/memory/working-set

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Thanks, Tarek and Noah! I'm getting the assembly count for the "current domain" `AppDomain.CurrentDomain.GetAssemblies().Length`, and the metric has to work for both .NET Framework and .NET Core. If the value could decrease in any case, I couldn't use `ObservableCounter` because a metric data point [could be dropped](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/383).

I found one [remark](https://docs.microsoft.com/en-us/dotnet/api/system.appdomain?view=net-6.0#remarks) about a special case when the assembly is in the default application domain:
> If an assembly is loaded into the default application domain, it cannot be unloaded from memory while the process is running.

If Noah's right about assemblies in AssemblyLoadContext would appear in `AppDomain.CurrentDomain.GetAssemblies()`, then I should use `ObservableGauge` (and use `ObservableUpDownCounter` when available). Otherwise it might lead to loss of data.

src/OpenTelemetry.Instrumentation.Runtime/description.md
tarekgh: `VirtualMemorySize64 ` tells the total virtual memory available for the process to use (not necessarily be committed). While Working Set is `the set of pages in the virtual address space of the process that are currently resident in physical memory` from the doc that @reyang sent. 
These are two different things, and it is expected the Working Set always be less than the virtual memory size.

src/OpenTelemetry.Instrumentation.Runtime/description.md
tarekgh: > I'm getting the assembly count for the "current domain" AppDomain.CurrentDomain.GetAssemblies().Length

If we are reporting the number from the current appdomain only, I'll question the @reyang's scenario `certain online service / background job run out of memory after hours and the engineers can tell from the "assembly loaded" metrics that the process kept loading plugins without unloading them.` when running on .NET Framework. Looks what we report wouldn't be enough for that scenario. 

To clarify, .NET Framework can have multiple appdomains. I guess it will be more useful reporting the total count of assemblies in all appdomain and not necessary the current appdomain. 

In .NET Core, there is only one appdomain, so reporting the number from the current appdomain would work, I guess. We need to confirm loading and unloading AssemblyLoadContext is going to be reflected to the number we report. When I get some time, I can try to do that.

src/OpenTelemetry.Instrumentation.Runtime/description.md
reyang: Some raw notes from our discussion:

process.runtime.dotnet.gc.heapsize: live object (not necessarily hold by the user objects/threads, could be hold by the GC itself) + fragmentation

all the gen adding up should <= committed (since GC commits at the granularity of page, and it normally would ask for multiple pages instead of just one for caching purpose)

neither of the above includes GC's own memory usage

github.com/Maoni0/mem-doc/blob/master/doc/.NETMemoryPerformanceAnalysis.md

GC.GetGenerationSize(n) is not providing the transactional behavior since we have to call it multiple times for different gen, it is okay as long as we mention it in the description/doc

remove gc.totalmemorysize since it can be covered by adding up different gens

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+- [GC.CollectionCount](https://docs.microsoft.com/dotnet/api/system.gc.collectioncount):
+The number of times garbage collection has occurred for the specified generation
+of objects.
+
+### Additional GC metrics only available for NETCOREAPP3_1_OR_GREATER
+
+| Name                                              | Description                                      | Units | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|---------------------------------------------------|--------------------------------------------------|-------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.allocated.bytes**     | Bytes allocated over the lifetime of the process | `By`  | ObservableCounter | `Int64`    |                  |                  |
+| process.runtime.dotnet.**gc.fragmentation.ratio** | GC fragmentation ratio                           | `1`   | ObservableGauge   | `Double`   |                  |                  |
```
reyang, line 38: Discussed offline, change fragmentation ratio -> size (because ratio doesn't add up)

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,198 @@
...
+
+## GC related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsGcEnabled` switch.
+
+| Name                                          | Description              | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-----------------------------------------------|--------------------------|-----------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.totalmemorysize** | Total allocated bytes    | `By`      | ObservableGauge   | `Int64`    |                  |                  |
+| process.runtime.dotnet.**gc.count**           | Garbage Collection count | `{times}` | ObservableCounter | `Int64`    | gen              | gen0, gen1, gen2 |
```
reyang, line 17: This is suggested by @noahfalk - for folks who have been using the EventCounters, we might consider something like "legacy time in GC" and make it an explicit opt-in (with a clear API name indicating that it has an unclear definition, and we don't recommend it).

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: I've removed the metrics for `GC.GetTotalMemory(false)` because it includes the fragmentation and it is not very useful to customers.

Its return value is the size of: live object (not necessarily hold by the user objects/threads, could be hold by the GC itself) + fragmentation.

xiang17: Removing `gc.totalmemorysize` because it is duplicate and can be calculated from `gc.heapsize`.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Yes.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: It will just return zero. The values are monotonically increasing.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Removed.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Removed.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Updated the instrument type to `ObservableGauge`.

For implementation on getting all AppDomains, I found three articles [1](http://miteshsureja.blogspot.com/2012/02/hot-to-get-list-of-all-appdomains.html), [2](https://stackoverflow.com/questions/33890264/enumerate-all-appdomains-without-mscoree) and [3](CorRuntimeHost), but they all seem like very old school code depending on MSCorEE.dll.

I still need to figure out whether I could use those APIs, or find better alternatives.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Done.

src/OpenTelemetry.Instrumentation.Runtime/description.md
noahfalk: fwiw we should be able to give a precise definition for the legacy counter too, it just isn't always useful or intuitive. I'm pretty sure the existing counter is computed like this:

GC_n = most recent GC that has ended prior to computing the value
GC_n-1 = GC that ended prior to GC_n

T_end_n = time when GC_n ended
T_end_n-1 = time when GC_n-1 ended
T_start_n = time when GC_n started

inter_gc_time = T_end_n - T_end_n-1
gc_duration = T_end_n - T_start_n
counter value = 100 * gc_duration / inter_gc_time

As an example if GC_n-1 ran from 2.1 sec to 2.2 sec and GC_n ran from 2.5 sec to 3.0 sec then
inter_gc_time = 3.0 - 2.2 = 0.8
gc_duration = 3.0 - 2.5 = 0.5
counter_value = 100 * 0.5 / 0.8 = 62.5%

People usually assume that the value measures the fraction of GC time during a fixed sampling interval (say every 5 minutes) but really it is measuring the fraction of GC time during a variable interval between the most recent GC and the one before that. If the most recent two GCs happen to be close together the counter may appear unexpectedly high and this often causes confusion.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,175 @@
+# Runtime metrics description
```
CodeBlanch, line 1: Maybe "Runtime metrics overview" or "Runtime metric details" here instead of "description"? Also we should link to this from README.

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,175 @@
...
+| process.runtime.dotnet.**methods.jitted.count** | Number of Methods Jitted | `{methods}` | ObservableCounter | `Int64`    |                  |                  |
+| process.runtime.dotnet.**time.in.jit**          | Time spent in JIT        | `ns`        | ObservableCounter | `Int64`   |                  |                  |
+
+[JitInfo.GetCompiledILBytes](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledilbytes?view=net-6.0#system-runtime-jitinfo-getcompiledilbytes(system-boolean)):
+Gets the number of bytes of intermediate language that have been compiled.
+The scope of this value is global.
+
+[JitInfo.GetCompiledMethodCount](https://docs.microsoft.com/dotnet/api/system.runtime.jitinfo.getcompiledmethodcount?view=net-6.0#system-runtime-jitinfo-getcompiledmethodcount(system-boolean)):
+Gets the number of methods that have been compiled.
+The scope of this value is global.
```
CodeBlanch, line 68: Do we need to state `The scope of this value is global.` on all of these? Could probably just drop that since it seems obvious (IMO) or move it to the heading so we only state it once? 🤷

src/OpenTelemetry.Instrumentation.Runtime/description.md
```diff
@@ -0,0 +1,175 @@
...
+
+- [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
+Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+
+## JIT Compiler related metrics
+
+The metrics in this section can be enabled by setting the
+`RuntimeMetricsOptions.IsJitEnabled` switch.
+
+These metrics are only available for NET6_0_OR_GREATER.
```
CodeBlanch, line 54: Using the preprocessor directives in this doc like `NET6_0_OR_GREATER` is very technical. Maybe just say like...

`These metrics are available when targeting .NET6 and greater`.

cijothomas: I think this is opened to just solicit feedback only.

src/OpenTelemetry.Instrumentation.Runtime/description.md
xiang17: Created a PR to remove it: #446

damikun: I have the CPU metrics percentage collector...  and will also push that to new namespace... https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/437  thx @reyang  for ping about info of new namespace...

PR #408, opened 2022-06-08
AWS Lambda wrapper enhancements: Trace methods enhancements and Activity creation behaviour changes.

rypdal:
## Changes

1. Extended public Trace methods of the AWSLambdaWrapper class with additional optional parameter: parent context which is used further for ActivitySource.StartActivity(...) call.
2. If the parent context is not provided to one of Trace methods from the AWSLambdaWrapper then parent is extracted from the incoming request if it's type is either APIGatewayProxyRequest or APIGatewayHttpApiV2ProxyRequest (both from Amazon.Lambda.APIGatewayEvents namespace). Corresponding method ExtractParentContext has been added to the AWSLambdaUtils class.
3. Changed behaviour of the [OnFunctionStart](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs#L223) method: Activity is created even if the parent context isn't defined.
4. Breaking change: AWSLambdaWrapper.Trace overloads without ILambdaContext argument have been completely removed. All current AWS lambda triggers provide ILambdaContext hence it's always possible to pass it from customer's code. The ILambdaContext is used further for function default tags definition.
5.  Two additional AWSLambdaWrapper.Trace overloads without generic input arguments have been added for the case of functions with only ILambdaContext as input.


src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -37,10 +45,32 @@ public class AWSLambdaWrapper
...
+        /// </param>
         /// <returns>Instance of output result.</returns>
-        public static TResult Trace<TInput, TResult>(TracerProvider tracerProvider, Func<TInput, TResult> lambdaHandler, TInput input)
+        public static TResult Trace<TInput, TResult>(
+            TracerProvider tracerProvider,
+            Func<TInput, TResult> lambdaHandler,
+            TInput input,
+            ActivityContext parentContext = default)
         {
-            return Intercept(tracerProvider, () => lambdaHandler(input));
```
Oberon00, line 43: Why was the `Intercept` pattern changed? Couldn't that approach still be used to reduce code duplication between the overloads?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +123,36 @@ internal static string GetAccountId(string functionArn)
...
+                return tags;
+            }
+
+            tags.Add(new(AWSLambdaSemanticConventions.AttributeFaasName, context.FunctionName));
+            tags.Add(new(AWSLambdaSemanticConventions.AttributeFaasID, context.InvokedFunctionArn));
+
+            var functionParts = context.InvokedFunctionArn?.Split(':');
+            if (functionParts != null && functionParts.Length >= 5)
+            {
+                tags.Add(new(AWSLambdaSemanticConventions.AttributeCloudAccountID, functionParts[4]));
```
Oberon00, line 150: It seems this is already set directly in OnFunctionStart and is redundant here.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -61,6 +63,27 @@ internal static ActivityContext GetParentContext()
             return activityContext;
         }
 
+        internal static ActivityContext ExtractParentContext<TInput>(TInput input)
```
Oberon00, line 66: I think this is an extremely useful feature and should be mentioned in the PR description, and the XML docs of the AWSLambdaWrapper class.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -112,5 +165,31 @@ private static ActivityContext ParseXRayTraceHeader(string rawHeader)
...
+
+        private static IEnumerable<string> GetHeaderValues(APIGatewayHttpApiV2ProxyRequest request, string name)
+        {
+            if (request.Headers != null &&
+                request.Headers.TryGetValue(name, out var header))
+            {
+                // Multiple values for the same header will be separated by a comma.
+                if (!string.IsNullOrEmpty(header))
+                {
+                    return header.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
rypdal: But do empty headers have any meaning for parent context extraction ?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,7 +27,15 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
     public class AWSLambdaWrapper
     {
-        private static readonly ActivitySource AWSLambdaActivitySource = new ActivitySource(AWSLambdaUtils.ActivitySourceName);
+        private static readonly ActivitySource AWSLambdaActivitySource = new(AWSLambdaUtils.ActivitySourceName);
+
+        /// <summary>
+        /// Gets or sets a value indicating whether AWS X-Ray propagation should be ignored. Default value is false.
+        /// The flag was introduced as a workaround of this issue:
+        /// the ActivitySource.StartActivity method returns null if sampling decision of DROP (Sampled=0).
+        /// Flag can be removed as soon as the bug https://github.com/open-telemetry/opentelemetry-dotnet/issues/3290 is resolved.
```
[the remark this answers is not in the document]
rypdal: Updated doc.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: All moved to AWSLambdaUtils.GetFunctionDefaultTags(...) method.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Added to the PR description, and mentioned in the XML docs of the AWSLambdaWrapper class.

utpilla:
@srprash Could you please review this PR?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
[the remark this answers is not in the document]
Oberon00: That should be up to the propagator and not decided in the Getter. As said, it should not matter in practice though.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +118,55 @@ internal static string GetAccountId(string functionArn)
...
+
+            if (context.AwsRequestId != null)
+            {
+                tags.Add(new(AWSLambdaSemanticConventions.AttributeFaasExecution, context.AwsRequestId));
+            }
+
+            var functionArn = context.InvokedFunctionArn;
+            if (functionArn != null)
+            {
+                tags.Add(new(AWSLambdaSemanticConventions.AttributeFaasID, functionArn));
```
Oberon00, line 158: This should be stored in `aws.lambda.invoked_arn` https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#all-triggers

For the faas.id, you need to make sure that any component of the ARN after the function name (alias or version) is removed (spec currently says to add the faas.version, as read from envvar, instead but I don't think this makes much sense actually)

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +118,55 @@ internal static string GetAccountId(string functionArn)
...
+            var trigger = "other";
+            if (input is APIGatewayProxyRequest || input is APIGatewayHttpApiV2ProxyRequest)
+            {
+                trigger = "http";
+            }
+
+            return trigger;
+        }
+
+        internal static IEnumerable<KeyValuePair<string, object>> GetFunctionDefaultTags<TInput>(TInput input, ILambdaContext context)
```
Oberon00, line 132: Why "Default"?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -112,5 +179,31 @@ private static ActivityContext ParseXRayTraceHeader(string rawHeader)
...
+            return null;
+        }
+
+        private static IEnumerable<string> GetHeaderValues(APIGatewayHttpApiV2ProxyRequest request, string name)
+        {
+            if (request.Headers != null &&
+                request.Headers.TryGetValue(name, out var header))
+            {
+                // Multiple values for the same header will be separated by a comma.
+                if (!string.IsNullOrEmpty(header))
```
Oberon00, line 200:
```suggestion
                if (header != null)
```

And could then be simplified to `return header?.Split(',');` without `if`.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
Oberon00: Did you not push the updated doc yet?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
+        public static TResult Trace<TInput, TResult>(
+            TracerProvider tracerProvider,
+            Func<TInput, ILambdaContext, TResult> lambdaHandler,
+            TInput input,
+            ILambdaContext context,
+            ActivityContext parentContext = default)
         {
-            Intercept(tracerProvider, () => lambdaHandler(input));
+            TResult result = default;
+            Action action = () => result = lambdaHandler(input, context);
```
Oberon00, line 63: I would inline the action into the TraceInternal call.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
+        /// Tracing wrapper for async Lambda handler.
         /// </summary>
         /// <typeparam name="TInput">Input.</typeparam>
-        /// <typeparam name="TResult">Output result.</typeparam>
         /// <param name="tracerProvider">TracerProvider passed in.</param>
         /// <param name="lambdaHandler">Lambda handler function passed in.</param>
         /// <param name="input">Instance of input.</param>
-        /// <returns>Task of result.</returns>
-        public static async Task<TResult> Trace<TInput, TResult>(TracerProvider tracerProvider, Func<TInput, Task<TResult>> lambdaHandler, TInput input)
+        /// <param name="context">Instance of lambda context.</param>
```
Oberon00, line 99:
```suggestion
        /// <param name="context">Lambda context (optional, but strongly recommended).</param>
```

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
```diff
@@ -29,11 +31,19 @@ public static class TracerProviderBuilderExtensions
         /// Add AWS Lambda configurations.
         /// </summary>
         /// <param name="builder"><see cref="TracerProviderBuilder"/> being configured.</param>
+        /// <param name="configure">AWS lambda instrumentation options.</param>
         /// <returns>The instance of <see cref="TracerProviderBuilder"/> to chain the calls.</returns>
-        public static TracerProviderBuilder AddAWSLambdaConfigurations(this TracerProviderBuilder builder)
+        public static TracerProviderBuilder AddAWSLambdaConfigurations(
```
Oberon00, line 36: This method name is weird. But since it was there before, fine with me.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +123,36 @@ internal static string GetAccountId(string functionArn)
...
+            {
+                new(AWSLambdaSemanticConventions.AttributeFaasTrigger, trigger),
+            };
+
+            if (context == null)
+            {
+                return tags;
+            }
+
+            tags.Add(new(AWSLambdaSemanticConventions.AttributeFaasName, context.FunctionName));
```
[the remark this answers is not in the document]
Oberon00: I guess it's fine to be redundant here, but just asking you to confirm you have seen this comment and think so too?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -133,15 +215,92 @@ public static async Task Trace<TInput>(TracerProvider tracerProvider, Func<TInpu
...
         }
 
-        private static TResult Intercept<TResult>(TracerProvider tracerProvider, Func<TResult> method, ILambdaContext context = null)
+        internal static Activity OnFunctionStart(
+            ILambdaContext context = null,
+            ActivityContext parentContext = default,
+            IEnumerable<KeyValuePair<string, object>> tags = null)
         {
-            var lambdaActivity = OnFunctionStart(context);
+            if (parentContext == default && !IgnoreAWSXRayPropagation)
```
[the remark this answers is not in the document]
Oberon00: I think this is useful independently of the bug. See also https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/408#discussion_r913663298

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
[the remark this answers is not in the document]
rypdal: I'm totally agree that's it's redundant.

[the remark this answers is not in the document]
Oberon00: OK, but do you agree that that's OK in this case? Or do you think it should be removed from one of these locations?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// AWS lambda instrumentation options.
+    /// </summary>
+    public class AWSLambdaInstrumentationOptions
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether AWS X-Ray propagation should be ignored.
+        /// </summary>
+        public bool IgnoreAWSXRayPropagation { get; set; }
```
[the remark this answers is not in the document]
Oberon00: How about DisableAwsXRayContextInjection? That makes it clear what mechanism is used, and the "Aws" is a bit descriptive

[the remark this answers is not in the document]
rypdal: But judging by usage it's more disabling of extraction than injection: https://github.com/rypdal/opentelemetry-dotnet-contrib/blob/issue/AWS-Lambda-wrapper-and-parent-context/src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs#L193

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Initially I thought about adding optional tags list to Trace methods and merging them with these tags. That's I called them "Default" (means if optional are not provided then these will be added). But optional tags are is not needed anymore. I'll remove "Default".

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Yes, if we accept empty header as above then we even can do just `return header?.Split(',');` without `if`.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
+            // force flush before function quit in case of Lambda freeze.
+            tracerProvider.ForceFlush();
+        }
+
+        private static void OnException(Activity activity, Exception exception)
+        {
+            if (activity != null)
             {
-                OnFunctionStop(lambdaActivity, tracerProvider);
+                if (activity.IsAllDataRequested)
```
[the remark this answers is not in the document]
rypdal: This check was initially present: https://github.com/rypdal/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs#L270

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
```diff
@@ -29,11 +31,19 @@ public static class TracerProviderBuilderExtensions
...
+            this TracerProviderBuilder builder,
+            Action<AWSLambdaInstrumentationOptions> configure = null)
         {
             Guard.ThrowIfNull(builder);
 
+            var options = new AWSLambdaInstrumentationOptions();
+            configure?.Invoke(options);
+
+            AWSLambdaWrapper.IgnoreAWSXRayPropagation = options.IgnoreAWSXRayPropagation;
```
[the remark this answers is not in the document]
rypdal: 1. This is part of initialisation and not expected to be called multiple times.
2. Of course making wrapper non-static would add more flexibility and code safety, but it goes beyond initial scope of the PR. Lets keep it simple for now as it's just a wrapper around methods calls.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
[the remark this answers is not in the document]
rypdal: I would rather see it among other tags and remove it from resources. But we can't remove it from resources as it brings breaking change. Lets keep it redundant.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
rypdal: Now it's pushed.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Now, the 8th part (version) of arn and everything after that is removed. I added corresponding doc references in the method removing parts.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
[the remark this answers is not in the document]
rypdal: Renamed to `DisableAwsXRayContextExtraction`.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
Oberon00: Could still clean that up a bit, as the code around here is completely rewritten anyway. But OK for me to leave it as-is.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +151,35 @@ internal static string GetAccountId(string functionArn)
             return null;
         }
 
+        private static string GetFaasId(string functionArn)
+        {
+            var faasId = functionArn;
+
+            // According to faas.id description https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#all-triggers
+            // the 8th part of arn (version, see https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#versioning-versions-using)
```
Oberon00, line 159: The 8th part may also be an alias, which is actually the main reason for removing it. Maybe this link would be better since it links to both the Lambda version and alias docs while also providing a short description of the ARN format: https://docs.aws.amazon.com/lambda/latest/dg/lambda-api-permissions-ref.html

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -100,6 +151,35 @@ internal static string GetAccountId(string functionArn)
...
+            // should not be included into faas.id
+            var items = functionArn.Split(':');
+            if (items.Length >= 8)
+            {
+                var result = new string[7];
+                Array.Copy(items, result, 7);
+                faasId = string.Join(":", result);
```
Oberon00, line 166: This looks like C-code... Isn't there a cool C# 8slicing feature we could use to have less code and maybe even elide the copy? Would `string.Join(":", items[..7])` work?

rypdal: It requires at least NET Standard 2.1, but it still can be simplified with LINQ expression: `string.Join(":", items.Take(7));`.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
rypdal: I agree. It's useful in any case.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -112,5 +191,28 @@ private static ActivityContext ParseXRayTraceHeader(string rawHeader)
...
+            return null;
+        }
+
+        private static IEnumerable<string> GetHeaderValues(APIGatewayHttpApiV2ProxyRequest request, string name)
+        {
+            if (request.Headers != null &&
+                request.Headers.TryGetValue(name, out var header))
+            {
+                // Multiple values for the same header will be separated by a comma.
+                return header.Split(',');
```
Oberon00, line 212:
```suggestion
                return header?.Split(',');
```

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
Oberon00: Doesn't OTel take a dependency a backport of the Span features anyway?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
Oberon00: See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/408#discussion_r932954117

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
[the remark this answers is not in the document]
Oberon00: > This is part of initialisation and not expected to be called multiple times.

I wonder about that. Quite a few instrumentations seem to have that expectation, but does opentelemetry-dotnet document that you can only initialize each instrumentation once? E.g. you might be tempted to have separate tracer providers with different span processors etc., and initialize and instrumentation with each of them.

You are right that this is outside the scope probably, just wondering.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
rypdal: Yes, but unfortunately this is a part of public API.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
rypdal: I prefer to minimise the difference and do not change code not related to the scope of this PR.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Thanks. I used this link.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Applied together with other suggestions.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
+            TracerProvider tracerProvider,
+            Func<TInput, ILambdaContext, TResult> lambdaHandler,
+            TInput input,
+            ILambdaContext context,
+            ActivityContext parentContext = default)
         {
-            Intercept(tracerProvider, () => lambdaHandler(input));
+            TResult result = default;
+            Action action = () => result = lambdaHandler(input, context);
+            TraceInternal(tracerProvider, action, input, context, parentContext);
```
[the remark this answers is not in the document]
rypdal: I would prefer the method without returning value. Looks a bit redundant.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
rypdal: I wouldn't do it as it makes code less clear.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
-
-        private static TResult Intercept<TResult>(TracerProvider tracerProvider, Func<TResult> method, ILambdaContext context = null)
-        {
-            var lambdaActivity = OnFunctionStart(context);
-            try
+            if (parentContext == default)
             {
-                return method();
+                parentContext = AWSLambdaUtils.ExtractParentContext(input);
+                if (parentContext == default && !IgnoreAWSXRayPropagation)
```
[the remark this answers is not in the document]
rypdal: Could you clarify why ?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
         /// <param name="tracerProvider">TracerProvider passed in.</param>
         /// <param name="lambdaHandler">Lambda handler function passed in.</param>
         /// <param name="input">Instance of input.</param>
-        /// <returns>Task.</returns>
-        public static async Task Trace<TInput>(TracerProvider tracerProvider, Func<TInput, Task> lambdaHandler, TInput input)
+        /// <param name="context">Instance of lambda context.</param>
+        /// <param name="parentContext">
+        /// The optional parent context <see cref="ActivityContext"/> is used for Activity object creation.
+        /// If no parent context provided, incoming request is used to extract one.
+        /// If parent is not extracted from incoming request then X-Ray propagation is used to extract one.
```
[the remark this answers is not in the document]
rypdal: - I added statement about disabled X-Ray propagation. 
- We can document it later if users will get confused.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,216 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
-        public static async Task Trace<TInput>(TracerProvider tracerProvider, Func<TInput, ILambdaContext, Task> lambdaHandler, TInput input, ILambdaContext context)
+        public static async Task Trace(
+            TracerProvider tracerProvider,
+            Func<ILambdaContext, Task> lambdaHandler,
+            ILambdaContext context,
+            ActivityContext parentContext = default)
         {
-            await Intercept(tracerProvider, () => lambdaHandler(input, context), context);
+            Func<Task> action = async () => await lambdaHandler(context);
+            await TraceInternalAsync<object>(tracerProvider, action, null, context, parentContext);
```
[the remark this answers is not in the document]
rypdal: It's totally fine in this case. I applied this change along with other suggestions.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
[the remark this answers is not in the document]
Oberon00: Sorry, it seems I confused the function calls. I thought ExtractParentContext would use X-Ray. I think renaming GetParentContext to GetXRayParentContext could help.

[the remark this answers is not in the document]
rypdal: Good idea. Renamed the method.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
Oberon00: I think the name `action` doesn't make anything more clear. But OK for me to leave it as-is.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,219 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
-        /// <typeparam name="TInput">Input.</typeparam>
         /// <param name="tracerProvider">TracerProvider passed in.</param>
         /// <param name="lambdaHandler">Lambda handler function passed in.</param>
-        /// <param name="input">Instance of input.</param>
         /// <param name="context">Instance of lambda context.</param>
+        /// <param name="parentContext">
+        /// The optional parent context <see cref="ActivityContext"/> is used for Activity object creation.
+        /// If no parent context provided, incoming request is used to extract one.
+        /// If parent is not extracted from incoming request then X-Ray propagation is used to extract one
+        /// unless X-Ray propagation is disabled in the configuration for this wrapper.
```
Oberon00, line 178:
```suggestion
        /// unless X-Ray propagation is disabled in the configuration.
```

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -27,162 +27,219 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation
...
-                OnException(lambdaActivity, ex);
 
-                throw;
-            }
-            finally
-            {
-                OnFunctionStop(lambdaActivity, tracerProvider);
-            }
+            var tags = AWSLambdaUtils.GetFunctionTags(input, context);
+            var activityName = AWSLambdaUtils.GetFunctionName(context) ?? "<Unknown>";
```
Oberon00, line 203:
```suggestion
            var activityName = AWSLambdaUtils.GetFunctionName(context) ?? "AWS Lambda Invoke";
```

(although that should hopefully be never happen in practice)

Oberon00 (approved):
Thank you for addressing all my comments!

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
rypdal: Not the name `action` itself, but splitting call into 2 lines instead of in-lining.

Oberon00:
Hello! From my and @rypdal's perspective, this PR would be ready for merging.

utpilla:
Please add a CHANGELOG entry and we can merge this.

utpilla:
@rypdal Please follow the steps mentioned here if you want to request a package release: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#how-to-request-for-release-of-package

rypdal:
Thanks to @Oberon00 for the review !

PR #409, opened 2022-06-08
Fix some bugs in Runtime metrics

### LINKED ISSUE #383: NewRelic metric errors (monotonic values must be positive)

xiang17:
Fixes #383.

## Changes

- For `gc.fragmentation.ratio`, use ObservableGauge and change value from percentage to fraction in range of [0, 1]
- Rename `alloc.rate` to `gc.allocated.bytes`
- Updates units to use Bytes for data volume; use nanoseconds in Int64 rather than seconds or milliseconds in Double
- `gc.committed`: use ObservableGauge
- Update all value types from Int32 to Int64


src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -65,34 +65,34 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
             {
                 this.meter.CreateObservableCounter($"{metricPrefix}il.bytes.jitted", () => System.Runtime.JitInfo.GetCompiledILBytes(), "By", description: "IL Bytes Jitted");
                 this.meter.CreateObservableCounter($"{metricPrefix}methods.jitted.count", () => System.Runtime.JitInfo.GetCompiledMethodCount(), description: "Number of Methods Jitted");
-                this.meter.CreateObservableGauge($"{metricPrefix}time.in.jit", () => System.Runtime.JitInfo.GetCompilationTime().TotalMilliseconds, "ms", description: "Time spent in JIT");
+                this.meter.CreateObservableGauge($"{metricPrefix}time.in.jit", () => System.Runtime.JitInfo.GetCompilationTime().Ticks * 100, "ns", description: "Time spent in JIT");
```
reyang, line 68: Instead of using a hard-coded `100`, consider leveraging something like https://docs.microsoft.com/dotnet/api/system.timespan.tickspermillisecond.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -65,34 +65,34 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
...
-                this.meter.CreateObservableCounter($"{metricPrefix}threadpool.thread.count", () => ThreadPool.ThreadCount, description: "ThreadPool Thread Count");
+                this.meter.CreateObservableGauge($"{metricPrefix}threadpool.thread.count", () => (long)ThreadPool.ThreadCount, description: "ThreadPool Thread Count");
                 this.meter.CreateObservableGauge($"{metricPrefix}threadpool.completed.items.count", () => ThreadPool.CompletedWorkItemCount, description: "ThreadPool Completed Work Item Count");
-                this.meter.CreateObservableCounter($"{metricPrefix}threadpool.queue.length", () => ThreadPool.PendingWorkItemCount, description: "ThreadPool Queue Length");
-                this.meter.CreateObservableCounter($"{metricPrefix}active.timer.count", () => Timer.ActiveCount, description: "Number of Active Timers");
+                this.meter.CreateObservableGauge($"{metricPrefix}threadpool.queue.length", () => ThreadPool.PendingWorkItemCount, description: "ThreadPool Queue Length");
+                this.meter.CreateObservableGauge($"{metricPrefix}active.timer.count", () => Timer.ActiveCount, description: "Number of Active Timers");
```
reyang, line 79: For things that should be using `ObservableUpDownCounter`, consider leaving a TODO comment here so we can revisit once .NET 7 is shipped.

reyang (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I create a `const` as it's the definition of ticks at https://docs.microsoft.com/dotnet/api/system.timespan.ticks#remarks and I didn't find no such field in the library.

reyang: Sounds good, thanks!

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I think almost all of the `ObservableGauge` should be `ObservableUpDownCounter` except for CPU utilization and fragmentation.ratio. I've put a TODO comment in the beginning as a reminder.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -65,34 +68,34 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
...
+                this.meter.CreateObservableGauge($"{metricPrefix}time.in.jit", () => System.Runtime.JitInfo.GetCompilationTime().Ticks * NanosecondsPerTick, "ns", description: "Time spent in JIT");
             }
 #endif
 
 #if NETCOREAPP3_1_OR_GREATER
             if (options.IsThreadingEnabled)
             {
                 this.meter.CreateObservableGauge($"{metricPrefix}monitor.lock.contention.count", () => Monitor.LockContentionCount, description: "Monitor Lock Contention Count");
-                this.meter.CreateObservableCounter($"{metricPrefix}threadpool.thread.count", () => ThreadPool.ThreadCount, description: "ThreadPool Thread Count");
+                this.meter.CreateObservableGauge($"{metricPrefix}threadpool.thread.count", () => (long)ThreadPool.ThreadCount, description: "ThreadPool Thread Count");
```
alanwest, line 79: Just checking if the plan is to just use a gauge for now... that is, I think this should be an `ObservableUpDownCounter`, but understood we gotta wait for DiagnosticSource 7.

xiang17: Yes. I'm using `ObservableGauge` before `ObservableUpDownCounter` is available. I've put a TODO comment in the beginning of this method. Most of the gauge should be `ObservableUpDownCounter`.

alanwest: Ah ha, missed the TODO. Also just saw the comment here https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/404/files#r891813678.

alanwest (approved):

PR #412, opened 2022-06-08
Add GC heap size and count in Runtime metrics

### LINKED ISSUE #335: Bringing .NET Runtime metrics instrumentation to Stable

### LINKED ISSUE #383: NewRelic metric errors (monotonic values must be positive)

xiang17:
Fixes #335.

## Changes

Add `gc.heapsize` and change `gc.count` to multi-dimensional metrics


src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -102,6 +104,17 @@ public void Dispose()
             this.meter?.Dispose();
         }
 
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionCounts()
+        {
+            Measurement<long>[] measurements = new Measurement<long>[NumberOfGenerations];
+            for (int i = 0; i < measurements.Length; i++)
+            {
+                measurements[i] = new Measurement<long>(GC.CollectionCount(i), new KeyValuePair<string, object>("gen", HeapNames[i]));
```
cijothomas, line 112: "gen" might be too "generic" ?

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -110,6 +123,20 @@ private static double GetFragmentation()
         }
 #endif
 
+#if NET6_0_OR_GREATER
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionHeapSizes()
+        {
+            var generationInfo = GC.GetGCMemoryInfo().GenerationInfo;
```
cijothomas, line 129: we need some assert that the length is no more than 5, else HeapNames will outofboundexception

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * Updated OTel SDK package version to 1.3.0
   ([#411](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/411))
+* Add GC heap size and count in Runtime metrics
```
cijothomas, line 7: gc count was already there. this is a breaking change (which is okay), so need to describe this a bit more clearly.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I thought to use `i < measurements.Length && i < HeapNames.Length` as for loop termination check, or use `i < HeapName.Length ? HeapNames[i] : $"Gen{i}"` as generation name, but it feels like waste of checks **if** we can assume the garbage collection algorithm does not change.

Do you have any suggestion? Or a simple `Assert(generationInfo.Length <= HeapNames.Length, "There should not be more than 5 generations");` will do?

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
xiang17: Reworded it as "Add GC heap size and refactor GC count as multi-dimensional metrics in Runtime metrics"

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -102,6 +104,17 @@ public void Dispose()
             this.meter?.Dispose();
         }
 
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionCounts()
+        {
+            Measurement<long>[] measurements = new Measurement<long>[NumberOfGenerations];
```
utpilla, line 109: This would cause allocation on each `Collect` cycle. We could use a dedicated static array in the class to store these measurements and reuse in each `Collect` cycle.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I've added an additional check to work if the size goes beyond 5.

cijothomas: > the garbage collection algorithm does not change.

^ Can you leverage this, and build readonly array in the ctor?

xiang17: Done. I've moved them to a readonly array.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I've moved it to static field. It's only allocated once during initialization.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -110,14 +127,29 @@ private static double GetFragmentation()
...
+#if NET6_0_OR_GREATER
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionHeapSizes()
         {
-            var process = Process.GetCurrentProcess();
-            return new[]
+            var generationInfo = GC.GetGCMemoryInfo().GenerationInfo;
+
+            // TODO: Confirm 1. that the number of heaps for garbage collection will not change, at least during the lifetime of a process.
+            // and 2. that there will not be more than 5 heaps, at least for the existing .NET version that is supported (net6.0).
+            Debug.Assert(generationInfo.Length <= HeapNames.Length, "There should not be more than 5 heaps");
```
reyang, line 137:
```suggestion
            Debug.Assert(generationInfo.Length <= HeapNames.Length, "There should not be more than 5 generations");
```

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -33,6 +33,13 @@ internal class RuntimeMetrics : IDisposable
         internal static readonly AssemblyName AssemblyName = typeof(RuntimeMetrics).Assembly.GetName();
         internal static readonly string InstrumentationName = AssemblyName.Name;
         internal static readonly string InstrumentationVersion = AssemblyName.Version.ToString();
+        private static readonly string[] HeapNames = new string[] { "gen0", "gen1", "gen2", "loh", "poh" };
+        private static readonly int NumberOfGenerations = 3;
+        private static readonly Measurement<long>[] GarbageCollectionCountMeasurements = new Measurement<long>[NumberOfGenerations];
+#if NET6_0_OR_GREATER
+        private static readonly Measurement<long>[] HeapSizeMeasurements = new Measurement<long>[GC.GetGCMemoryInfo().GenerationInfo.Length];
```
reyang, line 40: Why do we need this variable? Is it thread safe?

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/412/files#r893986212

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
reyang: Would this work better?

https://github.com/open-telemetry/opentelemetry-dotnet/blob/c07917913cbe617ae05957f7023d5574ad7bf5d5/docs/metrics/learning-more-instruments/Program.cs#L54-L60

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I thought saying "generation" here might be confusing as there are 3 generations (generation 0, generation 1, generation 2), while there are 5 segments of heaps: generation 0, generation 1, generation 2, large object heap (which is sometimes referred to as generation 3) and [Pinned Object Heap](https://devblogs.microsoft.com/dotnet/internals-of-the-poh/): https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -47,16 +49,16 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
...
+                this.meter.CreateObservableGauge($"{metricPrefix}gc.count", () => GetGarbageCollectionCounts(), description: "GC Count for all generations");
+
 #if NETCOREAPP3_1_OR_GREATER
                 this.meter.CreateObservableCounter($"{metricPrefix}alloc.rate", () => GC.GetTotalAllocatedBytes(), "By", "Allocation Rate");
                 this.meter.CreateObservableCounter($"{metricPrefix}gc.fragmentation", GetFragmentation, description: "GC Fragmentation");
 #endif
 
 #if NET6_0_OR_GREATER
                 this.meter.CreateObservableCounter($"{metricPrefix}gc.committed", () => (double)(GC.GetGCMemoryInfo().TotalCommittedBytes / 1_000_000), "Mi", description: "GC Committed Bytes");
+                this.meter.CreateObservableGauge($"{metricPrefix}gc.heapsize", () => GetGarbageCollectionHeapSizes(), "By", "Heap Size for all generations");
```
reyang, line 61:
```suggestion
                this.meter.CreateObservableGauge($"{metricPrefix}gc.heapsize", () => GetGarbageCollectionHeapSizes(), "By", "Heap size for all generations");
```

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -82,7 +84,7 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
 
             if (options.IsProcessEnabled)
             {
-                this.meter.CreateObservableCounter("process.cpu.time", this.GetProcessorTimes, "s", "Processor time of this process");
+                this.meter.CreateObservableCounter("process.cpu.time", GetProcessorTimes, "ns", "Processor time of this process");
```
reyang, line 87: I think this should be `s` instead of `ns`?

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I've fixed it.

I've refactored the array allocations with `yield return`. It's working well for two of the functions. 

The compile will fail with GetGarbageCollectionHeapSizes because it needs a `Span<T>`: "CS4013 Instance of type 'ReadOnlySpan<GCGenerationInfo>' cannot be used inside a nested function, query expression, iterator block or async method".

For this one I just allocated a plain array.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: Yes, I fixed it in the other PR. This PR was mainly about GC until I came to "optimize" the array allocation.

xiang17: Yes, it should be `s`.

Both the [system metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemcpu---processor-metrics) and [process metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#process) use `s` as unit for CPU time.

So does [Go](https://github.com/open-telemetry/opentelemetry-specification/pull/2392#issuecomment-1058546909).

I'll modify it back to `s`.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: This metrics is an `Observable` and gets called once every 15~20 seconds. So it's not worth doing too much optimization for now. Most importantly, if there are multiple readers reading it at almost the same time, they could have incorrect values.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -102,6 +104,14 @@ public void Dispose()
             this.meter?.Dispose();
         }
 
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionCounts()
+        {
+            for (int i = 0; i < NumberOfGenerations; ++i)
+            {
+                yield return new(GC.CollectionCount(i), new KeyValuePair<string, object>("gen", HeapNames[i]));
```
reyang, line 111: The naming is a bit weird to me.
I suggest we keep them consistent, e.g. either use "gen" and "GenNames"; or "heap" and "HeapNames" (or "values") .

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -47,16 +49,16 @@ public RuntimeMetrics(RuntimeMetricsOptions options)
...
+                this.meter.CreateObservableGauge($"{metricPrefix}gc.count", () => GetGarbageCollectionCounts(), description: "GC Count for all generations");
+
 #if NETCOREAPP3_1_OR_GREATER
                 this.meter.CreateObservableCounter($"{metricPrefix}alloc.rate", () => GC.GetTotalAllocatedBytes(), "By", "Allocation Rate");
                 this.meter.CreateObservableCounter($"{metricPrefix}gc.fragmentation", GetFragmentation, description: "GC Fragmentation");
 #endif
 
 #if NET6_0_OR_GREATER
                 this.meter.CreateObservableCounter($"{metricPrefix}gc.committed", () => (double)(GC.GetGCMemoryInfo().TotalCommittedBytes / 1_000_000), "Mi", description: "GC Committed Bytes");
+                this.meter.CreateObservableGauge($"{metricPrefix}gc.heapsize", () => GetGarbageCollectionHeapSizes(), "By", "Heap size for all generations");
```
reyang, line 61: Do we put `.` at the end of the description text or not?

reyang (approved):
LGTM with some nonblocking clarification questions / suggestions.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: It is indeed weird. I don't have a position on it since I'm confused about what to name it myself AND because it's a private variable. Using the words in a blog about [Internals of the POH](https://devblogs.microsoft.com/dotnet/internals-of-the-poh/) in .NET Blog:

> It's a bit unfortunate that we overloaded the word "heap" here but it's been this way since V1.0. When I started it was already too late to make changing the name worthwhile. But fortunately AFAICS it didn't seem that confusing for our customers.

https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals

According to article [Fundamentals of garbage collection](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals) plus the blog above, there are 3 generations of garbage collection and 5 portions of the managed heap. I've listed them here.

| Name for the garbage collection generation | Name for a portion of the managed heap                             |
|--------------------------------------------|--------------------------------------------------------------------|
| Generation 0                               | Generation 0                                                       |
| Generation 1                               | Generation 1                                                       |
| Generation 2                               | Generation 2                                                       |
|                                            | Large object heap (which is sometimes referred to as generation 3) |
|                                            | Pinned object heap                                                 |

Speaking of which I've realized that I never saw that Pinned Object Heap is referred to as generation 4, nor anyone mentioned `GC.GetGCMemoryInfo().GenerationInfo[4].SizeAfterBytes` describes Pinned Object Heap. I'll need to confirm on that too.

xiang17: I've renamed it to "GenNames" just to keep them seem consistent.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I've added it, since the example of a Prometheus metric exposition has it https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-format-example

However, I will go over the description texts again after this PR is merged, because there are some inconsistences in upper/lower case and word choices.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -113,14 +123,27 @@ private static double GetFragmentation()
...
 
-        private IEnumerable<Measurement<long>> GetProcessorTimes()
+#if NET6_0_OR_GREATER
+        private static IEnumerable<Measurement<long>> GetGarbageCollectionHeapSizes()
         {
-            var process = Process.GetCurrentProcess();
-            return new[]
+            var generationInfo = GC.GetGCMemoryInfo().GenerationInfo;
+
+            Measurement<long>[] measurements = new Measurement<long>[generationInfo.Length];
```
utpilla, line 131: This will cause allocation. We could either use a pre allocated array or make use of the `yield return` semantic.

xiang17: This metrics is an Observable and gets called once every 15~20 seconds. So it's not worth doing too much optimization for now. Most importantly, if there are multiple readers reading it at almost the same time, they could have incorrect values.

I've tried to use `yield return` and it worked for two other methods, however, for this one I got an error. See discussion thread at https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/412#discussion_r894070227

> The compile will fail with GetGarbageCollectionHeapSizes because it needs a `Span<T>`: "CS4013 Instance of type 'ReadOnlySpan' cannot be used inside a nested function, query expression, iterator block or async method".

PR #416, opened 2022-06-10
Redis: Adding the ability to disable adding Events to Activities

stebet:
## Changes

Adds an option for OpenTelemetry.Instrumentation.StackExchangeRedis to disable adding the Redis command lifetime ActivityEvents to Activities. Added this because for example in Honeycomb, Span Events are billed and measured just like individual spans, so Activity Events can quickly gobble up telemetry budgets.


stebet:
Another option might be to convert these Events into Tags instead, in case users do not want a separate event for these but would still like the info to be part of the span.

src/OpenTelemetry.Instrumentation.StackExchangeRedis/StackExchangeRedisCallsInstrumentationOptions.cs
```diff
@@ -44,5 +44,10 @@ public class StackExchangeRedisCallsInstrumentationOptions
         /// <para><see cref="IProfiledCommand"/>: the profiled redis command from which additional information can be extracted to enrich the activity.</para>
         /// </remarks>
         public Action<Activity, IProfiledCommand> Enrich { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether or not the <see cref="StackExchangeRedisCallsInstrumentation"/> should enrich Activity with <see cref="ActivityEvent"/> entries about the Redis command processing/lifetime. Defaults to <see cref="bool">true</see>.
+        /// </summary>
+        public bool EnrichActivityWithEvents { get; set; } = true;
```
cijothomas, line 51: I am not sure why the Events were added in the 1st place. There is no spec requirement about adding Events (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#redis), so it may be better to turn this feature off completely. 

If a user wants to have events, `Enrich` could be used.

Will wait to see if anyone has opinions on this.
(Note that the package is non-stable, so we are allowed to have breaking changes (i.e disable events)).

cijothomas: @utpilla Can you add it to next SIG meeting agenda, and if no objections, we can remove the Events being populated in this instrumentation.

cijothomas: Thanks for additional clarification. I see the need for events. But without a spec/sem. conventions, this is going to be done differently in diff. implementations.
Let me check if the spec has an issue tracking this..

stebet: Events are a good fit i.m.o, but being able to turn them off would also be nice, since there is no way to remove events from Activities once added, so writing a processor doesn't work either.

cijothomas: For now, we can add this flag. The package is anyway non-stable, so we would be okay to make breaking changes, if required (due to spec). Ideally, semantic conventions would list the name of the events, so at that, we'll need to influence the spec or change this instrumentation.
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis

ejsmith:
This is exactly what I was about to create a PR for. Thank you @stebet! In my case, I'm using Elasticsearch APM and it is converting these activity events into log messages and flooding my app logs with literally millions and millions of useless log messages:

image

ejsmith:
It would be super appreciated if we could get this PR merged and a new nuget package pushed out because this behavior is killing my ES APM logs.

cijothomas:
we can merge once CI is green. @stebet could you fix the markdownlint issues?

ejsmith:
Awesome! Can we get a new release pushed out? Also, while we are at it, can we do a new release for the Elasticsearch package as well?

PR #421, opened 2022-06-13
[Hangfire OTEL] create parent context if traceparent is available

marcindruzgala:
## Changes

This change updates `OpenTelemetry.Instrumentation.Hangfire` package. 

Currently this instrumentation package will create new activity on the server side whenever job is being executed. But this trace is not attached to the trace from which the job was enqueued. This change provides that functionality - if `Activity.Current` is not null.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ## Unreleased
 
+* Added client side instrumentation for jobs ([#421](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/421))
+
 ## 1.0.0-beta.1
 
 * Updated OTel SDK package version to 1.2.0
   ([#353](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/353))
-* Added client side instrumentation for jobs ([#421](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/421))
 
 ## Initial Release
 


src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
```diff
@@ -30,8 +31,12 @@ public void OnPerforming(PerformingContext performingContext)
                 return;
             }
 
+            var traceParent = performingContext.GetJobParameter<string>(HangfireInstrumentationConstants.TraceParentKey);
+            var traceState = performingContext.GetJobParameter<string>(HangfireInstrumentationConstants.TraceStateKey);
+            ActivityContext.TryParse(traceParent, traceState, out var parentContext);
```
cijothomas, line 36: it is not recommended to do this by-hand, as there is a feature to achieve context propagation/extraction : https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Api#context-propagation

cijothomas (changes requested):
Recommend to use the Propagators API for injecting and extracting context, instead of doing it manually.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
marcindruzgala: Hey! Thanks for the comment, I started implementing the change and it was quite easy to do injection of activity context in `OnCreating` method.

```c#
ActivityContext contextToInject = default;
if (Activity.Current is not null)
{
    contextToInject = Activity.Current.Context;
}

Propagator.Inject(new PropagationContext(contextToInject, Baggage.Current), creatingContext.Parameters, InjectActivityProperties);

private static void InjectActivityProperties(IDictionary<string, object> jobParams, string key, string value)
{
    jobParams[key] = value;
}
```

The problem is that in `OnPerforming` the `PerformingContext` is not directly exposing `Parameters` dictionary. Instead if you want to get any of the params you have to call `GetJobParameter<T>` which will hit the underlying store (db, redis...whatever else). 

So I'm not really sure how to continue? Any ideas?

marcindruzgala: I tried using `Fields` from propagator instance to enumerate over them and get all params from db but that didn't work..

Screenshot 2022-06-13 at 19 38 32

marcindruzgala: Lastly..it raises another question. If it actually has to go and get the param from the db...won't that be inefficient? Let's say we have 10 params in future, that's 10 separate calls to the database 🤔

fred2u: Maybe could you use a Dictionary<string,string> for the context and put it in a single JobParameter ?

marcindruzgala: > Maybe could you use a Dictionary<string,string> for the context and put it in a single JobParameter ?

@fred2u What a simple and great idea :D I don't know why I didn't think of it. 

I updated code with your suggestion, please take a look. @cijothomas

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
```diff
@@ -65,5 +79,38 @@ public void OnPerformed(PerformedContext performedContext)
                 activity.Dispose();
             }
         }
+
+        public void OnCreating(CreatingContext creatingContext)
+        {
+            // Short-circuit if nobody is listening
+            if (!HangfireInstrumentation.ActivitySource.HasListeners() || Activity.Current?.Id is null)
```
cijothomas, line 86: why checking if Activity.Current is null here? The same check seems repeated few lines below as well.

marcindruzgala: Good catch, fixed it

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
```diff
@@ -30,8 +36,16 @@ public void OnPerforming(PerformingContext performingContext)
                 return;
             }
 
+            var activityContextData = performingContext.GetJobParameter<Dictionary<string, string>>(HangfireInstrumentationConstants.ActivityContextKey);
+            ActivityContext parentContext = default;
+            if (activityContextData is not null)
+            {
+                var propagationContext = Propagator.Extract(new PropagationContext(default, Baggage.Current), activityContextData, ExtractActivityProperties);
```
alanwest, line 43: `Propagators.DefaultTextMapPropagator` can change a runtime. I don't think you want to hold on to it in a field on this class and instead always call `Propagators.DefaultTextMapPropagator.Extract` here. Likewise for where you're calling `Inject`.

marcindruzgala: done

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
```diff
@@ -30,8 +34,16 @@ public void OnPerforming(PerformingContext performingContext)
                 return;
             }
 
+            var activityContextData = performingContext.GetJobParameter<Dictionary<string, string>>(HangfireInstrumentationConstants.ActivityContextKey);
+            ActivityContext parentContext = default;
+            if (activityContextData is not null)
+            {
+                var propagationContext = Propagators.DefaultTextMapPropagator.Extract(new PropagationContext(default, Baggage.Current), activityContextData, ExtractActivityProperties);
```
alanwest, line 41: Not sure if this is what you intended, but instantiating the context with `Baggage.Current` may prevent baggage from being extracted.

See:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/34efdbf4488afcaa6ade2ae74e4633e10c199514/src/OpenTelemetry.Api/Context/Propagation/BaggagePropagator.cs#L43-L49

You might want to consider something like the following
https://github.com/open-telemetry/opentelemetry-dotnet/blob/34efdbf4488afcaa6ade2ae74e4633e10c199514/examples/MicroserviceExample/Utils/Messaging/MessageReceiver.cs#L58-L62

In the example above, we set Baggage.Current to the extracted baggage. However, if it was your intent to preserve any existing values that may be in baggage, you might need to manually add the extracted baggage to Baggage.Current. I'm not familiar with the flow of Hangfire jobs, so I don't know which of the two options would be more correct.

marcindruzgala: Nice catch! I had no idea...

Fixed it :)

alanwest (approved):
LGTM

utpilla:
@marcindruzgala Please fix the failing build.

utpilla:
@fred2u As the component owner for Hangfire instrumentation, could you please review this PR? It would be great if you could approve it. Thanks!

fred2u (approved):

utpilla:
> @marcindruzgala Please fix the failing build.

Please update the CHANGELOG as well.

src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Updated OTel SDK package version to 1.2.0
   ([#353](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/353))
+* Added client side instrumentation for jobs ([#421](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/421))
 
 ## Initial Release
```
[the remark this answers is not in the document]
marcindruzgala: Yes of course...I thought I put it there 😬

marcindruzgala:
> > @marcindruzgala Please fix the failing build.
> 
> Please update the CHANGELOG as well.

Done :)

utpilla:
@marcindruzgala Please create a PR to update the CHANGELOG if you want a newer package with these changes to be released: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#how-to-request-for-release-of-package

PR #427, opened 2022-06-14
[Exporter.Geneva] Update OTel SDK version

utpilla:
## Changes
- Update OTel package versions

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -22,7 +22,6 @@
       Please sort alphabetically.
       Refer to https://docs.microsoft.com/en-us/nuget/concepts/package-versioning for semver syntax.
     -->
-    <InMemoryExporterLatestStable>1.3.0</InMemoryExporterLatestStable>
     <MinVerPkgVer>[2.4.0,3.0)</MinVerPkgVer>
     <MicrosoftCodeAnalysisNetAnalyzersPkgVer>[6.0.0]</MicrosoftCodeAnalysisNetAnalyzersPkgVer>
     <MicrosoftCodeCoveragePkgVer>[16.7.1]</MicrosoftCodeCoveragePkgVer>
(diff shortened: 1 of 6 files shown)


reyang (approved):

TimothyMothra (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/OpenTelemetry.Exporter.Geneva.Tests.csproj
```diff
@@ -33,7 +33,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
-    <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="1.2.0-rc2" />
```
reyang, line 36: Not blocking this PR, could these be extracted to one place?

utpilla: Done

build/Common.props
```diff
@@ -22,17 +22,18 @@
       Please sort alphabetically.
       Refer to https://docs.microsoft.com/en-us/nuget/concepts/package-versioning for semver syntax.
     -->
+    <InMemoryExporterLatestStable>1.3.0</InMemoryExporterLatestStable>
```
reyang, line 25: Consider `OpenTelemetryExporterInMemoryPkgVer`, the semantic here is "the version of in-memory exporter used in this project", not necessarily "the latest stable version" (think about this, if in-memory exporter shipped a new stable version, the semantic would be broken here).

utpilla: Yeah, we would need to update this when a newer stable version is released, and the packages are ready to onboard to it but if we need the packages to use a newer pre-release version, then the semantic meaning would break.

reyang (approved):

PR #442, opened 2022-06-20
[Extensions.PersistentStorage.Abstractions] Update PersistentStorage Abstractions workflow

utpilla:
Fixes the publish package workflow for PersistentStorage.Abstractions

## Changes
- Remove the test step from the workflow as there is no test project for this component.


.github/workflows/package-Extensions.PersistentStorage.Abstractions.yml
```diff
@@ -27,14 +27,11 @@ jobs:
         fetch-depth: 0 # fetching all
 
     - name: Install dependencies
-      run: dotnet restore
+      run: dotnet restore src/${{env.PROJECT}}
```
reyang, line 30: Consider having `cd "src/${{env.PROJECT}}"` extracted as a single command before running other steps?

reyang (approved):

PR #456, opened 2022-06-24
[WIP] Use bitwise combination for metrics options

xiang17:
Exploring option improvement after discussion here https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/431#discussion_r900354914

Still working in progress, but requesting for suggestions on whether it seems to be the right direction, and on naming.

## Changes

Use bitwise combination for metrics options.

Pro:
* More granular control on every single metrics.

Con:
* A ton of options to be maintained. Not much to worry about if using default config, which is enabling all flags.
* Old issue still remains: it might be counterintuitive to have `.AddRuntimeMetrics()` as enabling all options as default, and after adding an option, say `.AddRuntimeMetrics(options => { options.GcEnabled = true; })`, all options other than `GcEnabled` will be disabled.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/Options/GcMetricOptions.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/Options/GcMetricOptions.cs
@@ -1,0 +1,74 @@
+// <copyright file="GcMetricOptions.cs" company="OpenTelemetry Authors">
+// Copyright The OpenTelemetry Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+
+namespace OpenTelemetry.Instrumentation.Runtime.Options
+{
+    /// <summary>
+    /// Enum for garbage collection metrics.
+    /// </summary>
+    [Flags]
+    public enum GcMetricOptions : int
+    {
+        /// <summary>
+        /// Do not set flag for any metrics.
+        /// </summary>
+        None = 0x00,
+
+        /// <summary>
+        /// Set flag for all options.
+        /// </summary>
+        All = Count
+#if NETCOREAPP3_1_OR_GREATER
+            | AllocatedBytes
+#endif
+
+#if NET6_0_OR_GREATER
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.Instrumentation.Runtime/Options/GcMetricOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        All = Count
+#if NETCOREAPP3_1_OR_GREATER
+            | AllocatedBytes
+#endif
+
+#if NET6_0_OR_GREATER
+            | CommitedBytes | HeapSize | FragmentationSize
+#endif
+            ,
```
xiang17, line 43: Warning due to space before comma. Removing the space doesn't work either. Might have to define this enum option for each frameworks with `#if.. #elseif...#else...#endif`.

xiang17: Can't use "else if" because when it's `NET6_0_OR_GREATER`, `NETCOREAPP3_1_OR_GREATER` has to be enabled as well. Suppressed it with `#pragma warning disable SA1001`.

reyang:
This seems over complicated, especially considering some metrics' availability is dependent on the runtime version.

reyang (changes requested):
I wonder if we can simply start without any options - e.g. `AddRuntimeInstrumentation()` will just give you everything that's available to the current runtime/platform. Later if we want to introduce options, we can, and that's not a breaking change.

xiang17:
> I wonder if we can simply start without any options - e.g. `AddRuntimeInstrumentation()` will just give you everything that's available to the current runtime/platform. Later if we want to introduce options, we can, and that's not a breaking change.

Do you mean remove all the options?

In the current implementation, that's the default behavior if you don't specify any options: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime#step-2-enable-runtime-instrumentation-at-application-startup

reyang:
> > I wonder if we can simply start without any options - e.g. `AddRuntimeInstrumentation()` will just give you everything that's available to the current runtime/platform. Later if we want to introduce options, we can, and that's not a breaking change.
> 
> Do you mean remove all the options?
> 
> In the current implementation, that's the default behavior if you don't specify any options: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime#step-2-enable-runtime-instrumentation-at-application-startup

I asked this because I saw `ActiveTimerCount` under the threading options. I don't know if Timer has a direct relation to threading despite the fact that .NET grouped it under the System.Threading namespace.

Overall I think it could take time to design these options and it requires a good understanding of usage scenario. I feel a good starting point is to have nothing configurable.

cijothomas:
Agree to have no configurability in the 1st release. Adding the turn off flag later is not a breaking change.

PR #457, opened 2022-06-25
Adjust the semantic of gc count to make it exclusive

reyang:
Based on discussion with @noahfalk.

If you call:
```csharp
GC.Collect(0);
GC.Collect(1);
GC.Collect(2);
```

The current implementation would give:
```
process_runtime_dotnet_gc_collections_count{gen="gen2"} 1
process_runtime_dotnet_gc_collections_count{gen="gen1"} 2
process_runtime_dotnet_gc_collections_count{gen="gen0"} 3
```

This PR would change it to:
```
process_runtime_dotnet_gc_collections_count{gen="gen2"} 1
process_runtime_dotnet_gc_collections_count{gen="gen1"} 1
process_runtime_dotnet_gc_collections_count{gen="gen0"} 1
```

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -140,9 +140,15 @@ namespace OpenTelemetry.Instrumentation.Runtime
 
         private static IEnumerable<Measurement<long>> GetGarbageCollectionCounts()
         {
-            for (int i = 0; i < NumberOfGenerations; ++i)
+            long collectionsFromHigherGeneration = 0;
+
+            for (int gen = NumberOfGenerations - 1; gen >= 0; --gen)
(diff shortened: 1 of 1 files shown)


cijothomas (approved):
Please add a changelog entry as this is significant for enduser interpretting the values.

reyang:
> Please add a changelog entry as this is significant for enduser interpretting the values.

I think we've also dramatically renamed the metrics. I would suggest having a single changelog entry saying "major redesign of the runtime instrumentation" once all the major changes are done.

cijothomas:
> > Please add a changelog entry as this is significant for enduser interpretting the values.
> 
> I think we've also dramatically renamed the metrics. I would suggest having a single changelog entry saying "major redesign of the runtime instrumentation" once all the major changes are done.

I think we could go more verbose on the Instrument Description, to callout that the counts does not include the lower gen counts?

PR #462, opened 2022-06-27
refactor runtime instrumentation

reyang:
Rename `RuntimeMetricsOptions` to `RuntimeInstrumentOptions` - it doesn't make sense to have the "metrics" name in the options as the instrumentation library could have other signals (logs, traces).


src/OpenTelemetry.Instrumentation.Runtime/RuntimeInstrumentOptions.cs
```diff
@@ -19,7 +19,7 @@ namespace OpenTelemetry.Instrumentation.Runtime
     /// <summary>
     /// Options to define the runtime metrics.
     /// </summary>
-    public class RuntimeMetricsOptions
+    public class RuntimeInstrumentOptions
```
cijothomas, line 22: can we remove this whole class given we dont have anything configurable in v1?

PR #470, opened 2022-06-29
[Instrumentation.MySql] update CHANGELOG.md

chad122:
## Changes

Update the CHANGELOG to add a new version.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.MySqlData/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.MySqlData/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 1.0.0-beta.3
+
 * Fix incomplete db.statement when the length>300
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Instrumentation.MySqlData/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+## 1.0.0-beta.3
```
cijothomas, line 5: I'd suggest to put the date as well. Example : https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md

@utpilla Could we update the docs to say release request should put the date as well?

utpilla: This could be done in a follow-up PR.

utpilla (approved):

PR #475, opened 2022-06-29
[Instrumentation.Runtime] Add description for metrics in `README.md`

xiang17:
## Changes

Add description for metrics in `README.md`


src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,131 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
 Refer to [Program.cs](../../examples/runtime-instrumentation/Program.cs) for a
 complete demo.
 
+## Metrics
+
+### GC related metrics
+
+| Name                                            | Description                                                          | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------|----------------------------------------------------------------------|-----------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.collections.count** | Number of times garbage collection has occurred since process start. | `{times}` | ObservableCounter | `Int64`    | gen              | gen0, gen1, gen2 |
```
cijothomas, line 69: Do we need to specify that we internally use ObservableCounter? Which instrument we use should be implementation detail, so can be omitted.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,131 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+  of this application domain.
+
+### Exception counter metric
+
+| Name                                       | Description                                                                                                                                                                                                            | Units      | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|--------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**exception.count** | Count of exceptions that have been thrown in managed code, since the observation started. The value will be unavailable until an exception has been thrown after OpenTelemetry.Instrumentation.Runtime initialization. | `{timers}` | ObservableCounter | `Int64`    |                  |                  |
+
+The API called in this section is:
```
cijothomas, line 182: Since there is no API to get exception.count, i think this section is misleading. We are using this, and doing the count using Metrics Counter API here

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: I'm using it because all the semantic convention docs mention the instrument type, e.g. [JVM metrics](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics).

Plus, the Prometheus exporter does expose whether it's a counter or a gauge, so I think this information somewhat relevant:

image

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: I've added a note here as explanation.

src/OpenTelemetry.Instrumentation.Runtime/README.md
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-types spec notes.

Not a blocker, okay to keep it here.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
 Refer to [Program.cs](../../examples/runtime-instrumentation/Program.cs) for a
 complete demo.
 
+## Metrics
+
+### GC related metrics
```
cijothomas, line 65: I'd suggest to use heading for each of the metric, instead of the table. It'll be easier to link to individual metric that way. And we can write very long, details about the metric. (Description in this doc is not that important, as it is already part of metric descripption.) We could do more here, like "what value is considered typically okay, what is the action item if this value is high/low etc."

cijothomas: eventually "tutorials:" like this : https://docs.microsoft.com/en-us/dotnet/core/diagnostics/debug-threadpool-starvation

xiang17: I was more of thinking to make it easier for people who want to have a quick glance to know what to expect without having to dive into the code. I like your tutorial idea, and we can also include tools to use, etc. but it would be much more than the scope of this PR.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+|-------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**jit.il_compiled.size**       | Count of bytes of intermediate language that have been compiled since the process start. The value will be zero under ahead-of-time (AOT) compilation mode.                                                                                                                                                                                  | `bytes`     | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 123:
```suggestion
| process.runtime.dotnet.**jit.il_compiled.size**       | Count of bytes of intermediate language that have been compiled since the process start.                                                                                                                                                                                  | `bytes`     | ObservableCounter | `Int64`    |                  |                  |
```

AOT compilation doesn't guarantee to pre-compile everything so the counter probably won't be zero even in that case.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+| process.runtime.dotnet.**jit.methods_compiled.count** | The number of times the JIT compiler compiled a method since the process start. The JIT compiler may be invoked multiple times for the same method to compile with different generic parameters, or because tiered compilation requested different optimization settings. The value will be zero under ahead-of-time (AOT) compilation mode. | `{methods}` | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 124:
```suggestion
| process.runtime.dotnet.**jit.methods_compiled.count** | The number of times the JIT compiler compiled a method since the process start. The JIT compiler may be invoked multiple times for the same method to compile with different generic parameters, or because tiered compilation requested different optimization settings. | `{methods}` | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+| process.runtime.dotnet.**jit.compilation_time**       | The amount of time the JIT compiler has spent compiling methods since the process start. The value will be zero under ahead-of-time (AOT) compilation mode.                                                                                                                                                                                  | `ns`        | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 125:
```suggestion
| process.runtime.dotnet.**jit.compilation_time**       | The amount of time the JIT compiler has spent compiling methods since the process start.                                                                                                                                                                               | `ns`        | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
 Refer to [Program.cs](../../examples/runtime-instrumentation/Program.cs) for a
 complete demo.
 
+## Metrics
+
+### GC related metrics
+
+| Name                                            | Description                                                          | Units     | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------|----------------------------------------------------------------------|-----------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.collections.count** | Number of times garbage collection has occurred since process start. | `{times}` | ObservableCounter | `Int64`    | gen              | gen0, gen1, gen2 |
```
noahfalk, line 69:
```suggestion
| process.runtime.dotnet.**gc.collections.count** | Number of garbage collections that have occurred since process start. | `{collections}` | ObservableCounter | `Int64`    | generation              | gen0, gen1, gen2 |
```

I notice some units have curly braces around them and others don't, is there a meaning to that?

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+| process.runtime.dotnet.**gc.committed_memory.size**   | The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. Committed virtual memory may be larger than the heap size because it includes both memory for storing existing objects (the heap size) and some extra memory that is ready to handle newly allocated objects in the future. The value will be unavailable until garbage collection has occurred. | `bytes` | ObservableGauge | `Int64`    |                  |                            |
+| process.runtime.dotnet.**gc.heap.size**               | The heap size (including fragmentation), as observed during the latest garbage collection. The value will be unavailable until garbage collection has occurred.                                                                                                                                                                                                                                                    | `bytes` | ObservableGauge | `Int64`    | gen              | gen0, gen1, gen2, loh, poh |
```
noahfalk, line 99:
```suggestion
| process.runtime.dotnet.**gc.heap.size**               | The heap size (including fragmentation), as observed during the latest garbage collection. The value will be unavailable until garbage collection has occurred.                                                                                                                                                                                                                                                    | `bytes` | ObservableGauge | `Int64`    | generation              | gen0, gen1, gen2, loh, poh |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+| process.runtime.dotnet.**gc.heap.fragmentation.size** | The heap fragmentation, as observed during the latest garbage collection. The value will be unavailable until garbage collection has occurred.                                                                                                                                                                                                                                                                     | `bytes` | ObservableGauge | `Int64`    | gen              | gen0, gen1, gen2, loh, poh |
```
noahfalk, line 100:
```suggestion
| process.runtime.dotnet.**gc.heap.fragmentation.size** | The heap fragmentation, as observed during the latest garbage collection. The value will be unavailable until garbage collection has occurred.                                                                                                                                                                                                                                                                     | `bytes` | ObservableGauge | `Int64`    | generation              | gen0, gen1, gen2, loh, poh |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+|--------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**monitor.lock_contention.count**     | The number of times there was contention when trying to acquire a monitor lock since the process start. Monitor locks are commonly acquired by using the lock keyword in C#, or by calling Monitor.Enter() and Monitor.TryEnter().                                                  | `{times}`   | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 145:
```suggestion
| process.runtime.dotnet.**monitor.lock_contention.count**     | The number of times there was contention when trying to acquire a monitor lock since the process start. Monitor locks are commonly acquired by using the lock keyword in C#, or by calling Monitor.Enter() and Monitor.TryEnter().                                                  | `{contentions}`   | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+  Gets the number of work items that are currently queued to be processed.
+* [Timer.ActiveCount](https://docs.microsoft.com/dotnet/api/system.threading.timer.activecount?view=netcore-3.1):
+  Gets the number of timers that are currently active. An active timer is registered
+  to tick at some point in the future, and has not yet been canceled.
+
+### Assemblies related metrics
+
+| Name                                      | Description                                              | Units          | Instrument Type | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------|----------------------------------------------------------|----------------|-----------------|------------|------------------|------------------|
+| process.runtime.dotnet.**assembly.count** | The number of .NET assemblies that are currently loaded. | `{assemblies}` | ObservableGauge | `Int64`    |                  |                  |
```
noahfalk, line 170:
```suggestion
| process.runtime.dotnet.**assemblies.count** | The number of .NET assemblies that are currently loaded. | `{assemblies}` | ObservableGauge | `Int64`    |                  |                  |
```

I think the pattern elsewhere is that these counts are pluralized?

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+  of this application domain.
+
+### Exception counter metric
+
+| Name                                       | Description                                                                                                                                                                                                            | Units          | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|--------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**exception.count** | Count of exceptions that have been thrown in managed code, since the observation started. The value will be unavailable until an exception has been thrown after OpenTelemetry.Instrumentation.Runtime initialization. | `{exceptions}` | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 182:
```suggestion
| process.runtime.dotnet.**exceptions.count** | Count of exceptions that have been thrown in managed code, since the observation started. The value will be unavailable until an exception has been thrown after OpenTelemetry.Instrumentation.Runtime initialization. | `{exceptions}` | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+  The number of times garbage collection has occurred for the specified generation
+of objects.
+
+#### Additional GC metrics only available when targeting .NET Core 3.1 or later
+
+| Name                                           | Description                                                                                                                                                                                             | Units   | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.allocations.size** | Count of the bytes allocated on the managed GC heap since the process start. .NET objects are allocated from this heap. Object allocations from unmanaged languages such as C/C++ do not use this heap. | `bytes` | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 86:
```suggestion
| process.runtime.dotnet.**gc.allocations.size** | Count of bytes allocated on the managed GC heap since the process start. .NET objects are allocated from this heap. Object allocations from unmanaged languages such as C/C++ do not use this heap. | `bytes` | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Yes. It is the semantic convention for instruments that measure an integer count of something: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Indeed! As is recommended by the semantic convention: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#pluralization

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+|--------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**monitor.lock_contention.count**     | The number of times there was contention when trying to acquire a monitor lock since the process start. Monitor locks are commonly acquired by using the lock keyword in C#, or by calling Monitor.Enter() and Monitor.TryEnter().                                                  | `{contentions}`   | ObservableCounter | `Int64`    |                  |                  |
```
xiang17, line 145: I'll also update this to `lock_conventions.count`

```suggestion
| process.runtime.dotnet.**monitor.lock_contentions.count**     | The number of times there was contention when trying to acquire a monitor lock since the process start. Monitor locks are commonly acquired by using the lock keyword in C#, or by calling Monitor.Enter() and Monitor.TryEnter().                                                  | `{contentions}`   | ObservableCounter | `Int64`    |                  |                  |
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: The word "contention" here seem to be an uncountable noun. I don't think using plural form would be good.

1. Meaning no. 1 in [this dictionary](https://www.oxfordlearnersdictionaries.com/us/definition/american_english/contention#:~:text=2%5Bcountable%5D%20contention%20(that,I%20would%20reject%20that%20contention.)
2. The official documentation didn't add "a" in front of the word: "the number of times there was **contention**".
3. The word actually has a second meaning in the dictionary above. And it is countable.

xiang17: The semantic convention [for instrumentation units](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units) doesn't have recommendation on what unit to use for **uncountable** nouns.

I think "times" is not perfect but it's okayish. Any suggestion on what to use?

xiang17: I think technically we couldn't say a "count" of an uncountable noun (a water, five water), but we could say a countable unit of the uncountable noun (five bottles of water with "{bottles}" as the unit, or two drops of rain, or a ray of sunlight, or two times of contention with "{times}" as the unit). There might be a better choice than "times" but I hope someone who's better at the English language can help answer it.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+|--------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**monitor.lock_contention.count**     | The number of times there was contention when trying to acquire a monitor lock since the process start. Monitor locks are commonly acquired by using the lock keyword in C#, or by calling Monitor.Enter() and Monitor.TryEnter().                                                  | `{times}`   | ObservableCounter | `Int64`    |                  |                  |
```
noahfalk, line 145: you are right that I was playing a little loose by pluralizing it. How about "contended acquisitions"?

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+| process.runtime.dotnet.**thread_pool.queue.length**          | The number of work items that are currently queued to be processed by the thread pool.                                                                                                                                                                                              | `{items}`   | ObservableGauge   | `Int64`    |                  |                  |
+| process.runtime.dotnet.**timer.count**                       | The number of timer instances that are currently active. Timers can be created by many sources such as System.Threading.Timer, Task.Delay, or the timeout in a CancellationSource. An active timer is registered to tick at some point in the future and has not yet been canceled. | `{timers}`  | ObservableGauge   | `Int64`    |                  |                  |
+
+The APIs used to retrieve the values are:
+
+* [Monitor.LockContentionCount](https://docs.microsoft.com/dotnet/api/system.threading.monitor.lockcontentioncount?view=netcore-3.1):
```
twenzel, line 153: Was it intentional to link the .NET Core 3.1 version instead of the current NET 6.0?

twenzel: Maybe we should omit the version/view parameter, so that always the latest version will be shown.

xiang17: It was intentional, but I'll remove the `?view=*` part, as showing the latest makes more sense.

I intentionally picked to link to .NET Core 3.1 as it is the earliest supported platform. (as shown in "These metrics are only available when targeting .NET Core 3.1 or later.") Some APIs are only available in .NET 6 and later and I picked the link for .NET 6 for them as well. On second thought, removing it also makes sense and that way it's the reader's responsibility to mind which versions are supported.

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Updated to `{contended_acquisitions}`.

twenzel (approved):

noahfalk (approved):
LGTM!

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+  of this application domain.
+
+### Exception counter metric
+
+| Name                                        | Description                                                                                                                                                                                                            | Units          | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|---------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**exceptions.count** | Count of exceptions that have been thrown in managed code, since the observation started. The value will be unavailable until an exception has been thrown after OpenTelemetry.Instrumentation.Runtime initialization. | `{exceptions}` | ObservableCounter | `Int64`    |                  |                  |
```
cijothomas, line 182: Is this using ObservableCounter?

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+
+Note: The value is tracked by calling a counter whenever an AppDomain.FirstChanceException
```
cijothomas, line 184: incrementing a counter

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
...
+event occurs. The observation starts when the Runtime instrumentation library is
+initialized, so the value will be unavailable until the an exception has been
```
cijothomas, line 186:
```suggestion
initialized, so the value will be unavailable until an exception has been
```

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -60,6 +60,138 @@ using var meterProvider = Sdk.CreateMeterProviderBuilder()
 Refer to [Program.cs](../../examples/runtime-instrumentation/Program.cs) for a
 complete demo.
 
+## Metrics
+
+### GC related metrics
+
+| Name                                            | Description                                                           | Units           | Instrument Type   | Value Type | Attribute Key(s) | Attribute Values |
+|-------------------------------------------------|-----------------------------------------------------------------------|-----------------|-------------------|------------|------------------|------------------|
+| process.runtime.dotnet.**gc.collections.count** | Number of garbage collections that have occurred since process start. | `{collections}` | ObservableCounter | `Int64`    | generation       | gen0, gen1, gen2 |
```
xiang17, line 69: yes

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Yes. Updated.

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Updated to `{contended_acquisitions}`. (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/475#discussion_r915166232)

cijothomas (approved):

PR #495, opened 2022-07-08
[Instrumentation.Runtime] Change API for GC Heap Size

xiang17:
## Changes

Use internal method `GC.GetGenerationSize` for GC Heap Size.

.NET 6.0 has a bug in computing those generation sizes for the `GC.GetGCMemoryInfo()` API and it was fixed in .NET 7.

Note that the heap sizes are not retrieved in one transactional query with `GC.GetGCMemoryInfo()`. It'd be better to change it back to `GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes` for .NET 7 when appropriate.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/README.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/README.md
@@ -114,7 +114,7 @@ The APIs used to retrieve the values are:
   For .NET 6, heap size is retrieved with an internal method `GC.GetGenerationSize`,
   which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
   retrieve the values.
-  See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
+  See source code [here](https://github.com/dotnet/runtime/blob/b4dd16b4418de9b3af08ae85f0f3653e55dc420a/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
 
 * [GCGenerationInfo.FragmentationAfterBytes Property](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo.fragmentationafterbytes)
   Gets the fragmentation in bytes on exit from the reported collection.
--- a/src/OpenTelemetry.Instrumentation.Runtime/README.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/README.md
@@ -110,7 +110,7 @@ The APIs used to retrieve the values are:
 
 * Heap size is retrieved with an internal method `GC.GetGenerationSize`,
 which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
-retrieves the values.
+retrieve the values.
 See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
 
 * [GCGenerationInfo.FragmentationAfterBytes Property](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo.fragmentationafterbytes)
--- a/src/OpenTelemetry.Instrumentation.Runtime/README.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/README.md
@@ -108,10 +108,13 @@ The APIs used to retrieve the values are:
 * [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes):
   Gets the total committed bytes of the managed heap.
 
-* Heap size is retrieved with an internal method `GC.GetGenerationSize`,
-which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
-retrieve the values.
-See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
+* [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
+  Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+  Note that this API on .NET 6 has a [bug](https://github.com/dotnet/runtime/pull/60309).
+  For .NET 6, heap size is retrieved with an internal method `GC.GetGenerationSize`,
+  which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
+  retrieve the values.
+  See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
 
 * [GCGenerationInfo.FragmentationAfterBytes Property](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo.fragmentationafterbytes)
   Gets the fragmentation in bytes on exit from the reported collection.
--- a/src/OpenTelemetry.Instrumentation.Runtime/README.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/README.md
@@ -108,8 +108,10 @@ The APIs used to retrieve the values are:
 * [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes):
   Gets the total committed bytes of the managed heap.
 
-* [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
-  Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+* Heap size is retrieved with an internal method `GC.GetGenerationSize`,
+which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
+retrieves the values.
+See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
 
 * [GCGenerationInfo.FragmentationAfterBytes Property](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo.fragmentationafterbytes)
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -108,8 +108,10 @@ The APIs used to retrieve the values are:
 * [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes):
   Gets the total committed bytes of the managed heap.
 
-* [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
-  Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+* Heap size is retrieved with an internal method `GC.GetGenerationSize`,
+which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
+retrieves the values.
```
utpilla, line 113: retrieve*

xiang17: Fixed.

utpilla (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -76,6 +76,9 @@ static RuntimeMetrics()
                 unit: "bytes",
                 description: "The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. Committed virtual memory may be larger than the heap size because it includes both memory for storing existing objects (the heap size) and some extra memory that is ready to handle newly allocated objects in the future. The value will be unavailable until garbage collection has occurred.");
 
+            MethodInfo mi = typeof(GC).GetMethod("GetGenerationSize", BindingFlags.NonPublic | BindingFlags.Static);
+            Func<int, ulong> getGenerationSize = mi.CreateDelegate<Func<int, ulong>>();
```
noahfalk, line 80: as a defensive precaution it might be useful to avoid creating the gauge if the GetMethod call or the CreateDelegate call returns null. I don't know of any situation where that would happen but its good to be safe.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -76,6 +76,9 @@ static RuntimeMetrics()
                 unit: "bytes",
                 description: "The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. Committed virtual memory may be larger than the heap size because it includes both memory for storing existing objects (the heap size) and some extra memory that is ready to handle newly allocated objects in the future. The value will be unavailable until garbage collection has occurred.");
```
noahfalk, line 78: Can we add a comment and an issue tracking restoring the SizeAfterBytes code once this runs on .NET 7? Private reflection is fine for .NET 6 where we know it works, but it would be unfortunate to forget about it. Its possible in some future version of .NET we'll remove or rename the internal API and we wouldn't this code to suddenly break.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -108,8 +108,10 @@ The APIs used to retrieve the values are:
 * [GCMemoryInfo.TotalCommittedBytes](https://docs.microsoft.com/dotnet/api/system.gcmemoryinfo.totalcommittedbytes):
   Gets the total committed bytes of the managed heap.
 
-* [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
-  Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+* Heap size is retrieved with an internal method `GC.GetGenerationSize`,
```
noahfalk, line 111: I would add to this comment that we are doing this explicitly to workaround a known issue in .NET 6. Here is the link to the PR that fixed it: https://github.com/dotnet/runtime/pull/60309

noahfalk (approved):
Aside from the comments inline, LGTM : )

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I've added comment and an issue #496.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -76,28 +76,51 @@ static RuntimeMetrics()
...
+                    getGenerationSize = mi.CreateDelegate<Func<int, ulong>>();
+                }
+            }
 
-                    var generationInfo = GC.GetGCMemoryInfo().GenerationInfo;
-                    Measurement<long>[] measurements = new Measurement<long>[generationInfo.Length];
-                    int maxSupportedLength = Math.Min(generationInfo.Length, GenNames.Length);
-                    for (int i = 0; i < maxSupportedLength; ++i)
+            // Either Environment.Version is not 6 or (it's 6 but internal API GC.GetGenerationSize is valid)
+            if (!isCodeRunningOnBuggyRuntimeVersion || getGenerationSize != null)
```
cijothomas, line 92: when would this condition be not met?

xiang17: On .NET 6 and the reflection failed. It is just [a defensive precaution](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/495#discussion_r917194228), and .NET team might remove it in some version https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/495#discussion_r917194674.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: @noahfalk Will this internal API ever be removed/renamed in future .NET 6 minor/patch versions, or just possibly in .NET 7 and future versions? If it's the later, then I can safely assume it's always available until we support .NET 7.

noahfalk: I think its safe to assume that API will never go away in .NET 6 servicing versions.

src/OpenTelemetry.Instrumentation.Runtime/README.md
```diff
@@ -110,6 +110,11 @@ The APIs used to retrieve the values are:
 
 * [GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes](https://docs.microsoft.com/dotnet/api/system.gcgenerationinfo):
   Represents the size in bytes of a generation on exit of the GC reported in GCMemoryInfo.
+  Note that this API on .NET 6 has a [bug](https://github.com/dotnet/runtime/pull/60309).
+  For .NET 6, heap size is retrieved with an internal method `GC.GetGenerationSize`,
+  which is how the [well-known EventCounters](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
+  retrieve the values.
+  See source code [here](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114).
```
pellared, line 117: Can we change the hyperlink in a way to make it stable? E.g. https://github.com/dotnet/runtime/blob/b4dd16b4418de9b3af08ae85f0f3653e55dc420a/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L110-L114

Kielek (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -76,28 +76,51 @@ static RuntimeMetrics()
...
+            {
+                // TODO: change to ObservableUpDownCounter
+                MeterInstance.CreateObservableGauge(
+                    $"{metricPrefix}gc.heap.size",
```
pellared, line 96: I looked at `RuntimeMetricsTests.cs` and I have a question it would not be good to test the exported metrics

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -76,28 +76,51 @@ static RuntimeMetrics()
                 unit: "bytes",
                 description: "The amount of committed virtual memory for the managed GC heap, as observed during the latest garbage collection. Committed virtual memory may be larger than the heap size because it includes both memory for storing existing objects (the heap size) and some extra memory that is ready to handle newly allocated objects in the future. The value will be unavailable until at least one garbage collection has occurred.");
 
-            // TODO: change to ObservableUpDownCounter
-            MeterInstance.CreateObservableGauge(
-                $"{metricPrefix}gc.heap.size",
-                () =>
+            // GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes is better but it has a bug in .NET 6. See context in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/496
+            Func<int, ulong> getGenerationSize = null;
+            bool isCodeRunningOnBuggyRuntimeVersion = Environment.Version.Major == 6;
```
reyang, line 81: I understand that this code is guarded by `#if NET6_0_OR_GREATER`, I wonder if we should change `== 6` to `<= 6` (although both would give the exact same results) just to express the intention.

reyang (approved):

cijothomas:
Please add changelog entry

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: My intention was to "check to avoid the buggy API on certain Runtime version" (only 6 and forward has this new API and only 6 has bug, thus the variable name), but not to "use `GC.GetGenerationSize` whenever possible".

In fact, if I go with the intention of the latter, I'd think whether we could add the metric for more Runtime versions whenever `GC.GetGenerationSize` is available. However, I don't think it's a good idea, because it would be a change of the spec and this API is not an official API.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: I didn't quite understand the sentence. Are you suggesting to add unit tests? I think it'd be a good idea since a lot were deleted in the test file but none was added. I can do it in a follow up PR.

src/OpenTelemetry.Instrumentation.Runtime/README.md
xiang17: Done.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: Added in here #510.

xiang17:
> Please add changelog entry

Added.

cijothomas (approved):

Issue #496, opened 2022-07-09
Use `GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes` for .NET 7


xiang17:
# Issue with OpenTelemetry.Instrumentation.Runtime

* [x] Use `GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes` starting from .NET 7: Done in #495.
* [x] Add back `gc.heap.fragmentation.size` metrics starting from .NET 7: Done in #524.

See context in #495.

The values for `GC.GetGCMemoryInfo().GenerationInfo[i].SizeAfterBytes` keep increasing because of a bug https://github.com/dotnet/runtime/pull/60309.

xiang17:
Done. Updated inline with issue number for each item.

PR #509, opened 2022-07-15
[Instrumentation.Runtime] Remove `gc.heap.fragmentation.size` metrics due to buggy API on .NET 6

xiang17:
Removing this metrics because of the same issue #496. Unfortunately there is no convenient alternative API for this metric.

* [x] Design discussion issue #496


src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -99,28 +99,27 @@ static RuntimeMetrics()
...
-                    for (int i = 0; i < maxSupportedLength; ++i)
-                    {
-                        measurements[i] = new(generationInfo[i].FragmentationAfterBytes, new KeyValuePair<string, object>("generation", GenNames[i]));
-                    }
-
-                    return measurements;
-                },
-                unit: "bytes",
-                description: "The heap fragmentation, as observed during the latest garbage collection. The value will be unavailable until at least one garbage collection has occurred.");
+            // TODO: Uncomment this for .NET 7 where the bug is fixed. See context in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/496
```
cijothomas, line 102: Maybe put this under "#if NET7_0_OR_GREATER", so it'll just work when we add .NET 7 target?

cijothomas (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
xiang17: Done.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -99,6 +99,8 @@ static RuntimeMetrics()
                 unit: "bytes",
                 description: "The heap size (including fragmentation), as observed during the latest garbage collection. The value will be unavailable until at least one garbage collection has occurred.");
 
+#if NET7_0_OR_GREATER
```
utpilla, line 102: Could we make this a non-nested `#if`?

utpilla (approved):

cijothomas (approved):
Could we add net7.0 target to the tests project and validate this as well?

xiang17:
> Could we add net7.0 target to the tests project and validate this as well?

I validated locally for .NET 7.0. However both the runtime library and the test project have to have `net7.0` as target. If the runtime library doesn't have `net7.0`, the `src\OpenTelemetry.Instrumentation.Runtime\bin\Debug` folder won't have binaries for `net7.0`, and `#if NET7_0_OR_GREATER` is never true in any compiled binaries.

xiang17:
> > > Could we add net7.0 target to the tests project and validate this as well?
> > 
> > 
> > I validated locally for .NET 7.0. However both the runtime library and the test project have to have `net7.0` as target. If the runtime library doesn't have `net7.0`, the `src\OpenTelemetry.Instrumentation.Runtime\bin\Debug` folder won't have binaries for `net7.0`, and `#if NET7_0_OR_GREATER` is never true in any compiled binaries.
> 
> Then its best to not rely on compiler preprocessor directives. You should switch to using Environment.Version to detect .NET 6 vs .NET 7, so that this can be used in .NET 7 apps right now.

Added it back in #524.

PR #510, opened 2022-07-15
[Instrumentation.Runtime] Add unit tests

xiang17:
## Changes

Add unit tests


utpilla (approved):

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +38,90 @@ public void RuntimeMetricsAreCaptured()
             meterProvider.ForceFlush(MaxTimeToAllowForFlush);
             Assert.True(exportedItems.Count > 1);
             Assert.StartsWith(MetricPrefix, exportedItems[0].Name);
+
+            var gcCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.gc.collections.count");
```
cijothomas, line 42: I'd suggest to split the tests rather than having an single test covering entire instrumentations. That way, if things fail, its easy to tell from the name of the test which part is failing.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +38,90 @@ public void RuntimeMetricsAreCaptured()
...
+            Assert.True(GetLongSum(threadCountMetric) > 0);
+
+            var completedItemsCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.completed_items.count");
+            Assert.True(GetLongSum(completedItemsCountMetric) > 0);
+
+            var queueLengthMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.queue.length");
+            Assert.True(GetLongSum(queueLengthMetric) == 0);
+
+            var timerCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.timer.count");
+            Assert.True(GetLongSum(timerCountMetric) > 0);
```
cijothomas, line 79: we can assume that there is at least one timer? Maybe we create our own timers, and validate that atleast that many is reported by the metric itself?

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +38,90 @@ public void RuntimeMetricsAreCaptured()
...
+        public void RuntimeMetrics_GcOnlyAvailableAfterFirst()
+        {
+            var exportedItems = new List<Metric>();
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                 .AddRuntimeInstrumentation()
+                 .AddInMemoryExporter(exportedItems)
+                .Build();
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
```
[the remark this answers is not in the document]
xiang17: It is verified in the first unit test case at line 52, 53:

```
Assert.False(exportedItems.Exists(i => i.Name == "process.runtime.dotnet.gc.committed_memory.size"));
Assert.False(exportedItems.Exists(i => i.Name == "process.runtime.dotnet.gc.heap.size"));
```

In fact, this test is kind of shaky and the call `GC.Collect(1)` is unnecessary. There is an automatic garbage collection run between the first and second unit test case. It is out of my control. Before I call `GC.Collect(1)` in this unit test, these metric already exist.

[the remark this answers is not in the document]
xiang17: If runtime decides to make a garbage collection before any of the unit test case starts, these metrics will always exists during the lifetime of any unit tests. Unfortunately this is the case when I split up the unit tests.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
xiang17: I tried to split up the tests. However, in that case, a garbage collection would have occurred before any of the unit tests. I will not be able to test the case that "before garbage collection happens, certain metrics won't exists". Some context: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/510#discussion_r923963846

xiang17: I think I should skip testing that scenario, because it depends on Runtime very much. Even the current unit tests could fail if GC decides to run once before the unit tests.

cijothomas: There is no need to test for that scenario, as it is hard to simulate that and there is no value testing it.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
xiang17: Done.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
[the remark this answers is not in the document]
xiang17: Decided to not test this case: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/510#discussion_r923966607

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            var timerCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.timer.count");
+            Assert.True(GetLongSum(timerCountMetric) >= timerCount);
+            for (int i = 0; i < timers.Count; i++)
+            {
+                timers[i].Dispose();
+            }
+        }
+#endif
+
+        private static double GetLongSum(Metric metric)
```
pellared, line 152: 1. Why do we use `GetLongSum` for non-sum metrics?
2. I think it would be enough just to check for the existence of the metrics in most of the places e.g.

```csharp
var gcCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.assemblies.count");
Assert.NotNull(gcCountMetric);
```

pellared (approved):

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            var sumReceived = GetLongSum(gcCountMetric);
+            Assert.True(sumReceived >= 0);
+
+#if NETCOREAPP3_1_OR_GREATER
+            var gcAllocationSizeMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.gc.allocations.size");
+            Assert.True(GetLongSum(gcAllocationSizeMetric) > 0);
+#endif
+
+#if NET6_0_OR_GREATER
+            // Supposedly to pass if no garbage collection occurred before. However one occurred, which is out of control of the code.
```
cijothomas, line 53: we should remove this whole commented out section.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            // Assert.False(exportedItems.Exists(i => i.Name == "process.runtime.dotnet.gc.heap.size"));
+#endif
+
+            var assembliesCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.assemblies.count");
+            Assert.True(GetLongSum(assembliesCountMetric) > 0);
+        }
+
+#if NET6_0_OR_GREATER
+        [Fact]
+        public void RuntimeMetrics_GcAvailableAfterFirst()
```
cijothomas, line 64: nit: the class is already RuntimeMetrics - we can simpify the name to GcMetricsTest()/similar

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            var exportedItems = new List<Metric>();
+            using var meterProvider = Sdk.CreateMeterProviderBuilder()
+                 .AddRuntimeInstrumentation()
+                 .AddInMemoryExporter(exportedItems)
+                .Build();
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            var lockContentionCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.monitor.lock_contention.count");
+            Assert.True(GetLongSum(lockContentionCountMetric) >= 0);
```
cijothomas, line 120: how do we know for sure?

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+
+            var threadCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.threads.count");
+            Assert.True(GetLongSum(threadCountMetric) > 0);
+
+            var completedItemsCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.completed_items.count");
+            Assert.True(GetLongSum(completedItemsCountMetric) > 0);
+
+            var queueLengthMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.queue.length");
+            Assert.True(GetLongSum(queueLengthMetric) == 0);
```
cijothomas, line 129: same qn as above. how can we know for sure tha TP queue is 0?

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            }
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            var timerCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.timer.count");
+            Assert.True(GetLongSum(timerCountMetric) >= timerCount);
+            for (int i = 0; i < timers.Count; i++)
+            {
+                timers[i].Dispose();
+            }
```
cijothomas, line 148: dispose in finally block.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
cijothomas: +1. For places where we can "cause" the metric value to be something, we can assert there. Eg: if we have 10 timers, we know the time count should be at least 10.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
xiang17: Changed to only validate the existence of the metrics.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -37,6 +39,133 @@ public void RuntimeMetricsAreCaptured()
...
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            var lockContentionCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.monitor.lock_contention.count");
+            Assert.True(GetLongSum(lockContentionCountMetric) >= 0);
+
+            var threadCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.threads.count");
+            Assert.True(GetLongSum(threadCountMetric) > 0);
+
+            var completedItemsCountMetric = exportedItems.First(i => i.Name == "process.runtime.dotnet.thread_pool.completed_items.count");
+            Assert.True(GetLongSum(completedItemsCountMetric) > 0);
```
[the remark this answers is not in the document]
xiang17: Changed to only validate the existence of the metrics.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
xiang17: Updated to only check for existence.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -34,9 +37,150 @@ public void RuntimeMetricsAreCaptured()
...
+                .Build();
+
+            GC.Collect(1);
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            var gcCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.collections.count");
+            Assert.NotNull(gcCountMetric);
+
+#if NETCOREAPP3_1_OR_GREATER
```
cijothomas, line 78: this is already inside net6_0 or greater, right?

xiang17: That's right. I did a mistake by moving it and `gcCountMetric` to this method, which is all net6 or greater. I'll adjust the `NET6_0_OR_GREATER` to cover only the applicable metrics items.

xiang17: Done. I've also added a test for `gc.heap.fragmentation.size` on condition `if (Environment.Version.Major >= 7)`.

Kielek (approved):

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -34,9 +37,156 @@ public void RuntimeMetricsAreCaptured()
...
+            var gcCommittedMemorySizeMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.committed_memory.size");
+            Assert.NotNull(gcCommittedMemorySizeMetric);
+
+            var gcHeapSizeMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.heap.size");
+            Assert.NotNull(gcHeapSizeMetric);
+
+            if (Environment.Version.Major >= 7)
+            {
+                var gcHeapFragmentationSizeMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.heap.fragmentation.size");
+                Assert.NotNull(gcHeapFragmentationSizeMetric);
```
xiang17, line 92: I manually tested this with .NET 7.0 preview: installing Visual Studio 17.3.0 Preview 4.0 and adding `net7.0` in `TargetFrameworks` tag of OpenTelemetry.Instrumentation.Runtime.Tests.csproj file.

Because .NET 7.0 is still in Preview status, I'm not making the change for csproj and CI process in this PR.

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
```diff
@@ -34,9 +37,156 @@ public void RuntimeMetricsAreCaptured()
...
+
+            meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+            var lockContentionCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.monitor.lock_contention.count");
+            Assert.NotNull(lockContentionCountMetric);
+
+            var threadCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.thread_pool.threads.count");
+            Assert.NotNull(threadCountMetric);
+
+            var completedItemsCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.thread_pool.completed_items.count");
```
cijothomas, line 138: its possible to cause this metric to be atleast N, if you queue N items to TP, and wait for it to complete. 
Similar to  Exception.Count
Timer.Count etc.

Not blocker for the PR.

cijothomas (approved):

test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
xiang17: Addressed in #559.

PR #514, opened 2022-07-19
[ExporterGeneva]Remove ConvertToJson option

### LINKED ISSUE #395: [Exporter.Geneva] Remove `ConvertToJson` option
Remove `ConvertToJson` from `GenevaTraceExporter` and `GenevaLogExporter`.

### Problem
We want to validate the data types for `PrepopulatedFields`. As of today, if someone passes in a value with an unsupported data type, we serialize that into a very unhelpful [string](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L98).

Here's the design for one of the possible solutions:

### Solution
The best thing to do here would be to validate the data upfront when the user actually provides the `PrepopulatedFields` option. The best place to do that would be the [setter ](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L87) of the `PrepopulatedFields`. We have a set of types that we support as a valid data type for the values. These data types are mentioned as various cases in the `switch`. We need a `default` case which will be hit for any unsupported type. For the `default` case, we throw an `ArgumentException` with the message: `$"Type: {value.GetType()} is not supported. Only bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double, and string are the supported types for PrepopulatedFields values";` .

xyq175com:
### [ExporterGeneva]Remove ConvertToJson option
Fix issue with #395 
## Changes
 Remove ConvertToJson option


src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -303,9 +303,6 @@ internal int SerializeLogRecord(LogRecord logRecord)
                     case double vd:
                     case string vs:
                         break;
-                    default:
```
utpilla, line 306: I think we still need a `default` case. We should probably take the `.ToString()` value here.

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -303,9 +303,6 @@ internal int SerializeLogRecord(LogRecord logRecord)
...
-                        value = this.convertToJson(value);
-                        break;
                 }
```
utpilla, line 307: Remove other references of `convertToJson` from this file.

xyq175com: updated according to the comments

utpilla:
Please update the PR title to reflect the work being done.

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
utpilla: After thinking some more about this, I believe that we should move this validation check to the GenevaExporterOptions class in the setter of `PrepopulatedFields`. We can then get rid of the switch case construct here in the exporter classes.

utpilla:
Could you please update you branch with the latest changes from main branch?

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
reyang: What's the design here? E.g. what would happen if the user triggered this code path?

reyang (changes requested):
Need to understand the design.

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
utpilla: Here's the design for one of the possible solutions:

### Problem
We want to validate the data types for `PrepopulatedFields`. As of today, if someone passes in a value with an unsupported data type, we serialize that into a very unhelpful [string](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L98).

### Solution
The best thing to do here would be to validate the data upfront when the user actually provides the `PrepopulatedFields` option. The best place to do that would be the [setter ](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L87) of the `PrepopulatedFields`. We have a set of types that we support as a valid data type for the values. These data types are mentioned as various cases in the `switch`. We need a `default` case which will be hit for any unsupported type. For the `default` case, we throw an `ArgumentException` with the message: `$"Type: {value.GetType()} is not supported. Only bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double are the supported types for PrepopulatedFields values";` .

We can then get rid of the switch case checks in the individual exporters as we would not get any unsupported data type.

xyq175com: @utpilla Agree with the design/solution, would you mind create another issue, then we can track change from there?

utpilla: I don't think we need to create another issue.

@reyang I have updated the existing issue #395 with this [comment](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/514#discussion_r928041421) detailing the problem description and one of the possible solutions. Could you please check if that looks good? Thanks!

reyang: > Only bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double are the supported types

and string

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -88,6 +88,28 @@ public class GenevaExporterOptions
...
+                    case int vi32:
+                    case uint vui32:
+                    case long vi64:
+                    case ulong vui64:
+                    case float vf:
+                    case double vd:
+                    case string vs:
+                        break;
+                    default:
+                        val = val.ToString();
```
reyang, line 108: I suggest taking a different approach here by explicitly asking the user to do the string conversion. Similar to what @utpilla explained in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/395#issue-1262454425.

Main reason of not using `ToString` - for arbitrary types, this would introduce localization/globalization issue (e.g. `val` might be converted to different string depending on the user language/locale preference). We don't know if the user's intention is to have an InvariantCulture string or not, only the user knows what they want.

In a nutshell, I think the spirit here should be giving a string saying "hey, you've provided a non-primitive type which we do not support, you probably want to convert it to a string (invariant or local culture - you decide)?".

xyq175com: @reyang  Thank you so much! I have updated the code, can you help review again?

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -88,6 +88,25 @@ public class GenevaExporterOptions
...
+                    case int vi32:
+                    case uint vui32:
+                    case long vi64:
+                    case ulong vui64:
+                    case float vf:
+                    case double vd:
+                    case string vs:
+                        break;
+                    default:
+                        throw new ArgumentException($"Type: {entry.Value.GetType()} is not supported. Only bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double, and string are the supported types for PrepopulatedFields values.");
```
reyang, line 108: would this throw NullReferenceException if value is null?

reyang, line 108: "for PrepopulatedFields values" can be removed since this is self-explanatory.

reyang (approved):
LGTM with minor suggestions.

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -88,7 +88,27 @@ public class GenevaExporterOptions
             var copy = new Dictionary<string, object>(value.Count + 1) { [Schema.V40.PartA.Ver] = schemaVersion };
             foreach (var entry in value)
             {
-                copy[entry.Key] = entry.Value; // shallow copy
+                var val = entry.Value;
```
utpilla, line 91: You can delete the internal `Func` ConvertToJson from this file.

utpilla:
Add a unit test for this.

xyq175com:
> Add a unit test for this.

There is no existing Unit test for this file. Add unit test is Complex for me.

utpilla:
> > Add a unit test for this.
> 
> There is no existing Unit test for this file. I have to see that add unit test is hard for me.;), Can you help approve or give some suggestion for this, Thank you so much!

You're right! Currently, there is no dedicated test class for this file. However, you could still add a unit test in `GenevaTraceExporter` for validating this feature. You can refer to this existing unit test: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs#L39-L56 and pass in some unsupported type in `PrepopulatedFields` and ensure that it throws an exception. Add a case for happy path as well.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -53,6 +53,34 @@ public void GenevaTraceExporter_constructor_Invalid_Input()
                     ConnectionString = null,
                 });
             });
+            Assert.Throws<ArgumentNullException>(() =>
```
utpilla, line 56: What's the purpose of this test?

xyq175com: When i worked on the unit test, doing some test, for this i saw this is throwing the exception for invalid connect string as well, so leave it here. I have deleted this one, Thanks!

utpilla (approved):

cijothomas (approved):

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
reyang: Is this addressed?

reyang: Is this addressed?

reyang: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/537

reyang: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/537

Issue #394, opened 2022-06-06
[Exporter.Geneva] Add data validation for MetricExportIntervalMilliseconds


utpilla:
- `MetricExportIntervalMilliseconds` should be at least 1 second.

xyq175com:
Is this still in progress, if not i can help with this.

PR #527, opened 2022-07-21
[Exporter.Geneva] Add data validation for MetricExportIntervalMilliseconds

Yun-Ting:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/394.


src/OpenTelemetry.Exporter.Geneva/GenevaMetricExporterOptions.cs
```diff
@@ -34,7 +35,28 @@ public class GenevaMetricExporterOptions
...
+        get
+        {
+            return this._metricExporterIntervalMilliseconds;
+        }
+
+        set
+        {
+            if (value < 0)
+            {
+                throw new ArgumentException($"MetricExportIntervalMilliseconds: {value} should be a greater than 0 integer.");
```
Kielek, line 49: Consider to use `ArgumentOutOfRangeException`. It inherits from `ArgumentException` but it is more descriptive.

reyang: Follow https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/dcf0156af8697c0f4d95e4fdae8d3b4b778e4810/src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs#L37.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterOptionsTests.cs
```diff
@@ -69,5 +69,24 @@ public void InvalidPrepopulatedDimensions()
...
+                    MetricExportIntervalMilliseconds = 500,
+                };
+            });
+
+#if NET6_0 || NETCOREAPP3_1_OR_GREATER
+            var negativeInputExpectedErrorMessage = "Must be in the range: [1000, 2147483647] (Parameter 'value')\r\nActual value was 500.";
+#else
+            var negativeInputExpectedErrorMessage = "Must be in the range: [1000, 2147483647]\r\nParameter name: value\r\nActual value was 500.";
+#endif
+            Assert.Equal(negativeInputExpectedErrorMessage, negativeInputMetricExportIntervalMillisecondsException.Message);
```
reyang, line 89: Do we really want this? My gut feeling is that the actual error message is owned by `Guard`, and if we really want to validate it, it should be in `GuardTests.cs`.

BTW, the use of EOL (CRLF vs. LF) might break depending on the operating system.

Yun-Ting: So maybe only validating whether it's an ArgumentOutofRange sufficient here?
I am not sure whether we really need this test.

reyang (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterOptionsTests.cs
utpilla: I think we should definitely test that we throw an `ArgumentOutOfRangeException` when we provide a value less than 1000 ms.

@reyang How do we test that we have called the right `Guard` method with the right set of parameters? I think we could assert the message here as well to ensure that.

reyang: I think the following tests should be good enough:
1. use 999ms, the ctor should complain
2. use 1000ms, the ctor should succeed

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterOptionsTests.cs
```diff
@@ -69,5 +69,27 @@ public void InvalidPrepopulatedDimensions()
...
         }
+
+        [Fact]
+        public void MetricExportIntervalValidationTest()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
+                var exporterOptions = new GenevaMetricExporterOptions
+                {
+                    MetricExportIntervalMilliseconds = 500,
```
reyang, line 80: Minor - consider testing on the boundary:

```suggestion
                    MetricExportIntervalMilliseconds = 999,
```

utpilla (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterOptionsTests.cs
utpilla: @Yun-Ting I have committed this suggestion.

Kielek (approved):

cijothomas (approved):

PR #534, opened 2022-07-26
[Instrumentation.AWSLambda] Updated AWSLambda ActivitySource Name

samimusallam:
Fixes #484 
## Changes
Use the assembly name as the ActivitySource name

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #

PR #535, opened 2022-07-26
[Instrumentation.AWS] Update aws activitysourcename

### LINKED ISSUE #484: [Instrumentation] Standardize ActivitySource names for instrumentation projects
Use the assembly name as the `ActivitySourceName` in the instrumentation projects:

For e.g.: OpenTelemetry.Instrumentation.StackExchangeRedis

Currently, the following projects do not follow this convention:

1. StackExchangeRedis
2. EntityFrameworkCore
3. AWS
4. AWS.Lambda

samimusallam:
Fixes #484 

## Changes
Use the assembly name without ".Contrib" as the ActivitySource name


src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog - OpenTelemetry.Contrib.Instrumentation.AWS
 
+## Unreleased
+
+* Update the `ActivitySource` name to the assembly name:
```
cijothomas, line 5: might be a breaking change, since this package is already stable.

samimusallam: > might be a breaking change, since this package is already stable.

Does it make sense to remove the unreleased section?

samimusallam: Can I get an update on this?

utpilla: Since this is a breaking change, you should wait for the `2.0.x` release to make this change. This change shouldn't go into `1.x.x` versions.

samimusallam: There was a decision to remove ".Contrib" from all packages. Only AWS packages do not follow this change as mentioned in [this comment](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/534#discussion_r934135417)

Oberon00: This package was only ever released as a beta, so it is not "already stable", or am I missing something @cijothomas? 1.1.0-beta1 is also what I see at NuGet: https://www.nuget.org/packages/OpenTelemetry.Contrib.Instrumentation.AWSLambda/

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/Utils.cs
```diff
@@ -21,6 +21,9 @@ namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation
 {
     internal class Utils
     {
+        internal static string ActivitySourceName =
```
Oberon00, line 24: Looking at how this played out with Instrumentation.AwsLambda so far (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590#issuecomment-1219119832, #534) I think we should either:
* Rename the whole package instead of patching the activity source name (disruptive but consistent); or
* Leave everything as-is

Oberon00 (changes requested):

PR #538, opened 2022-07-28
[Exporter.Geneva] Simplify exporter options for PrepopulatedFields

utpilla:
## Changes
- Simplify exporter options for PrepopulatedFields
- Add a test case for supported types to not throw an exception


reyang (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -81,6 +81,32 @@ public void GenevaTraceExporter_constructor_Invalid_Input()
                     },
                 });
             });
+
+            // Supported types for PrepopulatedFields should not throw an exception
+            var exception = Record.Exception(() =>
+            {
+                new GenevaExporterOptions
```
utpilla, line 88: This is not using an exporter ctor anymore. The tests need refactoring. We need to add a dedicated test class for `GenevaExporterOptions`. I'll not cover them in this PR.

PR #546, opened 2022-07-29
[Exporter.Geneva] One-time optimizations for GenevaTraceExporter and GenevaLogExporter

utpilla:
## Changes

- Reuse the buffer created in the ctor
- Mark epilogue buffer static
- Remove prologue buffer from `GenevaTraceExporter`


src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -36,8 +36,8 @@ public class GenevaLogExporter : GenevaBaseExporter<LogRecord>
     private readonly string m_defaultEventName = "Log";
     private readonly IReadOnlyDictionary<string, object> m_prepopulatedFields;
     private readonly List<string> m_prepopulatedFieldKeys;
-    private static readonly ThreadLocal<byte[]> m_buffer = new ThreadLocal<byte[]>(() => null);
-    private readonly byte[] m_bufferEpilogue;
+    private static readonly ThreadLocal<byte[]> m_buffer = new ThreadLocal<byte[]>();
+    private static readonly byte[] m_bufferEpilogue = new byte[25]; // This size was determined by precomputing the number of bytes required to serialize the epilogue.
```
reyang, line 40: Given it's now static instead of member variable:

```suggestion
    private static readonly byte[] s_bufferEpilogue = new byte[25]; // This size was determined by precomputing the number of bytes required to serialize the epilogue.
```

reyang (approved):

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -132,10 +132,10 @@ public GenevaLogExporter(GenevaExporterOptions options)
             this.m_customFields = customFields;
         }
 
+        _ = MessagePackSerializer.Serialize(s_bufferEpilogue, 0, new Dictionary<string, object> { { "TimeFormat", "DateTime" } });
```
reyang, line 135: Why is this needed in ctor instead of cctor?

reyang (changes requested):
The logic in ctor which updates static variable could cause race-condition (e.g. if multiple ctor invocations happened simultaneously).

PR #334, opened 2022-04-29
Use OTEL 1.2.0

seanat280:
## Changes

Migrated all projects to OTEL 1.2.0.  As .Net 4.6.1 and lower are now past end of life, and OTEL has upped the lowest supported version of .Net, I migrated all projects to be at a minium of .Net4.6.2.  More details on the supported versions can be found here:
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

PR #580, opened 2022-08-12
[Instrumentation.AWS] Use OpenTelemetry 1.3.0, update AWS SDK and drop EOLed dotnet versions

seanat280:
## Changes

A follow to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/334 to break out the AWS specific changes.  Sorry about dropping the ball and taking so long on this, I was pulled away from this by other issues at work for a while.

Migrated all AWS projects to OTEL 1.3.0.  As .Net 4.6.1 and lower are now past end of life, and OTEL has upped the lowest supported version of .Net, I migrated all projects to be at a minium of .Net4.6.2.  More details on the supported versions can be found here:
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/


src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
```diff
@@ -1,18 +1,21 @@
...
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\OpenTelemetry.Internal\Guard.cs" Link="Includes\Guard.cs" />
-    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Contrib.Shared\Api\SemanticConventions.cs" Link="Includes\SemanticConventions.cs"/>
+    <Compile Include="$(RepoRoot)\src\OpenTelemetry.Contrib.Shared\Api\SemanticConventions.cs" Link="Includes\SemanticConventions.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\OpenTelemetry.Contrib.Extensions.AWSXRay\OpenTelemetry.Contrib.Extensions.AWSXRay.csproj" />
```
Kielek, line 19: I think that it is not safe change. Each of AWS* package is released independently based on the tags `Extensions.AWSXRay-`, `Instrumentation.AWSLambda-`, `Instrumentation.AWS-`.

I suppose that releasing process is pretty manual and requires at release AWSXRay and then PR + additional release for other libs.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/OpenTelemetry.Contrib.Instrumentation.AWSLambda.csproj
```diff
@@ -8,14 +8,17 @@
...
+    <PackageReference Include="Amazon.Lambda.Core" Version="2.1.0" />
+    <PackageReference Include="OpenTelemetry" Version="1.3.0" />
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\OpenTelemetry.Internal\Guard.cs" Link="Includes\Guard.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\OpenTelemetry.Contrib.Extensions.AWSXRay\OpenTelemetry.Contrib.Extensions.AWSXRay.csproj" />
```
Kielek, line 21: Same issue as for AWS.csproj

Kielek (commented):
I think that you need to extend CHANGELOG.md for all packages.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
```diff
@@ -1,18 +1,21 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
Oberon00, line 4: I think as long as we are not using any features from the new versions, we should not upgrade yet.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
Oberon00: Right. With this change, things will break unless all of the cross-referenced packages are always released at the same time, which is not something that is enforced now.

Also the Lambda package has a different component owner from the other packages, and it's likely we'll request a release soon, without wanting to (or even having the responsibility to) release the other packages.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
Kielek: OpenTelemetry 1.3.0 is no longer compatible with .NET Framework 4.5.2. The latest compatible version is 1.1.0.

Ref: https://www.nuget.org/packages/OpenTelemetry/1.3.0#supportedframeworks-body-tab

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
seanat280: Alright so I reverted that part of the PR.  It builds and test pass, but I was a little concerned because the XRay package referenced causes a transitive dependency to a lower version of OpenTelemetry.  As such, it would probably be worthwhile to the referenced version of XRay just as soon as it is released.  In the mean time though it doesn't seem to break anything at least.

Oberon00: In these cases, the higher version should be used, and if that breaks anything, it would be a bug, because 1.x versions are guaranteed to be backwards compatible.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
Oberon00: OK, but is there any advantage from updating OTel from 1.1 yet? Any new features used?

Kielek: I do not see any direct feature, but I see some benefits:
 - promote newer OTel release. In case of any issues, the user will be asked to verify how it is working with the newer version.
 - Drop support for not supported .NET (Framework) versions https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework
  
 BTW my previous comment should suggest that we take both changes or any of them.

Kielek (approved):

Oberon00 (approved):

Issue #196, opened 2022-01-28
Rename all projects to remove contrib word


cijothomas:
Example: OpenTelemetry.Contrib.Instrumentation.MassTransit --> OpenTelemetry.Instrumentation.MassTransit
Changelog must be updated to indicated breaking name change, as nugets will be getting new name.

cijothomas:
Contributing guide must be updated. Similarly, search if anything else might be broken by this and address them as well.

reyang:
Couple things to consider:

Do we force existing packages to rename? It's probably not going to work for packages that are already marked as stable. 👀

How would the user know which package is backed by OpenTelemetry maintainers vs contrib maintainers? Are they have the same bar / recommendation?

cijothomas:
> Do we force existing packages to rename? It's probably not going to work for packages that are already marked as stable. 👀

For non-stable packages, we can rename. For stable packages, we can leave as is. Will ask the owners to see if they want to rename and unlist old packages.

cijothomas:
> How would the user know which package is backed by OpenTelemetry maintainers vs contrib maintainers? Do they have the same bar / recommendation?

The nuget shows the url of the repo, and user can infer which repo the package is coming from. But this topic requires a detailed discussion.

utpilla:
We have already released two pre-release packages from the repo without the word "Contrib" in the name:

1. [OpenTelemetry.Instrumentation.Quartz](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Quartz/1.0.0-alpha.1)
2. [OpenTelemetry.Exporter.Stackdriver](https://www.nuget.org/packages/OpenTelemetry.Exporter.Stackdriver/1.0.0-beta.2)

We plan to rename and publish all the other components that have not had a stable release.

@srprash What's your take on this? Would you want to rename the AWS packages?

cijothomas:
@srprash All packages except AWS packages are renamed. Could you comment here if there is a plan to rename AWS packages? If not, we can close this issue.

srprash:
Hi @cijothomas Allow me to discuss this within the AWS X-Ray team and I'll get back to you.

srprash:
Hi @cijothomas 
We will stick with the current package names until we have a major version to bump to. Feel free to close this issue.
Thanks.

PR #590, opened 2022-08-17
[Instrumentation.AWSLambda] Request release of 1.1.0-beta2, reword CHANGELOG

Oberon00:
@open-telemetry/dotnet-contrib-maintainers: This is a request for release of version 1.1.0-beta2 of the `OpenTelemetry.Instrumentation.AWSLambda` for which I am an owner.

~~Blocked by #593✅, #608✅, #622✅~~


src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,23 +2,22 @@
 
 ## Unreleased
 
-* Updated the `ActivitySource` name to the assembly name:
-  `OpenTelemetry.Instrumentation.AWSLambda`
+## 2.0.0
+
+* BREAKING (behavior): Update the `ActivitySource` name to not include `Contrib`.
+  The new activity source name is `OpenTelemetry.Instrumentation.AWSLambda`.
```
Oberon00, line 8: This rewording should fix https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/534#discussion_r947204413

I hope the activity source name actually conforms to the current guidelines? If not, I would like to revert #534 before the release naturally.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,23 +2,24 @@
 
 ## Unreleased
 
-* Updated the `ActivitySource` name to the assembly name:
-  `OpenTelemetry.Instrumentation.AWSLambda`
+## 2.0.0
```
utpilla, line 5: Do you want to do a stable release now? You could have another beta version (`1.1.0-beta.2`) released.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/CHANGELOG.md
utpilla: Yes, the ActivitySource name matches the standard practice.

Kielek:
As you have a plan to bump major version, can we make the change described here https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/196#issuecomment-1092161046 before the release?

Oberon00:
@Kielek I didn't realize only the AWS packages still had `Contrib` in the name. Yes, if the goal is to change every package to remove Contrib, then we should do that. I thought this was only about the Activity Source name originally.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/CHANGELOG.md
Oberon00: I will check back on that question internally. Are there any formal consequences of (not) using a beta release that I should be aware of for opentelemetry-dotnet-contrib (apart from signalling that it might not work properly yet and the need to specify `--prerelease` on the command line)?

utpilla: In general, it's a good practice to have a beta version released, so that people can use it for a good amount of time and provide feedback, if any.

Since, this package never had a stable release, so it should be fine to another beta version release with these breaking changes.

Oberon00: I changed it to 2.0.0-beta1 in 9acd7e1

utpilla: This package has never had a stable `1.*.*` release. Is there any particular reason to jump to `2.*.*` version without having a stable `1.*.*` version release?

Oberon00: I wouldn't expect such a large API change between betas. A beta is something that I would expect to be relatively stable API-wise, but maybe that's just my interpretation?

cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/VERSIONING.md#pre-releases if we go with this, its okay to have breaking changes in a beta release.
 With rc also it is allowed, but has a much higher bar than beta.

Oberon00: OK, I'm fine with 1.0.0-beta2 as well, @rypdal what do you think?

Oberon00:
After #622 is merged, this is ready to be merged (though there will be conflicts in the CHANGELOG)

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,26 +2,33 @@
...
+This is the first release with the new package name `OpenTelemetry.Instrumentation.AWSLambda`.
+
+* BREAKING (API, behavior): Rename package to OpenTelemetry.Instrumentation.AWSLambda
+  (remove ".Contrib") ([#593](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/593)).
+  This also affects the `ActivitySource` name (superseding [#534](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/534)).
+* BREAKING (API): Move public class AWSLambdaWrapper out of Implementation subnamespace
   ([#593](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/593))
+* BREAKING (API): Rename overloads of `AWSLambdaWrapper.Trace` that take an async
+  handler to `TraceAsync`, to emphasize that they (usually) need to be awaited.
+  ([#608](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/608))
```
Oberon00, line 18: I was made aware of an issue caused by #608: By naming the former overloads of Trace that took a Task-returning handler to TraceAsync, it is still possible to call Trace with a Task-returning handler. This will then cause the trace method to not await the result, ending the span probably immediately (and not reporting any exceptions). This seems like a bit of a pitfall to me and I'm currently considering reverting #608. Any thoughts?

Oberon00: I guess either way has it's issues. Let's keep it how it is now (with separate Trace/TraceAsync).

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,26 +2,33 @@
...
+## 1.1.0-beta2
+
+Release PR: [#590](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590)
+
+This is the first release with the new package name `OpenTelemetry.Instrumentation.AWSLambda`.
+
+* BREAKING (API, behavior): Rename package to OpenTelemetry.Instrumentation.AWSLambda
+  (remove ".Contrib") ([#593](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/593)).
+  This also affects the `ActivitySource` name (superseding [#534](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/534)).
+* BREAKING (API): Move public class AWSLambdaWrapper out of Implementation subnamespace
```
rypdal, line 14: Let's be consistent and use quotes also here: `AWSLambdaWrapper`, `Implementation `

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,26 +2,33 @@
...
+* Rewrite of parent context handling and related changes
+  ([#408](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/408)):
+  * BREAKING (API): Remove `AWSLambdaWrapper.Trace` overloads
+    without `ILambdaContext` parameter.
+  * BREAKING (behavior): Add automatic parent extraction from HTTP triggers
+    (API Gateway Proxy events), using the configured global textmap propagator.
+  * BREAKING (behavior): An activity is now also created if no parent context
+    could be extracted (previously this package would only create activities if
+    a valid parent span context could be extracted with X-Ray).
+  * Add optional parent context (`ActivityContext`) to `AWSLambdaWrapper.Trace`.
```
rypdal, line 28: Also to `AWSLambdaWrapper.TraceAsync`  after renaming of async `Trace` methods.

rypdal (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
Oberon00: Done in d4127a9

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
rypdal: I think this is more preferable that you kept `TraceAsync`. It's in sync with [MS naming convention for async methods](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/task-asynchronous-programming-model).

Oberon00:
This is now finally actually ready from my perspective, I know of no open points.

cijothomas (approved):

utpilla:
@Oberon00 @rypdal I noticed that the AWSLambda project uses `1.1.0` version of OTel SDK. Is this intentional? If not, you could update this to the latest version of OTel SDK like some of the other projects in the repo: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Wcf/OpenTelemetry.Instrumentation.Wcf.csproj#L11

Kielek:
@utpilla, please check #580.

Oberon00:
Hi, yes my idea is to only upgrade the OpenTelemetry SDK when there is a new feature we need. As the SDK is guaranteed to be backward-compatible, this should not prevent users from using any newer OpenTelemetry SDK, so I see no reason to upgrade.

utpilla:
@Oberon00 My apologies! I just realized after merging this PR that the package name does not follow the naming convention that we follow.

Could you please rename the version from `1.1.0-beta2` -> `1.1.0-beta.2`?

This is required because the build system sorts them alphabetically when retrieving packages, therefore, later in future `1.1.0-beta10` would be seen older than `1.1.0-beta9`.

Oberon00:
Hi, this follows the same convention as the beta1, but since we renamed the whole package it would of course be a good opportunity to also fix this versioning. You mean the CHANGELOG heading? I think that's the only part where we specify a version (and I have to double-check the unit test, I think I regex-match somewhere there).

utpilla:
> You mean the CHANGELOG heading?

Yes, that's the only thing to change.

PR #591, opened 2022-08-17
Add exception remarks for FileBlobProvider

vishweshbankwar:


src/OpenTelemetry.Extensions.PersistentStorage/FileBlobProvider.cs
```diff
@@ -62,6 +62,27 @@ public class FileBlobProvider : PersistentBlobProvider, IDisposable
     /// Controls the timeout when writing a buffer to blob.
     /// Default is 1 minute.
     /// </param>
+    /// <exception cref="ArgumentNullException">
+    /// path is null.
```
reyang, line 66: Consider some best practices like https://www.pluralsight.com/guides/using-param-paramref-tag-in-csharp-xml-documentation-comments.

reyang (approved):

src/OpenTelemetry.Extensions.PersistentStorage/FileBlobProvider.cs
vishweshbankwar: will follow up separately.

Issue #447, opened 2022-06-21
Add a package for process related metrics 


xiang17:
# Issue with OpenTelemetry.Instrumentation.Runtime

Process related metrics were removed from the Runtime metrics package and they should be added in a separate package and with OpenTelemetry.Instrumentation.Process as a potential name. See the context here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/404#discussion_r894123445

xiang17:
@cijothomas Can you remove the runtime label? It doesn't belong to the runtime instrumentation library now.

reyang:
@Yun-Ting FYI I've assigned this issue to you.

Yun-Ting:
Alpha version was released: https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Process.
This issue will be closed after the stable release.

Yun-Ting:
Dashboard built with Grafana for 0.5.0 (version number under discussion) release:
![image](https://user-images.githubusercontent.com/16979322/219503264-6226713d-e135-45cc-aebd-1f62545ed241.png)

cijothomas:
Could you send a PR to the repo with the json file?

Yun-Ting:
> Could you send a PR to the repo with the json file?

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1016

Kielek:
0.5.0 version is out, json configuration for graphana included.

Closing.

PR #595, opened 2022-08-19
[Instrumentation.Process] Added memory related metrics.

Yun-Ting:
## Changes

Adding 2 memory related metrics following the semantic conventions for OS process metrics as starters.
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#metric-instruments

Will be working on adding more metrics (cpu related and others) in the upcoming PRs.


src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,8 +25,23 @@ internal class ProcessMetrics
...
+    private const string MetricPrefix = "process.dotnet.";
+
     static ProcessMetrics()
     {
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            $"{MetricPrefix}memory.usage.physical",
+            () => Diagnostics.Process.GetCurrentProcess().WorkingSet64,
+            unit: "bytes",
+            description: "The amount of physical memory allocated for the current process.");
```
reyang, line 37: Consider aligning with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md, and if there is something in the spec which should be improved, propose a change in the spec.

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -13,9 +13,36 @@
...
+        var exportedItems = new List<Metric>();
+        using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddProcessInstrumentation()
+            .AddInMemoryExporter(exportedItems)
+            .Build();
+
+        meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+        Assert.True(exportedItems.Count == 2);
+        Assert.StartsWith("process", exportedItems[0].Name);
```
cijothomas, line 39: why testing just startswith? the actual name is already covered by tests below.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,8 +25,23 @@ internal class ProcessMetrics
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
 
+    private const string MetricPrefix = "process.runtime.dotnet.";
```
reyang, line 28:
```suggestion
    private const string MetricPrefix = "process.";
```

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -13,9 +13,33 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 // </copyright>
+using System.Collections.Generic;
```
reyang, line 16: nit: put an empty line between Ln15/16

reyang (approved):

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,8 +25,23 @@ internal class ProcessMetrics
...
+
     static ProcessMetrics()
     {
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            $"{MetricPrefix}memory.usage",
+            () => Diagnostics.Process.GetCurrentProcess().WorkingSet64,
```
reyang, line 35: Maybe the `Diagnostics.Process.GetCurrentProcess()` part can be pre-calculated and reused?

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -13,9 +13,33 @@
...
+        var exportedItems = new List<Metric>();
+        using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddProcessInstrumentation()
+            .AddInMemoryExporter(exportedItems)
+            .Build();
+
+        meterProvider.ForceFlush(MaxTimeToAllowForFlush);
+
+        Assert.True(exportedItems.Count == 2);
+        var physicalMemoryMetric = exportedItems.FirstOrDefault(i => i.Name == $"{metricPrefix}memory.usage");
```
reyang, line 40: I guess the goal is to test the exact name? Consider using string literal `"process.memory.usage"`.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,6 +16,7 @@
 
 using System.Diagnostics.Metrics;
 using System.Reflection;
+using Diagnostics = System.Diagnostics;
```
utpilla, line 19: Do we need an alias for the namespace?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: I saw from this [doc](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#attributes) indicating that  resource attributes SHOULD be included on runtime metric events as appropriate. And java uses this convention, for example: process.runtime.jvm.memory.usage.
Which one should we follow?

reyang: This is about process instrumentation instead of runtime instrumentation? no?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -25,21 +24,22 @@ internal class ProcessMetrics
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
 
-    private const string MetricPrefix = "process.runtime.dotnet.";
+    private const string MetricPrefix = "process.";
```
reyang, line 27: Curious - do we really need this const? (what benefit do we gain)

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -25,21 +24,22 @@ internal class ProcessMetrics
...
+    private static readonly System.Diagnostics.Process CurrentProcess = System.Diagnostics.Process.GetCurrentProcess();
```
reyang, line 28: using System.Diagnostics?

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
Yun-Ting: I've removed this assertion.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: I've created a static var for the current process and removed the alias.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: I guess not much given currently it is just a 8 char string.... (not even for readability)

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Compiler seems to be confused between System.Diagnostics.Process vs OpenTelemetry.Instrumentation.Process if we use using the System.Diagnostics.

reyang: Ah I see, thanks! In this way the [alias](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/595#discussion_r954405557) does seem to be a good fit.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,16 +16,31 @@
...
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
+    private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
 
     static ProcessMetrics()
     {
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            "process.memory.usage",
+            () => CurrentProcess.WorkingSet64,
```
alanwest, line 34: Does `Refresh()` need to be called here? See: https://docs.microsoft.com/dotnet/api/system.diagnostics.process.workingset64#remarks

alanwest, line 34: It's been awhile since I've studied the difference between `WorkingSet64` and `PrivateMemorySize64`, so I'll admit I'm a bit fuzzy on the semantics of one over the other.

This seems to suggest that PrivateMemorySize64 may be useful as well https://stackoverflow.com/a/1986486/2052722.

alanwest: Ah ha, thanks, yea I agree.

reyang: I would say all of these are useful (otherwise the NT OS kernel folks wouldn't invent these concepts):

image

I guess the struggle is do we want to just want to do a .NET specific thing (similar to what we ended up with https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime) or try to align with / wait for the spec (or maybe help the spec to move forward). I can also see that the spec might struggle when it comes to some verbose memory management concepts because those could be OS specific (or even specific to certain versions of the NT kernel).

alanwest: Yes, I suspect this would be hard to generalize at the spec level. Maybe something like a `process.dotnet.*` namespace would be useful. Though, yea `process.{os}.{os_distribution}.*` might be more accurate but still hard to generalize across windows, linux, and whatever else.

Yun-Ting: I've added Refresh() before collecting the info; thank you Alan!

alanwest: There was an effort a few years back to port these properties to equivalents for linux https://github.com/dotnet/corefx/pull/41122. So at least for .NET and the distros of linux that it supports there should be some level of generalization across OSes.

alanwest (approved):
I'm not certain what the overhead of calling Refresh() for each metric is. If we find out it's expensive, we might consider this strategy in the future https://github.com/dotnet/runtime/issues/62027#issuecomment-1114374751.

Though... if Refresh is deemed cheap enough, then 👍 to keeping things simple.

cijothomas (approved):

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,16 +16,39 @@
...
 using System.Reflection;
+using Diagnostics = System.Diagnostics;
 
 namespace OpenTelemetry.Instrumentation.Process;
 
 internal class ProcessMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
+    private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
```
reyang, line 27: This, combined with `Refresh`, look like a source of race condition.

image

CodeBlanch: @Yun-Ting You could probably just use a closure a la:

```csharp
    static ProcessMetrics()
    {
        Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();

        // TODO: change to ObservableUpDownCounter
        MeterInstance.CreateObservableGauge(
            "process.memory.usage",
            () =>
            {
                currentProcess.Refresh();
                return currentProcess.WorkingSet64;
            },
            unit: "By",
            description: "The amount of physical memory in use.");

        // TODO: change to ObservableUpDownCounter
        MeterInstance.CreateObservableGauge(
            "process.memory.virtual",
            () =>
            {
                currentProcess.Refresh();
                return currentProcess.VirtualMemorySize64;
            },
            unit: "By",
            description: "The amount of committed virtual memory.");
    }
```

Is `Process.Refresh()` expensive? We'll be invoking it twice here anytime metrics are collected. If it is expensive we might want to create a state object we can refresh once and then return the result to the two counters in the callback.

Yun-Ting: Thank you very much for pointing it out. Please refer to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/612 for the update.

Yun-Ting:
> I'm not certain what the overhead of calling Refresh() for each metric is. If we find out it's expensive, we might consider this strategy in the future [dotnet/runtime#62027 (comment)](https://github.com/dotnet/runtime/issues/62027#issuecomment-1114374751).
> 
> Though... if Refresh is deemed cheap enough, then 👍 to keeping things simple.

Thank you for the code pointer. I've made some modifications here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/612

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Please refer to the latest PR https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/624 instead. Thanks!

Yun-Ting:
> I'm not certain what the overhead of calling Refresh() for each metric is. If we find out it's expensive, we might consider this strategy in the future [dotnet/runtime#62027 (comment)](https://github.com/dotnet/runtime/issues/62027#issuecomment-1114374751).
> 
> Though... if Refresh is deemed cheap enough, then 👍 to keeping things simple.

Please refer to the latest PR: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/624 instead. Thanks!

PR #593, opened 2022-08-18
Instrumentation.AWSLambda: Rename to remove .Contrib & minor cleanups.

Oberon00:
Required before the release #590.

## Changes
* BREAKING: Rename package (& namespaces) to remove .Contrib.
* BREAKING: Move public class AWSLambdaWrapper out of implementation subnamespace
* BREAKING (technically): Make AWSLambdaWrapper a static class (found through #607).
* Add version to ActivitySource

For significant contributions please make sure you have completed the following items:

* [X] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [X] Design discussion issue: #196 (see also <https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590#issuecomment-1219051107>, CC @Kielek)

CC @srprash @rypdal

PR #607, opened 2022-08-24
[Instrumentation.AwsLambda] Enable public API check.

Oberon00:
Enable public API check for Instrumentation.AwsLambda, as I think that's a very useful tool.

No CHANGELOG entry.


src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,14 @@
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.AWSLambdaInstrumentationOptions
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.AWSLambdaInstrumentationOptions.AWSLambdaInstrumentationOptions() -> void
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.AWSLambdaInstrumentationOptions.DisableAwsXRayContextExtraction.get -> bool
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.AWSLambdaInstrumentationOptions.DisableAwsXRayContextExtraction.set -> void
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation.AWSLambdaWrapper
```
Oberon00, line 5: What is the policy for moving to Shipped? For all kinds of "shippings", or do pre-releases (beta) not count?

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,14 @@
...
+OpenTelemetry.Contrib.Instrumentation.AWSLambda.Implementation.AWSLambdaWrapper.AWSLambdaWrapper() -> void
```
Oberon00, line 6: This public ctor is actually unintentionally exposed, fixed in #593 after noticing through this line.

src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
utpilla: The entries are moved to Shipped from Unshipped when doing a stable release.

src/OpenTelemetry.Instrumentation.AWSLambda/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,13 @@
...
+static OpenTelemetry.Instrumentation.AWSLambda.AWSLambdaWrapper.Trace<TInput, TResult>(OpenTelemetry.Trace.TracerProvider tracerProvider, System.Func<TInput, Amazon.Lambda.Core.ILambdaContext, TResult> lambdaHandler, TInput input, Amazon.Lambda.Core.ILambdaContext context, System.Diagnostics.ActivityContext parentContext = default(System.Diagnostics.ActivityContext)) -> TResult
+static OpenTelemetry.Instrumentation.AWSLambda.AWSLambdaWrapper.Trace<TInput>(OpenTelemetry.Trace.TracerProvider tracerProvider, System.Action<TInput, Amazon.Lambda.Core.ILambdaContext> lambdaHandler, TInput input, Amazon.Lambda.Core.ILambdaContext context, System.Diagnostics.ActivityContext parentContext = default(System.Diagnostics.ActivityContext)) -> void
+static OpenTelemetry.Instrumentation.AWSLambda.AWSLambdaWrapper.TraceAsync(OpenTelemetry.Trace.TracerProvider tracerProvider, System.Func<Amazon.Lambda.Core.ILambdaContext, System.Threading.Tasks.Task> lambdaHandler, Amazon.Lambda.Core.ILambdaContext context, System.Diagnostics.ActivityContext parentContext = default(System.Diagnostics.ActivityContext)) -> System.Threading.Tasks.Task
```
Oberon00, line 10: It always clarifies things for me to look at this condensed API listing:

Apparently, we have overloads for (very unusual) void-returning handlers with and without `TInput input` parameter. For the (much more common) TResult-returning handlers, we only have the overload with input.

That seems weird to me. Should we remove the non-input overload also for void? Or add the non-input overload for TResult?

rypdal (approved):
Changes are looking good and can be merged.

src/OpenTelemetry.Instrumentation.AWSLambda/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
rypdal: I would go with "non-input overload for TResult". In case of "removing non-input overload" option we seems force customer to have function handler with redundant unused argument. For example: `TResult Handler(ILambdaContext context, object unusedArgument)`, where `TInput = typeof(object)`.

Oberon00: I discovered that the non-input overloads were not originally present but only added in #408 (unreleased).
That makes me think we should remove them after all. My reason being, that having a Lambda without input argument is basically useless, but having the overload may make customers use them nevertheless, since they can just access the parameter passed to the actual handler without passing it through the Trace methods. But that would prevent us from doing special handling for certain inputs

rypdal: Seems handlers without input parameters even not expected:  https://docs.aws.amazon.com/lambda/latest/dg/csharp-handler.html. So, I agree that input-less Trace overloads can be removed.

Oberon00: I created #622 to remove the overloads.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
```diff
@@ -5,6 +5,7 @@
     <Description>AWS Lambda tracing wrapper for OpenTelemetry .NET</Description>
     <PackageTags>$(PackageTags);AWS Lambda</PackageTags>
     <MinVerTagPrefix>Instrumentation.AWSLambda-</MinVerTagPrefix>
+    <EnablePublicApi>true</EnablePublicApi>
```
utpilla, line 8: You might want to enable code analysis as well.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj#L9

utpilla (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
Oberon00: That is disabled by default? Weird, how should I know 😅
I will try to enable it.
Maybe it would be better to enable it by default and explicitly disable it in projects where it causes problems?

Oberon00: I tried it locally, the tool only suggested adding some ConfigureAwait(false) in some places. I will submit a PR for that eventually, but I don't want to hold up the release longer, so I will do that after #590

PR #608, opened 2022-08-24
[Instrumentation.AwsLambda] Rename Trace to TraceAsync where appropriate

Oberon00:
Sorry for another renaming proposal, but it seems really bad that we have methods that you cannot await and methods that you need to await with the same name. Also, the overload resolution between some of them could get tricky.

Note: The diff looks really strange here, please look at the two commits separately to see what's actually going on.


src/OpenTelemetry.Contrib.Instrumentation.AWSLambda/Implementation/AWSLambdaWrapper.cs
```diff
@@ -89,80 +89,80 @@ public class AWSLambdaWrapper
         }
 
         /// <summary>
-        /// Tracing wrapper for async Lambda handler.
+        /// Tracing wrapper for Lambda handler.
```
Oberon00, line 92: Please look at the commits in the diff separately, I had to move a function around and that causes to the diffing algorithm to go off the rails apparently.

utpilla:
@rypdal Could you please review this PR?

Oberon00:
@rypdal Will probably not be available before Thursday.

rypdal:
Changes are looking good to me.

rypdal (approved):

rypdal (approved):
Changes are looking good and can be merged.

utpilla:
Please update the branch with the latest changes in `main` and we can merge this PR. I do not have access to update your branch.

Oberon00:
@utpilla I think I enabled maintainer access now.

Kielek (approved):

cijothomas (approved):

Oberon00:
Forgot the changelog entry, will add it to the release PR #590.

PR #612, opened 2022-08-26
[Instrumentation.Process] Added Cpu related metrics and addressed comments.

Yun-Ting:
## Changes

Following OTel spec:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#metric-instruments

1. Addressed some comments:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/595#pullrequestreview-1086234964
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/595#discussion_r956231814

2. Added CPU time and CPU utilization metrics. It is not clear from the spec that whether breakdown by CPU states (System, User, Wait) option should be implemented. Currently, this PR is not providing an option to specify that and uses no-breakdown gauge and counter by default.


src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -57,5 +58,74 @@ static ProcessMetrics()
...
+        {
+           MeterInstance.CreateObservableCounter(
+           $"process.cpu.time",
+           () =>
+           {
+               CurrentProcess.Refresh();
+               return CurrentProcess.TotalProcessorTime.Seconds;
+           },
+           unit: "s",
+           description: "Total CPU seconds broken down by different states.");
```
cijothomas, line 106: is it broken down by states?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -57,5 +58,74 @@ static ProcessMetrics()
...
+        {
+           MeterInstance.CreateObservableCounter(
+           $"process.cpu.time",
+           () =>
+           {
+               Measurement<long>[] measurements = new Measurement<long>[3];
+
+               CurrentProcess.Refresh();
+               var priviledgedCpuTime = CurrentProcess.PrivilegedProcessorTime.Seconds;
+               var userCpuTime = CurrentProcess.PrivilegedProcessorTime.Seconds;
```
cijothomas, line 85: both seems same?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -57,5 +58,74 @@ static ProcessMetrics()
...
+        // labels
+        // state, if specified, SHOULD be one of: system, user, wait. A process SHOULD be characterized either by data points with no state labels, or only data points with state labels.
+        // IEnumberable,
+        // List<Measurement<long>>
+        // wait time = totalprocessortime - Process.PrivilegedProcessorTime - Process.userProcessorTime
+
+        if (options.ExpandOnCpuStates == true)
+        {
+           MeterInstance.CreateObservableCounter(
+           $"process.cpu.time",
```
cijothomas, line 78: consider adding link to the relevant OTel semantic convention doc.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: fixed

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
```diff
@@ -21,4 +21,5 @@ namespace OpenTelemetry.Instrumentation.Process;
 /// </summary>
 public class ProcessInstrumentationOptions
 {
+    public bool? ExpandOnCpuStates { get; set; }
```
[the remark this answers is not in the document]
Yun-Ting: Would CpuStatesEnabled be better? :grin:

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: added link to the spec

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: updated the description

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
```diff
@@ -21,4 +21,9 @@ namespace OpenTelemetry.Instrumentation.Process;
 /// </summary>
 public class ProcessInstrumentationOptions
 {
+    /// <summary>
+    /// Gets the flag indicating whether Cpu time should be further broken down by its states.
+    /// The Cpu state could be one of the following type: system, user, or wait.
```
reyang, line 26:
```suggestion
    /// The CPU state could be one of the following type: system, user, or wait.
```

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +26,91 @@ internal class ProcessMetrics
...
+            return this.currentProcess.TotalProcessorTime.Seconds;
+        }
+
+        // See spec:
+        // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#metric-instruments
+        public Measurement<long>[] GetProcessorCpuTimeWithBreakdown()
+        {
+            Measurement<long>[] measurements = new Measurement<long>[Enum.GetNames(typeof(CpuState)).Length];
+            var priviledgedCpuTime = this.currentProcess.PrivilegedProcessorTime.Seconds;
+            var userCpuTime = this.currentProcess.UserProcessorTime.Seconds;
```
reyang, line 107: Do we want integer or double?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +26,91 @@ internal class ProcessMetrics
...
+
+            measurements[(int)CpuState.System] = new(priviledgedCpuTime, new KeyValuePair<string, object>("state", CpuState.System.ToString()));
```
reyang, line 109: Would the `ToString()` here cause allocation each time, or just once?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +26,91 @@ internal class ProcessMetrics
...
+            measurements[(int)CpuState.User] = new(userCpuTime, new KeyValuePair<string, object>("state", CpuState.User.ToString()));
+            measurements[(int)CpuState.Wait] = new(this.currentProcess.TotalProcessorTime.Seconds - priviledgedCpuTime - userCpuTime, new KeyValuePair<string, object>("state", CpuState.Wait.ToString()));
```
reyang, line 111: Will this give a non-zero value?

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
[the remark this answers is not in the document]
reyang: I would vote for not having the optional at all. (and later if folks **really** need that, we can add it in a non-breaking way)

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
```diff
@@ -21,4 +21,9 @@ namespace OpenTelemetry.Instrumentation.Process;
 /// </summary>
 public class ProcessInstrumentationOptions
 {
+    /// <summary>
+    /// Gets or sets the flag indicating whether CPU time should be further broken down by its states.
+    /// The CPU state could be one of the following type: system, user, or wait.
+    /// </summary>
+    public bool? CpuStatesEnabled { get; set; } = false;
```
CodeBlanch, line 28: Just curious, why is this a `Nullable<bool>`? Is `null` state meaningful?

nit:

```csharp
    /// <summary>
    /// Gets or sets a value indicating whether or not CPU time should be further broken down by its states.
    /// </summary>
   /// <remarks>The CPU state could be one of the following types: system, user, or wait.</remarks>
```

Yun-Ting: Yeah... I agree that Null state is not really meaningful here; removed.
And I've updated the XML as suggested. Thanks!

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Updated to use int.

reyang: Why int?

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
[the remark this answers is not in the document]
Yun-Ting: So what should we have as the default behavior?
Breaking down by the states or not?
The spec is not very clear about this currently.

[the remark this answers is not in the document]
reyang: What do you think? (I'll hold my answer for now 😄)

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -57,5 +58,74 @@ static ProcessMetrics()
...
+           $"process.cpu.time",
+           () =>
+           {
+               Measurement<long>[] measurements = new Measurement<long>[3];
+
+               CurrentProcess.Refresh();
+               var priviledgedCpuTime = CurrentProcess.PrivilegedProcessorTime.Seconds;
+               var userCpuTime = CurrentProcess.PrivilegedProcessorTime.Seconds;
+
+               measurements[(int)CPUState.System] = new(priviledgedCpuTime, new KeyValuePair<string, object>("state", CPUState.System.ToString()));
```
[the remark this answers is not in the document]
Yun-Ting: yeah... that's the issue; maybe hardcoding the indices here is not that of a bad choice?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: well, because the `.Seconds` from the Timespan struct returns int?

src/OpenTelemetry.Instrumentation.Process/ProcessInstrumentationOptions.cs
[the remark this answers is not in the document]
Yun-Ting: Maybe default to not having the breakdown unless people really want it.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: The browser does not show me a very well-defined highlight for this comment. May I ask which part do you refer to?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Declared string array as the member variable of ProcessMetrics for the CPU states to avoid unnecessary alloc.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
reyang: What does `TotalProcessorTime - PriviledgedCpuTime - UserCpuTime` actually mean?

reyang: If you read this https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.totalprocessortime?view=net-6.0#property-value:

`TotalProcessorTime - PriviledgedCpuTime - UserCpuTim` is always going to give zero, unless .NET or the underlying native API has bugs.

image

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
reyang: > well, because the `.Seconds` from the Timespan struct returns int?

How is this related?

REscobar: Shouldn't it be `.TotalSeconds`?

reyang (changes requested):
This looks heading towards a wrong direction.

PR #620, opened 2022-09-01
[Instrumentation.EventCounters] Stabilize and Enable Testing

mic-max:
Towards #215 since spec is not stable

## Changes
- Cleans up code for EventCounters instrumentation
- Fixes tests
- Only create counters for `EventCounters` named events by default and that are from a configured EventSource name, which includes nothing by default.

## Resources

- https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters
- [Similar class](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Diagnostics/AzureEventSourceListener.cs)
- https://github.com/dotnet/diagnostics/blob/main/documentation/design-docs/eventcounters.md
- [OpenTelemetry.Instrumentation.Wcf/TracerProviderBuilderExtensions](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Wcf/TracerProviderBuilderExtensions.cs#L38) copied implementation of only allowing a single addition of the instrumentation.

## Follow-up PR Plans
- Multiple counters from the same source with different names
- Multiple counters from the same source with the same name
- Call `meterProvider.AddEventCounterListener` multiple times.
- Verify stress testing the instrumentation does not incur memory leaks, or slowdowns.
- Consider making a Meter for each of the EventSources listened to. (this might make any future wildcard functionality being added to `AddEventSources` more complicated...)
- Add tests that ensure EventSource warning logs are created when expected. (use InMemoryEventListener vs. TestEventListener?)
- 


noahfalk (commented):
Thanks! I like this feature for OTel : ) I left some suggestions inline

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -14,15 +14,32 @@
...
-/// EventCounterMetrics Options.
+/// EventCounterListener Options.
 /// </summary>
-public class EventCounterMetricsOptions
+public class EventCountersInstrumentationOptions
 {
     /// <summary>
     /// Gets or sets the subscription interval in seconds.
     /// </summary>
     public int RefreshIntervalSecs { get; set; } = 60;
```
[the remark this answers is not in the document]
reyang: +1 1s makes sense to me considering Prometheus defaults to 15s and some other systems use 20s and 5s.

src/OpenTelemetry.Instrumentation.EventCounters/README.md
```diff
@@ -1,70 +1,74 @@
...
+EventCounter eventCounter = new("MyEventCounterName", eventSource);
+eventCounter.WriteMetric(0);
+eventCounter.WriteMetric(1000);
 
+PollingCounter pollingCounter = new("MyPollingCounterName", eventSource, () => new Random().NextDouble());
 ```
 
+There are some
+[Well-known EventCounters in .NET](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/available-counters)
+that are filtered out by this library.
```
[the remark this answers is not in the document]
mic-max: Yes, I agree, poor wording that I forgot to update from an earlier version of the change.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -14,15 +14,32 @@
...
+
+    /// <summary>
+    /// Gets or sets the names of <c>EventSource</c>s to listen to.
+    /// </summary>
+    public HashSet<string> Sources { get; set; } = new();
```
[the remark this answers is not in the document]
mic-max: That's a great point about keeping the data structure hidden for freedom to change and keeping the same API.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -14,15 +14,32 @@
...
+
+    /// <summary>
+    /// Gets or sets the name of <c>EventCounters</c> to listen to.
+    /// </summary>
+    public HashSet<string> Names { get; set; } = new() { "EventCounters" };
+
+    /// <summary>
+    /// Gets the arguments object used for the EventListener.EnableEvents function.
+    /// </summary>
+    public Dictionary<string, string> EnableEventsArguments => new() { { "EventCounterIntervalSec", this.RefreshIntervalSecs.ToString() } };
```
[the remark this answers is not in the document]
mic-max: And make that private field available to the `EventCountersMetrics.cs` class with a public property?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        var payload = eventData.Payload[0] as IDictionary<string, object>;
+        var name = payload["Name"].ToString();
+        var isGauge = payload.ContainsKey("Mean");
+        Tuple<string, string> metricKey = new(eventData.EventSource.Name, name);
+
+        this.values[metricKey] = isGauge
+            ? Convert.ToDouble(payload["Mean"])
+            : Convert.ToDouble(payload["Increment"]) / this.options.RefreshIntervalSecs;
```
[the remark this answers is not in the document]
mic-max: The `Increment` is used for created an `CreateObservableCounter`. Is your suggestion still valid? It would also require a change to `EventCounterMetricsTests.cs:IncrementingPollingCounter` making the expected value 2 instead of 1.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,90 @@
...
+
+namespace OpenTelemetry.Instrumentation.EventCounters;
+
+/// <summary>
+/// .NET EventCounters Instrumentation.
+/// </summary>
+internal class EventCountersMetrics : EventListener
+{
+    internal static readonly AssemblyName AssemblyName = typeof(EventCountersMetrics).Assembly.GetName();
+    internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
```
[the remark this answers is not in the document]
mic-max: Hmm 🤔 that's a cool idea, I'll look into this :)

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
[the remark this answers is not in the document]
noahfalk: Sorry instead of 'private' what I really intended to convey was 'not part of the public API that 3rd party developers will use'. For example the `internal` keyword would satisfy my goal while still letting EventCounterMetrics invoke it.

 The factoring I would choose personally is moving this property to the EventCounterMetrics type, for example:
```C#
class EventCounterMetrics
{
   ...
   public EventCountersMetrics(EventCountersInstrumentationOptions options)
    {
        this.options = options;

        while (this.preInitEventSources.TryTake(out EventSource source))
        {
            if (this.options.ShouldListenToSource(source.Name))
            {
                this.EnableEvents(source, EventLevel.LogAlways, EventKeywords.None, GetEnableEventsArguments(options));
            }
        }
    }

    private static Dictionary<string,string> GetEnableEventsArguments(EventCountersInstrumentationOptions options) =>
        new() { { "EventCounterIntervalSec", options.RefreshIntervalSecs.ToString() } };
}
```

Not having it be part of the public API is my main recommendation though, regardless whether you use this factoring or the internal keyword or something else.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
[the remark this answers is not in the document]
noahfalk: Its always possible I made a mistake, but I don't think that I did :) 

For example imagine that we have an IncrementingEventCounter, we call Increment() once per second, and the refresh interval reports an event back to us every 3 seconds:

| Time                 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |
| ---------------------|---|---|---|---|---|---|---|---|---|
| Increment() argument | 2 | 3 | 1 | 4 | 3 | 2 | 2 | 1 | 2 | 
| payload["Increment"] | - | - | 6 | - | - | 8 | - | - | 5 |

The ObservableCounter instrument expects that you provide the running total in the callback. So in this example you would need to provide 6 at t_3, 14 at t_6, and 19 t_9.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,77 @@
...
+    private readonly ConcurrentDictionary<Tuple<string, string>, double> values = new();
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EventCountersMetrics"/> class.
+    /// </summary>
+    public EventCountersMetrics()
+    {
+    }
+
+    public static EventCountersInstrumentationOptions Options { get; set; }
```
mic-max, line 44: If this gets called twice, it will throw an exception. Is that bad?

```csharp
Sdk.CreateMeterProviderBuilder()
            .AddEventCountersInstrumentation()
            .Build();
```

mic-max: Yes. I see that other instrumentation don't have this kind of protection so I will forego being the first to try especially with the goal of this PR being to stabilise the instrumentation library :)

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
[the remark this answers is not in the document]
mic-max: I cannot add `Meter`s after the `MeterProvider` has been built unfortunately. Got this code, but after reading [this ](https://github.com/open-telemetry/opentelemetry-dotnet/blob/cd6043bbf72571e5fdf83a1bb2cababc3c0a5a95/docs/metrics/customizing-the-sdk/README.md#meter)and seeing the tests fail I realise it was incorrect.

image

image

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,99 @@
...
+    private readonly ConcurrentDictionary<Tuple<string, string>, Instrument> instruments = new();
+    private readonly ConcurrentDictionary<Tuple<string, string>, double> values = new();
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="EventCountersMetrics"/> class.
+    /// </summary>
+    /// <param name="options">The options to define the metrics.</param>
+    public EventCountersMetrics(EventCountersInstrumentationOptions options)
+    {
+        this.options = options;
```
reyang, line 45: What if folks changed the public part of `options` (e.g. `RefreshIntervalSecs`) after the `EventCountersMetrics` is instantiated?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        this.eventSourceNames.Add(eventSourceName);
+    }
+
+    /// <summary>
+    /// Returns whether or not an EventSource should be enabled on the EventListener.
+    /// </summary>
+    /// <param name="eventSourceName">The EventSource name.</param>
+    /// <returns><c>true</c> when an EventSource with the name <paramref name="eventSourceName"/> should be enabled.</returns>
+    public bool ShouldListenToSource(string eventSourceName)
```
reyang, line 47: Does this have to be public?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    private readonly HashSet<string> eventSourceNames = new();
+
+    /// <summary>
+    /// Gets or sets the subscription interval in seconds.
+    /// </summary>
+    public int RefreshIntervalSecs { get; set; } = 1;
+
+    /// <summary>
+    /// Listens to EventCounters from the given EventSource name.
```
reyang, line 34: I didn't fully understand this, it seems `AddEventSource(eventSourceName)` is more intuitive?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+        else
+        {
+            _ = this.values.TryAdd(metricKey, 0);
+            this.values[metricKey] += Convert.ToDouble(payload["Increment"]);
+        }
+
+        _ = this.instruments.TryAdd(
+            metricKey,
+            isGauge ? MeterInstance.CreateObservableGauge(name, () => this.values[metricKey]) : MeterInstance.CreateObservableCounter(name, () => this.values[metricKey]));
```
reyang, line 94: What if the `name` already exists under the `MeterInstance`?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: With how the interface is setup I don't think they really can. All they have access to is `AddEventCountersInstrumentation` and a callback function to make any changes to the `EventCountersInstrumentationOptions` defined in the `AddEventCountersInstrumentation` method.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
mic-max: `internal` works!

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
reyang: If this is no way that an end user could use `builder.AddInstrumentation(() => new something(options))`, then it's fine.

Btw, now I see that the ctor is for an internal class 👍

reyang (approved):
LGTM.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: Currently it will not crash, but the value of the metric will be `0`.

I could add a warning that an instrument named that already exists?
Or probably better, is the name of the created instrument could be $"{eventSourceName}.${eventCounterName}"

reyang: > I could add a warning that an instrument named that already exists? Or probably better, is the name of the created instrument could be "eventSourceName.{eventCounterName}"

The latter sounds good to me if we make sure `eventSourceName` is properly escaped (so we have a valid instrument name which follows the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-naming-rule)).

Doesn't have to block this PR though.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
noahfalk: I think they could change it, but I also think you'd be well within your rights to say that is undefined behavior if they do:
```C#
var globalOptions;
builder.AddEventCounterInstrumentation(opts =>
{
    globalOptions = opts;
}).Build();
globalOptions.RefreshIntervalSecs = 1_000;
```

If you want to be very defensive you can copy the options in the constructor so that you never again refer to the mutable options object that was exposed in the callback.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
noahfalk: Even if name doesn't already exist, if the EventSource name isn't going to be the Meter name it should probably be in the instrument name. Some counter names may be very ambiguous such as "size", "length", or "requests."

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
noahfalk, line 45: I think they could change it, but I also think you'd be well within your rights to say that is undefined behavior if they do:
```C#
var globalOptions;
builder.AddEventCounterInstrumentation(opts =>
{
    globalOptions = opts;
}).Build();
globalOptions.RefreshIntervalSecs = 1_000;
```
If you want to be very defensive you can copy the options in the constructor so that you never again refer to the mutable options object that was exposed in the callback.

reyang: In general, OpenTelemetry .NET tries to be super conservative/defensive, `options` are only used during the ctor (and if someone trying to change the options while the ctor is running, it's considered their fault of introducing a race condition), all the necessary information in the `options` will be captured as a snapshot, so later modifications on the options won't affect anything that has been created already.

https://github.com/open-telemetry/opentelemetry-dotnet/blob/ed0450ef0ef5178cb29cb8908c3ff120a053caad/src/OpenTelemetry.Exporter.Prometheus.HttpListener/PrometheusHttpListener.cs#L47

Given https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/620#discussion_r978138219, I guess it's probably okay for now.

reyang (approved):

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+/// EventCounterListener Options.
+/// </summary>
+public class EventCountersInstrumentationOptions
+{
+    private readonly HashSet<string> eventSourceNames = new();
+
+    /// <summary>
+    /// Gets or sets the subscription interval in seconds.
+    /// </summary>
+    public int RefreshIntervalSecs { get; set; } = 1;
```
cijothomas, line 32: might need a better description to convey the fact that this is really the interval at which we refresh/read values from eventcounter.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    /// <summary>
+    /// Listens to EventCounters from the given EventSource name.
+    /// </summary>
+    /// <param name="eventSourceName">The EventSource name.</param>
+    public void AddEventSource(string eventSourceName)
```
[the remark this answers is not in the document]
mic-max: Originally had something similar, but it was suggested to abstract away the data type so we can make fewer breaking changes to the public interface

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: Sounds good, yes I will include a warning and simply not add an additional instrument. The instrument for event source `a` with counter `b.c` will go to the `a.b.c` instrument, same as the event source `a.b` with counter `c`

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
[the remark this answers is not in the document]
noahfalk: I'd favor keeping Add() like it is :)

I would think the strongest API precedents would be things like AddFilter() and AddProvider() from the ILoggingBuilder interface or the AddMeter() API in OTel. I expect Add() will feel more consistent to API users.

Aside from the value of consistency I think another benefit of Add() here is in the future if we want to support operations like adding a single counter from an EventSource. New overloads of Add() could handle that relatively easily whereas with the IEnumerable<string> property I'm not sure how we'd extend it in a way that is intuitive for API users.

utpilla, line 38: Use `AddSource(params string[] names)` instead?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        this.eventSourceNames.Add(eventSourceName);
+    }
+
+    /// <summary>
+    /// Returns whether or not an EventSource should be enabled on the EventListener.
+    /// </summary>
+    /// <param name="eventSourceName">The EventSource name.</param>
+    /// <returns><c>true</c> when an EventSource with the name <paramref name="eventSourceName"/> should be enabled.</returns>
+    internal bool ShouldListenToSource(string eventSourceName)
```
utpilla, line 53: This method feels like it should belong to `EventCountersMetrics` class.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Diagnostics.Metrics;
+using System.Diagnostics.Tracing;
+using System.Reflection;
+
+namespace OpenTelemetry.Instrumentation.EventCounters;
+
+/// <summary>
+/// .NET EventCounters Instrumentation.
+/// </summary>
+internal class EventCountersMetrics : EventListener
```
utpilla, line 29: nit: Good practice to add sealed to private and internal classes.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+{
+    internal static readonly AssemblyName AssemblyName = typeof(EventCountersMetrics).Assembly.GetName();
+    internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
+
+    private readonly EventCountersInstrumentationOptions options;
+    private readonly ConcurrentBag<EventSource> preInitEventSources = new();
```
utpilla, line 35: Should we be using `ConcurrentQueue` instead if we want to respect the order of creation of the EventSources?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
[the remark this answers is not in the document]
utpilla: Why is this not possible? You could go through the sources and create meters with the prefix "EventCounters." so if the user is interested in "Microsoft.AspNetCore" you could create a meter for it with the name: "EventCounters.Microsoft.AspNetCore" and then add the meter to the builder.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// <inheritdoc />
+    protected override void OnEventWritten(EventWrittenEventArgs eventData)
+    {
+        if (this.options == null)
+        {
+            return;
+        }
+
+        var payload = eventData.Payload[0] as IDictionary<string, object>;
```
cijothomas, line 77: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.EventCounters/EventCounterListener.cs#L85  Need all of these checks

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+        var name = payload["Name"].ToString();
```
utpilla, line 78: Use `TryGetValue` method to check if the dictionary contains "Name" and return if it doesn't.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+        var isGauge = payload.ContainsKey("Mean");
+        Tuple<string, string> metricKey = new(eventData.EventSource.Name, name);
+
+        if (isGauge)
+        {
+            this.values[metricKey] = Convert.ToDouble(payload["Mean"]);
```
utpilla, line 84: Use `ConcurrentDictionary.AddOrUpdate`.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        else
+        {
+            _ = this.values.TryAdd(metricKey, 0);
```
utpilla, line 88: Use `ConcurrentDictionary.GetOrAdd`.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+            this.values[metricKey] += Convert.ToDouble(payload["Increment"]);
+        }
+
+        var instrumentName = $"{eventData.EventSource.Name}.{name}";
+        _ = this.instruments.TryAdd(
```
utpilla, line 93: This is going to create an instrument on every `OnEventWritten` call. Use `ConcurrentDictionary.ContainsKey` instead.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
utpilla: Do we care about the order of creation of EventSources here?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        if (this.options == null)
+        {
+            return;
+        }
+
+        var payload = eventData.Payload[0] as IDictionary<string, object>;
+        var name = payload["Name"].ToString();
+        var isGauge = payload.ContainsKey("Mean");
+        Tuple<string, string> metricKey = new(eventData.EventSource.Name, name);
```
utpilla, line 80: This will cause allocation. Use ValueTuple instead.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
mic-max: The user code can loop through their array of names and call AddEventSource multiple times, I'd rather keep it simple

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
mic-max: Then the `eventSouceNames` HashSet would also move and then this class would be empty. I'll keep here but I kinda see your point

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: Sure that sounds like a better thing to do than possible introduce some randomness with the bag!

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
utpilla: `EventSourceNames` is also another good candidate to belong to `EventCountersMetrics` class.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
utpilla: As opposed to just passing their array to `AddEventSource` method? I think accepting `params` makes it simpler.

Refer to `AddSource` and `AddMeter` methods of the SDK:
- https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Trace/Builder/TracerProviderBuilderBase.cs#L84
- https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/Builder/MeterProviderBuilderBase.cs#L83

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
utpilla: In general, it falls more into the responsibility of the `EventCountersMetrics` class to determine what it should listen rather than an options class.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersInstrumentationOptions.cs
mic-max: Ah, if there's a precedence of this then I will make that suggested change :) wasn't aware if that's a common practice for either C# or OTel .NET

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: It will only create an instrument if `!this.instruments.ContainsKey(metricKey)` is `false`
I've just pushed some updated code though so please check again

utpilla: We have the `Meter.CreateObservableCounter`/ `Meter.CreateObservableGauge` in the `TryAdd` which would create an instrument on every `TryAdd` call.

I created a simple Benchmark with simpler data types: `int` and `string` to verify this behavior. It indeed would create the value every time as we have allocation here:

| Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|------- |---------:|---------:|---------:|-------:|----------:|
| TryAdd | 61.76 ns | 1.745 ns | 5.144 ns | 0.0357 |     224 B |

<details>

<summary>Expand for code</summary>

```csharp
using System.Collections.Concurrent;
using BenchmarkDotNet.Attributes;

namespace Benchmarks
{
    [MemoryDiagnoser]
    public class ConcurrentDictionaryBenchmarks
    {
        private readonly ConcurrentDictionary<int, string> dictionary = new();

        public ConcurrentDictionaryBenchmarks()
        {
            _ = this.dictionary.TryAdd(100, new string('A', 100));
        }

        [Benchmark]
        public void TryAdd()
        {
            _ = this.dictionary.TryAdd(100, new string('A', 100));
        }
    }
}
```

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
twenzel, line 77: payload might be null

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            if (this.options.ShouldListenToSource(source.Name))
+            {
+                this.EnableEvents(source, EventLevel.LogAlways, EventKeywords.None, GetEnableEventsArguments(this.options));
+            }
+        }
+    }
+
+    /// <inheritdoc />
+    protected override void OnEventSourceCreated(EventSource source)
```
twenzel, line 57: `source` should be `eventSource`

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,100 @@
...
+            this.values[metricKey] += Convert.ToDouble(payload["Increment"]);
+        }
+
+        var instrumentName = $"{eventData.EventSource.Name}.{name}";
+        _ = this.instruments.TryAdd(
+            metricKey,
+            isGauge ? MeterInstance.CreateObservableGauge(instrumentName, () => this.values[metricKey]) : MeterInstance.CreateObservableCounter(instrumentName, () => this.values[metricKey]));
+    }
+
+    private static Dictionary<string, string> GetEnableEventsArguments(EventCountersInstrumentationOptions options) =>
```
twenzel, line 98: Return type need to be changed to `Dictionary<string, string?>`

mic-max: Why, `EnableEvents` takes an `IDictionary<string, string>` as the `arguments` parameter?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -74,27 +78,80 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
+        this.EventWritten(hasMean, eventSourceName, name, value);
     }
 
     private static Dictionary<string, string> GetEnableEventsArguments(EventCountersInstrumentationOptions options) =>
         new() { { "EventCounterIntervalSec", options.RefreshIntervalSecs.ToString() } };
+
+    private static bool IsValidInstrumentName(string name) =>
+        !string.IsNullOrWhiteSpace(name) && InstrumentNameRegex.IsMatch(name);
+
+    private void EventWritten(bool isGauge, string eventSourceName, string name, double value)
```
twenzel, line 125: Could you rename this method because it hides the inherited member 'EventWritten'.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: Yes I agree with this. In the surrounding scope I do have a `if (!this.instruments.ContainsKey(metricKey))` check though which might have been hidden in the GitHub code preview?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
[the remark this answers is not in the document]
mic-max: Ya I guess that's true, I'll leave this as a potential future update after getting merged and released to some test customers and gather feedback :)

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,158 @@
...
+            return;
+        }
+
+        if (eventData.Payload == null || eventData.Payload.Count == 0 || eventData.Payload[0] is not IDictionary<string, object>)
+        {
+            EventCountersInstrumentationEventSource.Log.IgnoreEventWrittenEventArgsPayloadNotParseable(eventSourceName);
+            return;
+        }
+
+        var payload = eventData.Payload[0] as IDictionary<string, object>;
```
utpilla, line 95: nit: Update the `if` check above to cast it instead:
`.. || eventData.Payload[0] is not IDictionary<string, object> payload)`

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,158 @@
...
+    private void UpdateInstrumentWithEvent(bool isGauge, string eventSourceName, string name, double value)
+    {
+        try
+        {
+            ValueTuple<string, string> metricKey = new(eventSourceName, name);
+            _ = this.values.AddOrUpdate(metricKey, value, isGauge ? (_, _) => value : (_, existing) => existing + value);
+
+            var instrumentName = $"EventCounters.{eventSourceName}.{name}";
+
+            if (!IsValidInstrumentName(instrumentName))
```
utpilla, line 134: You can remove the checks for instrument name. It's already done at the SDK: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MeterProviderSdk.cs#L273-L282

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            if (!this.instruments.ContainsKey(metricKey))
+            {
+                Instrument instrument = isGauge
+                    ? MeterInstance.CreateObservableGauge(instrumentName, () => this.values[metricKey])
+                    : MeterInstance.CreateObservableCounter(instrumentName, () => this.values[metricKey]);
+                _ = this.instruments.TryAdd(metricKey, instrument);
+
+                if (!this.instrumentNames.TryAdd(instrumentName, 0))
```
utpilla, line 147: No need of checking duplicate instrument names either. These things are done centrally at the SDK setup:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MetricReaderExt.cs#L54-L61

utpilla:
The PR looks good. There are some duplicate checks with instrument names that should be avoided.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: True, I guess I was adding these checks prematurely before I figure out a good way to map EventCounter names into instrument names.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+        else if (this.options.ShouldListenToSource(source.Name))
+        {
+            this.EnableEvents(source, EventLevel.LogAlways, EventKeywords.None, GetEnableEventsArguments(this.options));
+        }
+    }
+
+    /// <inheritdoc />
+    protected override void OnEventWritten(EventWrittenEventArgs eventData)
+    {
```
utpilla, line 69: Could `eventData` be `null`?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,137 @@
...
+        var hasMean = payload.TryGetValue("Mean", out var mean);
+        var hasIncrement = payload.TryGetValue("Increment", out var increment);
+
+        if (!(hasIncrement ^ hasMean))
+        {
+            EventCountersInstrumentationEventSource.Log.IgnoreMeanIncrementConflict(eventSourceName);
+            return;
+        }
+
+        var value = Convert.ToDouble(hasMean ? mean : increment);
```
utpilla, line 108: You could move the `try-catch` block from `UpdateInstrumentWithEvent` to here to include the `Convert.ToDouble` method in the `try` block as that could also throw exceptions.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -0,0 +1,137 @@
...
+        else if (this.options.ShouldListenToSource(source.Name))
+        {
+            this.EnableEvents(source, EventLevel.LogAlways, EventKeywords.None, GetEnableEventsArguments(this.options));
+        }
+    }
+
+    /// <inheritdoc />
+    protected override void OnEventWritten(EventWrittenEventArgs eventData)
+    {
+        if (this.options == null)
```
utpilla, line 70: Is this check required?

utpilla (approved):

utpilla (approved):
Left some suggestions. They are not blocking this PR. We would also have to enable Public API Analyzer for this project.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
mic-max: Ya, just dropping an event if the constructor hasn't finished setting the `options` field. I'll add comments as this is an obscure situation

utpilla: We never enable events until the options are set, right? If options is `null` we just enqueue it and wait till the ctor has set options before proceeding to enable events for the enqueued EventSources.

mic-max: Ya this is to avoid creating instruments for event sources we don't care about.

PR #623, opened 2022-09-06
OpenTelemetry.Extensions.AzureMonitor - Add sampleRate attribute to SamplingResult.

rajkumar-rangaraj:
## Changes

On successful sampling, add `sampleRate` attribute to `SamplingResult`.

TODO
* Update `CHANGELOG.md` in a follow up PR.

### DIFF
--- a/src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
+++ b/src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 using System;
+using System.Collections.Generic;
 using OpenTelemetry.Internal;
 using OpenTelemetry.Trace;
 
@@ -26,6 +27,8 @@ namespace OpenTelemetry.Extensions.AzureMonitor;
 public class ApplicationInsightsSampler : Sampler
 {
     private readonly float samplingRatio;
+    private readonly SamplingResult dropSamplingResult;
+    private readonly SamplingResult recordAndSampleSamplingResult;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="ApplicationInsightsSampler"/> class.
@@ -38,6 +41,14 @@ public class ApplicationInsightsSampler : Sampler
 
         this.samplingRatio = samplingRatio;
         this.Description = "ApplicationInsightsSampler{" + samplingRatio + "}";
+        var sampleRate = (int)Math.Round(samplingRatio * 100);
+        this.dropSamplingResult = new SamplingResult(false);
+        this.recordAndSampleSamplingResult = new SamplingResult(
+            SamplingDecision.RecordAndSample,
+            new Dictionary<string, object>
+                {
+                    { "sampleRate", sampleRate },
+                });
     }
 
     /// <summary>
@@ -48,8 +59,26 @@ public class ApplicationInsightsSampler : Sampler
     /// <returns>Returns whether or not we should sample telemetry in the form of a <see cref="SamplingResult"/> class.</returns>
     public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
     {
+        if (this.samplingRatio == 0)
+        {
+            return this.dropSamplingResult;
+        }
+
+        if (this.samplingRatio == 1)
+        {
+            return this.recordAndSampleSamplingResult;
+        }
+
         double sampleScore = DJB2SampleScore(samplingParameters.TraceId.ToHexString().ToLowerInvariant());
-        return new SamplingResult(sampleScore < this.samplingRatio);
+
+        if (sampleScore < this.samplingRatio)
+        {
+            return this.recordAndSampleSamplingResult;
+        }
+        else
+        {
+            return this.dropSamplingResult;
+        }
     }
 
     private static double DJB2SampleScore(string traceIdHex)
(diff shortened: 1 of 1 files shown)


TimothyMothra (approved):

cijothomas (approved):
LGTM with a changelog

vishweshbankwar (approved):

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
```diff
@@ -48,8 +59,26 @@ public ApplicationInsightsSampler(float samplingRatio)
     /// <returns>Returns whether or not we should sample telemetry in the form of a <see cref="SamplingResult"/> class.</returns>
     public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
     {
+        if (this.samplingRatio == 0)
```
utpilla, line 62: Are we okay with the naive equality comparison for `float`?

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
```diff
@@ -26,6 +27,8 @@ namespace OpenTelemetry.Extensions.AzureMonitor;
 public class ApplicationInsightsSampler : Sampler
 {
     private readonly float samplingRatio;
+    private readonly SamplingResult dropSamplingResult;
```
utpilla, line 30: This could be a `static` field.

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
```diff
@@ -38,6 +41,14 @@ public ApplicationInsightsSampler(float samplingRatio)
 
         this.samplingRatio = samplingRatio;
         this.Description = "ApplicationInsightsSampler{" + samplingRatio + "}";
+        var sampleRate = (int)Math.Round(samplingRatio * 100);
```
utpilla, line 44: Not added by this PR, but is there any particular reason to prefer the type `float` over `double` for `samplingRatio`?

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
rajkumar-rangaraj: I don't see an issue, do you think this could cause an issue?

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
```diff
@@ -38,6 +41,14 @@ public ApplicationInsightsSampler(float samplingRatio)
...
+        this.dropSamplingResult = new SamplingResult(false);
+        this.recordAndSampleSamplingResult = new SamplingResult(
+            SamplingDecision.RecordAndSample,
+            new Dictionary<string, object>
+                {
+                    { "sampleRate", sampleRate },
```
cijothomas, line 50: `samplingRatio` would be nicer?

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
rajkumar-rangaraj: It's based on Application Insights Sampler Specification.

rajkumar-rangaraj:
> LGTM with a changelog

As specified in description TODO, want add changelog in a follow up PR as this project does not have any entry earlier.

src/OpenTelemetry.Extensions.AzureMonitor/ApplicationInsightsSampler.cs
utpilla: It depends on the sampler requirements. Do we want to sample and record for this value of sampling ratio `1.0000001`?

```csharp
float samplingRatio = 1.0000001F;
if (samplingRatio == 1) // <- This is not true
{

}
```

rajkumar-rangaraj: `1.0000001` is an invalid valid. Guard in the constructor - `Guard.ThrowIfOutOfRange((double)samplingRatio, min: 0.0, max: 1.0);` should throw an exception.

utpilla: In general, equality comparison of floating-point values is problematic: https://docs.microsoft.com/en-us/dotnet/api/system.double?view=net-6.0#testing-for-equality

rajkumar-rangaraj: This should **not** be treated as 1. Tested the behavior in a small test app, it works perfectly. I could understand the problem with float number, but it should not have impact on 0 and 1. As a follow up task we could add tests for this project to validate it.

utpilla: > This should not be treated as 1.

I was essentially asking this.

I know that the current code does not treat it as 1. I just wanted to know if we should be treating `0.9999999F` as `1.0` and `0.00000001F` as `0.0` as when the user provides the samplingRatio, it might have been computed as result of other floating-point operation which might cause precision loss.

utpilla (approved):

PR #625, opened 2022-09-07
[Instrumentation.Process] Added the metrics for CpuTime and CpuUtilization.

Yun-Ting:
## Changes

For the `process.cpu.time` metric, [Process.TotalProcessorTime](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.totalprocessortime?view=net-6.0) was used to get the value.

For the `process.cpu.utilization` metric, it was defined as the "difference in process.cpu.time since the last measurement, divided by the elapsed time and number of CPUs available to the process" in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#metric-instruments).
In this PR, the last measurement was defined as the last time `GetCpuUtilization()` callback function was invoked. 
At the collection phase of the metricReader of the Exporter, `GetCpuUtilization()` will be triggered.
The `lastCollectionTimestamp ` and `lastCollectionCpuTime ` will be updated before the current CPU utilization result was returned.

The unit defined in the spec for the above two mentioning CPU metrics are "s" (second). In this PR, the values were being calculated by converting [TimeSpan.Ticks](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.ticks?view=net-6.0)
to seconds for more granularity.

Please note that currently, broken down by different states , i.e. `user`, `system`, or `wait` for CPU related metrics is not implemented in this PR because of the following reasons.
1. There is no easy way to get the `wait` state from the existing Process releated APIs from .net.
2. I am unsure how beneficial to the user given this broken down by states information.

The plan is to release an alpha version of the `OpenTelemetry.Instrumentation.Process` package and collect customer feedback. And based on the feedback, we decide whether the division by states feature should be implemented for CPU-related metrics or not.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -14,6 +14,7 @@
 // limitations under the License.
 // </copyright>
 
+using System;
 using System.Diagnostics.Metrics;
 using System.Reflection;
 using Diagnostics = System.Diagnostics;
@@ -24,31 +25,40 @@ internal class ProcessMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
-    private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
 
     static ProcessMetrics()
     {
+        // Refer to the spec for instrument details:
+        // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/process-metrics.md#metric-instruments
+
+        InstrumentsValues values = new InstrumentsValues(() => (-1, -1, -1, -1));
+
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.usage",
-            () =>
-            {
-                CurrentProcess.Refresh();
-                return CurrentProcess.WorkingSet64;
-            },
+            () => values.GetMemoryUsage(),
             unit: "By",
-            description: "The amount of physical memory in use.");
+            description: "The amount of physical memory in use by the current process.");
 
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.virtual",
-            () =>
-            {
-                CurrentProcess.Refresh();
-                return CurrentProcess.VirtualMemorySize64;
-            },
+            () => values.GetVirtualMemoryUsage(),
             unit: "By",
-            description: "The amount of committed virtual memory.");
+            description: "The amount of committed virtual memory by the current process.");
+
+        MeterInstance.CreateObservableCounter(
+            $"process.cpu.time",
+            () => values.GetCpuTime(),
+            unit: "s",
+            description: "The amount of time that the current process has actively used a CPU to perform computations.");
+
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            $"process.cpu.utilization",
+            () => values.GetCpuUtilization(),
+            unit: "s",
+            description: "Difference in process.cpu.time since the last collection of observerble instruments from the MetricReader, divided by the elapsed time multiply by the number of CPUs available to the current process.");
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -58,4 +68,86 @@ static ProcessMetrics()
     public ProcessMetrics(ProcessInstrumentationOptions options)
     {
     }
+
+    private class InstrumentsValues
+    {
+        private const int TicksPerSecond = 10 * 7;
```
tarekgh, line 74: This doesn't look correct to me. 

https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs,61

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -58,4 +68,86 @@ static ProcessMetrics()
...
+        private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
+
+        private double? memoryUsage;
+        private double? virtualMemoryUsage;
+        private double? cpuTime;
+        private double? cpuUtlization;
+
+        private DateTime lastCollectionTimestamp = DateTime.Now;
```
tarekgh, line 82: Use `DateTime.UtcNow` instead of `DateTime.Now`. It is much faster.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -58,4 +68,86 @@ static ProcessMetrics()
...
+            if (!this.cpuUtlization.HasValue)
+            {
+                (this.memoryUsage, this.virtualMemoryUsage, this.cpuTime, this.cpuUtlization) = this.UpdateValues();
+            }
+
+            var elapsedTime = (DateTime.Now - this.lastCollectionTimestamp).Ticks * TicksPerSecond;
+            var currentCpuTime = this.currentProcess.TotalProcessorTime.Ticks * TicksPerSecond;
+            var deltaInCpuTime = currentCpuTime - this.lastCollectionCpuTime;
+
+            this.lastCollectionTimestamp = DateTime.Now;
```
tarekgh, line 145: This will be `DateTime.UtcNow` too.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -58,4 +68,86 @@ static ProcessMetrics()
...
+        }
+
+        public double GetCpuUtilization()
+        {
+            if (!this.cpuUtlization.HasValue)
+            {
+                (this.memoryUsage, this.virtualMemoryUsage, this.cpuTime, this.cpuUtlization) = this.UpdateValues();
+            }
+
+            var elapsedTime = (DateTime.Now - this.lastCollectionTimestamp).Ticks * TicksPerSecond;
```
tarekgh, line 141: To convert ticks to seconds, you need to divide the ticks by the TicksPerSecond.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,31 +25,40 @@ internal class ProcessMetrics
...
+            () => values.GetVirtualMemoryUsage(),
             unit: "By",
-            description: "The amount of committed virtual memory.");
+            description: "The amount of committed virtual memory by the current process.");
+
+        MeterInstance.CreateObservableCounter(
+            $"process.cpu.time",
+            () => values.GetCpuTime(),
+            unit: "s",
+            description: "The amount of time that the current process has actively used a CPU to perform computations.");
```
tarekgh, line 54: would be good to specify the measurement time unit in the description

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,31 +25,40 @@ internal class ProcessMetrics
...
+
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            $"process.cpu.utilization",
+            () => values.GetCpuUtilization(),
+            unit: "s",
+            description: "Difference in process.cpu.time since the last collection of observerble instruments from the MetricReader, divided by the elapsed time multiply by the number of CPUs available to the current process.");
```
tarekgh, line 61: `CPU utilization percentage since last collection` would be better?

The unit should be a percentage here?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: right, this is embarrassing, thanks for pointing it out.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
reyang: In addition, these are not monotonic, I think for duration we should use something that measures elapsed time rather than wall clock.

PR #626, opened 2022-09-09
[Instrumentation.AWSLambda] Issue/aws lambda http server semantic conventions attributes

rypdal:
## Changes

Added [HTTP Server semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server)s required tags to `Activity `created in `OpenTelemetry.Instrumentation.AWSLambda`.


Oberon00:
Hi, I would propose we merge this only after the first release is cut in #590.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+                    var hostHeaderV2 = GetHeaderValue(requestV2, HeaderHost);
+                    (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeaderV2);
+                    break;
+            }
+
+            tags.AddStringTagIfNotNull(SemanticConventions.AttributeHttpScheme, httpScheme);
+            tags.AddStringTagIfNotNull(SemanticConventions.AttributeHttpTarget, httpTarget);
+            tags.AddStringTagIfNotNull(SemanticConventions.AttributeHttpMethod, httpMethod);
+            tags.AddStringTagIfNotNull(SemanticConventions.AttributeNetHostName, hostName);
+            tags.AddStringTagIfNotNull(SemanticConventions.AttributeNetHostPort, hostPort);
```
Oberon00, line 63: The port must be an integer, not a string.

rypdal: Right. Fixed.

rypdal:
This PR was waiting for the next -beta release. The release was published and PR can be further reviewed and merged.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            if (targetKey != default)
+            {
+                value = dict[targetKey];
+                return true;
+            }
+
+            return false;
+        }
+
+        internal static void AddTagIfNotNull(this List<KeyValuePair<string, object>> tags, string tagName, object tagValue)
```
Oberon00, line 41: This looks very generic. Is there an utility function like that already somewhere in this repository or the SDK itself that we could use?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+                    httpScheme = GetHeaderValue(requestV2, HeaderXForwardedProto);
+                    httpTarget = requestV2?.RequestContext?.Http?.Path;
+                    httpMethod = requestV2?.RequestContext?.Http?.Method;
+                    var hostHeaderV2 = GetHeaderValue(requestV2, HeaderHost);
+                    (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeaderV2);
+                    break;
+            }
+
+            tags.AddTagIfNotNull(SemanticConventions.AttributeHttpScheme, httpScheme);
+            tags.AddTagIfNotNull(SemanticConventions.AttributeHttpTarget, httpTarget);
```
Oberon00, line 60: The target should include any query string, please check if that is the case, or if that is stored in a separate attribute.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+                case APIGatewayProxyRequest request:
+                    httpScheme = GetHeaderValue(request, HeaderXForwardedProto);
+                    httpTarget = request.Path;
+                    httpMethod = request.HttpMethod;
+                    var hostHeader = GetHeaderValue(request, HeaderHost);
+                    (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeader);
+                    break;
+                case APIGatewayHttpApiV2ProxyRequest requestV2:
+                    httpScheme = GetHeaderValue(requestV2, HeaderXForwardedProto);
+                    httpTarget = requestV2?.RequestContext?.Http?.Path;
```
Oberon00, line 52: There is also requestV2.RawPath. I wonder what the difference is. But unless we find out, that you currently use is probably fine.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+            tags.AddTagIfNotNull(SemanticConventions.AttributeHttpScheme, httpScheme);
+            tags.AddTagIfNotNull(SemanticConventions.AttributeHttpTarget, httpTarget);
+            tags.AddTagIfNotNull(SemanticConventions.AttributeHttpMethod, httpMethod);
+            tags.AddTagIfNotNull(SemanticConventions.AttributeNetHostName, hostName);
+            tags.AddTagIfNotNull(SemanticConventions.AttributeNetHostPort, hostPort);
+
+            return tags;
+        }
+
+        internal static void SetHttpTagsFromResult<TResult>(Activity activity, TResult result)
```
Oberon00, line 68:
```suggestion
        internal static void SetHttpTagsFromResult(Activity activity, object result)
```

It does not look like the generic buys us anything here.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+        internal static (string Host, int? Port) GetHostAndPort(string httpScheme, string hostHeaders)
+        {
+            if (hostHeaders == null)
+            {
+                return (null, null);
+            }
+
+            // In case of multiple headres we consider only the 1st.
+            var hostHeader = hostHeaders.Split(',').First();
+            var hostAndPort = hostHeader.Split(':');
```
Oberon00, line 90: For efficiency, you could specify the maximum number of splittings.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
```diff
@@ -0,0 +1,140 @@
...
+            if (hostAndPort.Length > 1)
+            {
+                return (hostAndPort[0], int.Parse(hostAndPort[1]));
```
Oberon00, line 93: We should probably be careful about int parsing errors here, maybe use `int.TryParse` and treat the port as not present in case of errors.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -197,51 +203,55 @@ private static void OnException(Activity activity, Exception exception)
...
         ActivityContext parentContext = default)
     {
-        var lambdaActivity = OnFunctionStart(input, context, parentContext);
+        var activity = OnFunctionStart(input, context, parentContext);
         try
         {
-            handler();
+            var result = handler(input, context);
+            HttpSemanticConventions.SetHttpTagsFromResult(activity, result);
+            return result;
```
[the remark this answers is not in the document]
rypdal: Usually we don't expect any return value from the method prefixed with `Set`.  May be another name ?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OpenTelemetry.Instrumentation.AWSLambda.Implementation
+{
+    internal static class CommonExtensions
+    {
+        internal static bool TryGetValueIgnoringCase<T>(this IDictionary<string, T> dict, string key, out T value)
```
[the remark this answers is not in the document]
rypdal: From other side in general HTTP headers are case-insensitive (see for example [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)).

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
rypdal: I've found such method.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            value = default;
+            var targetKey = dict.Keys
+                .Where(s => string.Equals(s, key, StringComparison.CurrentCultureIgnoreCase))
+                .FirstOrDefault();
```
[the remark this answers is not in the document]
rypdal: Not actual after changes. Method uses foreach and iterate only once.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            if (targetKey != default)
+            {
+                value = dict[targetKey];
```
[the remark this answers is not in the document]
rypdal: Method uses foreach and iterate only once after change.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
rypdal: Changed to object and committed along with other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
rypdal: Yes, lets use `int.TryParse `, although I can't imagine the case when port is not an integer.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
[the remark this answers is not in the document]
Oberon00: HTTP headers being case insensitive is what enables AWS to lowercase them actually. If they were case-sensitive, they could not / would not need to do that.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
Oberon00: This is untrusted input. So maybe someone who wants to bypass monitoring could use this for their advantage 🤷

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                if (string.Equals(kvp.Key, key, StringComparison.OrdinalIgnoreCase))
+                {
+                    value = kvp.Value;
+                    foundTargetKey = true;
+                    break;
+                }
+            }
+
+            return foundTargetKey;
```
Oberon00, line 38:
```suggestion
            foreach (var kvp in dict)
            {
                if (string.Equals(kvp.Key, key, StringComparison.OrdinalIgnoreCase))
                {
                    value = kvp.Value;
                    return true;
                }
            }

            value = default;
            return false;
```

Adding boolean control variables makes the code much harder to read IMHO.

rypdal: I was thinking to keep only the `value` variable and compare with the `default`, but what if value stored in `dict` is equal to default ?

Oberon00: `default` is `null` here, isn't it? So that would work as well.

rypdal: This method has been removed.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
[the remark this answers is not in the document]
rypdal: This method has been removed.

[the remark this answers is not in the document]
rypdal: Now, I'm a little bit in doubts. If you consider an example of HTTP triggered AWS Lambda function and go to the 'Test' tab where there is prepared example of JSON Event, you may see that header names are not lower case in the example.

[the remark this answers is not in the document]
rypdal: Added ignoring of header name case for both API Gateway versions.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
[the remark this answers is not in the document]
rypdal: New method transforming dictionary keys to lower case has been introduced.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
rypdal: Checked both API Gateway versions: `APIGatewayProxyRequest.Path` and `APIGatewayHttpApiV2ProxyRequest.RequestContext.Http.Path` do not contain query parameters. Added them to resulting `httpTarget`.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            return tags;
+        }
+
+        internal static void SetHttpTagsFromResult(Activity activity, object result)
+        {
+            switch (result)
+            {
+                case APIGatewayProxyResponse response:
+                    activity.SetTag(SemanticConventions.AttributeHttpStatusCode, response.StatusCode.ToString());
```
Oberon00, line 74: BLOCKING: The status code should be an integer, not a string: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        internal static (string Host, int? Port) GetHostAndPort(string httpScheme, string hostHeaders)
+        {
+            if (hostHeaders == null)
+            {
+                return (null, null);
+            }
+
+            // In case of multiple headres we consider only the 1st.
```
Oberon00, line 89:
```suggestion
            // In case of multiple headers we consider only the 1st.
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+            var hostHeader = hostHeaders.Split(',').First();
```
Oberon00, line 90: You could limit the split again for efficiency (split into at most 2, since we are only interested in the first). Although it would probably be even better (at least more readable) to use the multi-header getter instead, so we don't need to split at all here.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+                int? defaultPort = null;
+                switch (httpScheme)
+                {
+                    case "http":
+                        defaultPort = 80;
+                        break;
+                    case "https":
+                        defaultPort = 443;
+                        break;
+                }
```
Oberon00, line 110:
```suggestion
                int? defaultPort = httpScheme == "https" ? 443 : httpScheme == "http" ? 80 : null;
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+                return string.Empty;
+            }
+
+            var items = new List<string>();
+            foreach (string name in requestParameters.Keys)
+            {
+                items.Add(string.Concat(name, "=", HttpUtility.UrlEncode(requestParameters[name])));
+            }
+
+            return string.Concat("?", string.Join("&", items.ToArray()));
```
Oberon00, line 144: It seems more efficient to add directly to a StringBuilder instead of constructing an intermediate `List<string>`.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if (tagValue != null)
+            {
+                tags.Add(new(tagName, tagValue));
+            }
+        }
+
+        internal static T GetValueByKeyIgnoringCase<T>(this IDictionary<string, T> headers, string key)
+        {
+            var headersWithLowerCaseNames = headers?.KeysToLowerCase();
```
Oberon00, line 34: This is less efficient than the old approach, since we convert the dictionary each time. We need to reuse the result of KeysToLowerCase. If we cannot do that, the old approach with the linear case-insensitive lookup is more efficient.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+                headersWithLowerCaseNames.TryGetValue(key, out var value))
+            {
+                return value;
+            }
+
+            return default;
+        }
+
+        private static Dictionary<string, T> KeysToLowerCase<T>(this IDictionary<string, T> input) =>
+            input?.ToDictionary(x => x.Key.ToLower(), x => x.Value);
```
Oberon00, line 46:
```suggestion
            input?.ToDictionary(x => x.Key.ToLowerInvariant(), x => x.Value);
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+                case APIGatewayProxyRequest request:
+                    httpScheme = GetHeaderValue(request, HeaderXForwardedProto);
+                    httpTarget = string.Concat(request.Path ?? string.Empty, ConstructQueryString(request.QueryStringParameters));
+                    httpMethod = request.HttpMethod;
+                    var hostHeader = GetHeaderValue(request, HeaderHost);
+                    (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeader);
+                    break;
+                case APIGatewayHttpApiV2ProxyRequest requestV2:
+                    httpScheme = GetHeaderValue(requestV2, HeaderXForwardedProto);
+                    httpTarget = string.Concat(requestV2?.RequestContext?.Http?.Path ?? string.Empty, ConstructQueryString(requestV2.QueryStringParameters));
```
Oberon00, line 53: Let's use the `rawQueryString` property of the V2 request instead.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
```diff
@@ -0,0 +1,147 @@
...
+            string httpTarget = null;
+            string httpMethod = null;
+            string hostName = null;
+            int? hostPort = null;
+
+            switch (input)
+            {
+                case APIGatewayProxyRequest request:
+                    httpScheme = GetHeaderValue(request, HeaderXForwardedProto);
+                    httpTarget = string.Concat(request.Path ?? string.Empty, ConstructQueryString(request.QueryStringParameters));
```
Oberon00, line 46: I think it would be more correct to use `multiValueQueryStringParameters`. If you prefer, we can defer this to a follow-up issue though.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -198,6 +193,11 @@ private static IEnumerable<string> GetHeaderValues(APIGatewayProxyRequest reques
         {
             return values;
         }
+        else if (request.Headers != null &&
```
Oberon00, line 196: If we think we need to be case-insensitive, we need to be case-insensitive in the multi-header variant as well.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaHttpUtilTests.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        [Theory]
+        [InlineData(null, null, null, null)]
+        [InlineData("", "", "", null)]
+        [InlineData(null, "localhost:4321", "localhost", 4321)]
+        [InlineData(null, "localhost:4321, myhost.com:9876", "localhost", 4321)]
+        [InlineData(null, "localhost", "localhost", null)]
+        [InlineData("http", "localhost", "localhost", 80)]
+        [InlineData("https", "localhost", "localhost", 443)]
+        public void GetHostAndPort_HostHeader_ReturnsCorrectHostAndPort(string httpSchema, string hostHeader, string expectedHost, int? expectedPort)
```
Oberon00, line 152: Please also test the handling of unparseable (non-integer) ports.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
rypdal: Introduced `GetDefaultPort` method and moved this simplification there.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
rypdal: Using StringBuilder requires some extra code for considering "&" for the first (or the last) parameter. With this code all work is done by the Join method.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
rypdal: Reverted approach and simplified previous code.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
rypdal: Not actual.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/HttpSemanticConventions.cs
rypdal: In the new implementation the  `requestV2.RawPath` is used.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
rypdal: The new implementation uses the `RawQueryString` property as is and appends it to the `RawPath` after `?`.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtil.cs
rypdal: The new implementation uses the `MultiValueQueryStringParameters` property. Multiple values for the same parameter will be added to query as ampersand separated: `name=value1&name=value2`.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Introduced a single method GetHeaderValues for each type of HTTP trigger and used in both places: as the propagator getter and here for header values extraction.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaHttpUtilTests.cs
rypdal: Add non-integer ports case.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            foreach (var kvp in dict)
+            {
+                if (string.Equals(kvp.Key, key, StringComparison.OrdinalIgnoreCase))
+                {
+                    value = kvp.Value;
+                    break;
+                }
+            }
+
+            return value;
```
Oberon00, line 49:
```suggestion
            foreach (var kvp in dict)
            {
                if (string.Equals(kvp.Key, key, StringComparison.OrdinalIgnoreCase))
                {
                    return kvp.Value;
                }
            }

            return default;
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            return items.Count > 0
+                ? string.Concat("?", string.Join("&", items.ToArray()))
+                : string.Empty;
+        }
+
+        internal static string ConstructQueryString(APIGatewayHttpApiV2ProxyRequest request) =>
+            !string.IsNullOrEmpty(request.RawQueryString)
+            ? string.Concat("?", request.RawQueryString)
+            : string.Empty;
```
Oberon00, line 108:
```suggestion
            string.IsNullOrEmpty(request.RawQueryString) ? string.Empty : "?" + request.RawQueryString;
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,134 @@
...
+                // Multiple values for the same parameter will be added to query
+                // as ampersand separated: name=value1&name=value2
+                foreach (var value in request.MultiValueQueryStringParameters[name])
+                {
+                    items.Add(string.Concat(name, "=", HttpUtility.UrlEncode(value)));
+                }
+            }
+
+            return items.Count > 0
+                ? string.Concat("?", string.Join("&", items.ToArray()))
```
Oberon00, line 101:
```suggestion
                ? string.Concat("?", string.Join("&", items))
```

There is no need to use ToArray here.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,134 @@
...
+            var items = new List<string>();
+            foreach (var name in request.MultiValueQueryStringParameters.Keys)
+            {
+                // Multiple values for the same parameter will be added to query
+                // as ampersand separated: name=value1&name=value2
+                foreach (var value in request.MultiValueQueryStringParameters[name])
+                {
+                    items.Add(string.Concat(name, "=", HttpUtility.UrlEncode(value)));
+                }
+            }
```
Oberon00, line 98: This nested foreach looks like it could be a SelectMany. Although performance might suffer, so it's up to you.

```suggestion
            var items = request.MultiValueQueryStringParameters.SelectMany(kv => kv.Value.Select($"{name}={HttpUtility.UrlEncode(value)});
```

If you keep using a foreach, you should avoid the `[]`-lookup and instead iterate key-value-pairs in the outer loop.

(also consider the previous comment about encoding the key/name which I did not address here)

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,134 @@
...
+                case APIGatewayProxyResponse response:
+                    activity.SetTag(SemanticConventions.AttributeHttpStatusCode, response.StatusCode);
+                    break;
+                case APIGatewayHttpApiV2ProxyResponse responseV2:
+                    activity.SetTag(SemanticConventions.AttributeHttpStatusCode, responseV2.StatusCode);
+                    break;
+            }
+        }
+
+        internal static string ConstructQueryString(APIGatewayProxyRequest request)
```
Oberon00, line 82: Maybe just:

```suggestion
        internal static string GetQueryString(APIGatewayProxyRequest request)
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -137,6 +137,30 @@ internal static string GetFunctionVersion()
...
+
+        return headerValue != null ? new[] { headerValue } : null;
+    }
+
+    internal static IEnumerable<string> GetHeaderValues(APIGatewayHttpApiV2ProxyRequest request, string name)
+    {
+        var headreValue = GetHeaderValue(request, name);
+
+        // Multiple values for the same header will be separated by a comma.
+        return headreValue?.Split(',');
```
Oberon00, line 158:
```suggestion
        var headerValue = GetHeaderValue(request, name);

        // Multiple values for the same header will be separated by a comma.
        return headerValue?.Split(',');
```

Oberon00:
I think this now looks mostly done, only a few nitpicks left. 👍 Thank you for addressing my comments!

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
rypdal: Then tests have to be renamed too.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,134 @@
...
+            }
+
+            var items = new List<string>();
+            foreach (var name in request.MultiValueQueryStringParameters.Keys)
+            {
+                // Multiple values for the same parameter will be added to query
+                // as ampersand separated: name=value1&name=value2
+                foreach (var value in request.MultiValueQueryStringParameters[name])
+                {
+                    items.Add(string.Concat(name, "=", HttpUtility.UrlEncode(value)));
```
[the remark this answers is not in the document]
rypdal: Lets keep it simple and do not encode both. Removed encoding of the value.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
rypdal: I would prefer to keep the nested loop as not sure about performance of  `SelectMany`. Made another optimisation: got rid of the List<T> and used StringBuilder instead.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
rypdal: Suggestion committed along with tests renaming and other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
rypdal: Committed along other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            var queryString = new StringBuilder();
+            var separator = '?';
+            foreach (var parameterKvp in request.MultiValueQueryStringParameters)
+            {
+                // Multiple values for the same parameter will be added to query
+                // as ampersand separated: name=value1&name=value2
+                foreach (var value in parameterKvp.Value)
+                {
+                    queryString.Append(string.Concat(separator, parameterKvp.Key, "=", value));
```
Oberon00, line 97: If we are using a string builder, I think we should rely on it for concatenation and not allocate additional string. It looks like there is even an overload for interpolated strings <https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.append?view=net-6.0#system-text-stringbuilder-append(system-text-stringbuilder-appendinterpolatedstringhandler@)>

So this should be among the most efficient ways:

```suggestion
                    queryString.Append($"{separator}{parameterKvp.Key}={value}");
```

As the string builder will internally evaluate the $-string by calling Append multiple times. See here: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/stringbuilder-append-evaluation-order#new-behavior

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -0,0 +1,135 @@
...
+                    if (separator == '?')
+                    {
+                        separator = '&';
+                    }
```
Oberon00, line 101:
```suggestion
                    separator = '&';
```

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
Oberon00: Sorry, on second thought, while this is a very cool feature, it requires not only C# 10, but also .NET 6 (which we can't rely on). So it would probably be better to spell out the `Append` manually.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
Oberon00: I think this conversation can be resolved as it became obsolete

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -141,17 +147,14 @@ public static class AWSLambdaWrapper
...
-    public static async Task<TResult> TraceAsync<TInput, TResult>(
+    public static Task<TResult> TraceAsync<TInput, TResult>(
         TracerProvider tracerProvider,
         Func<TInput, ILambdaContext, Task<TResult>> lambdaHandler,
         TInput input,
         ILambdaContext context,
         ActivityContext parentContext = default)
     {
-        TResult result = default;
-        Func<Task> action = async () => result = await lambdaHandler(input, context);
```
Oberon00, line 152: d

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/CommonExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        internal static void AddTagIfNotNull(this List<KeyValuePair<string, object>> tags, string tagName, object tagValue)
+        {
+            if (tagValue != null)
+            {
+                tags.Add(new(tagName, tagValue));
+            }
+        }
+
+        internal static T GetValueByKeyIgnoringCase<T>(this IDictionary<string, T> dict, string key)
```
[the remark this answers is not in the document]
Oberon00: Thank you! I think this can be resolved.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -141,17 +147,14 @@ public static class AWSLambdaWrapper
...
-        await TraceInternalAsync(tracerProvider, action, input, context, parentContext);
```
Oberon00, line 153: (test comment please ignore)

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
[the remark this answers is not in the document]
Oberon00: While simplicity is good, correctness is more important. We have to check which of these are already URL-decoded and re-encode them. Otherwise this will break if there are special characters like #/?/=/& in the name or value.

Oberon00 (changes requested):
We need make sure the URL-encoding of the query string parts is correct, see
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/626#discussion_r983450941

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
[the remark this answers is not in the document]
rypdal: AWS checks show that values not encoded and we need to encode `name` and `value`. So, now both `name` and `value `are encoded.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
rypdal: Done

Oberon00 (approved):

rypdal:
> Hi, I would propose we merge this only after the first release is cut in #590.

Hi@utpilla , the corresponding release already happened and this PR can be merged. The PR requires one more approval.

utpilla (approved):
@rypdal Update the CHANGELOG.

Oberon00:
> this property shouldn't be set if neither path nor query string is provided

Is this possible? I don't think that would be a HTTP request then. Maybe we should fall back to `/` or empty string in this case?

rypdal:
> I think I would prefer always treating Path as if it was empty if it is actually null. But do not add extra code to try to get something sensible in these situations; just ensure we do not crash. Garbage-in/garbage-out applies here. (BTW: Before the next release we should ensure we catch any errors that happen before the traced handler is called and call it without tracing in the worst case; the code is now complex enough for a good chance of bugs)



> > this property shouldn't be set if neither path nor query string is provided
> 
> Is this possible? I don't think that would be a HTTP request then. Maybe we should fall back to `/` or empty string in this case?

@Oberon00 : I rolled back to the previous code and updated unit test. So, even if neither path nor query string is set the http.target is still set to an empty string.

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Implemented minimal HTTP Server semantic conventions attributes set
```
```suggestion
* Add HTTP server span attributes for API Gateway triggers
```

Oberon00 (approved):

rypdal:
Seems build fails because of failing test not related to changes from this PR: 
Failed OpenTelemetry.Instrumentation.Wcf.Tests.TelemetryClientMessageInspectorTests.OutgoingRequestInstrumentationTest(instrument: True, filter: False, suppressDownstreamInstrumentation: True, includeVersion: True, enrich: False, enrichmentException: False, emptyOrNullAction: False) [1 ms]
  Error Message:
   System.ObjectDisposedException : Cannot access a disposed object.

Details can be found here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/3188299150/jobs/5200797125

Oberon00:
I think this PR can be merged now. (The build failure was indeed unrelated, it became green after changing the totally unrelated CHANGELOG)

PR #639, opened 2022-09-14
Instrumentation.AwsLambda: Fix versioning scheme.

Oberon00:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590#issuecomment-1245999476, CC @utpilla

## Changes

Changes version numbering scheme of the Instrumentation.AwsLambda which was previously released as 1.1.0.beta1 (& different package name) to use a dot before the pre-release number.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -2,15 +2,17 @@
 
 ## Unreleased
 
-## 1.1.0-beta2
+## 1.1.0-beta.2
 
 Release PR: [#590](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590)
+& [#???](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/???).
 
 This is the first release with the new package name `OpenTelemetry.Instrumentation.AWSLambda`.
 
 * BREAKING (API, behavior): Rename package to `OpenTelemetry.Instrumentation.AWSLambda`
   (remove `.Contrib`) ([#593](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/593)).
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,15 +2,17 @@
...
 
 Release PR: [#590](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590)
+& [#639](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/639).
 
 This is the first release with the new package name `OpenTelemetry.Instrumentation.AWSLambda`.
 
 * BREAKING (API, behavior): Rename package to `OpenTelemetry.Instrumentation.AWSLambda`
   (remove `.Contrib`) ([#593](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/593)).
   This also affects the `ActivitySource` name (superseding [#534](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/534)).
+* Pre-release version numbering scheme changed from `.betaN` to `beta.N` ([#639](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/639))
```
Oberon00, line 15: I noticed this does *not* require a change in the unit tests since we use the Assembly version which has a fixed format of four integers. IMHO this is wrong and we should use the assembly product version instead. But consistency between different packages is at least as important, so this should be discussed in general, not just for this package.

Kielek (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,15 +2,17 @@
 
 ## Unreleased
 
-## 1.1.0-beta2
+## 1.1.0-beta.2
 
 Release PR: [#590](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590)
```
utpilla, line 7: nit: We don't need an entry for Release PR.

utpilla (approved):

PR #649, opened 2022-09-19
[Exporter.Geneva] Refactor GenevaExporter

utpilla:
## Changes
- Refactor GenevaTraceExporter and GenevaLogExporter to make their behavior configurable in the ctor
- This would allow us switch from MsgPack to TLD later on
- Moved the classes with extension methods outside of the `MsgPackExporter` folder


src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackLogExporter.cs
```diff
@@ -21,12 +21,11 @@
 using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.Extensions.Logging;
-using OpenTelemetry.Internal;
 using OpenTelemetry.Logs;
 
 namespace OpenTelemetry.Exporter.Geneva;
 
-public class GenevaLogExporter : GenevaBaseExporter<LogRecord>
+internal sealed class MsgPackLogExporter : IDisposable
```
reyang, line 28: Curious - why isn't this derived from `GenevaBaseExporter<LogRecord>`?

utpilla: There is no strong reason. We could do that as well. I don't see any major benfits to doing that either. In fact, I think we could move all of those internal fields and methods of `GenevBaseExporter` to a new abstract class called `MsgPackExporter` as all of those internal data is related to to MsgPack approach.

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackLogExporter.cs
reyang: Got it, makes sense 👍

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackLogExporter.cs
```diff
@@ -154,31 +150,6 @@ public override ExportResult Export(in Batch<LogRecord> batch)
...
-        {
-            return;
-        }
-
-        if (disposing)
-        {
-            // DO NOT Dispose m_buffer as it is a static type
-            try
-            {
-                (this.m_dataTransport as IDisposable)?.Dispose();
```
cijothomas, line 169: who is now responsible for disposing transport?

cijothomas: ignore.. i see now that the code just moved!

cijothomas (approved):

PR #651, opened 2022-09-20
[ASP.NET] Migrate to native Activity Status and Status Desc

vishweshbankwar:
Fixes [3112](https://github.com/open-telemetry/opentelemetry-dotnet/issues/3112)
## Changes


src/OpenTelemetry.Contrib.Shared/Api/SpanHelper.cs
```diff
@@ -61,4 +61,24 @@ public static Status ResolveSpanStatusForHttpStatusCode(ActivityKind kind, int h
 
         return Status.Error;
     }
+
+#if NET462 ||  NET6_0_OR_GREATER
```
utpilla, line 65: Is this check required?

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Migrate to native Activity `Status` and
+  `StatusDesciption`.([#651](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/651))
```
utpilla, line 6: nit: Add the PR link in a new line.

src/OpenTelemetry.Contrib.Shared/Api/SpanHelper.cs
vishweshbankwar: yes - some projects are still targeting older OTel library where it is not available. That is why using this workaround.

utpilla: This would not compile if we added a new project which targets `net462` and pulls in an older OTel/DiagnosticSource package, right?

vishweshbankwar: That is right. Do you anticipate this case? I would expect the new project to use latest sdk/ diag src.

utpilla: I think the better thing to do here would be to make that check on the DS version instead of target framework. (if that's feasible). Otherwise, we can just move this method to AspNet instrumentation instead of `SpanHelper` as it's only called by AspNetInstrumentation for now.

We currently have some projects in the repo that target the old version of OTel and therefore get the older version of DS. A few of the component owners want to use the lowest stable version of OTel as they don't use any features from the newer versions of the SDK. Look at this conversation for instance: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/590#issuecomment-1245027824

@open-telemetry/dotnet-contrib-maintainers Thoughts?

Yun-Ting (approved):

src/OpenTelemetry.Contrib.Shared/Api/SpanHelper.cs
cijothomas: Since aspnet instrumentation is guaranteed to have this (as it requires ds6.0), we should not need any check. We can move this to aspnet instrumentation. Helper can be have this, once everyone who uses this helper is updated to new ds6.0 or newer.

vishweshbankwar: Moved it to aspnet for now.

utpilla (approved):

PR #659, opened 2022-09-26
[Exporter.Geneva] Optimize GenevaLogExporter

utpilla:
`AddPartAField` method expects an `object` as a parameter. We can avoid calling this method to serialize `timestamp` as it would convert `DateTime` to an `object`.

## Changes
- Avoid allocation due to boxing


cijothomas (approved):
Great!
Optional : Might be good to add a changelog entry advertising the lower allocation log exporter.

reyang (approved):

PR #662, opened 2022-09-27
[Exporter.Geneva] Add TLDTraceExporter

utpilla:
## Changes
- Add TLDTraceExporter
- GenevaTraceExporter to use TLDExporter when the user provides "UseTLD=true" in the connectionString

### TODOs:
- Use a single `ThreadLocal` instead of multiple
- Avoid potential unbounded memory usage for `List<KeyValuePair<string, object>> keyValuePairs`

### Setup and usage:

```csharp
            .SetSampler(new AlwaysOnSampler())
            .AddSource(Source.Name)
            .AddGenevaTraceExporter(options =>
            {
                options.ConnectionString = "EtwSession=OpenTelemetry;UseTLD=true"; // Add UseTLD=true in the connection string
                options.PrepopulatedFields = new Dictionary<string, object>
                {
                    ["cloud.role"] = "BusyWorker",
                    ["cloud.roleInstance"] = "CY1SCH030021417",
                    ["cloud.roleVer"] = "9.0.15289.2",
                };
            })
            .Build();

//Emit activity
using (var activity = Source.StartActivity("SayHello"))
        {
            activity?.SetTag("foo", 1);
            activity?.SetTag("bar", "Hello, World!");
            activity?.SetTag("baz", new int[] { 1, 2, 3 });
            activity?.SetStatus(ActivityStatusCode.Ok);
        }
```

You could us any ETW collection tools such as WPA, PerfView etc. to view the data:

Example with PerfView:
For the connectionString = `"EtwSession=OpenTelemetry;UseTLD=true"`, you can run the following command in Command Prompt: 
```cmd
PerfView.exe collect -MaxCollectSec:3600 -NoGui /onlyProviders=*OpenTelemetry
```


src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
```diff
@@ -0,0 +1,309 @@
...
+    private readonly string partAName = "Span";
+    private readonly byte partAFieldsCount = 3; // At least three fields: time, ext_dt_traceId, ext_dt_spanId
+    private readonly IReadOnlyDictionary<string, object> m_customFields;
+    private readonly Tuple<byte[], byte[]> repeatedPartAFields;
+    private static readonly string INVALID_SPAN_ID = default(ActivitySpanId).ToHexString();
+
+    private readonly EventProvider eventProvider;
+    private static readonly ThreadLocal<EventBuilder> eventBuilder = new(() => new(UncheckedASCIIEncoding.SharedInstance));
+    private static readonly ThreadLocal<List<KeyValuePair<string, object>>> keyValuePairs = new(() => new());
+    private static readonly ThreadLocal<KeyValuePair<string, object>[]> partCFields = new(() => new KeyValuePair<string, object>[120]); // This is used to temporarily store the PartC fields from tags
```
utpilla, line 38: This might lead to `IndexOutOfRange` exceptions if there are too many Part C tags. I chose the size as 120 as we cannot exceed 127 fields otherwise the event gets dropped and we have at least around 7 fields from Part A and Part B itself.

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
```diff
@@ -0,0 +1,309 @@
...
+{
+    private readonly string partAName = "Span";
+    private readonly byte partAFieldsCount = 3; // At least three fields: time, ext_dt_traceId, ext_dt_spanId
+    private readonly IReadOnlyDictionary<string, object> m_customFields;
+    private readonly Tuple<byte[], byte[]> repeatedPartAFields;
+    private static readonly string INVALID_SPAN_ID = default(ActivitySpanId).ToHexString();
+
+    private readonly EventProvider eventProvider;
+    private static readonly ThreadLocal<EventBuilder> eventBuilder = new(() => new(UncheckedASCIIEncoding.SharedInstance));
+    private static readonly ThreadLocal<List<KeyValuePair<string, object>>> keyValuePairs = new(() => new());
```
utpilla, line 37: This is used for serializing ActivityLinks and `env_properties`. I chose a `List` here instead of an array as we don't know how many such fields there could be.

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TraceLoggingDynamic.cs
```diff
@@ -0,0 +1,2623 @@
...
+/// initialization and one instance should be shared by all code within the
+/// component that needs to write events for the provider. In some cases, it
+/// might be used in a smaller scope, in which case it should be closed via
+/// Dispose().
+/// </summary>
+
+#pragma warning disable CA5350
+#pragma warning disable CA5392
+
+internal sealed class EventProvider
```
utpilla, line 98: I have simply copied the contents of this file here: https://github.com/microsoft/tracelogging/blob/main/etw/cs/traceloggingdynamic/TraceLoggingDynamic.cs

Only additional changes that I made are that I have marked the types internal and internal sealed.

.editorconfig
```diff
@@ -147,3 +147,6 @@ generated_code = true
 
 [*.csproj]
 indent_size = 2
+
+[TraceLoggingDynamic.cs]
```
reyang, line 151: Maybe have a general pattern?

```text
[external/**.cs]
generated_code = true
```

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
```diff
@@ -0,0 +1,309 @@
...
+internal sealed class TLDTraceExporter : TLDExporter, IDisposable
+{
+    private readonly string partAName = "Span";
+    private readonly byte partAFieldsCount = 3; // At least three fields: time, ext_dt_traceId, ext_dt_spanId
+    private readonly IReadOnlyDictionary<string, object> m_customFields;
+    private readonly Tuple<byte[], byte[]> repeatedPartAFields;
+    private static readonly string INVALID_SPAN_ID = default(ActivitySpanId).ToHexString();
+
+    private readonly EventProvider eventProvider;
+    private static readonly ThreadLocal<EventBuilder> eventBuilder = new(() => new(UncheckedASCIIEncoding.SharedInstance));
```
reyang, line 36: Which one would be better?
1. Having 3 ThreadLocals
2. Having a single ThreadLocal which points to a complex struct/object with 3 fields

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
reyang: What's the initial capacity?

utpilla: > What's the initial capacity?

I let it use the default behavior.

> (and what's the reallocation behavior? e.g. would this give OOM?)

I understand that this might lead to unbounded memory usage and possibly even lead to OOM in extreme cases but I didn't know if we could enforce some sort of limitation on the number of links or env_properties allowed. It's straightforward for PartC fields as the TLD spec enforces a limit on the number of fields allowed. In this case though, links or env_properties would only count towards one field. If we could enforce a similar limitation here, we could very well use an array and just tell the user that they have too many links or env_properties.

Could we enforce such a limitation? If not, I guess we would have to go back to the two iterations approach.

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
utpilla: I have no strong preference. I suppose there's no performance implications of using one over the other. If the second approach looks more readable, I could make that change.

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
utpilla: Added a TODO for this.

src/OpenTelemetry.Exporter.Geneva/TLDExporter/TLDTraceExporter.cs
utpilla: Added a TODO for this.

reyang (approved):

PR #664, opened 2022-09-28
[Instrumentation.Process] Implement threadSafe Refresh()

Yun-Ting:
## Changes

Assumptions:
1. For callbacks of the instruments from the same collection of an exporter, the instrument values should come from the same snapshot of the current process. In this PR, this was achieved by calling Process.Refresh() exactly once for each collection and use the idea of a state machine to check whether the snapshot is expired or not.
Without this guarantee, there will be issues with the following cases:
If users are trying to use the data from the dashboard to debug OOM of their application, it will be misleading if physical memory usage and virtual memory usage were not taken from a single snapshot.
(Also, not in the scope of this PR yet, when calculating CPU utilization:
`Process.GetTotalProcessorTime() / elapsedTimeSinceTheProcessStart * (number of processors)`
Process.GetTotalProcessorTime() should be coming from the same snapshot for accuracy.

2. ProcessMetrics should be threadSafe for multi-reader scenarios.

3. Since .NET currently does not support multiple-instrument callback https://github.com/dotnet/runtime/issues/62027#issuecomment-1113762370, this PR uses the state machine workaround to keep track of whether the instrument callbacks were called or not within one collection of a reader.

Attached the console output for 2-collection-cycle from  https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/663  to elaborate the flow.

Console output for 1 reader case:
```console
threadId: 11 invoke GetMemoryUsage() without the latest snapshot.
threadId: 11 Refresh()
threadId: 11 retrieve memory usage from snapshot.
threadId: 11 retrieve virtual memory usage from snapshot.
Resource associated with Metric:
    service.name: unknown_service:runtime-instrumentation

Export process.memory.usage, The amount of physical memory in use., Unit: By, Meter: OpenTelemetry.Instrumentation.Process/0.0.0.0
(2022-09-28T22:21:35.3619754Z, 2022-09-28T22:21:45.3608093Z] DoubleGauge
Value: 29290496

Export process.memory.virtual, The amount of committed virtual memory., Unit: By, Meter: OpenTelemetry.Instrumentation.Process/0.0.0.0
(2022-09-28T22:21:35.3643951Z, 2022-09-28T22:21:45.3608338Z] DoubleGauge
Value: 14835712
threadId: 11 invoke GetMemoryUsage() without the latest snapshot.
threadId: 11 Refresh()
threadId: 11 retrieve memory usage from snapshot.
threadId: 11 retrieve virtual memory usage from snapshot.

Export process.memory.usage, The amount of physical memory in use., Unit: By, Meter: OpenTelemetry.Instrumentation.Process/0.0.0.0
(2022-09-28T22:21:35.3619754Z, 2022-09-28T22:21:55.3418369Z] DoubleGauge
Value: 29736960

Export process.memory.virtual, The amount of committed virtual memory., Unit: By, Meter: OpenTelemetry.Instrumentation.Process/0.0.0.0
(2022-09-28T22:21:35.3643951Z, 2022-09-28T22:21:55.3418375Z] DoubleGauge
Value: 14942208
```

Console output for 2 readers (I preprocessed the output with the correct time sequence from console for eye-friendly reading experience)

image

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -25,62 +25,51 @@ internal sealed class ProcessMetrics
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
 
-    private readonly InstrumentsValues instrumentsValues;
+    private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
+    private double? memoryUsage;
+    private double? virtualMemoryUsage;
 
     public ProcessMetrics(ProcessInstrumentationOptions options)
     {
-        this.instrumentsValues = new InstrumentsValues();
-
         // TODO: change to ObservableUpDownCounter
         this.MeterInstance.CreateObservableGauge(
             "process.memory.usage",
-            () => this.instrumentsValues.GetMemoryUsage(),
+            () =>
+            {
+                if (!this.memoryUsage.HasValue)
+                {
+                    this.Snapshot();
+                }
+
+                var value = this.memoryUsage.Value;
+                this.memoryUsage = null;
+                return value;
+            },
             unit: "By",
             description: "The amount of physical memory allocated for this process.");
 
         // TODO: change to ObservableUpDownCounter
         this.MeterInstance.CreateObservableGauge(
             "process.memory.virtual",
-            () => this.instrumentsValues.GetVirtualMemoryUsage(),
+            () =>
+            {
+                if (!this.virtualMemoryUsage.HasValue)
+                {
+                    this.Snapshot();
+                }
+
+                var value = this.virtualMemoryUsage.Value;
+                this.virtualMemoryUsage = null;
+                return value;
+            },
             unit: "By",
             description: "The amount of virtual memory allocated for this process that cannot be shared with other processes.");
     }
 
-    private sealed class InstrumentsValues
+    private void Snapshot()
     {
-        private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
-        private double? memoryUsage;
-        private double? virtualMemoryUsage;
-
-        internal double GetMemoryUsage()
-        {
-            if (!this.memoryUsage.HasValue)
-            {
-                this.Snapshot();
-            }
-
-            var value = this.memoryUsage.Value;
-            this.memoryUsage = null;
-            return value;
-        }
-
-        internal double GetVirtualMemoryUsage()
-        {
-            if (!this.virtualMemoryUsage.HasValue)
-            {
-                this.Snapshot();
-            }
-
-            var value = this.virtualMemoryUsage.Value;
-            this.virtualMemoryUsage = null;
-            return value;
-        }
-
-        private void Snapshot()
-        {
-            this.currentProcess.Refresh();
-            this.memoryUsage = this.currentProcess.WorkingSet64;
-            this.virtualMemoryUsage = this.currentProcess.PrivateMemorySize64;
-        }
+        this.currentProcess.Refresh();
+        this.memoryUsage = this.currentProcess.WorkingSet64;
+        this.virtualMemoryUsage = this.currentProcess.PrivateMemorySize64;
     }
 }
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -14,8 +14,11 @@
 // limitations under the License.
 // </copyright>
 
+using System;
+using System.Collections.Generic;
 using System.Diagnostics.Metrics;
 using System.Reflection;
+using System.Threading;
 using Diagnostics = System.Diagnostics;
 
 namespace OpenTelemetry.Instrumentation.Process;
@@ -24,38 +27,97 @@ internal class ProcessMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
-    private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
 
-    static ProcessMetrics()
+    public ProcessMetrics(ProcessInstrumentationOptions options)
     {
+        ThreadSafeInstrumentValues threadSafeInstrumentValues = new();
+
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.usage",
-            () =>
-            {
-                CurrentProcess.Refresh();
-                return CurrentProcess.WorkingSet64;
-            },
+            () => threadSafeInstrumentValues.GetMemoryUsage(),
             unit: "By",
             description: "The amount of physical memory in use.");
 
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.virtual",
-            () =>
-            {
-                CurrentProcess.Refresh();
-                return CurrentProcess.VirtualMemorySize64;
-            },
+            () => threadSafeInstrumentValues.GetVirtualMemoryUsage(),
             unit: "By",
             description: "The amount of committed virtual memory.");
     }
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="ProcessMetrics"/> class.
-    /// </summary>
-    /// <param name="options">The options to define the metrics.</param>
-    public ProcessMetrics(ProcessInstrumentationOptions options)
+    private class ThreadSafeInstrumentValues
     {
+        private readonly ThreadLocal<Dictionary<int, InstrumentsValues>> threadIdToInstrumentValues = new(() =>
+        {
+            return new Dictionary<int, InstrumentsValues>();
+        });
+
+        internal double GetMemoryUsage()
+        {
+            if (!this.threadIdToInstrumentValues.Value.ContainsKey(Environment.CurrentManagedThreadId))
+            {
+                this.threadIdToInstrumentValues.Value.Add(Environment.CurrentManagedThreadId, new InstrumentsValues());
+            }
+
+            this.threadIdToInstrumentValues.Value.TryGetValue(Environment.CurrentManagedThreadId, out var instrumentValues);
+            return instrumentValues.GetMemoryUsage();
+        }
+
+        internal double GetVirtualMemoryUsage()
+        {
+            if (!this.threadIdToInstrumentValues.Value.ContainsKey(Environment.CurrentManagedThreadId))
+            {
+                this.threadIdToInstrumentValues.Value.Add(Environment.CurrentManagedThreadId, new InstrumentsValues());
+            }
+
+            this.threadIdToInstrumentValues.Value.TryGetValue(Environment.CurrentManagedThreadId, out var instrumentsValues);
+            return instrumentsValues.GetVirtualMemoryUsage();
+        }
+    }
+
+    private class InstrumentsValues
+    {
+        private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
+        private double? memoryUsage;
+        private double? virtualMemoryUsage;
+
+        internal InstrumentsValues()
+        {
+            this.memoryUsage = null;
+            this.virtualMemoryUsage = null;
+        }
+
+        internal double GetMemoryUsage()
+        {
+            if (!this.memoryUsage.HasValue)
+            {
+                this.Snapshot();
+            }
+
+            var value = (double)this.memoryUsage;
+            this.memoryUsage = null;
+            return value;
+        }
+
+        internal double GetVirtualMemoryUsage()
+        {
+            if (!this.virtualMemoryUsage.HasValue)
+            {
+                this.Snapshot();
+            }
+
+            var value = (double)this.virtualMemoryUsage;
+            this.virtualMemoryUsage = null;
+            return value;
+        }
+
+        private void Snapshot()
+        {
+            this.currentProcess.Refresh();
+            this.memoryUsage = this.currentProcess.WorkingSet64;
+            this.virtualMemoryUsage = this.currentProcess.PagedMemorySize64;
+        }
     }
 }
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -14,8 +14,6 @@
 // limitations under the License.
 // </copyright>
 
-using System;
-using System.Collections.Generic;
 using System.Diagnostics.Metrics;
 using System.Reflection;
 using System.Threading;
@@ -23,63 +21,33 @@ using Diagnostics = System.Diagnostics;
 
 namespace OpenTelemetry.Instrumentation.Process;
 
-internal class ProcessMetrics
+internal sealed class ProcessMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
     internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
 
+    private static readonly ThreadLocal<InstrumentsValues> CurrentThreadInstrumentsValues = new(() => new InstrumentsValues());
+
     public ProcessMetrics(ProcessInstrumentationOptions options)
     {
-        ThreadSafeInstrumentValues threadSafeInstrumentValues = new();
-
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.usage",
-            () => threadSafeInstrumentValues.GetMemoryUsage(),
+            () => CurrentThreadInstrumentsValues.Value.GetMemoryUsage(),
             unit: "By",
             description: "The amount of physical memory in use.");
 
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.virtual",
-            () => threadSafeInstrumentValues.GetVirtualMemoryUsage(),
+            () => CurrentThreadInstrumentsValues.Value.GetVirtualMemoryUsage(),
             unit: "By",
             description: "The amount of committed virtual memory.");
     }
 
-    private class ThreadSafeInstrumentValues
+    private sealed class InstrumentsValues
     {
-        private readonly ThreadLocal<Dictionary<int, InstrumentsValues>> threadIdToInstrumentValues = new(() =>
-        {
-            return new Dictionary<int, InstrumentsValues>();
-        });
-
-        internal double GetMemoryUsage()
-        {
-            if (!this.threadIdToInstrumentValues.Value.ContainsKey(Environment.CurrentManagedThreadId))
-            {
-                this.threadIdToInstrumentValues.Value.Add(Environment.CurrentManagedThreadId, new InstrumentsValues());
-            }
-
-            this.threadIdToInstrumentValues.Value.TryGetValue(Environment.CurrentManagedThreadId, out var instrumentValues);
-            return instrumentValues.GetMemoryUsage();
-        }
-
-        internal double GetVirtualMemoryUsage()
-        {
-            if (!this.threadIdToInstrumentValues.Value.ContainsKey(Environment.CurrentManagedThreadId))
-            {
-                this.threadIdToInstrumentValues.Value.Add(Environment.CurrentManagedThreadId, new InstrumentsValues());
-            }
-
-            this.threadIdToInstrumentValues.Value.TryGetValue(Environment.CurrentManagedThreadId, out var instrumentsValues);
-            return instrumentsValues.GetVirtualMemoryUsage();
-        }
-    }
-
-    private class InstrumentsValues
-    {
-        private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
+        private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
         private double? memoryUsage;
         private double? virtualMemoryUsage;
 
@@ -115,9 +83,9 @@ internal class ProcessMetrics
 
         private void Snapshot()
         {
-            this.currentProcess.Refresh();
-            this.memoryUsage = this.currentProcess.WorkingSet64;
-            this.virtualMemoryUsage = this.currentProcess.PagedMemorySize64;
+            CurrentProcess.Refresh();
+            this.memoryUsage = CurrentProcess.WorkingSet64;
+            this.virtualMemoryUsage = CurrentProcess.PagedMemorySize64;
         }
     }
 }
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -16,7 +16,6 @@
 
 using System.Diagnostics.Metrics;
 using System.Reflection;
-using System.Threading;
 using Diagnostics = System.Diagnostics;
 
 namespace OpenTelemetry.Instrumentation.Process;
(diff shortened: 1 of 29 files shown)


src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +27,97 @@ internal class ProcessMetrics
...
             unit: "By",
             description: "The amount of committed virtual memory.");
     }
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="ProcessMetrics"/> class.
-    /// </summary>
-    /// <param name="options">The options to define the metrics.</param>
-    public ProcessMetrics(ProcessInstrumentationOptions options)
+    private class ThreadSafeInstrumentValues
```
Yun-Ting, line 50: This layer can be removed as threadID should be universal across threads.

CodeBlanch: Are you saying we could do something like...

```csharp
internal class ProcessMetrics
{
   private static readonly ThreadLocal<InstrumentsValues> currentThreadInstrumentValues = new(() => new InstrumentsValues());
}
```

I think that would work.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +27,97 @@ internal class ProcessMetrics
...
+            {
+                this.threadIdToInstrumentValues.Value.Add(Environment.CurrentManagedThreadId, new InstrumentsValues());
+            }
+
+            this.threadIdToInstrumentValues.Value.TryGetValue(Environment.CurrentManagedThreadId, out var instrumentsValues);
+            return instrumentsValues.GetVirtualMemoryUsage();
+        }
+    }
+
+    private class InstrumentsValues
```
CodeBlanch, line 80: nit: Good practice to add `sealed` to `private` and `internal` classes.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Thanks for the hint. I also learned why it is a good practice to make ThreadLocal object static!

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -24,38 +27,97 @@ internal class ProcessMetrics
...
     {
+        private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
```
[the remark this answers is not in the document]
Yun-Ting: Yup, I don't see a problem with making the currentProcess static. Updated. Thanks!

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
     {
         // TODO: change to ObservableUpDownCounter
         MeterInstance.CreateObservableGauge(
             "process.memory.usage",
-            () =>
-            {
-                CurrentProcess.Refresh();
-                return CurrentProcess.WorkingSet64;
-            },
+            () => CurrentThreadInstrumentsValues.Value.GetMemoryUsage(),
```
tarekgh, line 36: Thread local is not guaranteed to be initialized on all threads. So, you may need to check and initialize before using it.

tarekgh: never mind, I thought you are using the instance directly. Please ignore my comment.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
+        }
+
+        internal double GetMemoryUsage()
+        {
+            if (!this.memoryUsage.HasValue)
+            {
+                this.Snapshot();
+            }
+
+            var value = (double)this.memoryUsage;
```
utpilla, line 67: You could use `this.memoryUsage.Value` here.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
+        }
+
+        internal double GetVirtualMemoryUsage()
+        {
+            if (!this.virtualMemoryUsage.HasValue)
+            {
+                this.Snapshot();
+            }
+
+            var value = (double)this.virtualMemoryUsage;
```
utpilla, line 79: You could use `this.virtualMemoryUsage.Value` here.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
+            this.virtualMemoryUsage = null;
+            return value;
+        }
+
+        private void Snapshot()
+        {
+            CurrentProcess.Refresh();
+            this.memoryUsage = CurrentProcess.WorkingSet64;
+            this.virtualMemoryUsage = CurrentProcess.PagedMemorySize64;
```
utpilla, line 88: Did you intend to use `CurrentProcess.VirtualMemorySize64` here?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
-    public ProcessMetrics(ProcessInstrumentationOptions options)
+    private sealed class InstrumentsValues
     {
+        private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
+        private double? memoryUsage;
+        private double? virtualMemoryUsage;
+
+        internal InstrumentsValues()
+        {
+            this.memoryUsage = null;
```
tarekgh, line 56: do you need this explicit initialization?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
+            this.virtualMemoryUsage = null;
+        }
+
+        internal double GetMemoryUsage()
+        {
+            if (!this.memoryUsage.HasValue)
```
tarekgh, line 62: nit: is it a code style to always use `this`? it is just more reading. In .NET repos we just define the fields prefixed with `_` and not using `this` ate all. it is easier to read.

utpilla: Yes. This is coming from the StyleCop Analyzer: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1101.md

> The advantage of using 'this.' is that it applies equally to all element types including methods, properties, etc., and not just fields, making all calls to class members instantly recognizable, regardless of which editor is being used to view the code. Another advantage is that it creates a quick, recognizable differentiation between instance members and static members, which are not prefixed.

> A final advantage of using the 'this.' prefix is that typing this. will cause Visual Studio to show the IntelliSense popup, making it quick and easy for the developer to choose the class member to call.

tarekgh: Yes, but not mandatory rule to apply :-) as I mentioned we don't do that in the .NET code itself :-) we have other styling we apply. I am not asking to change anything more than wondering.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -16,46 +16,76 @@
...
+            {
+                this.Snapshot();
+            }
+
+            var value = (double)this.memoryUsage;
+            this.memoryUsage = null;
+            return value;
+        }
+
+        internal double GetVirtualMemoryUsage()
```
tarekgh, line 72: What You think instead of having two methods `GetMemoryUsage()` and `GetVirtualMemoryUsage()`, you can have only one method returning both values as value tuple? something like the following:

```C#
public static (double memoryUsage, double virtualMmeoryUsage) GetProcessMeasurements() => (CurrentProcess.WorkingSet64, CurrentProcess.VirtualMemorySize64);
```

and you can get rid of the whole class and the static local variable. Or do you think users will need to get each individual value at a different time?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
utpilla, line 36: How does this work with multiple `MeterProvider` setup?

In the scenario where you have multiple meterProviders and all of them add Process instrumentation, this would create the same set of instruments for each `MeterProvider` since they are all using the same `Meter` instance.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
utpilla: Yes, I was also not used to using `this.` prefix earlier 😄

reyang: I used to write `_foo` or `m_bar` before was treated as a C++ nerd who brings bad habits to .NET 🤣

tarekgh: In runtime code we use `_foo` style, `s_bar` for static. This is actually the recommended style by us :-)

reyang: You can tell that I was coming from DevDiv https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/215bc921a233200e2acfafc3420e1b763b4ea385/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L31 🤣

reyang: To my knowledge, these came from https://en.wikipedia.org/wiki/Hungarian_notation, looks like .NET decided to replace `m_` with `_`.

image

tarekgh: could be but Hungarian notation was listing the variable type in the field name too. like `arr`, 'l', 'i'...etc.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: My understanding is that the flexibility of this level of customization is deliberately provided by OTel SDK with the use of [View](https://github.com/open-telemetry/opentelemetry-specification/blob/31c81e87d697d5fc6dcff29eee24d6a0b0dbba26/specification/metrics/sdk.md#view).

But this grouping idea could potentially be used on instruments that would not make much sense if being decoupled. A good case for ProcessMetrics is actually process physical memory usage and process virtual memory usage. 

I am still debating myself whether to add this "limitation"(?) to ProcessMetrics. And if added, what should be the behavior of ProcessMetrics when there are users who only care about process virtual memory usage.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: I am not sure for what intent would the user want to do `.AddProcessInstrumentation()` for multiple meterProviders to a given application/process. Everything in ProcessMetrics should ideally be static in the scope of a Process.
 
Could you share with me the use case you have in mind that ProcessMetrics was being added to more than one meterProviders?

utpilla: One use case could be where the user wants to send data from different meters to different exporters along with the process metrics.

Yun-Ting: I think there will not be a problem if we mark the ctor of ProcessMetrics as static.
I still don't quite get why would the user want to monitor a static object via different exporters?
Once case I could think of now is role-based access control - authorized users of the application can see a certain field like for example process.disk.io while non-authorized users cannot?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Nope, this update was because I found out that PagedMemorySize64 should be the more suitable value for this gauge.
Please refer to this [PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/673) for my rationale.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: removed; thank you

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Updated to use instance members for meter and InstrumentValues with this commit 13434f599fcfc618d00ad5bb4a3872f5823c5c95.

One caveat with this implementation is in `Snapshot()`, threadA calls Process.Refresh(), and before threadA does the assignment to this.memoryUsage, threadB calls Process.Refresh(). The odds are small as assignments should be very fast.
Even if we hit this situation, it should not be a big deal because the time difference between these 2 refreshes is tiny.

Yun-Ting: Attached the discussion I had with Blanch and Utkarsh below.
[discussion.txt](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/files/9711796/discussion.txt)

reyang: > One caveat with this implementation is in `Snapshot()`, threadA calls Process.Refresh(), and before threadA does the assignment to this.memoryUsage, threadB calls Process.Refresh(). The odds are small as assignments should be very fast. Even if we hit this situation, it should not be a big deal because the time difference between these 2 refreshes is tiny.

When would this happen? Is there a concrete example?

Yun-Ting: image

The case I have in mind is:
At t3, the value got assigned to `this.virtualMemoryUsage` for threadA is from this.currentProcessRefresh() from t2 by threadB; 
because [GetCurrentProcess()](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.getcurrentprocess?view=net-7.0) is a static method that returns "A new [Process](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process?view=net-7.0) component associated with the process resource that **is** running the calling application." 

Even though we maintain different copies for currentProcess here:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/13434f599fcfc618d00ad5bb4a3872f5823c5c95/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs#L51 
Those are referencing the same object.

But again, I don't think this is a case that could be hit frequently. And even it did occur, the time difference between the 2 snapshots are negligible.

Yun-Ting: yeah.... `Snapshot()` will not be accessed concurrently because it is an instance method.
What I was worried about is that whether those 2 instances are refreshing on the same object.

After further investigation, that is not the case:
image

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -20,42 +20,67 @@
 
 namespace OpenTelemetry.Instrumentation.Process;
 
-internal class ProcessMetrics
+internal sealed class ProcessMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(ProcessMetrics).Assembly.GetName();
-    internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
-    private static readonly Diagnostics.Process CurrentProcess = Diagnostics.Process.GetCurrentProcess();
+    internal readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
```
tarekgh, line 26: I am wondering, if creating 2 instances of `ProcessMetrics` class, do we expect every instance create its own meter with the same name?

tarekgh (approved):

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: My understanding is that in the SDK, [MetricStreamIdentity](https://github.com/open-telemetry/opentelemetry-dotnet/blob/879d3c91a0032eee75061871851e2d4febd6899f/src/OpenTelemetry/Metrics/MetricStreamIdentity.cs) is maintained, so there will be only **one** `MetricStreamIdentity` in this case.

What would happen is both of the callbacks from the instruments would be firing, but there will only be one export.
The second one will hit this https://github.com/open-telemetry/opentelemetry-dotnet/blob/7f71283c0bf2ee0e2c5e3c9617295ddaf20c7a0c/src/OpenTelemetry/Metrics/MetricReaderExt.cs#L51

Please refer to the [testcases](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/6bfac03ec17a111d21342d6954feab1f696b9b59/test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs) for the experiments I've done.
Thank you.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: > How does this work with multiple `MeterProvider` setup?
> 
> In the scenario where you have multiple meterProviders and all of them add Process instrumentation, this would create the same set of instruments for each `MeterProvider` since they are all using the same `Meter` instance.

Please take a look:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/664/files#r989475625

CodeBlanch (approved):
LGTM

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -42,4 +43,90 @@ public void ProcessMetricsAreCaptured()
...
+        Meter m1 = new("myMeter");
+        Meter m2 = new("myMeter");
+
+        m1.CreateObservableCounter(
+            "myGaugeName",
+            () => { return 1D; },
+            unit: "1",
+            description: "test");
+
+        m1.CreateObservableCounter(
```
utpilla, line 88: Did you mean to use `m2` here?

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -42,4 +43,90 @@ public void ProcessMetricsAreCaptured()
...
+            .AddInMemoryExporter(exportedItemsB)
+            .Build();
+
+        meterProviderA.ForceFlush();
+        meterProviderB.ForceFlush();
+
+        var metricA = exportedItemsA.FirstOrDefault(i => i.Name == "myGaugeName");
+        var metricB = exportedItemsB.FirstOrDefault(i => i.Name == "myGaugeName");
+
+        Assert.Equal(GetValue(metricA), GetValue(metricB));
```
utpilla, line 110: Do we expect them to be equal?

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
utpilla: > there will be only one MetricStreamIdentity in this case.

I doubt that. `MetricStreamIdentity` is maintained at a `MetricReader` level within the SDK. In the scenario, where we have two instances of `ProcessMetrics` class due to multiple MeterProviders we would also have multiple readers each maintaining their own set of `MetricStreamIdentity`.

utpilla (approved):
LGTM overall.

The case of verifying correctness for multiple MeterProviders could be investigated further. We don't have to block this PR for that.

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
Yun-Ting: yup and fixed!

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
Yun-Ting: yes, for counters as the MetricStreamIdentity will be the same if the meter name matches, the value will be equal.


While, for gauges, my understanding is that since they do a spontaneous measurement at the time they were being collected, so it is hard to use InMemoryExporter to create a unit test for that.

But I checked that the running values in the aggregationStore match:
image


```
((OpenTelemetry.Metrics.Metric)(new System.Linq.SystemCore_EnumerableDebugView<System.Diagnostics.Metrics.ListenerSubscription>((new System.Collections.Generic.ICollectionDebugView<System.Diagnostics.Metrics.Instrument>(((OpenTelemetry.Instrumentation.Process.ProcessMetrics)(new System.Collections.Generic.ICollectionDebugView<object>(((OpenTelemetry.Metrics.MeterProviderSdk)meterProviderA).Instrumentations).Items[0])).MeterInstance._instruments).Items[0])._subscriptions).Items[0]).State).aggStore.metricPoints[0].runningValue.AsDouble



((OpenTelemetry.Metrics.Metric)(new System.Linq.SystemCore_EnumerableDebugView<System.Diagnostics.Metrics.ListenerSubscription>((new System.Collections.Generic.ICollectionDebugView<System.Diagnostics.Metrics.Instrument>(((OpenTelemetry.Instrumentation.Process.ProcessMetrics)(new System.Collections.Generic.ICollectionDebugView<object>(((OpenTelemetry.Metrics.MeterProviderSdk)meterProviderB).Instrumentations).Items[0])).MeterInstance._instruments).Items[0])._subscriptions).Items[0]).State).aggStore.metricPoints[0].runningValue.AsDouble
```

utpilla: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/664/files#r991558434

You have two ObservableCounters here: one always returns `1` and the other always returns `2` yet this line: `Assert.Equal(GetValue(metricA), GetValue(metricB));` succeeds. This means that only one of the ObservableCounters callbacks is being called.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
Yun-Ting: Hmm... I think each reader has its own storage with the current implementation - usage of instance for `ProcessMetrics`; 
so it may not be required to maintain MetricStreamIdentity at the reader level.

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
Yun-Ting: Hmm... what I see is that both callbacks will be triggered for the meterProviders, but the second callback will hit this:
image

test code here:
https://github.com/open-telemetry/opentelemetry-dotnet/compare/main...Yun-Ting:yunl/instrumentNames?expand=1

utpilla:
Merging this PR to unblock progress. Still need to verify the correctness in case of multiple MeterProviders.

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -20,42 +20,56 @@
...
+                    this.Snapshot();
+                }
+
+                var value = this.virtualMemoryUsage.Value;
+                this.virtualMemoryUsage = null;
+                return value;
             },
             unit: "By",
-            description: "The amount of committed virtual memory.");
+            description: "The amount of virtual memory allocated for this process that cannot be shared with other processes.");
```
noahfalk, line 66: I was expecting "process.memory.virtual" to be Process.VirtualMemorySize64. I saw you linked to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/673 which had a very nice explanation for why you made the change. I think you were attempting to be very thorough and precise in making this choice so I want to offer something equally detailed to give an alternate perspective on the problem.

When we get into details, I think 'committed memory' is unfortunately an ambiguous term and the best way to understand what the OTel spec writers intended will be for us to ask them and clarify the spec. In Mark Russinovich's blog post he repeatedly refers to memory that "counts against the commit limit", but he never actually says "Memory is 'committed memory' if and only if it counts against the commit limit". I think that would be an intuitive definition in context of his article and probably it is the definition he intended the reader to apply despite never stating it pedantically. For simplicity we can just call that his intended definition. His definition is probably technically precise for how some portion of the Windows Kernel uses the term, but I suspect it is not how most people understand and use it.

IMO the common definition of committed memory that most developers would use is that it is all the memory addresses a process can read from without triggering an access violation. This includes (at least) three categories of memory:
 - memory backed solely by RAM
 - memory backed by a swapfile/pagefile
 - memory backed by some other mapped file on disk

The first two agree with Mark's definition above so the difference is entirely whether the third bullet is included. To justify that it is common for most people to include file backed memory as part of 'committed memory' [this article](https://dzone.com/articles/windows-process-memory-usage-demystified) has a screenshot of the tool VMMap at the top. I consider VMMap to be the de-facto standard for investigating and categorizing virtual memory usage (incidentally its also a tool MarkR made). Looking at the table in that image is a column called 'Committed' and the first two rows are 'Image' and 'Mapped File'. If committed virtual memory did not include memory backed by files on disk then we'd expect those columns to be 0, but they aren't.

As a second example, [this link](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/-address) is a help page for the !address command in windbg which shows information about the virtual address space. There is an example output part way down the page showing:
```
0:000> !address 75831234
Usage:                  Image
Base Address:           75831000
End Address:            758f6000
Region Size:            000c5000
Type:                   01000000MEM_IMAGE
State:                  00001000MEM_COMMIT
Protect:                00000020PAGE_EXECUTE_READ
More info:              lmv m kernel32
More info:              !lmi kernel32
More info:              ln 0x75831234
```

The debugger says that this address is part of the kernel32 image and it is on a memory page that is marked both as being MEM_IMAGE and MEM_COMMIT. Mark's blog post says specifically that "The virtual memory in Testlimit's address space where its executable and system DLL images are mapped therefore don't count toward the commit limit" and yet windbg describes this exact type of memory as being MEM_COMMIT.

I think the way VMMap and Windbg refer to committed memory is the way most developers would use the term. That definition corresponds to using Process.VirtualMemorySize64 as the metric value. Of course we can always ask the OTel spec writers if that is the definition they intended. Hope that helps!

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -20,42 +20,56 @@
...
     {
         // TODO: change to ObservableUpDownCounter
-        MeterInstance.CreateObservableGauge(
+        this.MeterInstance.CreateObservableGauge(
             "process.memory.usage",
             () =>
             {
-                CurrentProcess.Refresh();
-                return CurrentProcess.WorkingSet64;
+                if (!this.memoryUsage.HasValue)
```
noahfalk, line 39: For these callbacks to be thread-safe they need to be inside a lock. For example:
```C#
() =>
{
    lock(this)
    {
        if (!this.memoryUsage.HasValue)
        {
            this.Snapshot();
        }
        var value = this.memoryUsage.Value;
        this.memoryUsage = null;
        return value;
    }
}
```

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -20,42 +20,56 @@
...
-            description: "The amount of committed virtual memory.");
+            description: "The amount of virtual memory allocated for this process that cannot be shared with other processes.");
     }
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="ProcessMetrics"/> class.
-    /// </summary>
-    /// <param name="options">The options to define the metrics.</param>
-    public ProcessMetrics(ProcessInstrumentationOptions options)
+    private void Snapshot()
```
noahfalk, line 69: Have you seen any practical examples where the lack of a snapshot caused trouble? For comparison, the CPU usage, GC memory usage, and total VM usage EventCounters have been present in .NET Core since .NET Core 3.1, there is nothing that forces them to be read as a single snapshot, and I am not aware of any user feedback saying that it has ever caused someone trouble.

Yun-Ting: Addressing the comment here:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/718

PR #669, opened 2022-09-29
[Instrumentation.MassTransit] Add MassTransit ActivitySource to configuration (#326)

### LINKED ISSUE #326: MassTransit instrumentation is using the incorrect diagnostic source
When using MassTransit 8.0.1 and OpenTelemetry.Instrumentation.Masstransit 1.0.0-beta3 I can't find the traces. To fix it I added the the MassTransit source `builder.AddSource("MassTransit");`.

This is because MassTransit instrumentation is not using the correct diagnostic source. 

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.MassTransit/TracerProviderBuilderExtensions.cs#L45

```
builder.AddSource(MassTransitDiagnosticListener.ActivitySourceName);
```

`MassTransitDiagnosticListener.ActivitySourceName` value is `"OpenTelemetry.Instrumentation.MassTransit"` but it should be `"MassTransit"`.

Bartizan:
Fixes #326.

## Changes

Added MassTransit source to AddMassTransitInstrumentation extension to make it traceable for MassTransit v8.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.MassTransit/TracerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.MassTransit/TracerProviderBuilderExtensions.cs
@@ -43,6 +43,7 @@ public static class TracerProviderBuilderExtensions
 
         builder.AddInstrumentation(() => new MassTransitInstrumentation(options));
         builder.AddSource(MassTransitDiagnosticListener.ActivitySourceName);
+        builder.AddSource(MassTransitInstrumentation.MassTransitDiagnosticListenerName);
 
         builder.AddLegacySource(OperationName.Consumer.Consume);
         builder.AddLegacySource(OperationName.Consumer.Handle);


utpilla:
Adding @alexvaluyskiy who is the owner of MassTransit instrumentation library to review this PR.

utpilla:
@Bartizan From the #326, it looks like this package is not needed for MassTransit 8.0. For MassTransit 7.0, this package just works fine. Is this change really needed?

Bartizan:
> @Bartizan From the #326, it looks like this package is not needed for MassTransit 8.0. For MassTransit 7.0, this package just works fine. Is this change really needed?
> 
> I think the right thing to do here is mentioned in this comment: [#326 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/326#issuecomment-1195655638)

@utpilla I see many of instrumentation packages are basically providing the name of a custom source like in that suggestion for MT (_AddSource("MassTransit")_) So as a user I would expect the MT instrumentation package work the same way without need to remember how custom source named.

utpilla:
> @utpilla I see many of instrumentation packages are basically providing the name of a custom source like in that suggestion for MT (_AddSource("MassTransit")_) So as a user I would expect the MT instrumentation package work the same way without need to remember how custom source named.

Before MT8.0, the MassTransit library used DiagnosticSource to write events and was using the legacy way of creating activities. The instrumentation library listens to these DiagnositceSource events and gets hold of the correct activity after which it sets updates the activity appropriately to be processed by TracerProvider.

It looks like starting MT 8.0, the MassTransit library itself is using the recommended ActivitySource approach to create activities. So, we don't need to listen to DiagnosticSource events anymore. That's why the instrumentation library shouldn't be used for MT 8.0 and above.

Kielek:
@utpilla, I think that the instrumentation library could be useful even for Mass Transit 8+.

Npgsql is using ActivitySource in the recommended way and also creates https://github.com/npgsql/npgsql/tree/main/src/Npgsql.OpenTelemetry. The only task for this library is to call `builder.AddSource("Npgsql");`.

As this library: 
 - does not have dependency on MassTransit
 - does not have stable release

We could have two methods:
 - `AddMassTransitInstrumentation`,
 - [Obsolete] `AddLegacyMassTransitInstrumentation`.

In the future just drop the legacy method.

utpilla:
> @utpilla, I think that the instrumentation library could be useful even for Mass Transit 8+.

I think it could add more confusion to users. If MT 8.0+ is not using DiagnosticSource events anymore, then none of the [instrumentation options](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.MassTransit/MassTransitInstrumentationOptions.cs) make sense. The users might be configuring some options thinking it might be applied to their instrumentation, but it won't for MT 8.0+.

I also don't think that `npgsql` is doing it right. 😄.  They also have the issue of allowing the user to provide an `Action` to configure options but they just drop it in the extension method.

PR #683, opened 2022-10-10
[Instrumentation.Runtime] Reintroduce heap size metric

### LINKED ISSUE #682: Reintroduce process.runtime.dotnet.gc.heap metric

* [x] Feature Request
* [ ] Bug

Add an instrument that reports current heap size at the time of observation. 

## Additional Context

It is beneficial to have an instrument that reports value returned by `GC.GetTotalMemory(false)` as it returns current size at the time of observation, rather than size as observed after the latest gc.

`process.runtime.dotnet.gc.heap` was initially added and then [removed](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/430).

Past discussion: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/404#discussion_r894116834

Runtime publishes counters for both [current size](https://github.com/dotnet/runtime/blob/c325a5a273c242b418100828d0ce0c6d2f670b71/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L89), and sizes as observed [after the latest gc](https://github.com/dotnet/runtime/blob/c325a5a273c242b418100828d0ce0c6d2f670b71/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L106).

lachmatt:
Fixes #682.

## Changes

Add an instrument to report current heap size at the time of observation.
Rename from previous `process.runtime.dotnet.gc.heap` to `process.runtime.dotnet.gc.heap.live` to disambiguate from `process.runtime.dotnet.gc.heap.size`

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -51,6 +51,13 @@ internal class RuntimeMetrics
             () => GetGarbageCollectionCounts(),
             description: "Number of garbage collections that have occurred since process start.");
 
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            "process.runtime.dotnet.gc.heap.live",
+            () => GC.GetTotalMemory(false),
+            "bytes",
+            "Count of bytes currently in use by live objects in the GC heap.");
+
 #if NETCOREAPP3_1_OR_GREATER
         MeterInstance.CreateObservableCounter(
             "process.runtime.dotnet.gc.allocations.size",
--- a/test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
+++ b/test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs
@@ -75,6 +75,9 @@ public class RuntimeMetricsTests
         var gcCountMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.collections.count");
         Assert.NotNull(gcCountMetric);
 
+        var currentHeapSizeMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.heap.live");
+        Assert.NotNull(currentHeapSizeMetric);
+
 #if NETCOREAPP3_1_OR_GREATER
         var gcAllocationSizeMetric = exportedItems.FirstOrDefault(i => i.Name == "process.runtime.dotnet.gc.allocations.size");
         Assert.NotNull(gcAllocationSizeMetric);


pellared (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -51,6 +51,13 @@ static RuntimeMetrics()
             () => GetGarbageCollectionCounts(),
             description: "Number of garbage collections that have occurred since process start.");
 
+        // TODO: change to ObservableUpDownCounter
+        MeterInstance.CreateObservableGauge(
+            "process.runtime.dotnet.gc.heap.live",
+            () => GC.GetTotalMemory(false),
+            unit: "bytes",
+            description: "Count of bytes currently in use by live objects in the GC heap.");
```
reyang, line 59: I guess this is the estimated bytes allocated, not necessarily "used by live objects" (e.g. I think it will contain fragmentation/gaps that are not connected to any live objects").

reyang: Maybe we can borrow some wording from https://learn.microsoft.com/en-us/dotnet/api/system.gc.gettotalmemory?view=net-6.0#returns, e.g. "A number that is the best available approximation of the number of bytes currently allocated in managed memory."

cijothomas: +1 to improve wording to indicate this is an approximation.

lachmatt: Current wording comes from [here](https://github.com/dotnet/runtime/blob/4822e3c3aa77eb82b2fb33c9321f923cf11ddde6/src/coreclr/System.Private.CoreLib/src/System/GC.cs#L327).

> I guess this is the estimated bytes allocated, not necessarily "used by live objects" (e.g. I think it will contain fragmentation/gaps that are not connected to any live objects").

Unless I'm looking at the wrong place, looking at the implementation (e.g `gc.cpp`, line 44841 for 6.0.0), fragmentation is specifically excluded.

With that in mind, if wording from docs is preferred, I will use it instead.

reyang: Thanks @lachmatt! Let me consult the owner of GC and see if we can improve the description here.

reyang: @noahfalk looking at the actual runtime implementation, I think @lachmatt is correct.
Guess there are two things:
1. I think the description "Count of bytes currently in use by live objects in the GC heap" is heading towards the right direction. In addition, it'll be great to call out whether it covers objects that are no longer referenced by the root but haven't yet been recycled.
2. Looks like we need to fix the description in https://learn.microsoft.com/en-us/dotnet/api/system.gc.gettotalmemory as the current one is misleading.

noahfalk: Sorry I've had this on the todo list since last week and just now getting to it. I know that Maoni (the GC owner) told me in the past that GetTotalMemory() had some issues and she was discouraging its use. I don't recall exactly what those issues were but I'll sync with her and find out.

Aside from any API specific issues, I see that the goal of this was to have a measurement that updates frequently rather than one that is snapshotted at the last GC. If we think that updates intra-GC are important, ideally I'd like to get the existing counters to update more frequently rather than have some GC counters that are frequent and some not. I'm not certain we'll be able to achieve that, but I'd like to investigate that possibility as the preferred option and make adding a new counter as a fallback option. I'll chat with Maoni and loop back here, hopefully within a few days.

noahfalk: I chatted with Maoni and now I'm back with some more questions :)

There are two major differences between the existing [gc-committed](https://github.com/dotnet/runtime/blob/c325a5a273c242b418100828d0ce0c6d2f670b71/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L103) counter and the newly proposed counter:
a. gc-committed only updates at GC collection time whereas GC.GetTotalMemory() attempts to capture an instantaneous value
b. gc-committed includes all fragmentation and yet-to-be-used committed VM whereas GC.GetTotalMemory() excludes those portions. This means if you call GC.GetTotalMemory() the instant a GC occurs it could still report a very different number than "gc-committed" counter. For example "gc-committed" might say 3GB and GC.GetTotalMemory() would say 2GB and both of them would be correct for what they intended to measure.

I want to learn which of these two differences do we care about and what scenario makes it important?

lachmatt: Motivation was to emit metric (which uses API available on all frameworks targeted by this package) similar to currently provided by the runtime with `gc-heap-size` counter, which could be used as a top-level gc size value for debugging scenarios similar to described in this [tutorial](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/debug-memory-leak).
The counter mentioned in the debugging tutorial uses the same [API](https://github.com/dotnet/runtime/blob/50f7ab606775184eeb6d1a7eeb6e5c4e7814ef24/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/RuntimeEventSource.cs#L89) as proposed in this PR.

Kielek: @noahfalk, do you have any recommendation based on @lachmatt comment or it will be _good enough_ to use metric from this PR?

noahfalk: Sorry for the delay, I've been jumping between a number of different issues. I've been chatting more with Maoni and we've got some misgivings about using for using GC.GetTotalMemory() in runtime's own counter. My plan for the moment is to write up a proposal and open a runtime GitHub issue to discuss deprecating the gc-heap-size counter. As part of that we would change the tutorial to reference the gc-committed counter instead. If the guidance in that tutorial changed would there be any other motivation to add this counter or would it now be moot?

noahfalk: This is the issue I just opened in the runtime repo discussing moving/removing/hiding the gc-heap-size counter: https://github.com/dotnet/runtime/issues/77530

pjanotti: As @lachmatt  mentioned the only downside to `gc-committed` is that it requires .NET 6.0 so no .NET Framework support and we still see many users on .NET Framework. I'm not concerned about .NET Core 3.1 given its upcoming EOL.

In the same line with proposals 1 & 5 on https://github.com/dotnet/runtime/issues/77530 we could document that the counter is available for legacy purposes and that `gc-committed` is preferred to track actual memory used by GC.

pjanotti: Thinking out loud: how much `gc-committed` diverges from the native perf counter tracking MEM_COMMIT for the typical .NET Framework application?

If the overall consensus is that adding `gc-heap-size` is not desirable. The docs could guide users that lack the `gc-committed` to the proper native counter, and explain what is expected to be different with it.

noahfalk: Depending on the app, total committed memory and GC committed memory could be very different numbers. I've been continuing to noodle on this and the feedback from https://github.com/dotnet/runtime/issues/77530. I think the main commonality from proposals (1) and (5) is that neither of them gets rid of the heap size metric, they just try to re-position it to reduce accidental misuse. If this is will be the only metric we have for .NET Framework then maybe rather than avoiding it we just be careful with naming it. From my perspective, the key distinction of GetTotalMemory() isn't that it updates between GCs, it is that it doesn't include fragmentation and other GC committed memory pools. Perhaps if we named it process.runtime.dotnet.gc.heap.total_objects_size that would make it clearer what this metric is really measuring.

Kielek (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -48,6 +48,12 @@ static RuntimeMetrics()
             () => GetGarbageCollectionCounts(),
             description: "Number of garbage collections that have occurred since process start.");
 
+        MeterInstance.CreateObservableUpDownCounter(
+            "process.runtime.dotnet.gc.heap.total_objects_size",
```
reyang, line 52: This is very inconsistent with other metrics names (e.g. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/005fdb71b709d5eefbc8c5b7fa5a2c3f14fddecf/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs#L53) in the existing library. Consider:

```suggestion
            "process.runtime.dotnet.gc.live_objects.size",
```

lachmatt: renamed in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/683/commits/25e9cf80f5cffbd2e81960003220bc0445ef7632

noahfalk: 'live' usually means that an object is reachable via references from a root, but that isn't necessarily true for these objects at the current instant in time. We know these objects haven't been collected yet, but they may already be unreachable. We can brainstorm different names but I would avoid "live". For example "process.runtime.dotnet.gc.objects.size"?

reyang: > 'live' usually means that an object is reachable via references from a root, but that isn't necessarily true for these objects at the current instant in time. We know these objects haven't been collected yet, but they may already be unreachable. We can brainstorm different names but I would avoid "live". For example "process.runtime.dotnet.gc.objects.size"?

Sounds good to me, and the name "process.runtime.dotnet.gc.objects.size" is aligned with the existing ones.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -48,6 +48,12 @@ static RuntimeMetrics()
             () => GetGarbageCollectionCounts(),
             description: "Number of garbage collections that have occurred since process start.");
 
+        MeterInstance.CreateObservableUpDownCounter(
+            "process.runtime.dotnet.gc.live_objects.size",
+            () => GC.GetTotalMemory(false),
+            unit: "bytes",
+            description: "Count of bytes currently in use by live objects in the GC heap. Fragmentation and other GC committed memory pools are excluded.");
```
reyang, line 55: "live objects" is inaccurate/misleading per @noahfalk's comment https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/683#discussion_r1018055369.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
lachmatt: renamed in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/683/commits/48e021b49143b68fa59ad9f95c677b72ee71ca78

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
lachmatt: description updated in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/683/commits/48e021b49143b68fa59ad9f95c677b72ee71ca78

reyang (approved):
LGTM.

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
```diff
@@ -34,6 +34,9 @@
 
   This does not affect applications targeting .NET Framework.
 
+* Add "process.runtime.dotnet.gc.objects.size" metric
```
cijothomas, line 37: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Runtime#processruntimedotnetgcheapsize 
Could you update the readme file as well?

cijothomas: would be also add a note about how is it different from the gc.heap.size (the fact that this exclude fragmentation, and works anytime even before the 1st GC event itself etc.) A lot such useful info was in this PR as comments. Its good to get the gist into the readme file itself to not lose them :)

cijothomas (approved):
Thanks.
The changes look good. Requested to update readme file as well. (It could be done as a follow PR as well)

cijothomas:
@lachmatt could you also update PR desc to reflect the final content of the PR?

lachmatt:
> @lachmatt could you also update PR desc to reflect the final content of the PR?

Updated.

cijothomas:
Merging. A follow up is needed to update readme.
Also, if there are any feedbacks about the instrument description, please reply here/open a new issue.

noahfalk (commented):
LGTM 👍

PR #686, opened 2022-10-11
[Instrumentation.StackExchangeRedis] StackExchangeRedis Instrumentation to leverage native Activity Status

Yun-Ting:
Related to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/707


vishweshbankwar:
Need changelog entry

src/OpenTelemetry.Instrumentation.StackExchangeRedis/Implementation/RedisProfilerEntryToActivityConverter.cs
```diff
@@ -107,8 +107,6 @@ public static Activity ProfilerCommandToActivity(Activity parentActivity, IProfi
             // Total:
             // command.ElapsedTime;             // 00:00:32.4988020
 
-            activity.SetStatus(Status.Unset);
```
cijothomas, line 110: there was never a case where we set status to error?

PR #703, opened 2022-10-17
[Instrumentation.EventCounters] Avoid race condition and performance trap in EventCountersMetrics

MihaZupan:
This PR fixes two issues with the current implementation:
- Uses `EventLevel.Critical` instead of `EventLevel.LogAlways`
  - Counters don't require a specific log level to work. Depending on the `EventSource` implementor, this change can avoid a lot of overhead (preparing and serializing event arguments, potentially taking more expensive code paths).
- Avoids a race condition when initializing the `EventCountersMetrics` instance
  - As `OnEventSourceCreated` can run before the constructor is done, you can run into a situation where you fail to enable an `EventSource` instance. A simple lock avoids this race.
  - Example course of events:
    - Thread A calls into `OnEventSourceCreated` and sees options aren't set yet
    - Thread B executes the constructor, processing any entries in `preInitEventSources`
    - Thread A adds the `EventSource` into `preInitEventSources`
    - Nothing will process and enable that `EventSource` instance

### DIFF
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -71,11 +71,6 @@ internal sealed class EventCountersMetrics : EventListener
             }
         }
     }
-    
-    private void EnableEvents(EventSource eventSource)
-    {
-         this.EnableEvents(eventSource, EventLevel.Critical, EventKeywords.None, GetEnableEventsArguments(this.options));
-    }
 
     /// <inheritdoc />
     protected override void OnEventWritten(EventWrittenEventArgs eventData)
@@ -124,6 +119,11 @@ internal sealed class EventCountersMetrics : EventListener
 
     private static Dictionary<string, string> GetEnableEventsArguments(EventCountersInstrumentationOptions options) =>
         new() { { "EventCounterIntervalSec", options.RefreshIntervalSecs.ToString() } };
+
+    private void EnableEvents(EventSource eventSource)
+    {
+        this.EnableEvents(eventSource, EventLevel.Critical, EventKeywords.None, GetEnableEventsArguments(this.options));
+    }
 
     private void UpdateInstrumentWithEvent(bool isGauge, string eventSourceName, string name, double value)
     {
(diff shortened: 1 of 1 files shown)


utpilla:
@MihaZupan Thanks for sending this PR! Could you please fix the build errors?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -40,27 +40,35 @@ internal sealed class EventCountersMetrics : EventListener
...
+            this.preInitEventSources.Clear();
         }
     }
 
     /// <inheritdoc />
-    protected override void OnEventSourceCreated(EventSource source)
+    protected override void OnEventSourceCreated(EventSource eventSource)
     {
-        if (this.options == null)
+        lock (this.preInitEventSources)
```
utpilla, line 62: Use double-checked locking and remove the else condition from inside the lock. Something like this:

```csharp
        if (this.options == null)
        {
            lock (this.preInitEventSources)
            {
                if (this.options == null)
                {
                    this.preInitEventSources.Enqueue(source);
                }
            }
        }
        else if (this.options.ShouldListenToSource(source.Name))
        {
            this.EnableEvents(source, EventLevel.LogAlways, EventKeywords.None, GetEnableEventsArguments(this.options));
        }
```

MihaZupan: Why?
That would just reintroduc the race condition:
- Thread A sees `this.options == null`
- Thread B runs the constructor and sets `options`
- Thread A checks `this.options == null` again but that now fails as the options are set
- The `EventSource` thread A just processed gets dropped silently

utpilla: > Why?

This avoids the thread having to acquire a lock on every `OnEventSourceCreated` call. We don't need the threads to acquire a lock once the options are set.


> Thread A checks this.options == null again but that now fails as the options are set

Umm yeah you could add the else condition inside the lock statement as well to fix that.

MihaZupan: > This avoids the thread having to acquire a lock on every OnEventSourceCreated call.

I don't see why that's a problem though, this method is only called a dozen times in a given process. Complicating the locking algorithm is not worth it in this case, uncontended lock acquisitions are extremely cheap anyway.

utpilla: The change doesn't seem that complicated to me. Having said that, I don't have a strong opinion on this. I'm okay with always locking in this case.

utpilla (approved):

mic-max (approved):
LGTM Thanks!

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
cijothomas: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Internal/SelfDiagnosticsEventListener.cs similar pattern being followed already

cijothomas (approved):

PR #718, opened 2022-10-19
[Instrumentation.Process] Dropping Snapshot

Yun-Ting:
Update to use Process.Refresh() on each callback after benchmarking the perf https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/717 and found it cheap.
Also, addressing https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/664#discussion_r993403498.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
@@ -28,12 +28,6 @@ internal sealed class ProcessMetrics
     internal readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
 
     private readonly Diagnostics.Process currentProcess = Diagnostics.Process.GetCurrentProcess();
-    private double? memoryUsageBytes;
-    private double? virtualMemoryUsageBytes;
-    private double? userProcessorTimeSeconds;
-    private double? privilegedProcessorTimeSeconds;
-    private int? numberOfThreads;
-    private IEnumerable<Measurement<double>> cpuUtilization;
 
     // vars for calculating CPU utilization
     private DateTime lastCollectionTimeUtc;
@@ -51,14 +45,8 @@ internal sealed class ProcessMetrics
             "process.memory.usage",
             () =>
             {
-                if (!this.memoryUsageBytes.HasValue)
-                {
-                    this.Snapshot();
-                }
-
-                var value = this.memoryUsageBytes.Value;
-                this.memoryUsageBytes = null;
-                return value;
+                this.currentProcess.Refresh();
+                return this.currentProcess.WorkingSet64;
             },
             unit: "By",
             description: "The amount of physical memory allocated for this process.");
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
```diff
@@ -68,14 +56,8 @@ public ProcessMetrics(ProcessInstrumentationOptions options)
...
             {
-                if (!this.virtualMemoryUsageBytes.HasValue)
-                {
-                    this.Snapshot();
-                }
-
-                var value = this.virtualMemoryUsageBytes.Value;
-                this.virtualMemoryUsageBytes = null;
-                return value;
+                this.currentProcess.Refresh();
```
reyang, line 59: Do we really need to keep using the same instance and do `Refresh` each time, or just creating a new Process instance is sufficient? Would the `Refresh` call here on the same instance cause race condition?

Yun-Ting: Race condition is possible here. I don't know what's the guarantee of the .Refresh() API so if we want to do Refresh on each callback, locking it might be the safest.

Not sure which one is cheaper, refresh + lock or creating a new instance each time. I'll do some experiments.

Yun-Ting: I've benchmarked the above 2 options https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/717#issue-1415674535.
There is no significance difference in time. I updated the code to use the latter option.

noahfalk (approved):
LGTM!

(I still have concerns about the cpu utilization metric and using PrivateMemory for the process.memory.virtual metric, but neither of those concerns needs to hold up this PR)

reyang (approved):

src/OpenTelemetry.Instrumentation.Process/ProcessMetrics.cs
reyang: Thanks!

Kielek (approved):

test/OpenTelemetry.Instrumentation.Process.Tests/ProcessMetricsTests.cs
```diff
@@ -158,7 +158,7 @@ private static double GetValue(Metric metric)
         {
             if (metric.MetricType.IsGauge())
             {
-                sum += metricPoint.GetGaugeLastValueDouble();
+                sum += metricPoint.GetGaugeLastValueLong();
```
utpilla, line 161: Why this change? Also, the if-else condition seems incorrect. Were you trying to do something similar to this: 

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/test/OpenTelemetry.Instrumentation.Runtime.Tests/RuntimeMetricsTests.cs#L187-L198

Yun-Ting: process.memory.usage is using .NET API WorkingSet64 which is `long`.
Previously, we used `double?` to cache the value. Thus, the change.

I've removed the unnecessary branching.

CodeBlanch (approved):
LGTM

utpilla (approved):

PR #726, opened 2022-10-21
Fix using version 3.7.100 of AWS SDK causing exception not finding EndpointResolver

normj:
## Changes
Starting with version 3.7.100 of the AWS .NET SDK the EndpointResolver is replaced with a service specific version of the handler. This breaks the OpenTelemetry integration because OpenTelemetry was inserting itself just after EndpointResolver in the SDK's request pipeline.

The fix is for OpenTelemetry to inject itself before the RetryHandler which happens after endpoint resolutions. So the position will be the same in the request pipeline.

This fix will is compatible with both version 3.7.100 and versions before 3.7.100 so it will not break users that haven't upgrade their SDK version. 

This change is very similar to the AWS X-Ray change which this code is based off: https://github.com/aws/aws-xray-sdk-dotnet/pull/268 

### DIFF
--- a/src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog - OpenTelemetry.Contrib.Instrumentation.AWS
 
+## Unreleased
+
+* Fixed issue when using version 3.7.100 of the AWS SDK for .NET triggering an
+EndpointResolver not found exception.
+
 ## 1.0.1
 
 Released 2021-Feb-24
--- a/src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## Unreleased
 
 * Fixed issue when using version 3.7.100 of the AWS SDK for .NET triggering an
-EndpointResolver not found exception.
+  EndpointResolver not found exception.
+  ([#726](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/726))
 
 ## 1.0.1
 
--- a/src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineCustomizer.cs
+++ b/src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineCustomizer.cs
@@ -44,6 +44,6 @@ internal class AWSTracingPipelineCustomizer : IRuntimePipelineCustomizer
             return;
         }
 
-        pipeline.AddHandlerAfter<EndpointResolver>(new AWSTracingPipelineHandler(this.options));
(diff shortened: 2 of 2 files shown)


src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineCustomizer.cs
```diff
@@ -44,6 +44,6 @@ public void Customize(Type serviceClientType, RuntimePipeline pipeline)
             return;
         }
 
-        pipeline.AddHandlerAfter<EndpointResolver>(new AWSTracingPipelineHandler(this.options));
+        pipeline.AddHandlerAfter<RetryHandler>(new AWSTracingPipelineHandler(this.options));
```
boblodgett, line 47: "The fix is for OpenTelemetry to inject itself before the RetryHandler"

But this is injecting AFTER. Which is correct? The summary or the code in this case?

normj: Fixed, good catch.

boblodgett (approved):

utpilla:
@normj Thanks for sending this PR! Could you please sign the CLA?

src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog - OpenTelemetry.Contrib.Instrumentation.AWS
 
+
+## 1.0.2
+
+* Fixed issue when using version 3.7.100 of the AWS SDK for .NET triggering an EndpointResolver not found exception.
```
klementi, line 6: AssemblyVersion change ?

normj:
> @normj Thanks for sending this PR! Could you please sign the CLA?

Thanks and done

src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog - OpenTelemetry.Contrib.Instrumentation.AWS
 
+## 1.0.2
+
+Released 2022-Oct-21
```
srprash, line 5: We don't yet know when this will be released. Let's just put this change under an `Unreleased` heading.

normj: Done

srprash (approved):

src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog - OpenTelemetry.Contrib.Instrumentation.AWS
 
+## Unreleased
+
+* Fixed issue when using version 3.7.100 of the AWS SDK for .NET triggering an
+EndpointResolver not found exception.
```
Kielek, line 6: Please add link to the PR in the changelog:

```suggestion
* Fixed issue when using version 3.7.100 of the AWS SDK for .NET triggering an
  EndpointResolver not found exception.
  ([#726](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/726))
```

Kielek (approved):

PR #735, opened 2022-10-26
Combine Windows and Linux CI Workflows

mic-max:
## Changes

Merged the Windows and Linux specific build test CI workflows into a single one.

- Sister PR https://github.com/open-telemetry/opentelemetry-dotnet/pull/3786/files


Kielek (approved):
Before merge it required builds have to updated in repo settings.

.github/workflows/ci-md.yml
```diff
@@ -19,7 +19,11 @@ jobs:
 
     strategy:
       matrix:
-        version: [netcoreapp3.1,net6.0]
+        os: [ windows-latest, ubuntu-latest ]
```
utpilla, line 22: Update `runs-on` for this workflow as well?

mic-max: Running it on ubuntu regardless of the value of `matrix.os` was done intentionally. The matrix was setup in a way just to mirror the `ci.yml` `strategy` so that GitHub would categorize it as the same check but it doesn't care what the `runs-on` value is. More info here: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

Actually it doesn't even know that the `os` field is for operating systems 😅

PR #737, opened 2022-10-27
Link fixes

Kielek:
Fixes # N/A

## Changes

~~Propagate changes from https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/1503~~
Fix links.


.github/workflows/validate-documentation.yml
```diff
@@ -16,7 +16,13 @@ jobs:
...
     steps:
     - name: check out code
-      uses: actions/checkout@v3
+      uses: actions/checkout@v3.1.0
+
+    - name: verify links by lycheeverse
+      uses: lycheeverse/lychee-action@v1.5.1
+      with:
+        fail: true
+        args: "--threads 1 --max-concurrency 1 --verbose --no-progress './**/*.md' './**/*.html'"
```
Kielek, line 25: In this repository there is tons of links to the GitHub. It leads to throttling. One of the way is to use once request at a time.

pellared, line 25: we could consider  removing `'./**/*.html'` and reverting the workflow name to `markdownlint` if that would lead to any problem.

_For me it could stay as it is 😄 _

pellared (approved):

.github/workflows/validate-documentation.yml
```diff
@@ -16,7 +16,13 @@ jobs:
 
     steps:
     - name: check out code
-      uses: actions/checkout@v3
+      uses: actions/checkout@v3.1.0
+
+    - name: verify links by lycheeverse
+      uses: lycheeverse/lychee-action@v1.5.1
```
cijothomas, line 22: do you know if this is used widely? Any idea how opentelemetry.io folks deal with this ?

pellared: - https://github.com/open-telemetry/opentelemetry-go/blob/ad45631b53faa74191fcee37c8f010e520af67e1/.github/workflows/links.yml#L19
- https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/b0d4457610231e5e3a8af32f8856a9d0b0fbdf8d/.github/workflows/validate-documentation.yml#L28
- also proposed here: https://github.com/open-telemetry/opentelemetry-go-instrumentation/pull/11

in opentelemetry.io it looks like it checks hyperlinks in HTML using `htmltest`: https://github.com/open-telemetry/opentelemetry.io/blob/9fb34a099bd99fd37dc526cc3a6c09e7d037fd15/Makefile#L19

Some OTel repos also use https://github.com/tcort/markdown-link-check (AFAIK it is used in otel-spec, otel-demo and otel-collector).

Personally, I prefer https://github.com/lycheeverse/lychee as:

1. it can be used not only for markdown
1. you can run it "natively" (created in Rust so you can just grab a binary)
1. it is under a GitHub organization (not a user)

.github/workflows/validate-documentation.yml
Kielek: File filers are taken from default lychee configuration. There is no possibility to set other custom attributes and not setting file filters.

I would keep as it is.

Kielek:
@cijothomas , I think that it is worth at least to merge link fixes if you prefer not to check links every build. Should I remove CI-link validation from this PR?

Kielek:
@cijothomas, @utpilla - changes related to the ci-link validation are reverted. We can revisit it later. For now, I would like to merge only fixed links.

cijothomas (approved):

PR #740, opened 2022-10-28
[Instrumentation.EventCounters] Support long event source names better in EventCounters instrumentation

lmolkova:
EventCounters instrumentation does not work for several [well-known EventSources](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/available-counters?source=recommendations).

E.g. `Microsoft-AspNetCore-Server-Kestrel` or most of the counters in `Microsoft.AspNetCore.Http.Connections`.
It happens because of instrument length limitations (63 symbols).

End users can't do anything about it except republish all events, so we need to find a reasonable way to allow at least well-known event sources. Kestrel counters, for example, are extremely useful and can't be measured in other ways (DiagSource/ActivitySource instrumentation in hosting).

So, `EventCounters.Microsoft-AspNetCore-Server-Kestrel.` is already 50 chars and it doesn't even include event name.

This change proposes:
- use `ec.` prefix instead of `EventCounters.` - It seems less important than event source/event names
- When the instrument name in EventCounters is too long, I ~~suggest abbreviating~~ shorten the event source portion of it:
 `ec.Microsoft-AspNetCore-Server-Kestrel.tls-handshakes-per-second` -> `ec.Microsoft-AspNetCore-Server-Kestr.tls-handshakes-per-second`. It generates stable and common name for a single instrument.

Happy to consider other options for shortening the name. 

### DIFF
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -140,7 +140,15 @@ internal sealed class EventCountersMetrics : EventListener
             return string.Concat(Prefix, ".", sourceName, ".", eventName);
         }
 
-        var abbreviation = new StringBuilder(sourceName.Length);
+        var maxEventSourceLength = MaxInstrumentNameLength - Prefix.Length - 1 - eventName.Length;
+        if (maxEventSourceLength < 2) 
+        {
+            // event name is too long, there is not enough space for sourceName.
+            // let ec.<eventName> flow to metrics SDK and filter if needed.
+            return string.Concat(Prefix, ".", eventName);
+        }
+
+        var abbreviation = new StringBuilder(maxEventSourceLength);
         int ind = 0;
         while (ind >= 0 && ind < sourceName.Length)
         {
@@ -151,6 +159,11 @@ internal sealed class EventCountersMetrics : EventListener
 
             if (ind < sourceName.Length)
             {
+                if (abbreviation.Length + 2 >= maxEventSourceLength)
+                {
+                    break;
+                }
+
                 abbreviation.Append(char.ToLowerInvariant(sourceName[ind])).Append('.');
             }
 
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -126,10 +126,10 @@ internal sealed class EventCountersMetrics : EventListener
     /// <summary>
     /// If instrument name is too long, abbreviates event source name.
     /// E.g. instrument for `Microsoft-AspNetCore-Server-Kestrel`, `tls-handshakes-per-second`
-    /// would be too long (64 chars), so it's abbreviated to `ec.m.a.s.k.tls-handshakes-per-second`
-    /// instead of `ec.Microsoft-AspNetCore-Server-Kestrel.tls-handshakes-per-second`.
-    ///
-    /// If after that instrument name is still invalid, it will be validated and ignored later in the pipeline.
+    /// would be too long (64 chars), so it's shortened to `ec.Microsoft-AspNetCore-Server-Kestre.tls-handshakes-per-second`.
+    /// 
+    /// If there is no room for event source name, returns `ec.{event name}` and 
+    /// if it's still too long, it will be validated and ignored later in the pipeline.
     /// </summary>
     private static string GetInstrumentName(string sourceName, string eventName)
     {
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -37,6 +37,7 @@ internal sealed class EventCountersMetrics : EventListener
     private readonly List<EventSource> preInitEventSources = new();
     private readonly ConcurrentDictionary<(string, string), Instrument> instruments = new();
     private readonly ConcurrentDictionary<(string, string), double> values = new();
+
     /// <summary>
     /// Initializes a new instance of the <see cref="EventCountersMetrics"/> class.
     /// </summary>
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -124,11 +124,12 @@ internal sealed class EventCountersMetrics : EventListener
         new() { { "EventCounterIntervalSec", options.RefreshIntervalSecs.ToString() } };
 
     /// <summary>
-    /// If instrument name is too long, abbreviates event source name.
+    /// If the resulting instrument name is too long, it trims the event source name
+    /// to fit in as many characters as possible keeping the event name intact.
     /// E.g. instrument for `Microsoft-AspNetCore-Server-Kestrel`, `tls-handshakes-per-second`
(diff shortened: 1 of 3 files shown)


mic-max:
Thanks for the change, I hadn't realised that some of the well-known counters were going to fail this!
The main point of prefixing with `EventCounters.` is just to help avoid possible collisions, so I think changing to `ec.` is fine and saves on the character count.
I agree that if anything needs to be shortened in the instrument name it should be the event source and then if still too long after doing so the SDK can just fail and log its warning.

mic-max (approved):

Kielek (approved):
Approved, but please add entry to the changelog under unreleased section https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.EventCounters/CHANGELOG.md
(Short description + link to this PR).

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -132,7 +192,7 @@ private void UpdateInstrumentWithEvent(bool isGauge, string eventSourceName, str
             ValueTuple<string, string> metricKey = new(eventSourceName, name);
             _ = this.values.AddOrUpdate(metricKey, value, isGauge ? (_, _) => value : (_, existing) => existing + value);
 
-            var instrumentName = $"EventCounters.{eventSourceName}.{name}";
+            var instrumentName = GetInstrumentName(eventSourceName, name);
```
utpilla, line 195: Move this inside the `if` check below to only call `GetInstrumentName` method when `instruments` does not contain the given `metricKey`.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -120,6 +124,62 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
+        var abbreviation = new StringBuilder(maxEventSourceNameLength);
+        int ind = 0;
+        while (ind >= 0 && ind < sourceName.Length)
+        {
+            while (ind < sourceName.Length && (sourceName[ind] == '.' || sourceName[ind] == '-'))
+            {
+                ind++;
+            }
+
+            if (ind < sourceName.Length)
```
utpilla, line 153: Check for `abbreviation.Length < maxEventSourceNameLength` as well.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -120,6 +124,62 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
+    /// </summary>
+    private static string GetInstrumentName(string sourceName, string eventName)
+    {
+        int totalLength = Prefix.Length + 1 + sourceName.Length + 1 + eventName.Length;
+        if (totalLength <= MaxInstrumentNameLength)
+        {
+            return string.Concat(Prefix, ".", sourceName, ".", eventName);
+        }
+
+        int maxEventSourceNameLength = MaxInstrumentNameLength - Prefix.Length - 2 - eventName.Length;
```
utpilla, line 143: This could result in a negative number if `eventName` is long enough.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
lmolkova: sure, but what should we do if it's longer? Currently, it's suppressed by the metrics pipeline in SDK because of the name length in the same way as before.

utpilla: We could either let the Metrics SDK drop it or we could prioritize on Event name. That would mean we first count the number of characters for prefix and eventName and the two `.` characters. If this already exceeds 63 then we don't do anything about it and let the SDK drop it. If not, we try to squeeze in as much of the source name as we can. That might lead to omission of some of the initial characters of the source name and might make the name unintuitive. But we would still have the name suffixed with the full event name and that's probably better than dropping it entirely because the source name is long.

Having said that, this is already a step in the right direction. We could always extend functionality in follow-up PRs.

utpilla (approved):

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
lmolkova: Makes sense! I implemented it, but I preserve the initial characters in event source name and drop the rest once it's impossible to squeeze more of them. PTAL

Kielek (approved):

lmolkova:
@open-telemetry/dotnet-contrib-approvers, if there are no objections, can someone please merge it?

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
utpilla: We are still not utilizing the full 63 characters length limit.

Consider this example:

SourceName: "CustomEventSource" (17 characters)
EventName: "longEventName-abcdefghijklmnopqrstuvwxyz-1234567890" (51 characters)

What should we return as the instrument name for this example?

With the current changes in the PR, we would return:

"ec.**c**.longEventName-abcdefghijklmnopqrstuvwxyz-1234567890" (56 characters)

While we had room for 7 more characters which we didn't use at all. In that case, the result would have been:
"ec.**CustomEv**.longEventName-abcdefghijklmnopqrstuvwxyz-1234567890" (63 characters)

I guess we should try to add as many characters as we can, but I would like to know your thoughts.

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -120,6 +123,33 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
     private static Dictionary<string, string> GetEnableEventsArguments(EventCountersInstrumentationOptions options) =>
         new() { { "EventCounterIntervalSec", options.RefreshIntervalSecs.ToString() } };
 
+    /// <summary>
+    /// If instrument name is too long, abbreviates event source name.
```
utpilla, line 127:
```suggestion
    /// If the resulting instrument name is too long, it trims the event source name to fit in as many characters as possible keeping the event name intact.
```

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -120,6 +123,33 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
...
+
+        var maxEventSourceLength = MaxInstrumentNameLength - Prefix.Length - 1 - eventName.Length;
+        if (maxEventSourceLength < 2)
+        {
+            // event name is too long, there is not enough space for sourceName.
+            // let ec.<eventName> flow to metrics SDK and it will suppress it if needed.
+            return string.Concat(Prefix, ".", eventName);
+        }
+
+        return string.Concat(Prefix, ".", sourceName.Substring(0, maxEventSourceLength - 1), ".", eventName);
```
utpilla, line 150: nit: Probably more readable to have `var maxEventSourceLength = MaxInstrumentNameLength - Prefix.Length - 2 - eventName.Length;

and use `sourceName.Substring(0, maxEventSourceLength)`

utpilla (approved):

utpilla:
I have published the package: https://www.nuget.org/packages/OpenTelemetry.Instrumentation.EventCounters/1.0.0-alpha.2

PR #757, opened 2022-11-04
Table level customization of `CustomFields`

vidommet:
Fixes #.

## Changes

CustomFields are now supported at a table level
- `` : All fields are in separate columns
- `*`: All fields are in separate columns
- `A`: Fields with name `A` from all tables are in separate columns
- `A/B`: Fields with name `B` in a table `A` are in separate column
- `A/*`: All fields in a table `A` are in separate column.

### DIFF
--- a/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
@@ -683,7 +683,7 @@ public class GenevaLogExporterTests
             {
                 // The field "customField" of LogRecord.State should be present in the mapping as a separate key. Other fields of LogRecord.State which are not present
                 // in CustomFields should be added in the mapping under "env_properties"
-                exporterOptions.CustomFields = new string[] { "customField" };
+                exporterOptions.CustomFields = new string[] { "customField", "DefaultLogRecord/customField2" };
             }
 
             using var loggerFactory = LoggerFactory.Create(builder => builder
@@ -725,6 +725,7 @@ public class GenevaLogExporterTests
             // When the exporter options are configured with TableMappings only "customField" will be logged as a separate key in the mapping
             // "property" will be logged under "env_properties" in the mapping
             logger.Log(LogLevel.Trace, 101, "Log a {customField} and {property}", "CustomFieldValue", "PropertyValue");
+            logger.Log(LogLevel.Trace, 101, "Log a {customField2} and {property}", "CustomFieldValue", "PropertyValue");
             logger.Log(LogLevel.Trace, 101, "Log a {customField} and {property}", "CustomFieldValue", null);
             logger.Log(LogLevel.Trace, 101, "Log a {customField} and {property}", null, "PropertyValue");
             logger.Log(LogLevel.Debug, 101, "Log a {customField} and {property}", "CustomFieldValue", "PropertyValue");
@@ -740,10 +741,12 @@ public class GenevaLogExporterTests
 
             var loggerWithDefaultCategory = loggerFactory.CreateLogger("DefaultCategory");
             loggerWithDefaultCategory.LogInformation("Basic test");
+            loggerWithDefaultCategory.LogInformation("Log a {customField} and {property}", "CustomFieldValue", "PropertyValue");
+            loggerWithDefaultCategory.LogInformation("Log a {customField2} and {property}", "CustomFieldValue", "PropertyValue");
             loggerWithDefaultCategory.LogInformation("\u0418"); // Include non-ASCII characters in the message
 
             // logRecordList should have 14 logRecord entries as there were 14 Log calls
-            Assert.Equal(14, logRecordList.Count);
+            Assert.Equal(17, logRecordList.Count);
 
             var m_buffer = typeof(MsgPackLogExporter).GetField("m_buffer", BindingFlags.NonPublic | BindingFlags.Static).GetValue(exporter) as ThreadLocal<byte[]>;
 
@@ -1206,7 +1209,12 @@ public class GenevaLogExporterTests
                 {
                     Assert.Equal(item.Value.ToString(), mapping["body"]);
                 }
-                else if (exporterOptions.CustomFields == null || exporterOptions.CustomFields.Contains(item.Key))
+                else if (
+                    exporterOptions.CustomFields == null ||
+                    exporterOptions.CustomFields.Contains("*") ||
+                    exporterOptions.CustomFields.Contains(item.Key) ||
+                    exporterOptions.CustomFields.Contains(partAName + "/" + "*") ||
+                    exporterOptions.CustomFields.Contains(partAName + "/" + item.Key))
                 {
                     if (item.Value != null)
                     {
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+* Add support for table level customization of customFields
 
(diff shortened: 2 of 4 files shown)


test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
```diff
@@ -1206,7 +1209,12 @@ private void AssertFluentdForwardModeForLogRecord(GenevaExporterOptions exporter
                 {
                     Assert.Equal(item.Value.ToString(), mapping["body"]);
                 }
-                else if (exporterOptions.CustomFields == null || exporterOptions.CustomFields.Contains(item.Key))
+                else if (
+                    exporterOptions.CustomFields == null ||
+                    exporterOptions.CustomFields.Contains("*") ||
+                    exporterOptions.CustomFields.Contains(item.Key) ||
+                    exporterOptions.CustomFields.Contains(partAName + "/" + "*") ||
+                    exporterOptions.CustomFields.Contains(partAName + "/" + item.Key))
```
reyang, line 1217: I guess this is going to cause tons of allocations on the hot path.

reyang (changes requested):
Need to understand the scenario - why would such flexibility (which comes with perf cost) be desirable.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
vidommet: This is for tests. Hot Path code uses hashsets and is strictly zero allocations

vidommet:
> Need to understand the scenario - why would such flexibility (which comes with perf cost) be desirable.

When we have a lot of tables with a lot of columns, we only like to pick and choose the columns depending on the table. Otherwise, names reused across tables will become dedicated columns in all tables.

The way this is implemented is using HashSets with zero allocations and the hot path should not add a lot of overhead, if any.

reyang:
> When we have a lot of tables with a lot of columns, we only like to pick and choose the columns depending on the table. Otherwise, names reused across tables will become dedicated columns in all tables.

This sounds like a schema standardization issue. Would you consider defining some conventions regarding how folks are naming columns in your group and does it help to remove the need for making complex configurations?

A high order bit is that we don't intend to make the SDK components more and more complex, so if the demand is not high enough (e.g. needed by many users), it is likely that we'll treat it as below the bar. A viable path might be some extensibility point, so teams who want more complex things can unblock themselves by absorbing the complexity.

> The way this is implemented is using HashSets with zero allocations and the hot path should not add a lot of overhead, if any.

Could you provide benchmarks and stress test results (to compare before/after)? Here are some examples:
* https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/Benchmarks
* https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/test/OpenTelemetry.Tests.Stress

vidommet:
I can give a bit more on the scenario. 
We have several tables each for different libraries that we use. They are separated using the Category names. When we consider a single table, a particular field might make sense to move to a dedicated column but the same field in a different table might not make sense.

A simple example. Consider `key`. If the library is for a dictionary implementation, we want the key to be its own column. But if it is a generic log line in a different table, we don't want to use a new column for that scenario

vidommet:
I am able to run benchmarks from the benchmarks folder and here are the results on my machine
Main branch

```
BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.755)
12th Gen Intel Core i7-12700K, 1 CPU, 20 logical and 12 physical cores
.NET SDK=7.0.100
  [Host]     : .NET 6.0.11 (6.0.1122.52304), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.11 (6.0.1122.52304), X64 RyuJIT AVX2

|                    Method | IncludeFormattedMessage |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|-------------------------- |------------------------ |---------:|---------:|---------:|-------:|----------:|
| LoggerWithMessageTemplate |                   False | 628.0 ns | 12.16 ns | 10.78 ns | 0.0229 |     304 B |
| LoggerWithDirectLoggerAPI |                   False | 545.1 ns |  9.10 ns |  7.10 ns | 0.0334 |     440 B |
| LoggerWithSourceGenerator |                   False | 574.1 ns |  4.88 ns |  3.81 ns | 0.0219 |     288 B |
|        SerializeLogRecord |                   False | 350.3 ns |  4.68 ns |  3.90 ns | 0.0076 |     104 B |
|                    Export |                   False | 376.7 ns |  7.36 ns |  7.88 ns | 0.0076 |     104 B |
| LoggerWithMessageTemplate |                    True | 606.7 ns | 11.12 ns |  9.29 ns | 0.0229 |     304 B |
| LoggerWithDirectLoggerAPI |                    True | 541.8 ns | 10.77 ns | 13.62 ns | 0.0334 |     440 B |
| LoggerWithSourceGenerator |                    True | 568.4 ns |  5.12 ns |  4.27 ns | 0.0219 |     288 B |
|        SerializeLogRecord |                    True | 347.8 ns |  5.97 ns |  5.59 ns | 0.0076 |     104 B |
|                    Export |                    True | 373.1 ns |  4.23 ns |  3.75 ns | 0.0076 |     104 B |
```

My branch
```
|                    Method | IncludeFormattedMessage |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|-------------------------- |------------------------ |---------:|---------:|---------:|-------:|----------:|
| LoggerWithMessageTemplate |                   False | 631.3 ns | 12.51 ns | 13.90 ns | 0.0238 |     312 B |
| LoggerWithDirectLoggerAPI |                   False | 539.2 ns |  8.96 ns |  8.39 ns | 0.0334 |     448 B |
| LoggerWithSourceGenerator |                   False | 597.9 ns | 11.85 ns | 11.08 ns | 0.0219 |     296 B |
|        SerializeLogRecord |                   False | 344.7 ns |  1.71 ns |  1.43 ns | 0.0081 |     112 B |
|                    Export |                   False | 382.7 ns |  2.50 ns |  2.34 ns | 0.0081 |     112 B |
| LoggerWithMessageTemplate |                    True | 626.1 ns |  2.70 ns |  2.52 ns | 0.0238 |     312 B |
| LoggerWithDirectLoggerAPI |                    True | 571.9 ns |  5.56 ns |  4.92 ns | 0.0334 |     448 B |
| LoggerWithSourceGenerator |                    True | 590.5 ns |  6.65 ns |  6.22 ns | 0.0219 |     296 B |
|        SerializeLogRecord |                    True | 344.9 ns |  2.85 ns |  2.52 ns | 0.0081 |     112 B |
|                    Export |                    True | 381.4 ns |  2.45 ns |  2.29 ns | 0.0081 |     112 B |
```

cijothomas:
> Fixes #.
> 
> ## Changes
> CustomFields are now supported at a table level
> 
> * `` : All fields are in separate columns
> * `*`: All fields are in separate columns
> * `A`: Fields with name `A` from all tables are in separate columns
> * `A/B`: Fields with name `B` in a table `A` are in separate column
> * `A/*`: All fields in a table `A` are in separate column.
> 
> For significant contributions please make sure you have completed the following items:
> 
> * [x]  Appropriate `CHANGELOG.md` updated for non-trivial changes
> * [ ]  Design discussion issue #

This does not seem very intuitive for users. Very hesitant to offer any advanced configuration unless this is a very common ask.  The benchmarks do show non-trivial performance impact (~2% for Export time. The 8B allocation is not clear as to what causes it..)

vidommet:
The 2% overhead is only when people use the features i added in this PR. If they are not using those features, even those lookups will not be made. This ask is from a bunch of services under a couple of teams that serve over 300 B requests in a day. We have to log every single request and therefore have to split the logs into multiple tables depending on the type of log and their required configurations.

PR #767, opened 2022-11-10
[Instrumentation.Runtime] Add exception type to exception counter's tags

### LINKED ISSUE #562: Add exception.type attribute to metric process.runtime.dotnet.exceptions.count
# Issue with OpenTelemetry.Instrumentation.Runtime

OpenTelemetry.Instrumentation.Runtime 1.0.0-rc.3

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can

`net6.0`

* [x] Feature Request
* [ ] Bug

At the moment `process.runtime.dotnet.exceptions.count `tracks the global amount of exceptions.
I think it would be beneficial to be able to track exception count per exception type.

I propose to add the attribute `exception.type` to the exception metric. Metrics conventions in the specification
don't contain exception related metrics yet. So I took the attribute from traces conventions [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md).

danspark:
Fixes #562.

## Changes

Adds the exception type to the `process.runtime.dotnet.exceptions.count` metric tags.

* [x] Design discussion issue #

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added the exception type to the `process.runtime.dotnet.exceptions.count` metric tags.
+
 ## 1.1.0-beta.1
 
 Released 2022-Nov-22
--- a/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Reflection;
 #if NET6_0_OR_GREATER
@@ -200,7 +201,12 @@ internal class RuntimeMetrics
 
         AppDomain.CurrentDomain.FirstChanceException += (source, e) =>
         {
-            exceptionCounter.Add(1);
+            var tags = new TagList
+            {
+                new("type", e.Exception.GetType().Name),
+            };
+
+            exceptionCounter.Add(1, tags);
         };
     }
 
--- a/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -32,7 +32,9 @@ namespace OpenTelemetry.Instrumentation.Runtime;
 internal class RuntimeMetrics
 {
     internal static readonly AssemblyName AssemblyName = typeof(RuntimeMetrics).Assembly.GetName();
-    internal static readonly Meter MeterInstance = new(AssemblyName.Name, AssemblyName.Version.ToString());
+    internal static readonly string MeterName = AssemblyName.Name;
+
+    internal readonly Meter MeterInstance = new(MeterName, AssemblyName.Version.ToString());
 
 #if NET6_0_OR_GREATER
     private const long NanosecondsPerTick = 100;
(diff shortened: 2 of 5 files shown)


src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -194,7 +195,12 @@ static RuntimeMetrics()
 
         AppDomain.CurrentDomain.FirstChanceException += (source, e) =>
         {
-            exceptionCounter.Add(1);
+            var tags = new TagList
+            {
+                new("type", e.Exception.GetType().Name),
+            };
+
+            exceptionCounter.Add(1, tags);
```
xiang17, line 203: The main scenario is to get a rough idea of the total number and see if it is abnormally high.

For the purpose of counting exception types, there is a concern on the number of types of exceptions in customer's code. If there are too many, it may impact the performance. One key design principle of this library is to minimize resource usage (memory consumption and pressure on garbage collection).

danspark: Would it be okay to make it possible to customize which tags could be recorded?

cijothomas: Separate from the above qn.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/test/Benchmarks/Metrics/MetricsBenchmarks.cs#L98 Use TagList only if you have more than 3 tags to pass. Else use the `Add(1, tag1)` overloads.

danspark: @cijothomas done in 9a1ff99. Since there was already a options type for runtime metrics I've added an option to record exception types, defaulting to `false`. Had to make a few changes to the `RuntimeMetrics` class, such as making its static constructor an instance constructor to allow creating the counter taking the options into account.

danspark: @cijothomas @xiang17 should I close this?

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added the exception type to the `process.runtime.dotnet.exceptions.count` metric tags.
```
Kielek, line 5: Please add link to this PR.

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
cijothomas: @dpk83 Whats your take on this? Should exception type be collected by default vs make it opt-in?

dpk83: > @dpk83 Whats your take on this? Should exception type be collected by default vs make it opt-in?

I think opt-in is fine here as this gives users a little more flexibility.

EugeneKrapivin:
Any chance to revive this PR? It seems like #562 is not closed yet.

Real-life example: I have processes spamming exceptions, which I know because I see this metric. However, it is really hard to understand what those exceptions are, and where are they originating from. It'd be really nice to have a source and type attribute. Their addition could be opt-in via configuration (as discussed before), or a documentation/readme section on how to filter those out.

I do not believe having those attributes as `opt-out` would generally be a bad idea.
Assuming there is a finite set of exceptions thrown from the dotnet framework, adding to that all the possible domain exceptions an author might add to their domain, I wouldn't expect a decent time series database (e.g. `prometheus` or `influxDB`) to buckle under the load. However, if those attributes are `opt-out` there should be documentation on how to do that.

I could add this logic on my own as part of my infrastructure setup, but it does seem like something that would greatly increase the usefulness of the metric.

PR #773, opened 2022-11-17
Enable public api analyzer by default

Kielek:
Fixes # N/A.

## Changes

Enable Public API Analyzers for all production projects by default.
Opt-out by `<EnablePublicApi>false</EnablePublicApi>` in particular csproj.
Shipped/Unshipped created based on released version (non-stable versions in unshipped).


Kielek:
I know that the PR is bigger than usual, but most of the changes are autogenerated. I would like to avoid artificial split on smaller parts.

Oberon00 (approved):
I only checked the AwsLambda changes: Those seem fine.

pellared (approved):

cijothomas (approved):

utpilla:
We don't have to enable publicAPI analyzer for `OpenTelemetry.Contrib.Shared`. That project is not shipped as its own package.

utpilla:
Does enabling public API analyzer automatically add those files? Or do we have to manually add the Shipped and Unshipped files to the project? If we have to manually add these files, then we should update the [Contributing a new project](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#contributing-a-new-project) section with this step.

Kielek:
> We don't have to enable publicAPI analyzer for `OpenTelemetry.Contrib.Shared`. That project is not shipped as its own package.

We can disable it, but IMO it is valuable as a guard to avoid introducing any public API in this project.

Kielek:
> Does enabling public API analyzer automatically add those files? Or do we have to manually add the Shipped and Unshipped files to the project? If we have to manually add these files, then we should update the [Contributing a new project](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#contributing-a-new-project) section with this step.

It is semi automatic by Visual Studio, I will check if it can be handed also by dotnet fix tool, Will be handled in separate PR if you are fine with this solution.

utpilla:
> > We don't have to enable publicAPI analyzer for `OpenTelemetry.Contrib.Shared`. That project is not shipped as its own package.
> 
> We can disable it, but IMO it is valuable as a guard to avoid introducing any public API in this project.

Okay. I don't have a strong opinion here. However, if a public API is added to the shared project, then it would show up under the public API analysis for the consuming project so we wouldn't easily miss it.

utpilla (approved):

PR #782, opened 2022-11-21
feat: Add version to AWS activity source

chertby:
## Changes

Added Version to AWS Activity source.


utpilla:
@chertby How are you planning to use version? Do you plan to use it from `Activity.Current`? OpenTelemetry SDK does not use ActivitySource version. It currently only supports using the name of the ActivitySource.

chertby:
@utpilla On our side, we added a custom processor that adds a tag with a link to the context for the specific library and its version. But I didn't know that OpenTelemetry SDK does not use ActivitySource version.

utpilla:
@srprash Please review this PR.

PR #797, opened 2022-11-24
Handle DateTimeOffset in the same way as DateTime

reyang:
Many .NET users are adopting DateTimeOffset, there seem to be a common expectation that DateTimeOffset should be handled in the same way as DateTime.

Additional context:
* https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset
* https://github.com/open-telemetry/opentelemetry-dotnet/blob/943c856c1c6295a536bc8a1c89896febc0be6107/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/TimestampHelpers.cs

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MessagePackSerializer.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MessagePackSerializer.cs
@@ -563,6 +563,8 @@ internal static class MessagePackSerializer
                 return SerializeArray(buffer, cursor, v);
             case DateTime v:
                 return SerializeUtcDateTime(buffer, cursor, v.ToUniversalTime());
+            case DateTimeOffset v:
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MessagePackSerializer.cs
```diff
@@ -563,6 +563,8 @@ public static int Serialize(byte[] buffer, int cursor, object obj)
                 return SerializeArray(buffer, cursor, v);
             case DateTime v:
                 return SerializeUtcDateTime(buffer, cursor, v.ToUniversalTime());
+            case DateTimeOffset v:
```
utpilla, line 566: Do we want to include `TimeSpan` as well?

reyang: I guess not now.

utpilla (approved):

Issue #763, opened 2022-11-08
Flaky Test `EventCountersMetricsTests.IncrementingEventCounter`


TimothyMothra:
This test failed on a netcoreapp3.1 run:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/3422750930/jobs/5700528653#step:6:263
```
[xUnit.net 00:00:05.06]         --- End of stack trace from previous location where exception was thrown ---
  Failed OpenTelemetry.Instrumentation.EventCounters.Tests.EventCountersMetricsTests.IncrementingEventCounter [1 s]
  Error Message:
   Assert.NotNull() Failure
  Stack Trace:
     at OpenTelemetry.Instrumentation.EventCounters.Tests.EventCountersMetricsTests.IncrementingEventCounter() in D:\a\opentelemetry-dotnet-contrib\opentelemetry-dotnet-contrib\test\OpenTelemetry.Instrumentation.EventCounters.Tests\EventCountersMetricsTests.cs:line 103
```

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/005fdb71b709d5eefbc8c5b7fa5a2c3f14fddecf/test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs#L93-L103

My best guess is that the metric hasn't been exported yet, causing the `Assert.NotNull` to fail.
This could be mitigated using a SpinWait instead of the Task.Delay.
For example: https://github.com/open-telemetry/opentelemetry-dotnet/blob/d9e08386bb8038ae65a12719d5dcb83fe3f31898/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs#L881-L893

PR #798, opened 2022-11-24
[EventCounters] Fix flaky tests

mic-max:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/763

## Changes

Use `SpinWait` instead of `Task.Delay` for a metric item to be exported.


test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs
```diff
@@ -17,43 +17,34 @@
...
-
-    [Fact(Skip = "Other tests metrics are being exported here")]
-    public async Task NoMetricsByDefault()
+    [Fact]
+    public void NoMetricsByDefault()
     {
-        // Arrange
         List<Metric> metricItems = new();
 
         var meterProvider = Sdk.CreateMeterProviderBuilder()
```
TimothyMothra, line 33: I think all the MeterProviders in this test class are missing the `using`.

PR #862, opened 2022-12-30
[EventCounters] SpinWait for tests

mic-max:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/763 and https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/910

## Changes

Use `SpinWait` instead of `Task.Delay` for a metric item to be exported.
The key was to spin until the metric item with the expected name was in the collection since the collection is being populated by all other instruments so simply checking for a count being greater than 1 would not work. Especially in the `Theory` test case.


test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs
```diff
@@ -267,15 +235,20 @@ private static double GetActualValue(Metric metric)
...
+        Metric metric = null;
+
+        SpinWait.SpinUntil(
             () =>
             {
                 Thread.Sleep(100);
                 meterProvider.ForceFlush();
-                return exportedItems.Count == 1;
+                metric = exportedItems.Where(x => x.Name == expectedInstrumentName).FirstOrDefault();
+                return metric != null;
```
TimothyMothra, line 248: nit: This can be simplified as `return exportedItems.Any(x => x.Name == expectedInstrumentName)`

test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs
```diff
@@ -43,10 +45,9 @@ public void EventCounter()
 
         // Act
         counter.WriteMetric(1997.0202);
-        AwaitExport(meterProvider, metricItems);
+        var metric = AwaitExport(meterProvider, metricItems, "ec.a.c");
```
TimothyMothra, line 48: nit: used a named parameter for the string const.
That will make this easier to read/understand.

test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs
```diff
@@ -243,7 +209,9 @@ public void InstrumentNameTooLong()
 
         // Act
         connections.Increment(1);
-        AwaitExport(meterProvider, metricItems);
+
+        Task.Delay(1800).Wait();
```
TimothyMothra, line 213: Why does this test have a Task.Delay, but the others have the SpinWait?


Also, not your code, but I don't understand what this test is doing.
name is "InstrumentNameTooLong", and it creates an eventName with 100 "e" characters.
But the asserts don't inspect that name nor that name's length.

I think this test needs to be rewritten.
If the Task.Delay fails and no metrics are exported, this test will pass because the foreach on an empty collection.

vishweshbankwar: @mic-max - Could you open an issue for this? And disable this test for now.

mic-max: These tests are checking that nothing gets exported for the event counters created in it. But I do agree it is a little weird especially now that others will use `SpinWait`.

It's checking the last part "it will be validated and ignored later in the pipeline" of `GetInstrumentName`

```csharp
    /// <summary>
    /// If the resulting instrument name is too long, it trims the event source name
    /// to fit in as many characters as possible keeping the event name intact.
    /// E.g. instrument for `Microsoft-AspNetCore-Server-Kestrel`, `tls-handshakes-per-second`
    /// would be too long (64 chars), so it's shortened to `ec.Microsoft-AspNetCore-Server-Kestre.tls-handshakes-per-second`.
    ///
    /// If there is no room for event source name, returns `ec.{event name}` and
    /// if it's still too long, it will be validated and ignored later in the pipeline.
    /// </summary>
    private static string GetInstrumentName(string sourceName, string eventName)
```

mic-max: Skipping unit test for now: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/936

vishweshbankwar (approved):

test/OpenTelemetry.Instrumentation.EventCounters.Tests/EventCountersMetricsTests.cs
```diff
@@ -64,11 +45,9 @@ public void EventCounter()
 
         // Act
         counter.WriteMetric(1997.0202);
-        Task.Delay(Delay).Wait();
-        meterProvider.ForceFlush();
+        var metric = AwaitExport(meterProvider, metricItems, expectedInstrumentName: "ec.a.c");
```
cijothomas, line 48: why no flush first before await?

cijothomas (approved):
approving to make progress. some follow ups expected to fully stabilize this.
Thanks for the efforts so far!

PR #864, opened 2023-01-03
Remove Nullable from NoWarn

Kielek:
Revert NoWarn introduced in #114 

## Changes

Remove `<NoWarn>$(NoWarn);nullable</NoWarn>` from old frameworks.
I do not see this kind of record in the main repository

Fix issues detected by the compiler.


src/OpenTelemetry.Exporter.Stackdriver/Implementation/ActivityExtensions.cs
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 // </copyright>
 
+#nullable enable
```
Kielek, line 17: This switch can be removed later.
IMO it is better to have some annotation instead of removing it now and reintroducing in next PRs.

mic-max (approved):

src/OpenTelemetry.Extensions/Trace/AutoFlushActivityProcessor.cs
```diff
@@ -87,7 +87,7 @@ public override void OnEnd(Activity data)
         var shouldFlush = this.RunPredicate(data);
         if (shouldFlush)
         {
-            this.tracerProvider.ForceFlush(this.timeoutMilliseconds);
+            this.tracerProvider?.ForceFlush(this.timeoutMilliseconds);
```
utpilla, line 90: The checks above should ensure that `tracerProvider` should not be null at this line.

Kielek: Code analyzer is not smart enough to detect this so I have decided to check it once more time in case of future code changes. Changed `?` to `!`.

Fixed in a445ce05cbd21efad58aed8498a9173d355acc92

Issue #478, opened 2022-06-30
WCF: Activity not stopped on EndpointNotFoundException (client-side)


vyrp:
# Issue with OpenTelemetry.Instrumentation.Wcf

**Environment (csproj):**

```csproj
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net472</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <Reference Include="System.ServiceModel" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="OpenTelemetry.Instrumentation.Wcf" Version="1.0.0-rc.6" />
  </ItemGroup>
</Project>
```

**Code:**

```cs
using System;
using System.Diagnostics;
using System.ServiceModel;

using OpenTelemetry;
using OpenTelemetry.Trace;
using OpenTelemetry.Instrumentation.Wcf;

[ServiceContract]
public interface ITestService
{
    [OperationContract]
    void Ping();
}

public static class Program
{
    public static void Main()
    {
        ActivitySource.AddActivityListener(new ActivityListener
        {
            ShouldListenTo = _ => true,
            Sample = (ref ActivityCreationOptions<ActivityContext> options) => ActivitySamplingResult.AllDataAndRecorded,
            ActivityStarted = _ => Console.WriteLine("Start"),
            ActivityStopped = _ => Console.WriteLine("Stop"),
        });

        using (Sdk.CreateTracerProviderBuilder().AddWcfInstrumentation().Build())
        {
            try
            {
                var factory = new ChannelFactory<ITestService>(
                    new BasicHttpBinding(),
                    new EndpointAddress("http://localhost/testservice"));
                factory.Endpoint.EndpointBehaviors.Add(new TelemetryEndpointBehavior());
                var channel = factory.CreateChannel();
                channel.Ping();
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.GetType().Name);
            }

            Console.WriteLine($"Activity.Current is null? {Activity.Current is null}");
        }
    }
}
```

**Is this a feature request or a bug?**

* [ ] Feature Request
* [x] Bug

**What is the expected behavior?**

When running the code above without any service listening on "http://localhost/testservice", I expected the output to be:
```
Start
Stop
EndpointNotFoundException
Activity.Current is null? True
```

**What is the actual behavior?**

Actual output:
```
Start
[...]

PR #870, opened 2023-01-05
Add known issues to wcf documentation

dszmigielski:
Documentation for #478

## Changes

Add known issues section in WCF documentation.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Wcf/README.md
+++ b/src/OpenTelemetry.Instrumentation.Wcf/README.md
@@ -228,7 +228,7 @@ an application code by catching the exception and stopping the `Activity` manual
     catch (Exception)
     {
         var activity = Activity.Current;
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Instrumentation.Wcf/README.md
```diff
@@ -212,6 +212,25 @@ contracts you want to instrument:
...
+
+```csharp
+    StatusResponse? response = null;
+    try
+    {
+        response = await client.PingAsync(statusRequest).ConfigureAwait(false);
+    }
+    catch (Exception)
+    {
+        Activity.Current?.Stop();
```
Kielek, line 230: Please consider something like: `Activity.Current != null and SourceName == "wcf"`. Just to be sure that we do not close other activity by accident.

Kielek (approved):

PR #874, opened 2023-01-08
[Exporter.Geneva] Add TLDLogExporter

utpilla:
## Changes
- Add TLDLogExporter 
- GenevaLogExporter to use TLDExporter when the user provides "UseTLD=true" in the connectionString

### TODOs:
- Use a single `ThreadLocal` instead of multiple
- Avoid potential unbounded memory usage for `List<KeyValuePair<string, object>> keyValuePairs`
- Add the support for prefix-based table name mapping added in #796

### Setup and usage:

```csharp
using var loggerFactory = LoggerFactory.Create(builder => builder
        .AddOpenTelemetry(loggerOptions =>
        {
            loggerOptions.IncludeScopes = true;
            loggerOptions.AddGenevaLogExporter(exporterOptions =>
            {
                exporterOptions.ConnectionString = "EtwSession=OpenTelemetry;UseTLD=true";
                exporterOptions.PrepopulatedFields = new Dictionary<string, object>
                {
                    ["cloud.role"] = "BusyWorker",
                    ["cloud.roleInstance"] = "CY1SCH030021417",
                    ["cloud.roleVer"] = "9.0.15289.2",
                };
            });
        }));

        var logger = loggerFactory.CreateLogger<Program>();

        logger.LogInformation("Hello from {food} {price}.", "artichoke", 3.99);
```

You could us any ETW collection tools such as WPA, PerfView etc. to view the data:

Example with PerfView:
For the connectionString = `"EtwSession=OpenTelemetry;UseTLD=true"`, you can run the following command in Command Prompt: 
```cmd
PerfView.exe collect -MaxCollectSec:3600 -NoGui /onlyProviders=*OpenTelemetry
```


reyang (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
```diff
@@ -27,6 +27,7 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using OpenTelemetry.Logs;
+using OpenTelemetry.Trace;
```
reyang, line 30: Why is this (trace) needed?

utpilla: It's not needed. I have removed it.

PR #876, opened 2023-01-09
[Exporter.Geneva] Use default instead of string.Empty as format for TryFormat method

utpilla:
## Changes
- Use `default` instead of `string.Empty` as format for `TryFormat` method


mic-max (approved):
just curious, default(String) is null right, so putting null there would act the same?

PR #879, opened 2023-01-10
[Exporter.Geneva] Add the missing changelog entry

reyang:
There was a hotfix released and it didn't have a corresponding entry in the changelog, this PR is trying to fix it.


utpilla (approved):

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -28,6 +28,13 @@ Released 2022-Dec-09
 * Fix EventSource logging.
   ([#813](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/813))
 
+## 1.3.1
+
+Released 2022-Dec-07
+
+* Fix the overflow bucket value serialization for Histogram.
```
utpilla, line 35: We could probably add a statement saying, these changes are on top 1.3.0 and do not include the changes from any of the 1.4.0-beta releases.

cijothomas: Another option to consider:
Lets put 1.3.1 right above 1.3.0 release and include the changelog items there.

On the chronological ordered entry (i.e the one being added in this PR), we can simply link to the above, instead of listing the changelog items here.

That way, users will find the 1.3.1 (or any hotfix) by looking chronologically or by looking at version order. and wont miss anything.

reyang: I guess that would introduce a problem - users would think `## 1.4.0-beta.1` has the changes introduced in `1.3.1`.

reyang: My current thinking is to keep it as-is, the common expectation should be "1.3.1 doesn't have any fix in 1.4.x.y". (for example, a Windows XP hotfix shouldn't be taking changes from Windows 8 preview, unless it is explicitly called out)

cijothomas (approved):

PR #900, opened 2023-01-17
Use Array.Empty<T>()

martincostello:
## Changes

Use `Array.Empty<string>()` instead of allocating a new array.

* [ ] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #

Issue #894, opened 2023-01-12
Update all projects with `Nullable`


Kielek:
# Issue that does not fit into other categories

This issue is to track the effort of updating all projects to support Nullable and remove:

```xml
  <Nullable>disable</Nullable>
```

If possible enable projects one by one with corresponding Tests project to avoid huge PRs.

Ref: https://github.com/open-telemetry/opentelemetry-dotnet/issues/3958

FernandoNunes:
Hi @Kielek ,

I've started to pick this up. First PR is here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/920

Kielek:
@FernandoNunes, it is the goal, but I would expect more changes related to the code adjustment. See as an example: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/892/files.

Note: the tricky part could be OpenTelemetry.Contrib.Shared project which is compiled-in for multiple projects. What is more it is copied from main/SDK repository. It will be great to enable nullability there and then copy it here.

Kielek:
@zivaninstana, would you like to contribute to this issue? Exporter.Instana still does not implement this feature.

zivaninstana:
Yes sure, I will take a look

Kielek:
Assigning @CodeBlanch as he is working on last part - Geneva exporter.

Kielek:
Closing. See #1349

PR #908, opened 2023-01-19
Enable analysers

martincostello:
## Changes

Enable `<AnalysisLevel>latest-All</AnalysisLevel>` as suggested in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/900#issuecomment-1387222338.

Within the production shipping packages there are two changes that change the public API surface. As both packages are still pre-release on NuGet.org I assumed this was fine. These changes can be reverted and suppressed if preferred.

For the test projects I erred on the side of disabling rules rather than fixing them to reduce churn.

I'm leaving this as a draft for now as it touches a lot of files so might ping a lot of different reviewers. I can also split the commits up differently and/or create smaller more targeted PRs if that's preferred.

Let me know if any of these changes warrant a note in the CHANGELOG.

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -7,6 +7,7 @@
     <DefineConstants>$(DefineConstants);SIGNED</DefineConstants>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <NetFrameworkMinimumSupportedVersion>net462</NetFrameworkMinimumSupportedVersion>
+    <AnalysisLevel>latest-All</AnalysisLevel>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)'=='Debug'">
--- a/src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/AWSEKSResourceDetector.cs
+++ b/src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/AWSEKSResourceDetector.cs
@@ -41,19 +41,19 @@ public class AWSEKSResourceDetector : IResourceDetector
     public IEnumerable<KeyValuePair<string, object>> Detect()
     {
         var credentials = this.GetEKSCredentials(AWSEKSCredentialPath);
-        var httpClientHandler = Handler.Create(AWSEKSCertificatePath);
+        using var httpClientHandler = Handler.Create(AWSEKSCertificatePath);
 
         if (credentials == null || !this.IsEKSProcess(credentials, httpClientHandler))
         {
             return null;
         }
 
-        return this.ExtractResourceAttributes(
+        return ExtractResourceAttributes(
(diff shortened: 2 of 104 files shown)


build/Common.props
```diff
@@ -7,6 +7,7 @@
     <DefineConstants>$(DefineConstants);SIGNED</DefineConstants>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <NetFrameworkMinimumSupportedVersion>net462</NetFrameworkMinimumSupportedVersion>
+    <AnalysisLevel>latest-All</AnalysisLevel>
```
Oberon00, line 10: latest-All sound a bit too much, can we use a fixed version here?

src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/AWSEKSResourceDetector.cs
```diff
@@ -41,19 +41,19 @@ public class AWSEKSResourceDetector : IResourceDetector
     public IEnumerable<KeyValuePair<string, object>> Detect()
     {
         var credentials = this.GetEKSCredentials(AWSEKSCredentialPath);
-        var httpClientHandler = Handler.Create(AWSEKSCertificatePath);
+        using var httpClientHandler = Handler.Create(AWSEKSCertificatePath);
```
Oberon00, line 44: Nice catch

Oberon00:
All in all, this PR seems a bit too large. The nontrivial changes should probably go to separate PRs (separately dis/enabled before)

build/Common.props
Kielek: There is an agreement to use this approach on SDK repository. I think we could do the same here. See https://github.com/open-telemetry/opentelemetry-dotnet/issues/3958

Kielek:
> All in all, this PR seems a bit too large. The nontrivial changes should probably go to separate PRs (separately dis/enabled before)

I think that enabling this feature project by project (including tests) could be easier to review. If you agree, it will be great to create issue to link all changes, similar to #894.

@TimothyMothra, fyi as you working on similar issue in SDK repository.

Oberon00:
I was thinking of doing the trivial warnings (such as adding static) in one PR, and for the more involved changes, doing it per warning, or per project, however it clusters better.

build/Common.props
Oberon00: Makes sense.

martincostello:
As noted in the description, I'm happy to chunk this up into N commits/PRs (whatever works best for you), but doing _all_ the changes and pushing up a draft as a starting point was the easiest way for me to do the majority of the work in the first instance. The easiest way to fix with it turned on was per-file, per-project.

martincostello:
Have opened #4144 to track doing these changes in a series of separate pull requests.

PR #909, opened 2023-01-19
Do not assign people to draft pull requests

martincostello:
## Changes

I opened #908 as a draft to prevent spamming people, but as a custom workflow is used instead of `CODEOWNERS` it didn't have the desired effect.

This PR changes the workflow to only assign when the pull request is not a draft.

If the current behaviour is intentional, feel free to close this pull request.


cijothomas:
Force merge as the CI failure is unrelated.

Issue #934, opened 2023-01-27
Unable to make socket connection at startup causing application container to restart


arunmk-ms:
# Issue with OpenTelemetry.Exporter.Geneva
Application containers are restarting continuously whenever there is a delay in mounting the UDS mount points at application startup. I have made a PR with the changes. Please check the same and let me know if these can be merged in. 

List of [all OpenTelemetry NuGet packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are using (e.g. `OpenTelemetry 1.0.2`):

```
    <PackageReference Include="OpenTelemetry.Exporter.Geneva" Version="1.3.1" />
    <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="1.3.2" />
```

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can find this information from the `*.csproj` file):
net7.0

**Is this a feature request or a bug?**

* [ ] Feature Request
* [X] Bug

**What is the expected behavior?**
The application should startup even if the mount point is not available. 

What do you expect to see?
The application should startup without a socket connection to the mount point, reconnect later when the mount point becomes available.

**What is the actual behavior?**
The application fails fast when the uds socket connection could not be made and the containers keep on restarting.

## Additional Context

Add any other context about the feature request here.
Our proposal is to create only the socket in ctor and make the socket connection later while trying to send some data in Send() method.

UnixDomainSocketDataTransport.cs
```
    public UnixDomainSocketDataTransport(
        string unixDomainSocketPath,
        int timeoutMilliseconds = DefaultTimeoutMilliseconds)
    {
        this.unixEndpoint = new UnixDomainSocketEndPoint(unixDomainSocketPath);
        this.timeoutMilliseconds = timeoutMilliseconds;
        this.CreateSocket();
    }

    private void Reconnect()
    {
        this.socket.Close();
        try
        {
            this.CreateSocket();
            this.socket.Connect(this.unixEndpoint);
        }
        catch (Exception ex)
        {
            ExporterEventSource.Log.ExporterException("UDS Connect failed.", ex);

            // Re-throw the exception to fail in the Reconnect attempt.
            throw;
        }
    }

    private void CreateSocket()
    {
        this.socket = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.IP)
        {
            SendTimeout = this.timeoutMilliseconds,
        };
    }
```

PR #935, opened 2023-01-27
Handle exception while opening connection from constructor

arunmk-ms:
Fixes #.
[Issue 934](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/934)

## Changes
Add try/catch to the socket connection in constructor so that the application wont fail fast when the UDS is not available at startup.


reyang (approved):
LGTM with a changelog entry and clearing the CLA.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
   environment variables) used by GenevaTraceExporter in Linux.
   ([#925](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/925))
 
+* Fix socket connection issue at startup when UDS mount point is not available
```
reyang, line 18: I would suggest that we capture the expectation from the users' perspective. Maybe something like (to be frank I haven't spent much time polishing the message):

"Changed the behavior of during initialization on Linux. Before this change, the exporter initialization would throw exception if the target Unix Domain Socket does not exist; after this change, the exporter initialization would return success and the exporting background thread will try to establish the connection."

arunmk-ms: Updated change log as per your suggestion. Thank you!

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/Transport/UnixDomainSocketDataTransport.cs
```diff
@@ -43,7 +43,7 @@ internal class UnixDomainSocketDataTransport : IDataTransport, IDisposable
     {
         this.unixEndpoint = new UnixDomainSocketEndPoint(unixDomainSocketPath);
         this.timeoutMilliseconds = timeoutMilliseconds;
-        this.Connect();
+        this.CreateSocket();
```
cijothomas, line 46: we can keep the same code as before, with just one change: wrap this line inside try..catch, and log the fact that UDS was unavailable at startup.

cijothomas:
Thanks!

Given we are in RC state right now, we might hold off merging until 1.4.0 stable is out, so we'll get some time to test this.

wsugarman:
I am concerned a bit that this PR will further reduce any "fail-fast" behavior that may save telemetry from being dropped. If my understanding of the code is accurate, today if there are any errors when sending telemetry batches to the socket, then those batches will simply be dropped. However today, at the very least, the constructor will throw exceptions and fail the application if the socket is misconfigured or not ready.

I can create a separate issue for it, but I'd love to see more resiliency in the exporter with regards to transient failures -- perhaps retries or even some preservation of unsent batches while managing a reasonable amount of backpressure.

cijothomas:
> I am concerned a bit that this PR will further reduce any "fail-fast" behavior that may save telemetry from being dropped. If my understanding of the code is accurate, today if there are any errors when sending telemetry batches to the socket, then those batches will simply be dropped. However today, at the very least, the constructor will throw exceptions and fail the application if the socket is misconfigured or not ready.
> 
> I can create a separate issue for it, but I'd love to see more resiliency in the exporter with regards to transient failures -- perhaps retries or even some preservation of unsent batches while managing a reasonable amount of backpressure.

You concern is valid, and we did consider it before deciding to take this in.
1. In windows also, where we use ETW, the telemetry is simply lost if agent is not ready. This PR makes Linux behavior somewhat same - if agent is not available, telemetry is simply lost. 
2. This is consistent with OTLP Exporters - they dont fail at startup.

Agree about the data loss issues/retry etc - We are adding it to OTLP Exporters - https://github.com/open-telemetry/opentelemetry-dotnet/issues/4115 and could likely re-use idea from it for the Linux side.

(Long term, Linux would move away from UDS towards user_events (https://docs.kernel.org/trace/user_events.html) in Linux, once its get ready in linux distributions), so it'll be consistent with Windows (no batching etc!)

dustinburson:
> 

Any consideration to allowing this behavior to be configurable?  Also is there any recommendation on how to monitor this in the process?   For example, the code in question has this line ```ExporterEventSource.Log.ExporterException("UDS unavailable at startup.", ex);```.  I assume it is possible to look for these events and respond?  If so is the a similar mechanism to detect if the UDS is becomes available?  I am thinking of a scenario where we might be okay with this for a limited time but if we detect failures for an extended period (say 5 minutes) we might want to take some action to avoid a process running for an extended prior of time without telemetry.

cijothomas:
> > 
> 
> Any consideration to allowing this behavior to be configurable? Also is there any recommendation on how to monitor this in the process? For example, the code in question has this line `ExporterEventSource.Log.ExporterException("UDS unavailable at startup.", ex);`. I assume it is possible to look for these events and respond? If so is the a similar mechanism to detect if the UDS is becomes available? I am thinking of a scenario where we might be okay with this for a limited time but if we detect failures for an extended period (say 5 minutes) we might want to take some action to avoid a process running for an extended prior of time without telemetry.


_`ExporterEventSource.Log.ExporterException("UDS unavailable at startup.", ex);`. I assume it is possible to look for these events and respond?_

https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry#self-diagnostics is one option. Or any other tools like perfview, logman, or own listener. (We do not support routing this to iLogger, but thats something which might be added in the future)

_If so is the a similar mechanism to detect if the UDS is becomes available?_

The exporters keeps attempting for each Export, so if UDS is unavailable for some time, then becomes available, those telemetry from that down-window is lost, the rest will continue to flow.

_we might want to take some action to avoid a process running for an extended prior of time without telemetry._
The best option right now is to write own listener to the Exporter eventsource, and react to it (like sending some metrics saying UDS unavailable...). 

_Any consideration to allowing this behavior to be configurable?_
As mentioned in the previous comment, the goal for GenevaExporter long term is to only have native-to-OS transport. i.e ETW/TLD in windows and user_events in Linux. We are hesistant to keep adding more configs/switches which are not aligned with the long term direction.

reyang:
> Any consideration to allowing this behavior to be configurable? Also is there any recommendation on how to monitor this in the process? For example, the code in question has this line `ExporterEventSource.Log.ExporterException("UDS unavailable at startup.", ex);`. I assume it is possible to look for these events and respond? If so is the a similar mechanism to detect if the UDS is becomes available? I am thinking of a scenario where we might be okay with this for a limited time but if we detect failures for an extended period (say 5 minutes) we might want to take some action to avoid a process running for an extended prior of time without telemetry.

I suggest us solving this problem outside the SDK/Exporter. For example, using ETW there is no easy for the SDK to know whether the data is received by the downstream or not (e.g. ETW does not tell the sender if the data is dropped due to buffer full), and such design comes with good reason.

I think something like heartbeat/QoS is a better direction.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -33,6 +33,13 @@ Released 2023-Jan-30
   environment variables) used by GenevaTraceExporter in Linux.
   ([#925](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/925))
 
+* Changed the behavior of Unix domain socket connection at startup. Before this
```
Kielek, line 36: After merges with main it was moved to relesed section,. It should be under unreleased.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -359,7 +359,7 @@ public void GenevaTraceExporter_Constructor_Missing_Agent_Linux()
                         options.ConnectionString = "Endpoint=unix:" + path;
                     })
                     .Build();
-                Assert.True(false, "Should never reach here. GenevaTraceExporter should fail in constructor.");
+                Assert.True(true, "GenevaTraceExporter should not fail in constructor.");
```
utpilla, line 362: Could you add a comment here saying "GenevaExporter would not throw if it was not able to connect to the UDS socket in ctor. It would keep attempting to connect to the socket when sending telemetry."

utpilla (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -359,7 +359,9 @@ public void GenevaTraceExporter_Constructor_Missing_Agent_Linux()
                         options.ConnectionString = "Endpoint=unix:" + path;
                     })
                     .Build();
-                Assert.True(false, "Should never reach here. GenevaTraceExporter should fail in constructor.");
+                // GenevaExporter would not throw if it was not able to connect to the UDS socket in ctor. It would
```
cijothomas, line 362:
```suggestion

                // GenevaExporter would not throw if it was not able to connect to the UDS socket in ctor. It would
```

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Changed the behavior of Unix domain socket connection at startup. Before this
+  change, the exporter initialization would throw exception if the target Unix
+  Domain Socket does not exist. After this change, the exporter initialization
+  would return success and the exporting background thread will try to
+  establish the connection.
+  ([#935](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/935))
+
```
cijothomas, line 11:
```suggestion

```

utpilla:
@arunmk-ms We released the stable 1.4.0 version of GenevaExporter. Could you please resolve the merge conflicts, and we can get this PR merged?

arunmk-ms:
@utpilla done

utpilla:
Thanks for your contribution @arunmk-ms !

PR #939, opened 2023-01-31
Add support for TCP and UDP sockets to Geneva exporters

### LINKED ISSUE #937: Enable TCP and UDP transports for Geneva exporter

Only ETW and Unix sockets are supported as transport for Geneva exporters. Our team has a scenario which is better solved by sending data over TCP or UDP.

* [x] Feature Request
* [ ] Bug

Exporters are able to communicate with agents over TCP or UDP.

This is not possible today.

wasker:
Fixes #937.

## Changes

Added TCP and UDP transport support for Geneva exporters. Consumers can use `tcp://host:port` or `udp://host:port` schemes in the connection string.


cijothomas (changes requested):
Hi.
We do not plan to add more transport mechanisms to this Exporter. The long term goal is to use native-to-os technologies only - i.e Windows will be ETW/TLD and Linux will be user_events (https://docs.kernel.org/trace/user_events.html)

For TCP, the general guidance is to use OTLP Exporter.

wasker:
@cijothomas I've sent an invite offline to have a quick sync on that. That'd be great if we could find a middle ground here.

cijothomas:
Closing as this is not in line with the long term plan for this exporter, and OTLP is ready to be used!

PR #951, opened 2023-02-03
[Extensions.AWSXRay] Fix analysis warnings

martincostello:
## Changes

Fix/suppress code analysis warnings in OpenTelemetry.Contrib.Extensions.AWSXRay.

Contributes to #950.


cijothomas:
@srprash Could you review?

utpilla:
@srprash @lupengamzn Could you please review this PR?

martincostello:
Can someone else review this PR so it can be merged? I don't think the changes made are particularly contentious or specific to the provider's specific purpose (i.e. they don't need AWS knowledge).

src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/AWSEKSResourceDetector.cs
```diff
@@ -122,17 +127,17 @@ public class AWSEKSResourceDetector : IResourceDetector
...
     }
 
-    internal AWSEKSClusterInformationModel? DeserializeResponse(string response)
+    private static string GetEKSClusterInfo(string credentials, HttpClientHandler? httpClientHandler)
     {
-        return ResourceDetectorUtils.DeserializeFromString<AWSEKSClusterInformationModel>(response);
+        return ResourceDetectorUtils.SendOutRequest(AWSClusterInfoUrl, "GET", new KeyValuePair<string, string>("Authorization", credentials), httpClientHandler).Result;
     }
 
     private string? GetEKSClusterName(string credentials, HttpClientHandler? httpClientHandler)
```
utpilla, line 135: Update this method to static as well?

utpilla: Same suggestion for the method `IsEKSProcess`.

src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/ResourceDetectorUtils.cs
```diff
@@ -37,11 +39,13 @@ internal static async Task<string> SendOutRequest(string url, string method, Key
             httpRequestMessage.Method = new HttpMethod(method);
             httpRequestMessage.Headers.Add(header.Key, header.Value);
 
+#pragma warning disable CA2000 // Dispose objects before losing scope
             var httpClient = handler == null ? new HttpClient() : new HttpClient(handler);
```
utpilla, line 43: Wouldn't this be better here?

```suggestion
           using var httpClient = handler == null ? new HttpClient() : new HttpClient(handler);
```

src/OpenTelemetry.Contrib.Extensions.AWSXRay/Trace/AWSXRayPropagator.cs
```diff
@@ -303,7 +303,13 @@ internal static bool TryParseSampleDecision(ReadOnlySpan<char> sampleDecision, o
     internal static string ToXRayTraceIdFormat(string traceId)
     {
         var sb = new StringBuilder();
+#if NETFRAMEWORK
+#pragma warning disable CA1834
```
utpilla, line 307: I think it's better to simply add the version here as a string instead of char even for non-netframework targets. Currently, the version is just one character, it might be updated to something like `1.1` and we would have to update the code here again to make sure that we append the full version. 

Or we could have a check on the version length, if we want to use the optimized version of stringBuilder.Append for a char.

utpilla (commented):
Left a suggestion.

src/OpenTelemetry.Contrib.Extensions.AWSXRay/Trace/AWSXRayPropagator.cs
martincostello: The length check doesn't work because the analyser knows it's a constant so still emits a warning.

I've refactored the code to use a `char` for the constant, and then if it needs to change to something like `1.1` then it'll need to change to a `string` with appropriate code changes to keep the analyser happy then.

utpilla (approved):

PR #959, opened 2023-02-03
[Instrumentation.AWSLambda] Fix analysis warnings

martincostello:
## Changes

Fix/suppress code analysis warnings in OpenTelemetry.Contrib.Instrumentation.AWSLambda.

Contributes to #950.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
@@ -189,10 +189,7 @@ public static class AWSLambdaWrapper
 
     private static void OnFunctionStop(Activity activity, TracerProvider tracerProvider)
     {
-        if (activity != null)
-        {
-            activity.Stop();
-        }
+        activity?.Stop();
 
         // force flush before function quit in case of Lambda freeze.
         tracerProvider?.ForceFlush();
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
@@ -128,6 +128,8 @@ public static class AWSLambdaWrapper
         ILambdaContext context,
         ActivityContext parentContext = default)
     {
+        Guard.ThrowIfNull(lambdaHandler);
+
         async Task<object> Handler(TInput input, ILambdaContext context)
         {
             await lambdaHandler(input, context).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -69,6 +70,7 @@ public static class AWSLambdaWrapper
         ILambdaContext context,
         ActivityContext parentContext = default)
     {
+        Guard.ThrowIfNull(lambdaHandler);
```
Oberon00, line 73: Check is missing from the overload below.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -125,7 +127,7 @@ public static class AWSLambdaWrapper
     {
         Func<TInput, ILambdaContext, Task<object>> func = async (input, context) =>
```
Oberon00, line 128: We should also guard against null here.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -93,12 +95,15 @@ public static class AWSLambdaWrapper
         ILambdaContext context,
         ActivityContext parentContext = default)
     {
-        Func<TInput, ILambdaContext, object> func = (input, context) =>
+        Guard.ThrowIfNull(lambdaHandler);
+
+        object Handler(TInput input, ILambdaContext context)
```
Oberon00, line 100: Any particular reason for changing it to a local function?

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -123,12 +128,13 @@ public static class AWSLambdaWrapper
         ILambdaContext context,
         ActivityContext parentContext = default)
     {
-        Func<TInput, ILambdaContext, Task<object>> func = async (input, context) =>
+        async Task<object> Handler(TInput input, ILambdaContext context)
```
Oberon00, line 131: It looks like we are still missing a null-check for lambdaHandler here.

Oberon00 (approved):

Oberon00:
Thank you, LGTM!

Oberon00:
Just one thing: You intentionally didn't enable analysis per-project but want to do so in a follow-up PR for all projects?

Kielek (approved):

martincostello:
> Just one final question: You intentionally didn't enable analysis per-project but want to do so in a follow-up PR for all projects?

That's correct - I originally had one big PR, but I've broken it up into chunks. #969 will turn it on once all the projects are fixed-up.

Issue #791, opened 2022-11-23
remove NetCoreApp3.1 from CI


TimothyMothra:
Follow up from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/759#discussion_r1026765172

## .NET Core 3.1 is officially scheduled to reach end of life on December 3, 2022

- Official Statement: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

Kielek:
Effort handled in scope of #1031

PR #974, opened 2023-02-06
Remove netcore3.1 from CI jobs

cijothomas:
Towards https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/791 but only removing from CI. (to make CIs run a bit faster).

Actual removal from projects is to be separately done.

### DIFF
--- a/.github/workflows/ci-md.yml
+++ b/.github/workflows/ci-md.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ windows-latest, ubuntu-latest ]
-        version: [ net462, netcoreapp3.1, net6.0, net7.0 ]
+        version: [ net462, net6.0, net7.0 ]
         exclude:
         - os: ubuntu-latest
           version: net462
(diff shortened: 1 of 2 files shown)


CodeBlanch (changes requested):
I think this needs some SIG discussion. If we're going to allow code from this repo to stay on older runtimes then we should have the CI in place. To say that another way, removing the CI without first removing the targets feels backwards.

cijothomas:
> I think this needs some SIG discussion. If we're going to allow code from this repo to stay on older runtimes then we should have the CI in place. To say that another way, removing the CI without first removing the targets feels backwards.

Agree to discuss this before removal. Yes, some components still target older runtimes, so its best to 1st remove there.

cijothomas:
Discussed in SIG meeting on 02/07/2023
Will close this PR.
Will instead open an issue to collect feedback about requiring projects to only target supported .NET Versions.
Once the above is agreed, start by removing the unsupported targets from the projects itself,
Remove from CI as last step.

PR #979, opened 2023-02-08
[Instana Exporter] Wrong JSON format for serialized spans

zivaninstana:
Fixes #.

## Changes

Instana backend reported generated JSON by InstanaExporter as wrong. 
JSON should contains spans node as a root but it was not a case.

### DIFF
--- a/src/OpenTelemetry.Exporter.Instana/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Instana/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-* Wrong JSON for serialized spans was generated after last changes fixed
+* Fixes issue in span serialization process introduced in 1.0.2 version.
+  ([#979](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/979))
 * Update OTel SDK version to `1.3.2`.
   ([#917](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/917))
 
--- a/src/OpenTelemetry.Exporter.Instana/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Instana/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Wrong JSON for serialized spans was generated after last changes fixed
 * Update OTel SDK version to `1.3.2`.
   ([#917](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/917))
 
--- a/src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
+++ b/src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
@@ -26,8 +26,9 @@ namespace OpenTelemetry.Exporter.Instana.Implementation;
 
 internal class Transport
 {
-    private const int MultiSpanBufferLimit = 4096000;
-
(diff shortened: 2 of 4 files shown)


src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
```diff
@@ -26,8 +26,9 @@ namespace OpenTelemetry.Exporter.Instana.Implementation;
 
 internal class Transport
 {
-    private const int MultiSpanBufferLimit = 4096000;
-
+    private const int MULTI_SPAN_BUFFER_SIZE = 4096000;
```
Kielek, line 29: Could you please explain why this difference between buffer size and used limit?

src/OpenTelemetry.Exporter.Instana/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Wrong JSON for serialized spans was generated after last changes fixed
```
Kielek, line 5:
```suggestion
* Fixes issue in span serialization process introduced in 1.0.2 version.
  ([#979](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/979))
```

src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
 	<PackageTags>Instana Tracing APM</PackageTags>
+	  <LangVersion>latest</LangVersion>
```
Kielek, line 6: Common lang version is defined under https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/9877a3220a3977dce2fb29d7dc2742dee553a0a8/build/Common.props#L3

I think, it will be fine to create separate PR with bumping it to 11.0.
Please revert change here.

src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
```diff
@@ -26,8 +26,9 @@ namespace OpenTelemetry.Exporter.Instana.Implementation;
...
+    private const int MULTI_SPAN_BUFFER_LIMIT = 4070000;
+    private readonly byte[] tracesBuffer = new byte[MULTI_SPAN_BUFFER_SIZE];
```
Kielek, line 31: Please fix compilation errors.
(It should be easily reproducible locally if you switch Debug to Release).

src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
zivaninstana: Buffer size is a size of byte array but we should never reach that size, we should stop before, because of that we have limit config.

src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj
zivaninstana: Ok, I will

src/OpenTelemetry.Exporter.Instana/Implementation/Transport.cs
zivaninstana: 👍

Kielek (approved):

PR #986, opened 2023-02-09
Create GitHub releases

Kielek:
Fixes N/A

## Changes

Extend releasing process.
Automate creating GitHub releases for each package.
Current, tags-only approach does not create GitHub notifications about new releases. 
On the GitHub you can subscribe for the new releases
image

Tested on my fork with disabled nuget.org step:
image

### DIFF
--- a/build/RELEASING.md
+++ b/build/RELEASING.md
@@ -34,7 +34,7 @@ on the `main` branch is the one which added/updated the Changelog to
 the project being released. *This latest commit will be tagged on the release.*
 
 1. Create and push git tag for the project and the version of the project
-you want to release. The version shoud the one used in the **Pre-steps** to
+you want to release. The version should be the one used in the **Pre-steps** to
 update the Changelog.
 
     ```powershell
@@ -74,9 +74,4 @@ release page.
 3. Validate that the new version (as specified in step 1) of the project is
 successfully published to nuget.org under OpenTelemetry owner.
 
-4. Publish a release in GitHub:
-
-   - Use draft created by "Pack YOUR_PROJECT_NAME" GitHub workflow.
-   - Use the appropriate CHANGELOG.md content in the description.
-   - Set as a pre-release if it not stable release.
-   - Set as the latest release if you releasing stable version.
+4. Validate that the new version was published in GitHub.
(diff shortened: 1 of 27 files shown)


build/RELEASING.md
```diff
@@ -68,7 +68,15 @@ update the Changelog.
...
 tab on the repo and monitor the "Pack YOUR_PROJECT_NAME" workflow. The last step
-in the workflow is to publish to Nuget.org.
+in the workflow is to publish to Nuget.org and prepare corresponding github
+release page.
 
-3. Validate that the new version (as specified in step1) of the project is
+3. Validate that the new version (as specified in step 1) of the project is
 successfully published to nuget.org under OpenTelemetry owner.
+
+4. Publish a release in GitHub:
```
cijothomas, line 77: This is additional burden on maintainers and adds time needed for each release. (currently its as simpl as pushing the right tag only.)

Could you remove the draft one, so the release will be published? The contents can point to the changelog.md file, so there is no manual work needed?

TimothyMothra: The "draft' provides some comfort when onboarding to new automation. I took this same approach on another repo, but eventually removed it because of the extra manual step.

@Kielek, it sounds like you already verified the automation in your fork. 
I agree with @cijothomas here, we should aim to do this without "draft".

Kielek: Done in 9865ca16a6eacd3c63037a0ecbf173998d9982a2 and 829fdfd3d2c1bb28072ea6cc3b1a0257da7c30ad.

Results from the simplified workflow you can find under https://github.com/Kielek/opentelemetry-dotnet-contrib/releases
image

Unfortunately I do not found any better way to avoid duplication of code for usntable/stable releases.

pellared (approved):

cijothomas (approved):
Thanks @Kielek .
This looks very good. Will try this for the next release request and we can improve as needed.
(Still hoping to automate the main repo as well someday in the future!)

build/RELEASING.md
cijothomas: > Unfortunately I do not found any better way to avoid duplication of code for usntable/stable releases

Okay with the current one. we can try this and improve, as needed.

PR #993, opened 2023-02-13
[Instrumentation.Process, Instrumentation.Runtime] Release wtih OTel 1.4.0-rc.4

Kielek:
Fixes N/A

## Changes

Release request
[Instrumentation.Process, Instrumentation.Runtime] Release wtih OTel 1.4.0-rc.4

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.1.0-beta.5
+
+Released 2023-Feb-13
+
 * Update OpenTelemetry API to 1.4.0-rc.4
   ([#990](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/990))
 
--- a/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-## 1.1.0-beta.5
+## 1.1.0-rc.1
 
 Released 2023-Feb-13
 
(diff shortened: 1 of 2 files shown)


Kielek:
@xiang17, fyi

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.1.0-beta.5
```
cijothomas, line 5: @xiang17 Do you think we should no be doing an RC, and we do not anticipate any further changes to this package?

Kielek: I would consider rc with stable OTel.

Yun-Ting (approved):

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
cijothomas: why not do stable of this package, when OTel 1.4 is stable as well ?

Kielek: I am not against stable with stable OTel.
I do not like idea where RC version can be compiled only with exact package version, not even the range.

cijothomas: it should be range, given OTel .NET said its the last rc, so we should leave upper bound open.

Kielek: I am not complaining, but the same case was for rc.3.

If possible please merge #994 then this PR.

cijothomas (approved):

cijothomas (approved):

src/OpenTelemetry.Instrumentation.Runtime/CHANGELOG.md
xiang17: I think doing RC now is a good idea. We should release the RC before we release the stable anyways.

I agree with Piotr. The exact package version is annoying, especially it causes trouble when downstream wants to test against version upgrades.

PR #1002, opened 2023-02-14
Updated projects to build better with VS Code / OmniSharp on Mac

tillig:
## Changes

No actual _functional_ changes here, just changes to make VS Code / OmniSharp work a little better, particularly on MacOS.

- Update target framework monikers in projects to be descending - .NET Core, .NET Standard, .NET Desktop. OmniSharp uses the first one in the list and falls back. Ordering this way ensures the right analyzers, etc. will be in place. Includes standard comment to indicate the ordering is intentional.
- Fix IDE0001 warnings - The `.editorconfig` currently has these as warnings, so there were a few warnings showing up. Removing these warnings by simplifying names (basically removing unnecessary namespace prefixes) makes it easier to see what's actually going on.
- Spelling/dictionary updates - The `cSpell` dictionary got a couple additions so less is flagged as misspelled.
- .NET desktop framework tests (`net472`) only run on Windows - While MacOS and Linux can _compile_ stuff for `net472`, it can't _execute_. Updated the test fixtures for `net472` to only include a test adapter in those test projects if it's Windows.
- Enabled tests for non-Windows - For test projects that target .NET Core, you do need to include a test adapter (ie `xunit.runner.visualstudio`) or `dotnet test` doesn't know how to run the tests. Updated to remove the Windows-only testing on those projects.
- Fixed an `OpenTelemetry.Exporter.Geneva` test to pass on MacOS - The socket handling in these tests doesn't seem to work right on MacOS. I found one test that was failing because a socket was trying to be disposed but was getting disposed twice, which caused an exception. Doing a check before disposing allows the test to run and pass.

This fixes a lot, but doesn't fix everything.

- The `OpenTelemetry.Exporter.Geneva` tests are problematic - I got some to pass, but there is a lot of challenge with Unix sockets and MacOS. I haven't dived into that yet. These tests do fail on MacOS. I have seen a few issues with getting Unix sockets and .NET working with Mac, unsure if it overlaps with that.
- Hundreds of `info` messages - VS Code shows all the `info` messages ("suggested refactorings") by default. These are easy one-click fixes like "`using` statement can be simplified" or "Use compound assignment." It might be worth doing a bit of housekeeping to clean these up; or, alternatively, disable the info messages via `.editorconfig`.
- The `examples/AspNet` project is all red - At least on MacOS, this thing is all red because it's using the old project type and doesn't seem to include a reference to the .NET reference assemblies. I didn't dig in to see what can be done here, either.

### DIFF
--- a/test/OpenTelemetry.Exporter.Geneva.Benchmark/OpenTelemetry.Exporter.Geneva.Benchmark.csproj
+++ b/test/OpenTelemetry.Exporter.Geneva.Benchmark/OpenTelemetry.Exporter.Geneva.Benchmark.csproj
@@ -1,21 +1,22 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <PropertyGroup>
-		<OutputType>Exe</OutputType>
-    	<!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
-        <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
-        <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net48;net472;net471;net47;net462</TargetFrameworks>
-        <NoWarn>$(NoWarn),SA1201,SA1202,SA1204,SA1311,SA1123</NoWarn>
-	</PropertyGroup>
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
+    <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net48;net472;net471;net47;net462</TargetFrameworks>
+    <NoWarn>$(NoWarn),SA1201,SA1202,SA1204,SA1311,SA1123</NoWarn>
+  </PropertyGroup>
 
-	<ItemGroup>
-		<PackageReference Include="BenchmarkDotNet" Version="$(BenchmarkDotNetPkgVer)" />
-		<PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="$(OpenTelemetryExporterInMemoryLatestPreReleasePkgVer)" />
-		<PackageReference Include="Microsoft.Extensions.Logging" Version="[6.0.0,)" />
-	</ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="$(BenchmarkDotNetPkgVer)" />
+    <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="$(OpenTelemetryExporterInMemoryLatestPreReleasePkgVer)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="[6.0.0,)" />
+  </ItemGroup>
 
-	<ItemGroup>
-		<ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\OpenTelemetry.Exporter.Geneva.csproj" />
-	</ItemGroup>
+  <ItemGroup>
+    <ProjectReference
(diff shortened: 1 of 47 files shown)


tillig:
It looks like there are a few PRs hanging out with some analysis issue fixes that may help with some of the info messages I mentioned.

tillig:
Looks like it's both good and bad that I enabled tests on non-Windows... Good because the tests are running, bad because there are three not passing on Ubuntu. I'll dig in.

test/OpenTelemetry.Exporter.Geneva.Benchmark/OpenTelemetry.Exporter.Geneva.Benchmark.csproj
```diff
@@ -1,21 +1,22 @@
...
+    <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="$(OpenTelemetryExporterInMemoryLatestPreReleasePkgVer)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="[6.0.0,)" />
+  </ItemGroup>
 
-	<ItemGroup>
-		<ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\OpenTelemetry.Exporter.Geneva.csproj" />
-	</ItemGroup>
+  <ItemGroup>
+    <ProjectReference
+      Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\OpenTelemetry.Exporter.Geneva.csproj" />
```
Kielek, line 19: nit:
```suggestion
    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\OpenTelemetry.Exporter.Geneva.csproj" />
```

Kielek (approved):

cijothomas (approved):

cijothomas:
Failed OpenTelemetry.Contrib.Extensions.AWSXRay.Tests.Resources.Http.TestServerCertificateValidationProvider.TestValidCertificate [213 ms]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at OpenTelemetry.Contrib.Extensions.AWSXRay.Tests.Resources.Http.TestServerCertificateValidationProvider.TestValidCertificate() in /home/runner/work/opentelemetry-dotnet-contrib/opentelemetry-dotnet-contrib/test/OpenTelemetry.Contrib.Extensions.AWSXRay.Tests/Resources/Http/TestServerCertificateValidationProvider.cs:line 46

This test is failing. To unblock, we can disable it, and create an issue to track re-enabling.

tillig:
Yeah, it looks like it's specifically on Linux in `netcoreapp3.1` - Linux in `net7.0` and `net6.0` seems to pass.

I can add a `[Skip]` to the whole test, or I can do something like "if(OsIsWindows)" around it or something. Since it was only running in Windows before, the OS check would make sense; it'd probably be better to at least run it _sometimes_.

Let me know how you want to proceed. I hadn't gotten a chance to get back to this yet today so I'm not quite on top of it.

cijothomas:
> I can do something like "if(OsIsWindows)" around it or something. Since it was only running in Windows before, the OS check would make sense; it'd probably be better to at least run it sometimes

^ sounds good to me.

tillig:
Looks like that took care of it. Once this is merged I'll file an issue to track fixing that test.

cijothomas:
> Failed OpenTelemetry.Contrib.Extensions.AWSXRay.Tests.Resources.Http.TestServerCertificateValidationProvider.TestValidCertificate [213 ms] Error Message: Assert.True() Failure Expected: True Actual: False Stack Trace: at OpenTelemetry.Contrib.Extensions.AWSXRay.Tests.Resources.Http.TestServerCertificateValidationProvider.TestValidCertificate() in /home/runner/work/opentelemetry-dotnet-contrib/opentelemetry-dotnet-contrib/test/OpenTelemetry.Contrib.Extensions.AWSXRay.Tests/Resources/Http/TestServerCertificateValidationProvider.cs:line 46
> 
> This test is failing. To unblock, we can disable it, and create an issue to track re-enabling.

@srprash @lupengamzn Tagging owners of this component to revisit/fix this.

tillig:
Created #1007 to track that test.

PR #1011, opened 2023-02-16
[Instrumentation.Process] Release Process instrumentation version 0.5.0

Yun-Ting:
## Changes

Release Process instrumentation version 0.5.0

* [x] Design discussion issue https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/447

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 Released 2023-Feb-17
 
+> **Note**
+> The version number was lowered from 1.0.0 to 0.5.0 to better reflect the
+experimental state of Opentelemetry process metrics specification status.
+Packages that were older than this release will be delisted to avoid confusion.
+
 * Added `process.cpu.count` metric.
   ([#981](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/981))
 
--- a/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Unreleased
 
-## 0.5.0
+## 0.5.0-beta.1
 
-Released 2023-Feb-16
+Released 2023-Feb-17
 
 * Added `process.cpu.count` metric.
   ([#981](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/981))
--- a/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.5.0
+
+Released 2023-Feb-16
+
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.5.0
```
Kielek, line 5: This version can be confusing. I do not think that it is good idea to release 0.5.0 after 1.0 (alpha)

Yun-Ting: The spec is not stable yet and based on the following statement, I think 0.X.0 without any identifier may be a good fit.
https://semver.org/#spec-item-4

But I do see your concern.
Would 1.0.0-prerelease.1 makes more sense?

cijothomas: there is proposal, even in the main repo to release the instrumentation libraries as stable (i.e without prerelease identifiers), however with < 1.0.0 as version.

0.5.0 is doable, but lets first do 0.5.0-beta/rc and wait for weeks.

the 1.0.0-alpha out there could be delisted to avoid confusion?

Kielek: Delisting seems to be good compromise.
Why starting from some random value? Why not 0.1.0?
What should be order `1.0.0-prerelease.1`? alpha, beta, prerelease, then rc (I suppose that nuget will treat it alphabetically)?

IMO, delisting + start from some low number looks like the best option.

@cijothomas, it will be great to cleanup a bit also old contrib packages: https://www.nuget.org/packages?q=OpenTelemetry.Contrib and redirect them to version without `Contrib` infix. If not possible, also delist them.

cijothomas: > Why starting from some random value? Why not 0.1.0?

TBH, i do not have a preference if it is 0.1.0 or 0.5.0 as long as it is < 1.0.0.

We are planning to try this for instrumentaion library in main repo. This package is a good way to test the process. (as we will need to delist 1.* from nuget here too.)

cijothomas (approved):

CodeBlanch (approved):

src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.5.0-beta.1
+
+Released 2023-Feb-17
```
Kielek, line 7: Please make a note that old packages ware delisted in nuget org

Kielek (approved):

src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
cijothomas: @Yun-Ting could you make a note here? and then I'll push the new nuget and delist old.

Yun-Ting: updated the changelog; thank you

PR #1029, opened 2023-02-22
Added Azure Resource Detector

martinjt:
Draft right now... needs work, but I thought I'd put it out there for people to see.

### DIFF
--- a/src/OpenTelemetry.ResourceDetectors.Azure/AppServiceResourceDetector.cs
+++ b/src/OpenTelemetry.ResourceDetectors.Azure/AppServiceResourceDetector.cs
@@ -24,6 +24,7 @@ namespace OpenTelemetry.ResourceDetectors.Azure;
 /// <summary>
 /// Resource detector for Azure AppService environment.
 /// </summary>
+[Obsolete("Use AzureResourceDetector instead")]
 public sealed class AppServiceResourceDetector : IResourceDetector
 {
     /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


src/OpenTelemetry.ResourceDetectors.Azure/AppServiceResourceDetector.cs
```diff
@@ -24,6 +24,7 @@ namespace OpenTelemetry.ResourceDetectors.Azure;
 /// <summary>
 /// Resource detector for Azure AppService environment.
 /// </summary>
+[Obsolete("Use AzureResourceDetector instead")]
```
cijothomas, line 27: this package is not stable, so no need to obselete. we can just remove things.. assuming this is the direction agreed by the owners.

src/OpenTelemetry.ResourceDetectors.Azure/AzureResourceDetector.cs
```diff
@@ -0,0 +1,81 @@
...
+/// </summary>
+public class AzureResourceDetector : IResourceDetector
+{
+    private readonly List<Tuple<string, string>> envVarsToAdd = new()
+    {
+        new("az.appservice.site_name", "WEBSITE_SITE_NAME"),
+        new("az.resource_group", "WEBSITE_RESOURCE_GROUP"),
+        new("az.subscription_id", "WEBSITE_OWNER_NAME"),
+        new("az.region", "REGION_NAME"),
+        new("az.appservice.platform_version", "WEBSITE_PLATFORM_VERSION"),
```
cijothomas, line 33: are these part of all azure environems like vms, aks etc?

Issue #1034, opened 2023-02-23
Add SQS/SNS support.


rypdal:
# Issue with OpenTelemetry.Instrumentation.AWSLambda and OpenTelemetry.Contrib.Instrumentation.AWS

1. _OpenTelemetry.Instrumentation.AWSLambda_ 
The current AWS Lambda instrumentation SDK supports tracing for the following Amazon API Gateway events: [APIGatewayProxyRequest and APIGatewayHttpApiV2ProxyRequest](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/README.md). The SDK can be easily extended to support [SQS](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs-example.html) and [SNS](https://docs.aws.amazon.com/lambda/latest/dg/with-sns-example.html) events. Thus tracing for Lambda handlers receiving SQS and SNS messages can be added to SDK without API changes.
2. _OpenTelemetry.Contrib.Instrumentation.AWS_ 
Tracing of outgoing SQS and SNS calls can be easily added without API changes by extension of already existing [AWSTracingPipelineHandler class](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs).

rypdal:
The issue is resolved by this PR https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051 and can be closed.

PR #1051, opened 2023-02-28
#1034: AWS SQS and SNS support

rypdal:
## Changes

- incoming requests (OpenTelemetry.Instrumentation.AWSLambda) : implemented parent extraction for incoming SQS and SNS events (getters)
- outgoing requests (OpenTelemetry.Contrib.Instrumentation.AWS): implemented enrichment of outgoing message attributes by trace data (setters)
- no API changes


src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessageAttributeHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+            return false;
+        }
+        else if (index >= MaxMessageAttributes)
+        {
+            // TODO: Add logging (event source).
+            return false;
+        }
+
+        var attributePrefix = this.attributeFormatter.AttributeNamePrefix + $".{index.Value}";
+        parameters.Add(attributePrefix + ".Name", name.Trim());
```
Oberon00, line 52: I don't think it is safe to call `Trim` in this method. We should use the name and value as-is.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs
```diff
@@ -196,10 +196,18 @@ private void AddRequestSpecificInformation(Activity activity, IRequestContext re
...
         }
 
-        if (AWSServiceHelper.IsDynamoDbService(service))
+        if (AWSServiceType.IsDynamoDbService(service))
         {
             activity.SetTag(SemanticConventions.AttributeDbSystem, AWSSemanticConventions.AttributeValueDynamoDb);
         }
+        else if (AWSServiceType.IsSqsService(service))
+        {
+            Propagators.DefaultTextMapPropagator.Inject(new PropagationContext(activity.Context, Baggage.Current), requestContext, AWSMessagingUtils.SqsMessageAttributeSetter);
```
Oberon00, line 205: This approach has the problem that we might do a partial injection if the propagator sets multiple keys.

I would suggest to change this to inject into a temporary Dictionary, then check if any of the keys is already contained and if we have enough space in the message attributes to insert all keys. If any of that fails, we shouldn't modify the message attributes ("all-or-nothing").

While the temporary dictionary itself is additional overhead, the two-step-approach can be used to make finding insertion indices faster, since we only need to iterate once to find the highest existing index.

Alternatively, we could also rely on the OriginalRequest to get the list/dictionary of message attributes, maybe that would be easier?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,141 @@
...
+    internal static PropagationContext ExtractParentContext(SQSEvent.SQSMessage sqsMessage)
+    {
+        if (sqsMessage == null)
+        {
+            return default;
+        }
+
+        // SQS subscribed to SNS topic with raw delivery disabled case, i.e. SNS record serialized into SQS body.
+        // https://docs.aws.amazon.com/sns/latest/dg/sns-large-payload-raw-message-delivery.html
+        SNSEvent.SNSMessage snsMessage = GetSnsMessage(sqsMessage);
```
Oberon00, line 54: I think before checking for the embedded SNS message, we should first try the SQS MessageAttributes, not the other way round.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,141 @@
...
+        SNSEvent.SNSMessage snsMessage = null;
+
+        var body = sqsMessage.Body;
+        if (body != null &&
+            body.TrimStart().StartsWith("{") &&
+            body.Contains(SnsMessageAttributes))
+        {
+            try
+            {
+                snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
```
Oberon00, line 130: Instead of deserializing the full to a .NET type, have you considered using lower-level JSON APIs to just deserialize the attributes? I don't know if this would be more efficient though, so in doubt we can leave it as-is.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            catch (JsonException)
```
Oberon00, line 132: Probably we should catch any exception here, we might also get OOM exceptions or similar.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessageAttributeHelper.cs
rypdal: This code is moved to implementations of `IRequestContextAdapter.AddAttribute(...)` method. Trims have been removed.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs
rypdal: - I made some refactoring and now the code for the index determining and key presence relies on the OriginalRequest. 
- Also the code for adding trace data became transactional: "all-or-nothing". 
- I introduced `IRequestContextAdapter` in order to have common code for accessing SNS and SQS attributes properties.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: The check order has been swapped.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: Replaced by the common exception. Yes, it's bad. Probably it's a question of another issue.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+
+internal static class AWSMessagingUtils
+{
+    // SQS/SNS message attributes collection size limit according to
+    // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html and
+    // https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+    private const int MaxMessageAttributes = 10;
+
+    internal static void Inject(IRequestContextAdapter requestAdapter, PropagationContext propagationContext)
```
Oberon00, line 29:
```suggestion
    internal static void Inject(IRequestContextAdapter request, PropagationContext propagationContext)
```

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (!requestAdapter.HasMessageBody ||
+            !requestAdapter.HasOriginalRequest)
+        {
+            return;
+        }
+
+        var carrier = new Dictionary<string, string>();
+        Propagators.DefaultTextMapPropagator.Inject(propagationContext, carrier, Setter);
```
Oberon00, line 38:
```suggestion
        Propagators.DefaultTextMapPropagator.Inject(propagationContext, carrier, (c, k, v) => c[k] = v));
```

and remove Setter

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+            // TODO: Add logging (event source).
+            return;
+        }
+
+        int nextAttributeIndex = attributesCount + 1;
+        foreach (var param in carrier)
+        {
+            if (requestAdapter.ContainsAttribute(param.Key))
+            {
+                continue;
```
Oberon00, line 52: If any attribute is already contained, we should skip injection entirely and not inject any of the attributes. Partial injection should be avoided (e.g. we might end-up injecting a tracestate that is unrelated to an already existing traceparent)

(Alternatively, we could overwrite existing attributes, I think that's the principle most other instrumentations follow, though I am not sure it would be appropriate here)

See also my original comment https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051#discussion_r1119857473

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            requestAdapter.AddAttribute(param.Key, param.Value, nextAttributeIndex);
+            nextAttributeIndex++;
+
+            // Add trace data to message attributes dictionary of the original request.
```
Oberon00, line 58:
```suggestion
            // Add injected attributes to the original request as well.
```

Injection might contain more than trace context, theoretically it could even be configured to *only* propagate baggage

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,66 @@
...
+internal static class AWSMessagingUtils
+{
+    // SQS/SNS message attributes collection size limit according to
+    // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html and
+    // https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+    private const int MaxMessageAttributes = 10;
+
+    internal static void Inject(IRequestContextAdapter requestAdapter, PropagationContext propagationContext)
+    {
+        if (!requestAdapter.HasMessageBody ||
```
Oberon00, line 31: Why is a message body required?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    bool HasMessageBody { get; }
+
+    bool HasOriginalRequest { get; }
+
+    int AttributesCount { get; }
+
+    bool ContainsAttribute(string name);
+
+    void AddAttribute(string name, string value, int nextAttributeIndex);
```
Oberon00, line 28: Maybe `AddAttribute` should already include `AddAttributeToOriginalRequest` instead of having two separate methods?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+internal interface IRequestContextAdapter
+{
+    bool HasMessageBody { get; }
+
+    bool HasOriginalRequest { get; }
```
Oberon00, line 22: Probably this should be renamed to something like
```suggestion
    bool CanInject { get; }
```

if that is the purpose?

I am not sure what we need HasMessageBody  for.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
```diff
@@ -8,12 +8,14 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.7.101.15" />
```
Oberon00, line 11: I don't think we should upgrade this version. Wouldn't that raise the minimum required version?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    internal static PropagationContext ExtractParentContext(SNSEvent snsEvent)
+    {
+        if (snsEvent == null)
+        {
+            return default;
+        }
+
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
+        var record = snsEvent.Records.LastOrDefault();
```
Oberon00, line 76:
```suggestion
        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
        var record = snsEvent?.Records?.LastOrDefault();
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    internal static PropagationContext ExtractParentContext(SQSEvent sqsEvent)
+    {
+        if (sqsEvent == null)
+        {
+            return default;
+        }
+
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
+        var message = sqsEvent.Records.LastOrDefault();
```
Oberon00, line 42:
```suggestion
        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
        var message = sqsEvent?.Records?.LastOrDefault();
```

(The SQSEvent is parsed from JSON, and using the Lambda Invoke API arbitrary JSON can be passed, so in this structure we should be prepared for null anywhere.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
+        var record = snsEvent.Records.LastOrDefault();
+        return ExtractParentContext(record);
+    }
+
+    internal static PropagationContext ExtractParentContext(SNSEvent.SNSRecord record)
+    {
+        return (record?.Sns != null) ?
```
Oberon00, line 82:
```suggestion
        return (record?.Sns?.MessageAttributes != null) ?
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+            Propagators.DefaultTextMapPropagator.Extract(default, record.Sns.MessageAttributes, SnsMessageAttributeGetter) :
+            default;
+    }
+
+    internal static PropagationContext ExtractParentContext(SNSEvent.SNSMessage message)
+    {
+        return (message != null) ?
```
Oberon00, line 89:
```suggestion
        return (message?.MessageAttributes != null) ?
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+            Propagators.DefaultTextMapPropagator.Extract(default, message.MessageAttributes, SnsMessageAttributeGetter) :
+            default;
+    }
+
+    private static IEnumerable<string> SqsMessageAttributeGetter(IDictionary<string, SQSEvent.MessageAttribute> attributes, string attributeName)
+    {
+        SQSEvent.MessageAttribute attribute = attributes.GetValueByKeyIgnoringCase(attributeName);
```
Oberon00, line 96: I don't think we should need to ignore casing here, it's not HTTP. Isn't it rather expensive as well? Same for SNS.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    private static IEnumerable<string> SnsMessageAttributeGetter(IDictionary<string, SNSEvent.MessageAttribute> attributes, string attributeName)
+    {
+        SNSEvent.MessageAttribute attribute = attributes.GetValueByKeyIgnoringCase(attributeName);
+        if (attribute == null)
+        {
+            return null;
+        }
+
+        switch (attribute.Type)
```
Oberon00, line 115:
```suggestion
        switch (attribute?.Type)
```

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
```diff
@@ -10,13 +10,15 @@
   <ItemGroup>
     <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.1" />
     <PackageReference Include="Amazon.Lambda.Core" Version="1.2.0" />
+    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.0.0" />
+    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="2.1.0" />
```
Oberon00, line 14: We have all-different versions here. Probably we should stick to the lowest possible for max compatiblity? Or are these independently versioned?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Implemented logic: if at least one attribute is already present in the request then we skip the injection.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Not required anymore. Removed.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
+        var message = sqsEvent.Records.LastOrDefault();
+        return ExtractParentContext(message);
+    }
+
+    internal static PropagationContext ExtractParentContext(SQSEvent.SQSMessage sqsMessage)
+    {
+        if (sqsMessage == null)
```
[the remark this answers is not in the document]
rypdal: Applied along with other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: Applied along with other changes.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
rypdal: What do you mean ? It's not an upgrade. We introduced this reference the 1st time.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
rypdal: It's the latest stable versions available at the moment. There is no 2.0.1 available for the SNSEvents. 
https://www.nuget.org/packages/Amazon.Lambda.SQSEvents#versions-body-tab
https://www.nuget.org/packages/Amazon.Lambda.SNSEvents#versions-body-tab

Do you think we still have to keep versions equal ?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
Oberon00: Isn't the `if` above now redundant?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Collections.Generic;
+using OpenTelemetry.Context.Propagation;
+
+namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+
+internal static class AWSMessagingUtils
```
Oberon00, line 22: A separate class is probably overkill now, this Inject method could be defined inside AWServiceType now (BTW, I'm also not sure if the split of ServiceType from ServiceHelper is helpful)

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSServiceType.cs
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+
+namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+
+internal class AWSServiceType
```
Oberon00, line 21: Is splitting this in a separate file really worth it?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Collections.Generic;
+
+namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+internal interface IRequestContextAdapter
```
Oberon00, line 20: Not sure what the guidelines in this project are, but if there is no guideline against it, I would put this interface and the two implementations into a single .cs file.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
Oberon00: No, I agree if they aren't released together, it does not make any sense to keep them in sync.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/SnsRequestContextAdapter.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        this.parameters = context.Request?.ParameterCollection;
+        this.originalRequest = context.OriginalRequest as PublishRequest;
+    }
+
+    public bool CanInject => this.originalRequest?.MessageAttributes != null && this.parameters != null;
+
+    public void AddAttributes(IReadOnlyDictionary<string, string> attributes)
+    {
+        if (!this.CanInject)
```
Oberon00, line 44: This check feels redundant. If we are very worried about users forgetting to call CanInject, we could remove `CanInject` and instead create a factory method that returns null instead of a new instance if CanInject would be false.

At that point (maybe even now already), it wouldn't really be an adapter anymore but basically just a single method that could as well be static.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: I don't see benefits of just deserializing the attributes. I'm not sure about efficiency 'the whole class'  vs 'only attributes'. Let's leave it as-is.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSServiceType.cs
rypdal: With `AWSServiceType` the code looks more self-explanatory. I think tests parameterised with `[InlineData(AWSServiceType.SQSService)]` are looking more elegant. If for If you open the sub-folder you immediately see AWSServiceType and realise what services are covered at the moment. Strongly speaking it's may be had to be an enum, but it brings unnecessary conversions enum <-> string and I decided to keep string constants.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
rypdal: In this case I would prefer to keep definition per file. The interface implementations are not that small. Also looking at the project sub-folder you may guess just by names how many implementations the interface has.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
rypdal: Lets keep code more structured. Probably we will have to support even more AWS triggers with it's types. Otherwise the single method of this class will go into AWSTracingPipelineHandler as well as possible future methods for other AWS triggers. AWSTracingPipelineHandler will become big all-in-one class.

src/OpenTelemetry.Contrib.Instrumentation.AWS/OpenTelemetry.Contrib.Instrumentation.AWS.csproj
rypdal: I returned direct ref. to AWSSDK.Core 3.5.1.24 and downgraded AWSSDK.SimpleNotificationService and AWSSDK.SQS to the highest compatible with this version of AWSSDK.Core.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -14,22 +14,23 @@
...
 
 namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
 
 internal static class AWSMessagingUtils
 {
-    internal static void Inject(IRequestContextAdapter request, PropagationContext propagationContext)
+    internal static void Inject(Action<IReadOnlyDictionary<string, string>>? addAttributes, PropagationContext propagationContext)
     {
-        if (!request.CanInject)
+        if (addAttributes == null)
```
Oberon00, line 27: Is this check really necessary? I don't think `null` is possibly passed here, as one can easily verify. If null is actually passed, it will be accidental and this check will make the bug harder to find.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
```diff
@@ -14,22 +14,23 @@
...
+using System;
 using System.Collections.Generic;
 using OpenTelemetry.Context.Propagation;
 
 namespace OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
 
 internal static class AWSMessagingUtils
 {
-    internal static void Inject(IRequestContextAdapter request, PropagationContext propagationContext)
+    internal static void Inject(Action<IReadOnlyDictionary<string, string>>? addAttributes, PropagationContext propagationContext)
```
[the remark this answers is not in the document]
rypdal: I did further simplification, but not sure what do you mean by 'direct AddAttributes' ?

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/IRequestContextAdapter.cs
rypdal: This interface has been removed during code simplifications.

src/OpenTelemetry.Contrib.Instrumentation.AWS/Implementation/AWSMessagingUtils.cs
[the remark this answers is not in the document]
rypdal: I left just a single static method `AddAttributes` for each type of adapter (actually it became just a helper).

Oberon00 (approved):

rypdal:
Hi @srprash. Please, take a look and review/approve changes made in the OpenTelemetry.Contrib.Instrumentation.AWS project. Seems you are the only owner of the project :-)

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,137 @@
...
+        SNSEvent.SNSMessage snsMessage = null;
+
+        var body = sqsMessage.Body;
+        if (body != null &&
+            body.TrimStart().StartsWith("{") &&
+            body.Contains(SnsMessageAttributes))
+        {
+            try
+            {
+                snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
```
martincostello, line 126: Is it possible to use System.Text.Json to do this as it would probably be lighter weight?

rypdal: I checked other projects and seems they use the Newtonsoft serializer (for example, see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/e280873085ea96712cc636f00941c0e7acef7f3e/src/OpenTelemetry.Contrib.Extensions.AWSXRay/Resources/ResourceDetectorUtils.cs#L60). May be it's worth to keep Newtonsoft  for consistency and check/switch to the System.Text.Json everywhere. It may be a subject of a separate issue.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,137 @@
...
+internal class AWSMessagingUtils
+{
+    // SNS attribute types: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+    private const string SnsAttributeTypeString = "String";
+    private const string SnsAttributeTypeStringArray = "String.Array";
+    private const string SnsMessageAttributes = "MessageAttributes";
+
+    internal static PropagationContext ExtractParentContext(SQSEvent sqsEvent)
+    {
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
```
atshaw43, line 36: I don't think this assumption is correct.

Each message could be its own workflow.

The typical use case is that the parent span is the span created by the Lambda function.

A span link should be added to the trace context in X-Amzn-Trace-Id or AWSTraceHeader.

atshaw43 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
atshaw43, line 36: Shouldn't the parent be the trace context created by the Lambda function?

atshaw43 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,137 @@
...
+        var message = sqsEvent?.Records?.LastOrDefault();
+        return ExtractParentContext(message);
+    }
+
+    internal static PropagationContext ExtractParentContext(SQSEvent.SQSMessage sqsMessage)
+    {
```
atshaw43, line 42: Shouldn't the parent be the trace context created by the Lambda function?

Also, shouldn't we be creating span links to the trace contexts in the MessageAttributes?

atshaw43 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: This is the parent *of* the trace context created by the Lambda function.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: As above: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051#discussion_r1147241174


This is the parent *of* the span created by the Lambda function.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
atshaw43: It looks like we are using the span context in the last message's message attribute as a parent.

What happens in a batch case? Say we have multiple messages in a Lambda event. We may not want to use the last message's upstream trace context.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
atshaw43: It looks like we are using the span context in the message attribute as the parent. I believe this span context should be a span link, not the parent.

What happens in the fan-in case? Say there are multiple messages in the Lambda event that are using in a downstream call?

atshaw43 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: I think you are right and we should add links for all messages as well. However, I do think that adding one of the messages as parent (in this case the last one) is useful as well and should stay.

rypdal: I added links for SQS messages. Please, see the commit: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051/commits/e88edbe98534c65924834d96edc2b0b1a706fb8a

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        // We assume there can be only one parent that's why we consider only a single (the last) record as the carrier.
+        var parentRecord = sqsEvent.Records.LastOrDefault();
+        var parentContext = ExtractParentContext(parentRecord);
+
+        var links = new List<ActivityLink>();
+        foreach (var record in sqsEvent.Records)
+        {
+            var context = ExtractParentContext(record);
```
Oberon00, line 49: I think we should add an optimization here for the case of `object.ReferenceEquals(record, parentRecord)` where we reuse the `parentContext` instead of extracting again.
Especially big improvement for the common case of a single record.

rypdal: Thanks. I added this optimisation.

martinjt:
I've been taking a look through and I've seen the discussion around span links versus parents so thought I'd contribute.

This is a long standing thing that's discussed. Particularly, the issue is when 1 process generates hundreds of messages, and then the trace is flooded and unusable.

The TL;DR is that it needs to be configurable by the user (that's how Azure are doing this with the ServiceBus SDKs). How you achieve that, I'm not sure, but there isn't a one size fits all approach that works.

I also saw something around SNS Records, and the fact that we're assuming that they're all for the same tracecontext so we only take into account the last one, I'm not sure that's right, but I don't understand SNS enough to make an actual judgement on that.

martincostello:
SNS supports batching for publishing: [Amazon SNS now supports publishing batches of up to 10 messages in a single API request](https://aws.amazon.com/about-aws/whats-new/2021/11/amazon-sns-supports-publishing-batches-messages-single-api-request/)

Oberon00:
Batch-publishing doesn't imply batch-receiving.
The docs for AWS SQS and AWS Lambda use singular "a message" in https://docs.aws.amazon.com/sns/latest/dg/lambda-console.html and https://docs.aws.amazon.com/lambda/latest/dg/with-sns.html.

For SNS, you can configure things like a batch window size in the Lambda trigger, I don't think that is possible for SNS. Unfortunately the AWS docs are not super extensive...

jeastham1993:
I second @martinjt comments around making this configurable. I see two different use cases with SQS, one in which I want to trace over an entire batch of messages or I want to trace an individual message as part of it's original context. The context propagation (or event links) would be different in both.

I've got a pretty rudimentary example on my GitHub which either ['auto starts' the trace](
https://github.com/jeastham1993/dotnet-lambda-observability/blob/main/src/OTEL/src/ObservableLambda.Shared/ObservableFunction.cs) using the Lambda propagation context or allows a user to [start a trace as part of the individual message processing](https://github.com/jeastham1993/dotnet-lambda-observability/blob/main/src/OTEL/src/ObservableLambda.Processor.SingleMessage/Function.cs)

The comments about SNS and a single message is correct, you can see that in the SNS FAQS
https://aws.amazon.com/sns/faqs/

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
jeastham1993: I don't think using the last record is the right approach, given each message in the queue could have come from a different source with a completely different original trace. Using the last would mean all other traces wouldn't be linked/used.

martinjt:
So I'm having a look through the code and I'm struggling to understand what's happening. There looks to be a link being created for every record, and added to the lambda execution?

So what we have is 1 span/activity for the lambda execution, with a ParentContext set to the value of the Last SQS record's trace information, then links to all the other SQS records' trace information?

The issue is that this isn't a good debug experience in the majority of messaging scenarios.

What you happen is that there is a span for the lambda execution (root span, no parent, SpanKind = Server), then a span for each message process action (SpanKind = Consumer). that message span needs to have the ability to have a Parent Context that comes from the SQS record OR that context from the SQS record as an ActivityLink/SpanLink but no parent.  That's the configurability required.

Apologies if this is what it's doing, I'm struggling to see that it's doing that without running it up in SQS myself.

Perhaps we need to see a worked example, maybe with LocalStack and Jaegar?

martinjt:
So option 2 there is the least confusing, and similar to the way that Azure Functions are doing the same for service bus.

Option 1 would be confusing to me as I'd see the span for the lambda in Trace 2, not Trace 1 (links are parentage, so they're not necessarily part of the same view).

Oberon00:
With option 2, we get 3 completely separate traces. I know that we'd like to support option 1 at least for Dynatrace since it works very nicely for the probably not uncommon case that you only have a single message. And you could argue that the Lambda is triggered by the last message, since that is the one that made the batch full (or after which the batch wait window expired).

As you say, the Lambda execution is no the message processing, but it is still triggered by the sent message(s), so it makes sense to link them.

I think what one would probably want is additional spans inside the lambda that track the per-message processing. This should probably be designed once https://github.com/open-telemetry/oteps/pull/220 has made it to the spec.

martinjt:
Right now, I can't say that I support the idea that you would see the process of Message 1 inside the trace for Message 2. That feels wrong, and misleading.

Having a message trace that continues into the inside of the execution of the lambda, linked to a trace for the lambda execution would be the desired outcome. If that's not something we're willing to do now (It's part of that OTEP), I think it should be that the message processing is part of the Lambda execution trace that isn't linked back.

Failing that, add the optionality, perhaps as a setup in the lambda somewhere.

jeastham1993:
I agree with Martin. Including trace 1 within trace 2 and vica versa might be ok in some cases, but the majority of use cases I see with SQS each message is processed as an independent entity.

Displaying them together would be confusing, and I'd suggest there needs to be some way to control this behaviour.

martinjt:
How does this work in other language implementations of the SDK?

Oberon00:
I think .NET is the first to support something other than the X-Ray parent for SQS.

rypdal:
I added new option (SetParentFromMessageBatch) configuring setting parent behaviour. No parent is set when it's false (links only), and parent is set from the last received message in a batch when the option is true.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        return parentContext;
+    }
+
+    internal static PropagationContext ExtractParentContext(SNSEvent snsEvent)
+    {
+        // We assume there can be only a single SNS record (message) and records list is kept in the model consistency.
+        // See https://aws.amazon.com/sns/faqs/ for details.
```
Oberon00, line 90:
```suggestion
        // See https://aws.amazon.com/sns/faqs/#Reliability for details.
```

Oberon00 (commented):
Please add a unit test for the new feature to the AWS Lambda instrumentation unit tests, I think there is none yet?

rypdal:
I added a couple of unit tests for the new config option.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSMessagingUtilsTests.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        var tracerProvider = Sdk.CreateTracerProviderBuilder()
+            .Build();
+    }
+
+    [Fact]
+    public void ExtractParentContext_SetParentFromMessageBatchIsDisabled_ParentIsNotSet()
+    {
+        AWSMessagingUtils.SetParentFromMessageBatch = false;
+        var @event = CreateSqsEventWithMessages(new[] { SpanId1, SpanId2 });
```
Oberon00, line 45: Maybe use "input" to avoid the weird @-syntax

Oberon00 (approved):
Thank you!
I hope (and do think) this also addresses the previous [concerns](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051#issuecomment-1491860438) by @martinjt and @jeastham1993 about the potentially confusing trace structure, since the parent-setting is now behind an off-by-default option.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSMessagingUtilsTests.cs
rypdal: Renamed to sqsEvent.

rypdal:
Hi @srprash ! The final changes have been made after discussions about creating/not creating a parent for a message batch. The new configuration setting has been introduced for toggling the parent creation. Please, review these as well as previous changes. Thank you.

Oberon00:
EasyCLA issue seems to be https://github.com/open-telemetry/community/issues/1445

srprash:
Looping in @atshaw43 to take a look.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: @atshaw43 Your concern should already be addressed in the newest version of this PR, as @rypdal added an option to configure whether a parent should be set or not (if not, the current context, if any, will be the implicit parent). See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051#issuecomment-1505516804 and the diff since then (https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1051/files/24ae92dfdcdd1ed8914bce51f08698fb41ece7a5..0e4d3857c649b9c4f99570d1103fabd37b697779).

Also note that the typical usage context will be different here than in your Java PR. In this PR here, we know that we are inside an AWS Lambda, so there typically will not be any implicit parent, the Lambda invocation span that we create here will be the (local) root span. Still, if there would be any parent span, I think setting the SetParentFromMessageBatch configuration to false (which is even the default now) should use it.

Oberon00: As @rypdal made me aware, we actually never use the implicit / current context as parent, contrary to my previous comment. But this applies also to the existing old HTTP code paths. I think adjusting this should be done in a follow-up. You do have the possibility to pass in the current context explicitly when calling this library's API though.

atshaw43 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
rypdal: We may discuss using of the implicit / current parent in the follow-up PR. It may trigger further discussions in this already long playing PR.

src/OpenTelemetry.Contrib.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   ([#1095](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1095))
 * Removes `AddAWSInstrumentation` method with default configure default parameter.
   ([#1117](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1117))
+* Added tracing of outgoing SQS and SNS calls.
```
Oberon00, line 9: I think this is wrong, it was already traced.

```suggestion
* Global propagator is now used to inject into sent SQS and SNS message attributes (in addition to X-Ray propagation).
```

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
   ([#943](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/943))
 * BREAKING (behavior): `AddAWSLambdaConfigurations` no longer calls `AddService`
   ([#1080](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1080))
+* Added tracing of AWS Lambda handlers receiving SQS and SNS messages.
```
Oberon00, line 11: Same here, we were already tracing it, but:

```suggestion
* Tracing received SQS events will now create links to the contexts contained in the messages and optionally a parent. Receiving SNS events will now set the parent to the context contained in the notification.
```

srprash (approved):

martinjt:
Spent a bit of time thinking about this and chatting to messaging people about it, and I'm still not completely sure that this is the right approach. That said, I'm also not sure what the right approach should be here, and so, please take this as an observation and in no way a blocker.

The links that are being created could be misleading. I'm behind the fact that we shouldn't set a parent (a batch execution is a trace in itself), however, linking that batch execution to the parent spans of the messages it's processing isn't how I would see this being beneficial.

The issue is that you can't link to the spans that are created for the message processing as, by definition, they're started/created after the batch has started. Which leaves the dilemma of what to do?

I look at things through the lens of someone debugging a platform, rather than the code.

One issue I'm thinking of is that the batch processing crashes out half way through, and you end up with the latter half of the messages not being processed. In that scenario, the latter half of the messages' trace parent is linked to the batch execution, however, nothing happened. That could be really confusing to the person debugging.

The other part is that I would expect the message processing span to be linked to the batch execution span, while having a parent of the originating message. Which means we're ending up with twice the amount of links being created.

I haven't seen anything in the library like a helper method to create an activity from the item in the batch, that would be a nice addition as we would be expecting people to do that in batch processing mode.

All in all, I think we've come along way in the right direction now. I haven't reviewed the code for style or completeness etc. I've assumed someone else was doing that, but if I need to then please ping me.

martinjt:
I also didn't see anything in a README about this setting, I think that would be a strong suggestion from me as the setting will need a lot of explaining.

Oberon00:
Are you aware of the messaging semantic convention's group proposal about this topic? https://github.com/open-telemetry/oteps/pull/220
I think the implementation here is now in alignment with that.

Oberon00:
> I haven't seen anything in the library like a helper method to create an activity from the item in the batch, that would be a nice addition as we would be expecting people to do that in batch processing mode.

I agree on this, but that can be provided in a follow-up (this PR is already rather large).

utpilla (approved):

PR #1038, opened 2023-02-24
Bump OTel to 1.4.0

Kielek:
## Changes

* Bump OTel 1.4.0 together with non-Core packages.
* `Exporter.Instana` was pinned to 1.3.2. For now it is referencing net4.6.1
* Adjust tests.
* Tests executed only against netcoreapp3.1 are executed also on net7.0 and net6.0 (projects referencing OTel 1.4.0).

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* ~~[ ] Design discussion issue #~~
* ~~[ ] Changes in public API reviewed~~

PR #1069, opened 2023-03-10
[Exporter.Geneva] Update GenevaMetricExporter to export Exemplars

utpilla:
## Changes
- Update GenevaMetricExporter serialization
- Export exemplars
- GenevaExporter to depend on `1.5.0-alpha.1` version of OpenTelemetry SDK

TODO:
- Cleanup code
- Update unit tests
- Update benchmark


src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -181,52 +186,49 @@ public override ExportResult Export(in Batch<Metric> batch)
...
-                                MetricData min = ulongZero;
-                                MetricData max = ulongZero;
-                                if (metricPoint.TryGetHistogramMinMaxValues(out var minValue, out var maxValue))
-                                {
-                                    min = new MetricData { UInt64Value = Convert.ToUInt64(minValue) };
-                                    max = new MetricData { UInt64Value = Convert.ToUInt64(maxValue) };
-                                }
-
-                                var bodyLength = this.SerializeHistogramMetric(
+                                _ = metricPoint.TryGetHistogramMinMaxValues(out double min, out double max);
```
cijothomas, line 219: if min max is not available, dont we need ulongZero to be written?

build/Common.props
```diff
@@ -34,7 +34,7 @@
     <MicrosoftPublicApiAnalyzersPkgVer>[3.3.3]</MicrosoftPublicApiAnalyzersPkgVer>
     <MicrosoftSourceLinkGitHubPkgVer>[1.1.1,2.0)</MicrosoftSourceLinkGitHubPkgVer>
     <OpenTelemetryCoreLatestVersion>[1.4.0,2.0)</OpenTelemetryCoreLatestVersion>
-    <OpenTelemetryCoreLatestPrereleaseVersion>[1.4.0,2.0)</OpenTelemetryCoreLatestPrereleaseVersion>
+    <OpenTelemetryCoreLatestPrereleaseVersion>[1.5.0-alpha.1]</OpenTelemetryCoreLatestPrereleaseVersion>
```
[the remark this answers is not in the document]
utpilla: There are no other projects referencing this version. They were updated to depend on the latest stable version in #1038

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: Yes, it would serialize them as zero. If `TryGetHistogramMinMaxValues` returns false, min and max would be set to zero.

cijothomas: mm.. Is that a contract or relying on the internal implementation details on OTel sdk? I think its wrong to rely on the min, max if TryGet fails.

utpilla: That should be considered a breaking change from the SDK standpoint: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/MetricPoint.cs#L328-L329

cijothomas: I am not sure if that is the case. Once it returned false, callers should not be relying on the values on the out parameters.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -484,6 +486,384 @@ internal static void DisableOpenTelemetrySdkMetricNameValidation()
...
+            {
+                dimensionsWritten += this.prepopulatedDimensionsCount;
+            }
+
+            // Serialize MetricPoint Dimension keys
+            foreach (var tag in tags)
+            {
+                if (tag.Key.Length > MaxDimensionNameSize)
+                {
+                    // TODO: Data Validation
```
cijothomas, line 565: this is same as existing behavior right? i.e e drop whole metric due to excess key?

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -484,6 +486,384 @@ internal static void DisableOpenTelemetrySdkMetricNameValidation()
...
+
+            if (exemplars.Length > 0)
+            {
+                MetricSerializer.SerializeByte(this.buffer, ref bufferIndex, (byte)PayloadType.Exemplars);
+
+                // Get a placeholder to add the payloadType length
+                payloadTypeStartIndex = bufferIndex;
+                bufferIndex += 2;
+
+                MetricSerializer.SerializeByte(this.buffer, ref bufferIndex, 0); // version
```
cijothomas, line 814: nit: consider moving these to a const elsewhere.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -484,6 +486,384 @@ internal static void DisableOpenTelemetrySdkMetricNameValidation()
...
+
+                var exemplarsCount = 0;
+                foreach (var exemplar in exemplars)
```
cijothomas, line 817: lets add a TODO to avoid this dual iteration, and instead backfill the count

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: I can explicitly set them to zero if the method returns false, but I do think the SDK should not change the behavior of this public API when it returns false. Also, it doesn't make sense for the SDK to assign to them any value other than the default value of zero when it returns false.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -500,6 +880,99 @@ private void SerializeHistogramBucket(in HistogramBucket bucket, ref int bufferI
...
+            MetricSerializer.SerializeUInt64(this.buffer, ref bufferIndex, Convert.ToUInt64(lastExplicitBound + 1));
+        }
+
+        MetricSerializer.SerializeUInt32(this.buffer, ref bufferIndex, Convert.ToUInt32(bucket.BucketCount));
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void SerializeExemplar(Exemplar exemplar, ref int bufferIndex)
+    {
+        MetricSerializer.SerializeByte(this.buffer, ref bufferIndex, 0); // version
```
cijothomas, line 902: this is separate version? We write a version just before calling SerializeExemplar

utpilla: Yeah we have to serialize the same version multiple times. Once for the exemplar list, and then one for each individual exemplar. The version value is going to be zero for now. It might change in future.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -500,6 +880,99 @@ private void SerializeHistogramBucket(in HistogramBucket bucket, ref int bufferI
...
+        {
+            MetricSerializer.SerializeFloat64(this.buffer, ref bufferIndex, value); // serialize double value
+        }
+
+        var bufferIndexForNumberOfLabels = bufferIndex;
+        MetricSerializer.SerializeByte(this.buffer, ref bufferIndex, 0); // serialize zero as the count of labels; this would be updated later if the exemplar has labels
+        bool hasLabels = exemplar.FilteredTags != null && exemplar.FilteredTags.Count > 0;
+        byte numberOfLabels = 0;
+
+        var unixNanoSeconds = DateTime.FromFileTimeUtc(exemplar.Timestamp.ToFileTime())
```
cijothomas, line 933: good to extract to a helper method.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -500,6 +880,99 @@ private void SerializeHistogramBucket(in HistogramBucket bucket, ref int bufferI
...
+            MetricSerializer.SerializeInt64AsBase128(this.buffer, ref bufferIndex, valueAsLong); // serialize long value
+        }
+        else
+        {
+            MetricSerializer.SerializeFloat64(this.buffer, ref bufferIndex, value); // serialize double value
+        }
+
+        var bufferIndexForNumberOfLabels = bufferIndex;
+        MetricSerializer.SerializeByte(this.buffer, ref bufferIndex, 0); // serialize zero as the count of labels; this would be updated later if the exemplar has labels
+        bool hasLabels = exemplar.FilteredTags != null && exemplar.FilteredTags.Count > 0;
```
cijothomas, line 930: nit; move hasLabels to further below, closer to the actual place where it is used.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: I don't think we can avoid it. The count field doesn't have a predefined size. It could be anywhere from 1 to 9 bytes. So we need to know the count beforehand.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: This follows the existing behavior but the existing behavior is not to drop the metric. We simply serialize the tag key to the buffer and if the whole payload length is still under 64KB, we export it. It's then up to the downstream systems to handle it.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: This is the only method that needs it 😄

cijothomas (approved):

PR #1080, opened 2023-03-14
Removing automatic service name from Lambda instrumentation

ehornby:
## Changes

Adding a `ServiceNameOverride` property to `AwsLambdaInstrumentationOptions` for instances where desired service name is different than Lambda function name.

* [X] Changes in public API reviewed


Kielek:
@Oberon00, could you please review?
@ehornby, if the changes will be approved, we need record in CHANGELOG.md.

src/OpenTelemetry.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
```diff
@@ -55,7 +55,11 @@ public static class TracerProviderBuilderExtensions
         builder.AddSource(AWSLambdaWrapper.ActivitySourceName);
         builder.SetResourceBuilder(ResourceBuilder
```
Oberon00, line 56: I don't think this is necessary or even very useful: Using the new API introduced in open-telemetry/opentelemetry-dotnet#2909 + open-telemetry/opentelemetry-dotnet#3307, one can override the service more easily already.

But: It would be good if this call here could also be changed to use  `ConfigureResource` instead of `SetResourceBuilder`, so it does not reset any other resources set before it.

Oberon00 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs
ehornby: @Oberon00 Completely agree - there are fine existing ways to override the service name as is, I'm currently using the `ConfigureResource` API in existing implementations. 

Where I saw value here was in the lack of clarity that such an override is necessary in the first place, if one required their service name to differ from the Lambda function name. [Documentation](https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/) indicates that setting the `OTEL_SERVICE_NAME` environment variable (or injecting attributes via `OTEL_RESOURCE_ATTRIBUTES`) should set the `service.name` attribute as desired, but that didn't seem to be what happened in practice. 

I didn't find it immediately clear anywhere that behavior is different with Lambdas using this package, so this change wasn't intended to add functionality so much as just clarity.

Oberon00: This is currently still in beta, so we may break things. I think that this overrides an explicitly set OTEL_SERVICE_NAME is a bug. IMHO, it would be good to completely remove setting the service name here, and we can maybe later add it back under a dedicated off-by-default option (or even just provide an API that returns the function name for users to set as service name themselves)

ehornby: It certainly seems more like a bug than intended behaviour, for sure. 

I'm happy to repurpose this PR to remove the functionality that sets the service name, if you'd like?

Oberon00: I think that would be a better solution for the problem you describe. Thank you!

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   ([#626](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/626))
 * Removes `AddAWSLambdaConfigurations` method with default configure parameter.
   ([#943](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/943))
+* BREAKING (behavior): `AddAWSLambdaConfigurations` no longer initially calls `AddService`
```
Oberon00, line 9:
```suggestion
* BREAKING (behavior): `AddAWSLambdaConfigurations` no longer calls `AddService` 
```

Oberon00 (approved):

Kielek (approved):
LGTM, per discussion in this PR.

PR #1087, opened 2023-03-16
Add xray context as link, not parent

johnbley:
## Changes

Per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-lambda.md#aws-x-ray-environment-span-link we should be adding the AWS X-ray context as a span link, not setting the parent context.


Kielek:
@johnbley, please add record to CHANGELOG.md

Oberon00:
This spec  does not say that no parent should be created. It just says that a link should be created. IMHO, the X-Ray parent may be helpful in many cases. What should probably be changed is to _additionally_ add a link even if a different propagation mechanism was used for the parent (ideally with option to disable).

Oberon00:
A clarification would be appreciated, but actually I think it is very strange to forbid using X-Ray as parent. In case it isn't clear: That environment variable AFAIK is reset by the AWS Lambda runtime on every request and tells you the parent span. So it is not really some kind of "background context" but actually the parent you want if you use X-Ray.

tsloughter:
@Oberon00 since it is an environment variable I don't think it can be updated for each event/request. It can only be set on start, while executions of the handler can happen many times for each start.

Oberon00:
@tsloughter You are mistaken. The environment variable is updated in-process based on information read from the request. This is also explained in the docs https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime:

> _X_AMZN_TRACE_ID – The [X-Ray tracing header](https://docs.aws.amazon.com/lambda/latest/dg/services-xray.html). This environment variable is not defined for custom runtimes (for example, runtimes that use the provided or provided.al2 identifiers). You can set _X_AMZN_TRACE_ID for custom runtimes using the Lambda-Runtime-Trace-Id response header from the [Next invocation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-next).

E.g. here is the .NET implementation:

* Setter for the envvar: https://github.com/aws/aws-lambda-dotnet/blob/83fdc802f2b8a6db212600a1cbc0c83fac259076/Libraries/src/Amazon.Lambda.RuntimeSupport/Context/LambdaEnvironment.cs#L74-L77
* Call to the setter when a LambdaContext is constructed (per-request argument to the handler): https://github.com/aws/aws-lambda-dotnet/blob/83fdc802f2b8a6db212600a1cbc0c83fac259076/Libraries/src/Amazon.Lambda.RuntimeSupport/Context/LambdaContext.cs#L54

tsloughter:
@Oberon00 thanks. Well that just confuses me even more, haha. I was waiting on a response to find out in what cases it sets it. I'm guessing my spec update will be to say that it should be used as a link and optionally as the parent. AWS and Lambda are blowing my mind with the many ways context can be propagated :)

Oberon00:
Well yes. It would be simpler to not introduce the link, and if somebody wants X-Ray, they use it as parent, as it is now in this implementation.

srprash:
Hello @tsloughter and @Oberon00 
Sorry for being late to the party. 😉 

I can give a bit of background on how tracing works when a Lambda function is invoked. If there is an incoming trace context on the function invocation, then this trace context is set as the `_X_Amzn_Trace_Id` environment variable. This trace context is in X-Ray format and contains the traceId, parentId, and sampling flag. 
This trace context from the environment variable should be the parent for any span created within the function. And this is the expected behavior for the X-Ray users. Adding this context as a link instead of the parent would likely result in a broken trace for these users. It may work fine for the tracing solutions that can process links.
I think Christian's [suggestion](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1087#issuecomment-1473541471) is probably the way to go here.

Question: If the incoming context from the environment variable is set only as a link, then what would be the parent context? What if the link and parent context are different in some cases?

Oberon00:
@tsloughter As you were involved in the design of this, maybe you want to answer that question?

> Question: If the incoming context from the environment variable is set only as a link, then what would be the parent context? What if the link and parent context are different in some cases?

I would think that there would be no parent in that case, and that the link and parent context are different in some (or even most) cases would be expected and normal.

Oberon00:
> Yes, my understanding is they will always be different but share a trace id. 

I don't think they usually will. Consider a Lambda called via a function URL (or API gateway). The parent will be read from the HTTP traceparent+tracestate headers contained in the event. However, as far as I know, X-Ray will not interpret these headers and thus will create an independent root span that it injects into the X-Ray envvar.

I don't know if that assumption is true. If X-Ray does interpret the HTTP headers (either in the API gateway or Lambda monitoring), then they might have the same trace ID.

johnbley:
OK, where did the specification discussion here land?  Happy to update this in any direction that is generally agreed on.

tsloughter:
This isn't stale.

Working on an update to the spec and then this PR will be updated.

tylerbenson:
Looks like this got lost.  Is there still confusion in the spec?  To me it sounds like this was implementing the updated spec correctly.

Regarding @Oberon00's discussion above...
* if there is an existing trace propagated from the remote call, then the span link would share the same trace id
* if there is no existing trace, xray would start a new trace and otel would start a separate trace linking to xray's span.

PR #1092, opened 2023-03-18
[Extensions.AWSXray] Replace Newtonsoft.Json with System.Text.Json

wjrogers:
## Changes

Replace OpenTelemetry.Contrib.Extensions.AWSXRay's Newtonsoft.Json dependency with System.Text.Json. When targeting newer runtimes, this effectively removes the dependency because System.Text.Json is part of the base class library.

There are no changes to the public API, and test coverage of the affected code looks good.


src/OpenTelemetry.Contrib.Extensions.AWSXRay/OpenTelemetry.Contrib.Extensions.AWSXRay.csproj
Kielek, line 1: Changes should be documented in CHANGELOG.md

srprash (approved):

Kielek (approved):

PR #913, opened 2023-01-20
[Instrumentation.Process, Instrumentation.Process] make configure action nullable

Kielek:
Follow up #893 #892

## Changes

make configure action nullable
based on https://github.com/open-telemetry/opentelemetry-dotnet/pull/3653/files#r970049254 and conversation with @alanwest 

For significant contributions please make sure you have completed the following items:

* ~~[ ] Appropriate `CHANGELOG.md` updated for non-trivial changes~~
* ~~[ ] Design discussion issue #~~

PR #1094, opened 2023-03-20
[Instrumentation.MySqlData] Nullable

Kielek:
Towards #894.

## Changes

[Instrumentation.MySqlData] Nullable.
Part of shared files also updated.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.MySqlData/TracerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.MySqlData/TracerProviderBuilderExtensions.cs
@@ -41,7 +41,7 @@ public static class TracerProviderBuilderExtensions
     /// <returns>The instance of <see cref="TracerProviderBuilder"/> to chain the calls.</returns>
     public static TracerProviderBuilder AddMySqlDataInstrumentation(
         this TracerProviderBuilder builder,
-        Action<MySqlDataInstrumentationOptions> configure)
+        Action<MySqlDataInstrumentationOptions>? configure)
     {
         Guard.ThrowIfNull(builder);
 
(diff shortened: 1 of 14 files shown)


moonheart:
This LGTM 👍

src/OpenTelemetry.Instrumentation.MySqlData/TracerProviderBuilderExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static class TracerProviderBuilderExtensions
     /// <returns>The instance of <see cref="TracerProviderBuilder"/> to chain the calls.</returns>
     public static TracerProviderBuilder AddMySqlDataInstrumentation(
         this TracerProviderBuilder builder,
-        Action<MySqlDataInstrumentationOptions> configure)
+        Action<MySqlDataInstrumentationOptions>? configure)
```
utpilla, line 44: We should probably not put `?` to stay consistent with other instrumentation libraries.

Look at this for reference: https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/TracerProviderBuilderExtensions.cs#L59-L71

Kielek: I have discussion with @alanwest some time ago. The consensus is to make it nullable non-optional.

See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/913

utpilla: Ah okay. Thanks for sharing that.

PR #1101, opened 2023-03-23
[Extensions.Docker] Nullable

Kielek:
Towards #894

## Changes

Nullable for docker and tests.

### DIFF
--- a/src/OpenTelemetry.Extensions.Docker/.publicApi/net462/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Extensions.Docker/.publicApi/net462/PublicAPI.Shipped.txt
@@ -1,0 +1,1 @@
+#nullable enable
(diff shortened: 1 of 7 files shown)


Kielek:
@iskiselev, it will be great if you can check.

cijothomas (approved):

src/OpenTelemetry.Extensions.Docker/.publicApi/net462/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1 @@
+#nullable enable
```
iskiselev, line 1: I don't know much about PublicAPI.Shipped.txt and PublicAPI.Unshipped.txt, but it looks strange that there is
```
nullable enabled
```
here, but not in Unshipped. It feels that it should be either in both or in none.

iskiselev:
Added one question, but LGTM otherwise.

src/OpenTelemetry.Extensions.Docker/.publicApi/net462/PublicAPI.Shipped.txt
Kielek: It is the way how VS is generating it.
It is enough to make it in one place and usually we put it to Shipped.

PR #1091, opened 2023-03-17
[Sampler.AWS] AWS X-Ray remote sampler part 1

srprash:
**New feature: AWS X-Ray remote sampler.**

I am going to create a series of PRs for adding the X-Ray remote sampling support for OTel via the new`OpenTelemetry.Sampler.AWS` package.
For more details on the X-Ray remote sampling, please refer to this doc: https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-sampling

## Changes

This PR is the first one in the series and introduces the following:
- A basic builder and concrete class for the `AWSXRayRemoteSampler`. The builder provides the configuration for the polling interval and the Otel collector [xray proxy](https://aws-otel.github.io/docs/components/x-ray-receiver#proxy-server-configuration) endpoint.
- The `AWSXRayRemoteSampler` overrides the `ShouldSample` method but currently throws an `NotImplementedException` exception when an activity/span is started.
- The `AWSXRayRemoteSampler` upon initialization spins up a scheduled thread to fetch sampling rules from the X-Ray service. The sampling rules fetched are not stored anywhere in this current PR, but further implementation will build up on this logic.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #1124, opened 2023-03-30
[Sampler.AWS] Part-2: Add rules cache and rule matching logic

srprash:
This PR is a continuation of building the AWS X-Ray Remote Sampler.
See part 1 PR: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1091

## Changes
- Adding a Sampling Rules Cache. 
  - This cache holds the sampling rules which are periodically fetched from the X-Ray service.
  - When a new set of sampling rules are fetched, the stateful properties like the `Reservoir` and the `Statistics` are preserved from the old set of rules. These stateful properties will be used in making the sampling decisions and keeping the sampling records respectively throughout the duration of the application. Note: The `Reservoir` and `Statistics` classes are very bare-bone right now, but will be developed in further PRs.
  - The cache is also responsible for finding the sampling rule that matched a particular request. This sampling rule will be used to make the final sampling decision.

- Adding the sampling rule matching logic.
  - When an activity/span is created, the sampler will try and match a sampling rule based on the attributes like service name, http method, host, etc.
  - A rule definition can specify matching of these attributes using wildcard characters.

- The sampler now requires an OpenTelemetry Resource object.
  - The sampler allows users to specify service name and service type (for example AWS::EC2::Instance) in their sampling rule to control sampling on a service level. These attributes are currently obtainable from the OTel SDK `Resource` configured for the application.
  - There is an [open ask](https://github.com/open-telemetry/opentelemetry-specification/issues/1588) to make the OTel SDK Resource available to samplers. In the meantime, the users will have to provide the `Resource` instance to the sampler themselves.


src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
```diff
@@ -43,20 +50,41 @@ internal AWSXRayRemoteSampler(TimeSpan pollingInterval, string endpoint)
...
+
     /// <summary>
     /// Initializes a <see cref="AWSXRayRemoteSamplerBuilder"/> for the sampler.
     /// </summary>
+    /// <param name="resource">an instance of <see cref="Resources.Resource"/>
+    /// to identify the service attributes for sampling. This resource should
+    /// be the same as what the OpenTelemetry SDK is configured with.</param>
     /// <returns>an instance of <see cref="AWSXRayRemoteSamplerBuilder"/>.</returns>
-    public static AWSXRayRemoteSamplerBuilder Builder()
+    public static AWSXRayRemoteSamplerBuilder Builder(Resource resource)
```
Kielek, line 64: I am not sure if Resource is the best way how to configure this values.
Based on what I see, you need Service name and cloud platform.
Can it be passed here?

srprash: My aim of using a `Resource` here is the ease of use. Typically, an OTel user would configure the service name through either environment variable or a Resource. Also, we have a bunch of AWS resource detectors which can automatically detect the cloud platform. So I think passing in a resource here is better than explicitly passing service name and cloud platform values. 
Also, this way the sampler behavior will be consistent across other OTel SDK. For example OTel Java SDK: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java#L67

src/OpenTelemetry.Sampler.AWS/CHANGELOG.md
```diff
@@ -6,3 +6,4 @@ Initial release of `OpenTelemetry.Sampler.AWS`.
 
 * Feature - AWSXRayRemoteSampler - Add support for AWS X-Ray remote sampling
   ([#1091](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1091))
+  ([#1124](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1124))
```
Kielek, line 9:
```suggestion
  ([#1091](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1091),
   [#1124](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1124))
```

src/OpenTelemetry.Sampler.AWS/Clock.cs
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+
+namespace OpenTelemetry.Sampler.AWS;
+
+// A time keeper for the purpose of this sampler.
+internal sealed class Clock
```
Kielek, line 22: What is a benefit for using this additional abstraction layer, instread of just using directly DateTime.UtcNow?

src/OpenTelemetry.Sampler.AWS/Matcher.cs
```diff
@@ -0,0 +1,144 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace OpenTelemetry.Sampler.AWS;
+
+internal static class Matcher
+{
+    public static readonly Dictionary<string, string> XRayCloudPlatform = new Dictionary<string, string>()
```
Kielek, line 26:
```suggestion
    public static readonly IReadOnlyDictionary<string, string> XRayCloudPlatform = new Dictionary<string, string>()
```

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
```diff
@@ -43,20 +50,41 @@ internal AWSXRayRemoteSampler(TimeSpan pollingInterval, string endpoint)
...
     public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
     {
-        // TODO: add the actual functionality for sampling.
-        throw new System.NotImplementedException();
+        if (this.RulesCache.Expired())
+        {
+            return this.FallbackSampler.ShouldSample(samplingParameters);
+        }
+
+        SamplingRule? matchedRule = this.RulesCache.MatchRule(samplingParameters, this.Resource);
```
Kielek, line 77: SamplingParameters are passed as `in` reference. Is it worth to propagate this style for other methods?
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier

src/OpenTelemetry.Sampler.AWS/README.md
```diff
@@ -19,16 +19,28 @@ X-Ray remote sampling](https://aws-otel.github.io/docs/getting-started/remote-sa
...
+var resourceBuilder = ResourceBuilder
+    .CreateDefault()
+    .AddService(serviceName: serviceName)
+    .AddDetector(new AWSEC2ResourceDetector());
+
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(serviceName)
+    .SetResourceBuilder(resourceBuilder)
+    .AddConsoleExporter()
+    .AddOtlpExporter()
```
Kielek, line 38: Remove exporter from code listing or add some description similar to https://github.com/open-telemetry/opentelemetry-dotnet/blob/a61711d0c37346acfe0f3b096c6ac5a919c9c155/src/OpenTelemetry.Instrumentation.AspNetCore/README.md?plain=1#L52-L54

src/OpenTelemetry.Sampler.AWS/Statistics.cs
```diff
@@ -0,0 +1,68 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System;
+using System.Threading;
+
+namespace OpenTelemetry.Sampler.AWS;
+
+internal class Statistics : IDisposable
```
Kielek, line 22: For not, I do not see any meaningful usage for this. I suppose part 3 brings this?

test/OpenTelemetry.Sampler.AWS.Tests/TestMatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace OpenTelemetry.Sampler.AWS.Tests;
+
+public class TestMatcher
+{
+    [Fact]
+    public void TestWildcardMatching()
```
Kielek, line 25: Consider to use InLineData https://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/

adn have two separate test methods - WildcardMatch and WildcardDoesNotMatch.

test/OpenTelemetry.Sampler.AWS.Tests/TestMatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        Assert.True(Matcher.WildcardMatch(null, "*"));
```
Kielek, line 27: I am not sure that null should match anything. I think it needs some documentation for this case.

Kielek (commented):
It will be great to review code also by AWS experienced guys.

Maybe @Oberon00?

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
Kielek: @utpilla, FYI

Oberon00:
Sorry, I don't have any experience with X-Ray, so this PR is out of my area of expertise.

test/OpenTelemetry.Sampler.AWS.Tests/Utils.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        var resourceAttributes = new List<KeyValuePair<string, object>>()
+        {
+            new KeyValuePair<string, object>("service.name", serviceName),
+            new KeyValuePair<string, object>("cloud.platform", cloudPlatform),
+        };
+
+        return ResourceBuilder.CreateEmpty().AddAttributes(resourceAttributes).Build();
+    }
+}
```
atshaw43, line 60: What do you think about testing the locking mechanism? I think we are fine, but there are concerns about race conditions and dead locks.

Maybe simulate a few hundred/thousand changes here from a mocked backend service?

atshaw43 (changes requested):

src/OpenTelemetry.Sampler.AWS/FallbackSampler.cs
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.Sampler.AWS;
+
+internal class FallbackSampler : Trace.Sampler
```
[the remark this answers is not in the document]
srprash: This implementation of the FallbackSampler is just here for placeholder. It will eventually be a rate limited sampler with 1 req/sec and 5% of remaining requests. 
I will update it in the next and final PR.

test/OpenTelemetry.Sampler.AWS.Tests/Utils.cs
srprash: I realized that there were a lot of mutable states in this code which require careful locking. So in my recent code change, I got rid of some of those mutable states and instead create entirely new objects. This should reduce some of the possible race conditions.

src/OpenTelemetry.Sampler.AWS/Clock.cs
srprash: The sampler does quite a few things based on time, like the rules cache freshness and rate limited sampling (which would come in the next PR. The DateTime APIs can be used for such logic, but I think for testing the functionality, I need a mocked clock which can be passed in as a `Clock` type.

src/OpenTelemetry.Sampler.AWS/Statistics.cs
srprash: Yes. The `Statistics` class instance maintains the counter of the requests sampled. I will expand on it's use in the next PR.

src/OpenTelemetry.Sampler.AWS/SamplingRule.cs
```diff
@@ -72,14 +85,91 @@ internal class SamplingRule : IComparable<SamplingRule>
...
+
+        string serviceName = (string)resource.Attributes.FirstOrDefault(kvp =>
+                kvp.Key.Equals("service.name", StringComparison.Ordinal)).Value;
+
+        return Matcher.AttributeMatch(samplingParameters.Tags, this.Attributes) &&
+               Matcher.WildcardMatch(httpTarget, this.UrlPath) &&
+               Matcher.WildcardMatch(httpMethod, this.HttpMethod) &&
+               Matcher.WildcardMatch(httpHost, this.Host) &&
+               Matcher.WildcardMatch(serviceName, this.ServiceName) &&
+               Matcher.WildcardMatch(GetServiceType(resource), this.ServiceType);
```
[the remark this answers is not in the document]
srprash: Good catch. I missed it. I'll add the comparison for ResourceARN as well.

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
srprash: Sure.

test/OpenTelemetry.Sampler.AWS.Tests/TestMatcher.cs
```diff
@@ -0,0 +1,86 @@
...
+        Assert.True(Matcher.WildcardMatch(string.Empty, "*"));
+        Assert.True(Matcher.WildcardMatch("HelloWorld", "*"));
+
+        Assert.False(Matcher.WildcardMatch(null, "Hello*"));
+        Assert.False(Matcher.WildcardMatch("HelloWorld", null));
+
+        Assert.True(Matcher.WildcardMatch("HelloWorld", "HelloWorld"));
+        Assert.True(Matcher.WildcardMatch("HelloWorld", "Hello*"));
+        Assert.True(Matcher.WildcardMatch("HelloWorld", "*World"));
+        Assert.True(Matcher.WildcardMatch("HelloWorld", "?ello*"));
```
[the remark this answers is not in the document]
srprash: We don't need really complex testing here since the sampling rule options only allow for `*` and `?` for matching.
https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html#xray-console-sampling-options

src/OpenTelemetry.Sampler.AWS/SamplingRule.cs
```diff
@@ -72,14 +85,91 @@ internal class SamplingRule : IComparable<SamplingRule>
...
+        string? httpTarget = null;
+        string? httpUrl = null;
+        string? httpMethod = null;
+        string? httpHost = null;
+
+        if (samplingParameters.Tags is not null)
+        {
+            foreach (var tag in samplingParameters.Tags)
+            {
+                if (tag.Key.Equals(SemanticConventions.AttributeHttpTarget, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
srprash: The `samplingParameters.Tags` is not actually a dictionary rather an `IEnumerable<KeyValuePair>` type, so I can't use `samplingParameters.Tags.TryGetValue`.

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
srprash: done

test/OpenTelemetry.Sampler.AWS.Tests/TestMatcher.cs
srprash: a `null` input here would mean that the particular attribute is not set on the span, and so the `*` should match any value of an attribute whether it is set or not.

src/OpenTelemetry.Sampler.AWS/Matcher.cs
```diff
@@ -0,0 +1,144 @@
...
+            return text.Length == 0;
+        }
+
+        // it is faster to check if we need a regex comparison than
+        // doing always regex comparison, even where we may not need it.
+        foreach (char c in globPattern)
+        {
+            if (c == '*' || c == '?')
+            {
+                return Regex.IsMatch(text, ToRegexPattern(globPattern));
```
atshaw43, line 58: I do have concerns here about regex being slow. Especially since we are using a simplified version. I will leave it up to you on if we should keep it.

atshaw43 (approved):

src/OpenTelemetry.Sampler.AWS/Matcher.cs
srprash: Yes. A performance benchmark would be a good idea. I can think of adding it once the sampler implementation is complete.

Kielek (approved):
I still have some doubts about public API. @utpilla, could you please review it?

Other parts LGTM.

src/OpenTelemetry.Sampler.AWS/README.md
```diff
@@ -15,20 +15,34 @@ dotnet add package OpenTelemetry.Sampler.AWS
...
+var resourceBuilder = ResourceBuilder
+    .CreateDefault()
+    .AddService(serviceName: serviceName)
+    .AddDetector(new AWSEC2ResourceDetector());
+
+using var tracerProvider = Sdk.CreateTracerProviderBuilder()
+    .AddSource(serviceName)
+    .SetResourceBuilder(resourceBuilder)
+    .AddConsoleExporter()
+    .SetSampler(AWSXRayRemoteSampler.Builder(resourceBuilder.Build()) // you must provide a resource
```
utpilla, line 41: Just a suggestion:

This setup looks quite verbose. I think you could extract this into an extension method. Something like: `AddAWSXRayRemoteSampler(this TracerProviderBuilder, Action<AWSXRaySamplerOptions> configure)`.

That way you would only be exposing the required options and you could reduce the publicAPI surface by marking these methods as internal: `AWSXRayRemoteSampler.Builder`, `SetPollingInterval`, `SetEndpoint`, `Build`

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
```diff
@@ -25,38 +28,70 @@ namespace OpenTelemetry.Sampler.AWS;
...
     }
 
     /// <inheritdoc/>
     public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
     {
-        // TODO: add the actual functionality for sampling.
-        throw new System.NotImplementedException();
+        if (this.RulesCache.Expired())
+        {
+            return this.FallbackSampler.ShouldSample(in samplingParameters);
```
atshaw43, line 91: Is a warning log generated somewhere when the cache expires and we use the fallback?

This could be very useful in debugging with a customer.

src/OpenTelemetry.Sampler.AWS/README.md
srprash: I had earlier of thought of doing the extension method similar to `AddXRayTraceId()` we have for the custom trace id generator, but I think that is counterintuitive for adding a Sampler. Since the `AWSXRayRemoteSampler` is just another implementation of `Sampler`, it should follow the standard procedure of adding a sampler via the `SetSampler` method.

Regarding the setting of the sampler being verbose, I agree. The customers of OTel [Java](https://aws-otel.github.io/docs/getting-started/java-sdk/trace-manual-instr#using-x-ray-remote-sampling) and [GoLang](https://aws-otel.github.io/docs/getting-started/go-sdk/trace-manual-instr#sampling-using-aws-x-ray-remote-sampler) are already configuring the remote sampler in similar way, so I would like to keep the same behavior here as well.

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSampler.cs
srprash: Not yet. I can add one.

src/OpenTelemetry.Sampler.AWS/RulesCache.cs
```diff
@@ -0,0 +1,138 @@
...
+        this.Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    internal DateTime GetUpdatedAt()
+    {
+        this.rwLock.EnterReadLock();
+        try
+        {
+            return this.UpdatedAt;
```
atshaw43, line 123: Is locking around returning a variable needed?

Also, I don't think GetUpdatedAt is used.

atshaw43 (approved):

PR #1175, opened 2023-05-09
Fix indexing issue in GetSanitizedCategoryName

### LINKED ISSUE #1174: TldLogExporter event name sanitization issues
During event name sanitization in TldLogExporter when strings contain non alphanumeric characters the output is not correct or throws IndexOutOfRangeException.

* OpenTelemetry.Exporter.Geneva 1.4.0+

* All

* [x] Bug

All non-alphanumeric characters removed from given string, returning resulting string up to MaxSanitizedEventNameLength.

Given the code below:
- Debug.Fail was called. Expected: S12 but found S`00`1
- Index was outside the bounds of the array.

```cs
void Main()
{
    var arrange = new[] {
        ("S_12", "S12"),
        ("S_123", "S123"),
        ("S_12_34", "S123"),
    };
    
    foreach ((string,string) x in arrange)
    {
        try
        {
            var result = GetSanitizedCategoryName(x.Item1);

            Debug.Assert(result == x.Item2, $"Expected: {x.Item2} but found {result}");
        } catch (Exception e) {
            Console.WriteLine(e.Message);
        }
    }
}

const int MaxSanitizedEventNameLength = 4;

private static string GetSanitizedCategoryName(string categoryName)
{
    int validNameLength = 0;
    Span<char> result = stackalloc char[MaxSanitizedEventNameLength];

    // Special treatment for the first character.
    var firstChar = categoryName[0];
    if (firstChar >= 'A' && firstChar <= 'Z')
    {
        result[0] = firstChar;
        ++validNameLength;
    }
    else if (firstChar >= 'a' && firstChar <= 'z')
    {
        // If the first character in the resulting string is a lower-case alphabet,
        // it will be converted to the corresponding upper-case.
        result[0] = (char)(firstChar - 32);
        ++validNameLength;
    }
    else
    {
        // Not a valid name.
        return string.Empty;
    }

    for (int i = 1; i < categoryName.Length; i++)
    {
        if (validNameLength == MaxSanitizedEventNameLength)
        {
            break;
        }

        var cur = categoryName[i];
        if ((cur >= 'a' && cur <= 'z') || (cur >= 'A' && cur <= 'Z') || (cur >= '0' && cur <= '9'))
        {
            result[i] = cur;
            ++validNameLength;
        }
    }

    return result.Slice(0, validNameLength).ToString();
}
```

dariusclay:
Small trivial update. Fixes #1174.

## Changes

Change variable used to index result string.

* [x] Design discussion issue #

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/TLDExporter/TldLogExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/TLDExporter/TldLogExporter.cs
@@ -476,7 +476,7 @@ internal sealed class TldLogExporter : TldExporter, IDisposable
             var cur = categoryName[i];
             if ((cur >= 'a' && cur <= 'z') || (cur >= 'A' && cur <= 'Z') || (cur >= '0' && cur <= '9'))
             {
-                result[i] = cur;
+                result[validNameLength] = cur;
                 ++validNameLength;
             }
         }


Kielek:
@dariusclay, is it worth to add some test to cover this scenario? Just to prevent regression.

utpilla (approved):

utpilla:
@dariusclay Thanks for creating this PR!

PR #1182, opened 2023-05-10
Add azure vm resource detector

vishweshbankwar:
Fixes #.

## Changes


src/OpenTelemetry.ResourceDetectors.Azure/AzureVMResourceDetector.cs
```diff
@@ -0,0 +1,83 @@
...
+namespace OpenTelemetry.ResourceDetectors.Azure;
+
+/// <summary>
+/// Resource detector for Azure VM environment.
+/// </summary>
+public class AzureVMResourceDetector : IResourceDetector
+{
+    internal static readonly IReadOnlyCollection<string> ExpectedAzureAmsFields = new string[]
+    {
+        "azInst_location",
```
rajkumar-rangaraj, line 30: Shall we introduce a AzureConstants or AzureConventions to group these string.

vishweshbankwar: Yes, let me do that as follow up. Would do it for AppService too.

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetaDataRequestor.cs
```diff
@@ -0,0 +1,40 @@
...
+// </copyright>
+
+using System.Net.Http;
+using System.Text.Json;
+
+namespace OpenTelemetry.ResourceDetectors.Azure;
+
+internal class AzureVmMetaDataRequestor : IAzureVmMetaDataRequestor
+{
+    private const string AMSURL = "http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json";
```
rajkumar-rangaraj, line 24:
```suggestion
    private const string AzureMetadataEndpointURL = "http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json";
```

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetaDataRequestor.cs
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Net.Http;
+using System.Text.Json;
+
+namespace OpenTelemetry.ResourceDetectors.Azure;
+
+internal class AzureVmMetaDataRequestor : IAzureVmMetaDataRequestor
```
utpilla, line 22: nit: Mark internal class sealed.

src/OpenTelemetry.ResourceDetectors.Azure/OpenTelemetry.ResourceDetectors.Azure.csproj
```diff
@@ -10,6 +10,7 @@
 
     <ItemGroup>
         <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
+        <PackageReference Include="System.Text.Json" Version="7.0.2" />
```
rajkumar-rangaraj, line 13: Do we need this package?
If it is needed for .NET Framework, shall we add it only for it?

vishweshbankwar: we target netstandard2.0 only at the moment so this is needed.

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetaDataRequestor.cs
```diff
@@ -0,0 +1,40 @@
...
+// </copyright>
+
+using System.Net.Http;
+using System.Text.Json;
+
+namespace OpenTelemetry.ResourceDetectors.Azure;
+
+internal sealed class AzureVmMetaDataRequestor : IAzureVmMetaDataRequestor
+{
+    private const string AzureVmMetadataEndpointURL = "http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json";
```
rajkumar-rangaraj, line 24: hard-coding api version here, looks incorrect to me. Should we have an option to pass API version in the constructor as optional parameter. It' expected that these APIs are backward compactible and should not break our implementation.

vishweshbankwar: Leaving an option to specify API version means we have to make sure our deserialization works for all of them (depending on the data format that is returned by specific API version). I am not sure if we should allow that. In future we could just update this version to use the latest.

vishweshbankwar: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=windows#versioning

rajkumar-rangaraj (approved):

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetadataResponse.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    [JsonPropertyName("vmScaleSetName")]
+    public string? VmScaleSetName { get; set; }
+
+    [JsonPropertyName("vmSize")]
+    public string? VmSize { get; set; }
+
+    internal string GetValueForField(string fieldName)
+    {
+        string? aimsValue = null;
```
rajkumar-rangaraj, line 58: nit: avoid variable name starting with ai.

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetaDataRequestor.cs
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// </copyright>
+
+using System.Net.Http;
+using System.Text.Json;
+
+namespace OpenTelemetry.ResourceDetectors.Azure;
+
+internal sealed class AzureVmMetaDataRequestor : IAzureVmMetaDataRequestor
```
utpilla, line 22: This can be a static class. Is the interface added just for the unit test? If yes, then we can do the unit test even without the interface.

src/OpenTelemetry.ResourceDetectors.Azure/AzureVMResourceDetector.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// <inheritdoc/>
+    public Resource Detect()
+    {
+        List<KeyValuePair<string, object>> attributeList = new();
+        try
+        {
+            var vmMetaDataResponse = this.azureVmMetaDataRequestor.GetAzureVmMetaDataResponse();
+            if (vmMetaDataResponse != null)
+            {
+                foreach (var field in ExpectedAzureAmsFields)
```
utpilla, line 69: nit: Move the initialization of `attributeList` here.

```csharp
var attributeList = new List<KeyValuePair<string, object>>(ExpectedAzureAmsFields.Count);
```

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetaDataRequestor.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    private const string AzureVmMetadataEndpointURL = "http://169.254.169.254/metadata/instance/compute?api-version=2021-12-13&format=json";
+
+    public AzureVmMetadataResponse? GetAzureVmMetaDataResponse()
+    {
+        using var httpClient = new HttpClient();
+
+        httpClient.DefaultRequestHeaders.Add("Metadata", "True");
+        var res = httpClient.GetStringAsync(AzureVmMetadataEndpointURL);
```
utpilla, line 31: We should probably be using `ConfigureAwait(false)` and `GetAwaiter().GetResult()`.

https://learn.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development#:~:text=Figure%206%20Use%20HttpClient%20with%20ConfigureAwait(false)%20to%20Prevent%20Deadlock

```csharp
try
{
    var res = httpClient.GetStringAsync(AzureVmMetadataEndpointURL).ConfigureAwait(false);
    return JsonSerializer.Deserialize<AzureVmMetadataResponse>(res.GetAwaiter().GetResult());
}
catch (Exception)
{
    return null;
}
```

vishweshbankwar: Done

src/OpenTelemetry.ResourceDetectors.Azure/AzureVmMetadataResponse.cs
```diff
@@ -0,0 +1,104 @@
...
+                break;
+            case "azInst_resourceGroupName":
+                amsValue = this.ResourceGroupName;
+                break;
+            case "azInst_vmScaleSetName":
+                amsValue = this.VmScaleSetName;
+                break;
+        }
+
+        if (amsValue == null)
```
utpilla, line 97: nit: If you initialize `amsValue` to `string.Empty` instead of `null`, then you can get rid of this check.

utpilla (approved):

PR #1187, opened 2023-05-13
[Instrumenation.MySqlData] Refactor to use Harmony to instrument to fix compatibility in 8.0.33.

### LINKED ISSUE #1157: [Instrumentation.MySqlData] is not working with MySQL.Data 8.0.33
# Issue with OpenTelemetry.Instrumentation.MySqlData

* OpenTelemetry.Instrumentation.MySqlData 1.0.0-beta.6 also current main (eccfe8a6c1b4a5cb5710680bbe721914a2ee145e)

* net7.0, net6.0, probably also other

* [ ] Feature Request
* [x] Bug

Activities/Spans created as for 8.0.32.1 version

No Activities/Span.

Just execute application test from https://github.com/Kielek/opentelemetry-dotnet-instrumentation/commit/95c191b5d68bf1b591f50fe3041f6178c92e6239 with mysql avaialbe for port 3306.
It can be used to test also with older versions by modyfying csproj.

## Additional Context

Found by https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/2478

moonheart:
Refactor to use [Harmony](https://github.com/pardeike/Harmony/) to instrument to fix compatibility in 8.0.33.

Fixes #1157.

## Changes

`Activity.Current` uses `AsyncLocal` to store the current activity.

Mysql.Data uses full async methods in 8.0.33 which was introduced by this commit https://github.com/mysql/mysql-connector-net/commit/291c2f31fa9d7bf1b3c02bd7da19db1b66f89abd, which causes `Activity.Current` returns null.

Before this commit, the command is executed on a single thread:
``` csharp
        public Task<int> ExecuteQueryAsync()
        {
            OpenQuery();
            var r = Execute();
            CloseQuery();
            return Task.FromResult(r);
        }

        public void OpenQuery()
        {
            // Activity created here
        }
        public int Execute()
        {
            return 1;
        }

        public void CloseQuery()
        {
            // Activity can be retrieved from here
        }
```
The Activity can be retrieved because it's synchronous, no context switch happened.

After this commit, the code is asynchronous all the way through:
``` csharp
        public async Task<int> ExecuteQueryAsync()
        {
            await OpenQueryAsync();
            var r = await ExecuteAsync();
            await CloseQueryAsync();
            return r;
        }

        public async Task OpenQueryAsync()
        {
            // Activity created here
        }
        public async Task<int> ExecuteAsync()
        {
            return 1;
        }

        public async Task CloseQueryAsync()
        {
            // Activity can not be retrieved from here 
        }

```
The Activity can not be retrieved because when it left the `OpenQuery` method, the `ExecutionContext `was restored to default, and the Activity was lost.

This PR uses [Harmony](https://github.com/pardeike/Harmony/) to patch the original method and create the Activity in prefix and pass it to finalizer.


moonheart:
This PR "just works", but I think there should be a better way to instrument Mysql.Data.

moonheart:
> @moonheart, did you have chance to contact MySql.Data maintainers/owners? Maybe there will be an option for native support for Activities or some improvements related to current state?

I don't have chance to contact `MySql.Data` maintainers/owners. I'm trying to contact using their bug report system to create a feature request.

PR #1203, opened 2023-05-24
[Instrumentation.EntityFrameworkCore] Add Filter public API to enable filtering

akoken:
## Changes

This PR introduces the addition of a `Filter` public API, which enables filtering out activities based on the properties of the
`IDbCommand` object being executed. The `Filter` API provides users with the ability to apply custom filters to only collect specific activities, enhancing the flexibility and versatility of the system.


Kielek:
Did you considered using`EnrichWithIDbCommand`? It should allow you to modify activity in context of IDbCommand.

akoken:
Yes, we can modify activity with `EnrichWithIDbCommand`, but I don't think it's a right place to do that. `Encrich` and `Filter` should have their own purpose.

Kielek:
@akoken, I have checked it internally with other maintainers.

Please consider following `Filter` option `Func<string, IDbCommand, bool>` instead of `Func<IDbCommand, bool>`.

The string parameter could be potentially providerName.

@utpilla, @alanwest FYI

akoken:
@Kielek, I added `providerName` to the filter function and fixed the markdown issues.

Kielek (approved):

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/EntityFrameworkInstrumentationOptions.cs
```diff
@@ -44,4 +44,25 @@ public class EntityFrameworkInstrumentationOptions
     /// <para><see cref="IDbCommand"/>: db command to allow access to command.</para>
     /// </remarks>
     public Action<Activity, IDbCommand> EnrichWithIDbCommand { get; set; }
+
+    /// <summary>
+    /// Gets or sets a filter function that determines whether or not to
+    /// collect telemetry about a command.
```
utpilla, line 50:
```suggestion
    /// collect telemetry about a command from a provider name.
```

utpilla (approved):

PR #1213, opened 2023-05-31
Fix unit test

utpilla:
## Changes
- Fix a unit test failing the build
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/5126743062/jobs/9221624195#step:7:198


Kielek (approved):
Approving, but I think that some additional rework should be done for this test.

@srprash, could you please check and fix the test?

PR #1214, opened 2023-06-01
[Instrumentation.Runtime] Update CHANGELOG for 1.4.0 release

utpilla:
This is to avoid confusion and make it easy for users to understand what version of core packages this would work with.

## Changes
- Update CHANGELOG for 1.4.0 release


cijothomas (approved):

Kielek:
I am not fully against this, but it might be tricky to achieve common versioning across all packages.
Especially if component owner decide to introduce breaking changes and release 2.0 version in different timeframe than OTel API/SDK.

If it should be the only package with synced versions - why it is more important than the Process instrumentation, on any other package?

cijothomas:
This is already a stable package.

Separately, we should see if we can achieve same versioning for a set of curated packages. Probably all instrumentations maintained in the main repo as it may not be feasible for instrumentations from contrib repo with various owners.
(the runtime instrumentation can be maintained in the main repo as well.)

xiang17:
Maybe it'll be good to put a note in the changelog to explicitly point out the version bump solely for purpose of the version sync?

xiang17 (approved):

Kielek (approved):

PR #875, opened 2023-01-09
Add ECS and Logs attributes based on Metadata v4 endpoint

mmanciop:
## Changes

Implement `aws.ecs` and `aws.log` resource attributes with data from the ECS Metadata endpoint v4.

* [X] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #1224, opened 2023-06-07
* [Instrumentation.AWSLambda] Request release of 1.1.0-beta.3

rypdal:
@open-telemetry/dotnet-contrib-maintainers: this is a request for release of version 1.1.0-beta.3 of the OpenTelemetry.Instrumentation.AWSLambda for which I am an owner.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.1.0-beta.3
+
+Released 2023-Jun-14
+
 * Add HTTP server span attributes for API Gateway triggers
   ([#626](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/626))
 * Removes `AddAWSLambdaConfigurations` method with default configure parameter.


Kielek:
Current set of changes under OpenTelemetry.Contrib.Instrumentation.AWS are considered as breaking changes. Before raising version it should be renamed to OpenTelemetry.Instrumentation.AWS. Similar thing should be done for OpenTelemetry.Contrib.Extensions.AWSXRay (I am not sure how this packages should be called, OpenTelemetry.Propagator.AWSXRay?)

Can you create PR for this?

utpilla (approved):

Kielek:
@rypdal, do you want to have a release this package washout renaming and releasing dependencies - mainly `OpenTelemetry.Contrib.Extensions.AWSXRay`?

rypdal:
Hi @Kielek. You mention breaking changes in OpenTelemetry.Contrib.Instrumentation.AWS. But does it force somehow renaming or it should be enough just to rise the major release version ? Can't we do the renaming(s) in a special separate PR ?
But generally speaking I can create PR for both packages washout and renaming if owners (@srprash and @atshaw43) have no objections. I guess it should/can be split into 3 parts: 
1. **OpenTelemetry.Contrib.Extensions.AWSXRay** -> **OpenTelemetry.Propagator.AWSXRay** (looks like a good package name as the main functionality offered by the package is the propagator) -> **create and publish** 
2. Update of AWSXRay package in the current PR for **OpenTelemetry.Instrumentation.AWSLambda** -> publish
3. Update and renaming for **OpenTelemetry.Contrib.Instrumentation.AWS** (-> **OpenTelemetry.Instrumentation.AWS**) and then PR for publishing release.

I guess there is no need to release it as soon as possible, but also to not postpone the release for a long time. 2023-Jun-14 is an approximate date.

Kielek:
@rypdal,
it is not the only message about this, but for sure you can treat it as agreement to rename: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/875#issuecomment-1452338635

Details can be adjusted when you create PR. Thank you!

rypdal:
Hi @srprash, @atshaw43 and @Kielek. Unfortunately, we are running out of time and I would like to ask if it's possible to skip renaming and releasing of the OpenTelemetry.Contrib.Extensions.AWSXRay (the package is referenced from OpenTelemetry.Contrib.Instrumentation.AWS and OpenTelemetry.Instrumentation.AWSLambda) ?

Oberon00:
@rypdal IIUC, we don't need any release of other packages (even OpenTelemetry.Contrib.Extensions.AWSXRay) as a hard dependency before merging this PR. Unless the renaming is done in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1232, and then we'd also need to update the reference from here.

Oberon00:
It's independent in the sense that there will be no problems if we release one but not the other. However, to leverage the ability to use a non-XRay propagator, both need to be released eventually.

Kielek:
https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AWSLambda/1.1.0-beta.3

PR #1225, opened 2023-06-08
[Exporter.Geneva] Update EventSource implmentation of GenevaMetricExporter

utpilla:
## Changes
- Update `EventSource` implementation of `GenevaMetricExporter` to be a Singleton

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
@@ -87,7 +87,7 @@ public class GenevaMetricExporter : BaseExporter<Metric>
             case TransportProtocol.Unspecified:
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
-                    this.metricDataTransport = new MetricEtwDataTransport();
+                    this.metricDataTransport = MetricEtwDataTransport.Shared;
                     break;
                 }
                 else
--- a/src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
@@ -24,7 +24,14 @@ internal sealed class MetricEtwDataTransport : EventSource, IMetricDataTransport
 {
     private readonly int fixedPayloadEndIndex;
 
-    public MetricEtwDataTransport()
+    static MetricEtwDataTransport()
+    {
+        Shared = new();
+    }
+
+    public static readonly MetricEtwDataTransport Shared;
+
+    private MetricEtwDataTransport()
     {
         unsafe
         {


src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
   [RS0026](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md#rs0026-do-not-add-multiple-public-overloads-with-optional-parameters).
   ([#1218](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1218))
 
+* Update `MetricEtwDataTransport` which is  the `EventSource` implementation for
```
reyang, line 14: I guess the user would be very confused by this message since they are not supposed to understand the internals (e.g. MetricEtwDataTransport might mean nothing to them).

Do we really need a changelog entry?

utpilla: I added this CHANGELOG entry so that we can easily figure out which version got this fix.

Update the wording.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -11,6 +11,12 @@
   [RS0026](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md#rs0026-do-not-add-multiple-public-overloads-with-optional-parameters).
   ([#1218](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1218))
 
+* Fix the issue of running into the `ArgumentException`: `An instance of
+  EventSource with Guid edc24920-e004-40f6-a8e1-0e6e48f39d84 already exists.`
+  when using multiple instances of `GenevaMetricExporter` by updating
+  `MetricEtwDataTransport` to be a singleton.
```
reyang, line 17: I suggest that we remove "by updating ..." to make it concise (my take is that the reader of the changelog would want to focus on "what does it mean to me" rather than "how you've implemented it", and for folks who have curiosity to know the details, they can always go to the PR - which is why we put the PR link).

I don't feel strongly about this though, so not a blocker.

reyang (approved):
LGTM.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
reyang: > so that we can easily figure out which version got this fix.

I guess we want to look at the commit history (e.g. `git log | grep ...`) rather than CHANGELOG (which is for the users, not the maintainers/owners)?

utpilla: I think it's much quicker to first look for any noteworthy changes for a given release in the CHANGELOG, than going through git log which is the comprehensive source of truth.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
utpilla: > my take is that the reader of the changelog would want to focus on "what does it mean to me" rather than "how you've implemented it", and for folks who have curiosity to know the details, they can always go to the PR - which is why we put the PR link

I think CHANGELOG could be meant for developers as well? For example, look at this entry in the collector [CHANGELOG](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CHANGELOG.md#v0790) under v0.79.0:

- batchprocessor: Change multiBatcher to use sync.Map, avoid global lock on fast path (#7714)

This is too much detail for users but it gives a good reference point for developers when they quickly want to identify which version and above have that change.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@
   [RS0026](https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md#rs0026-do-not-add-multiple-public-overloads-with-optional-parameters).
   ([#1218](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1218))
 
+* Fix the issue of running into the `ArgumentException`: `An instance of
```
cijothomas, line 14: do we have same problem in log/trace exporter?

reyang: Similar but slightly different (e.g. most folks can just to two different log ETW providers, while metrics always use the same provider).

utpilla: @reyang @cijothomas I was wondering if we need to worry about scenarios where there are multiple instances of an application running each with the same GenevaExporter setup. They would run into this issue.

utpilla: Actually, on second thoughts, maybe not? It's probably just tracked at a process level?

PR #1234, opened 2023-06-14
[OpenTelemetry.Instrumentation.AWSLambda]: update README according to the recent code changes.

rypdal:
- Described tracing API split into synchronous and asynchronous.
- Described new configuration options.
- Described details about parent extraction.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/README.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/README.md
@@ -15,16 +15,42 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
 ```csharp
 TracerProvider tracerProvider = Sdk.CreateTracerProviderBuilder()
         // add other instrumentations
-        .AddAWSLambdaConfigurations()
+        .AddAWSLambdaConfigurations(options => options.DisableAwsXRayContextExtraction = true)
         .Build();
 ```
 
+### AWSLambdaInstrumentationOptions
+
+`AWSLambdaInstrumentationOptions` contains various properties to configure AWS lambda instrumentation.
+
+#### `DisableAwsXRayContextExtraction` (optional)
+
+Boolean value indicating whether AWS X-Ray context extraction should be disabled. Default value is false.
+
+#### `SetParentFromBatch` (optional)
+
+Boolean value indicating whether the parent Activity should be set when a potentially batched event is received where multiple parents are potentially available (e.g. SQS).
+If set to true, the parent is set using the last received record (e.g. last message). Otherwise the parent is not set. In both cases, links will be created for such events.
+Currently, the only event type to which this applies is SQS. Default value is false.
+
 ## Instrumentation
+
+`AWSLambdaWrapper` class contains tracing methods covering different types of function handler method signatures. `AWSLambdaWrapper.Trace()` and `AWSLambdaWrapper.TraceAsync()` methods are
+used for wrapping synchronious and asynchronious function handlers respectively. The `ActivityContext parentContext` parameter is optional and used to pass the explicitely extracted parent. If the parent
+is not passed from the customer's code then it's either extracted from the input parameter or uses AWS X-Ray headers if AWS X-Ray context extraction is enabled (see configuration property `DisableAwsXRayContextExtraction`).
+The sequence of the parent extraction: `explicit parent` -> `parent from input parameter` -> `parent from AWS X-Ray headers` -> `default context`
+The parent extraction is supported for the following input types:
+
+| Package | Types |
+|---------|-------|
+| `Amazon.Lambda.APIGatewayEvents` | `APIGatewayProxyRequest, APIGatewayHttpApiV2ProxyRequest` |
+| `Amazon.Lambda.SQSEvents` | `SQSEvent, SQSMessage` |
+| `Amazon.Lambda.SNSEvents` | `SNSEvent, SNSRecord` |
 
 ### Lambda Function
 
 1. Create a wrapper function with the same signature as the original Lambda function.
-Call `AWSLambdaWrapper.Trace()` API and pass `TracerProvider`, original Lambda function
+Call `AWSLambdaWrapper.Trace()` or `AWSLambdaWrapper.TraceAsync()` API and pass `TracerProvider`, original Lambda function
 and its inputs as parameters.
 
 2. Set the wrapper function as the Lambda handler input.
@@ -48,14 +74,14 @@ public string OriginalFunctionHandler(JObject input, ILambdaContext context)
 
 For using base classes from package [Amazon.Lambda.AspNetCoreServer](https://github.com/aws/aws-lambda-dotnet/tree/master/Libraries/src/Amazon.Lambda.AspNetCoreServer#amazonlambdaaspnetcoreserver),
 override the `FunctionHandlerAsync` function in `LambdaEntryPoint.cs` file. Call
-`AWSLambdaWrapper.Trace()` API and pass `TracerProvider`, original Lambda function
+`AWSLambdaWrapper.TraceAsync()` API and pass `TracerProvider`, original Lambda function
 and its inputs as parameters. Below is an example if using `APIGatewayProxyFunction`
 as base class.
 
 ```csharp
 public override async Task<APIGatewayProxyResponse> FunctionHandlerAsync(
     APIGatewayProxyRequest request, ILambdaContext lambdaContext)
-=> await AWSLambdaWrapper.Trace(tracerProvider, base.FunctionHandlerAsync,
+=> await AWSLambdaWrapper.TraceAsync(tracerProvider, base.FunctionHandlerAsync,
     request, lambdaContext);
 ```
 


src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+        .AddAWSLambdaConfigurations(options => options.DisableAwsXRayContextExtraction = true)
         .Build();
 ```
 
+### AWSLambdaInstrumentationOptions
+
+`AWSLambdaInstrumentationOptions` contains various properties to configure
+AWS lambda instrumentation.
+
+#### `DisableAwsXRayContextExtraction` (optional)
```
Oberon00, line 27: "(optional)" can be omitted, as all options are optional and this is as expected I think.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+record (e.g. last message). Otherwise the parent is not set. In both cases, links
+will be created for such events. Currently, the only event type to which this
+applies is SQS. Default value is false.
+
 ## Instrumentation
 
+`AWSLambdaWrapper` class contains tracing methods covering different types of
+function handler method signatures. `AWSLambdaWrapper.Trace()` and
+`AWSLambdaWrapper.TraceAsync()` methods are used for wrapping synchronious
+and asynchronious function handlers respectively. The `ActivityContext parentContext`
```
Oberon00, line 46:
```suggestion
`AWSLambdaWrapper.TraceAsync()` methods are used for wrapping synchronous
and asynchronous function handlers respectively. The `ActivityContext parentContext`
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+parameter is optional and used to pass the explicitly extracted parent. If the parent
```
Oberon00, line 47:
```suggestion
parameter is optional and used to pass a custom parent context. If the parent
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+is not passed from the customer's code then it's either extracted from the
```
Oberon00, line 48:
```suggestion
is not passed explicitly then it's either extracted from the
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+input parameter or uses AWS X-Ray headers if AWS X-Ray context extraction is
+enabled (see configuration property `DisableAwsXRayContextExtraction`).
+The sequence of the parent extraction:
+`explicit parent` -> `parent from input parameter` -> `AWS X-Ray headers` -> `default`
+The parent extraction is supported for the following input types:
+
+| Package | Types |
+|---------|-------|
+| `Amazon.Lambda.APIGatewayEvents` | `APIGatewayProxyRequest, APIGatewayHttpApiV2ProxyRequest` |
+| `Amazon.Lambda.SQSEvents` | `SQSEvent, SQSMessage` |
```
Oberon00, line 58: IMHO, we should not document this. Users should always pass the full record.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,52 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+        .AddAWSLambdaConfigurations(options => options.DisableAwsXRayContextExtraction = true)
         .Build();
 ```
 
+### AWSLambdaInstrumentationOptions
+
+`AWSLambdaInstrumentationOptions` contains various properties to configure
+AWS lambda instrumentation.
+
+- [`DisableAwsXRayContextExtraction`](/AWSLambdaInstrumentationOptions.cs#L25)
```
Oberon00: Linking to a particular line seems dangerous, one would need to remember to update the readme for any change in this file. It will surely get out of sync. I would just link the whole file once, and not link each option. The radical approach would be to not even list them but just point to the file for the list, but one could also list them with a brief one-sentence summary.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,54 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
 ```csharp
 TracerProvider tracerProvider = Sdk.CreateTracerProviderBuilder()
         // add other instrumentations
-        .AddAWSLambdaConfigurations()
+        .AddAWSLambdaConfigurations(options => options.DisableAwsXRayContextExtraction = true)
```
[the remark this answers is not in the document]
rypdal: Good idea. Added it to the default example.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
rypdal: Removed.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
rypdal: I removed individual record types and mentioned that users should always pass the full input object.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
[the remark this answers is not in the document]
rypdal: I added relative links to the xml doc.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
rypdal: Right. I replaced both links by the file reference.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,52 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
 
+### AWSLambdaInstrumentationOptions
+
+`AWSLambdaInstrumentationOptions` contains various properties to configure
+AWS lambda instrumentation.
+
+- [`DisableAwsXRayContextExtraction`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+- [`SetParentFromBatch`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+
+Boolean value indicating whether the parent Activity should be set when
```
Oberon00, line 30: Is this paragraph intentionally left? Seems to belong to the second point, or should be removed.

rypdal: No, unintentionally. Removed.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
         // add other instrumentations
-        .AddAWSLambdaConfigurations()
+        .AddAWSLambdaConfigurations(options => options.DisableAwsXRayContextExtraction = true)
         .Build();
 ```
 
+### AWSLambdaInstrumentationOptions
+
+`AWSLambdaInstrumentationOptions` contains various properties to configure
+AWS lambda instrumentation.
```
Oberon00, line 25:
```suggestion
AWS lambda instrumentation:
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+
+* [`DisableAwsXRayContextExtraction`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+* [`SetParentFromBatch`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+
 ## Instrumentation
 
+`AWSLambdaWrapper` class contains tracing methods covering different types of
```
Oberon00, line 32:
```suggestion
`AWSLambdaWrapper` contains tracing methods covering different types of
```

or

```suggestion
The `AWSLambdaWrapper` class contains tracing methods covering different types of
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+function handler method signatures. `AWSLambdaWrapper.Trace()` and
+`AWSLambdaWrapper.TraceAsync()` methods are used for wrapping synchronous
```
Oberon00, line 34:
```suggestion
function handler method signatures. `AWSLambdaWrapper.Trace()` and
`AWSLambdaWrapper.TraceAsync()` are used for wrapping synchronous
```

or

```suggestion
function handler method signatures. The methods `AWSLambdaWrapper.Trace()` and
`AWSLambdaWrapper.TraceAsync()` are used for wrapping synchronous
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+is not passed explicitly then it's either extracted from the
+input parameter or uses AWS X-Ray headers if AWS X-Ray context extraction is
+enabled (see configuration property `DisableAwsXRayContextExtraction`).
+The sequence of the parent extraction:
+`explicit parent` -> `parent from input parameter` -> `AWS X-Ray headers` -> `default`
+The parent extraction is supported for the input types listed in the table below.
+Although the `SQSEvent` and `SNSEvent` consist of multiple
+records (`SQSMessage` and `SNSRecord`) it's recommended to pass the full event.
+
+| Package | Types |
```
Oberon00, line 46: I wonder if there is any  value in listing the Package? Instead, it would be good to mention briefly where it is extracted from (e.g. for SQSEvent: Message attributes of last message; API Gateway events: HTTP headers)

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+|---------|-------|
+| `Amazon.Lambda.APIGatewayEvents` | `APIGatewayProxyRequest, APIGatewayHttpApiV2ProxyRequest` |
+| `Amazon.Lambda.SQSEvents` | `SQSEvent` |
+| `Amazon.Lambda.SNSEvents` | `SNSEvent` |
+
 ### Lambda Function
 
 1. Create a wrapper function with the same signature as the original Lambda function.
```
Oberon00, line 54: I think this is not true in all cases, as AFAIK we require the context parameter now.

```suggestion
1. Create a wrapper function with the same signature as the original Lambda function but an added ILambdaContext parameter if it was not already present.
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
-Call `AWSLambdaWrapper.Trace()` API and pass `TracerProvider`, original Lambda function
+Call `AWSLambdaWrapper.Trace()` or `AWSLambdaWrapper.TraceAsync()` API and pass
+`TracerProvider`, original Lambda function
 and its inputs as parameters.
```
Oberon00, line 57:
```suggestion
Call `AWSLambdaWrapper.Trace()` or `AWSLambdaWrapper.TraceAsync()` API and pass
`TracerProvider`, original Lambda function and its parameters.
```

rypdal:
Hello @open-telemetry/dotnet-contrib-maintainers. You may be interested in this README change too. Please, take a look at this PR when you have some time. Thank you.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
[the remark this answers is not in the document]
Oberon00: This was not meant as suggestion, but I am wondering why we show that option in the example. If the option would be good as default-on for some reason, probably we should change it in the implementation, otherwise I would recommend leaving the example as it was before with just

```suggestion
        .AddAWSLambdaConfigurations()
```

And maybe show a call with customized options in the section below.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -15,16 +15,45 @@ Add `AddAWSLambdaConfigurations()` to `TracerProvider`.
...
+and asynchronous function handlers respectively. The `ActivityContext parentContext`
+parameter is optional and used to pass a custom parent context. If the parent
+is not passed explicitly then it's either extracted from the
+input parameter or uses AWS X-Ray headers if AWS X-Ray context extraction is
+enabled (see configuration property `DisableAwsXRayContextExtraction`).
+The sequence of the parent extraction:
+`explicit parent` -> `parent from input parameter` -> `AWS X-Ray headers` -> `default`
+The parent extraction is supported for the input types listed in the table below.
+Although the `SQSEvent` and `SNSEvent` consist of multiple
+records (`SQSMessage` and `SNSRecord`) it's recommended to pass the full event.
```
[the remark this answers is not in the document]
rypdal: Better to remove it.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
rypdal: I replaced the 'Package' column by the 'Parent extraction source'.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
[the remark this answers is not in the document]
rypdal: But may be before we change implementation it's worth to keep DisableAwsXRayContextExtraction in the example ? Otherwise it may lead to some misunderstandings.

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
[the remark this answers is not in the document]
Kielek: In general looks good, I am not AWSLamda expert. If `DisableAwsXRayContextExtraction` set to `true` should be default option lets keep it on both examples. Otherwise remove it.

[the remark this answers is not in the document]
rypdal: Then lets keep it in both examples as is.

Kielek (approved):

Issue #1024, opened 2023-02-22
Deadlock while instrumenting EventCountersMetrics


besseb05:
# Issue with OpenTelemetry.Instrumentation.EventCounters

List of [all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.0.2`):

    <PackageReference Update="OpenTelemetry" Version="1.4.0-beta.3" />
    <PackageReference Update="OpenTelemetry.Api" Version="1.4.0-beta.3" />
    <PackageReference Update="OpenTelemetry.Exporter.CredScan" Version="2.0.12" />
    <PackageReference Update="OpenTelemetry.Extensions.Hosting" Version="1.0.0-rc9.9" />
    <PackageReference Update="OpenTelemetry.Exporter.Zipkin" Version="1.4.0-beta.3" />
    <PackageReference Update="OpenTelemetry.Instrumentation.AspNetCore" Version="1.0.0-rc9.9" />
    <PackageReference Update="OpenTelemetry.Instrumentation.EventCounters" Version="1.0.0-alpha.2" />
    <PackageReference Update="OpenTelemetry.Instrumentation.GrpcNetClient" Version="1.0.0-rc9.9" />
    <PackageReference Update="OpenTelemetry.Instrumentation.Http" Version="1.0.0-rc9.9" />
    <PackageReference Update="OpenTelemetry.Instrumentation.Process" Version="1.0.0-alpha.2" />
    <PackageReference Update="OpenTelemetry.Instrumentation.Runtime" Version="1.1.0-beta.1" />

* TBD

Runtime version: net7.0 

* TBD

**Is this a feature request or a bug?**

* [ ] Feature Request
* [X] Bug

**What is the expected behavior?**

When we are instrumenting our code, we expect this code for instrumenting the MeterProvider is executed when the ServiceFabric service settings are created. The code should execute and return normally, so that the service can start up and start serving calls.

```csharp
#pragma warning disable CA2000 // Dispose objects before losing scope
            EventSource gcEventSource = new("gc");
#pragma warning restore CA2000 // Dispose objects before losing scope
            // This Counter will be removed once OT team adds process.gc.percentageTimeinGC metrics in OT process instrumentation library
#pragma warning disable CA2000 // Dispose objects before losing scope
[...]

PR #1260, opened 2023-07-05
[EventCounters] fix deadlock

Yun-Ting:
Mitigates https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1024.

## Changes
Below is an example of when the deadlock can occur:

|                 | Thread 1 | Thread 2 |  locks acquired at this timestamp|
| ---------- | ----------|-----------|-----------|
| Time **↓**| call ctor of `EventCountersMetrics` | | |
|                 | [lock](https://github.com/Yun-Ting/opentelemetry-dotnet-contrib/blob/3e2015dd2a2884b1c89e596bcb53d599b6f961f4/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs#L50C1-L50C40) `preInitEventSources` |  | lock on `preInitEventSources` object by thread 1|
|                | iterate through the HashSet of `EventSources` [AND enable](https://github.com/Yun-Ting/opentelemetry-dotnet-contrib/blob/3e2015dd2a2884b1c89e596bcb53d599b6f961f4/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs#L58C1-L58C52) them  | | [lock](https://github.com/Yun-Ting/opentelemetry-dotnet-contrib/blob/3e2015dd2a2884b1c89e596bcb53d599b6f961f4/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs#L58C1-L58C52) on `EventSource`  by thread 1 |
|               |     | `EventSourceA` got enabled by thread 1 | |
|               |     | `OnEventWritten` callback fired on receiving an event | | 
|               |     | `EventCountersInstrumentationEventSource` got [lazy-init](https://github.com/Yun-Ting/opentelemetry-dotnet-contrib/blob/3e2015dd2a2884b1c89e596bcb53d599b6f961f4/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs#L127C13-L127C101) | lock on `EventSource` by thread 2 |
|               |     | `OnEventSourceCreated` callback fired  | |
|               |     |  [lock](https://github.com/Yun-Ting/opentelemetry-dotnet-contrib/blob/3e2015dd2a2884b1c89e596bcb53d599b6f961f4/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs#L96C8-L96C40) `preInitEventSouces` |  lock on `preInitEventSouces` by thread 2 | 

## Discussion

1. Better way to mitigate this issue could be, use an assistive dictionary to keep track of whether a particular eventSource should be enabled in a lock, and a flag (for example, shouldEnable) before releasing the lock. Outside of the lock, when the flag was set to true earlier, enable that eventSource. This approach should be more performant as compared to the current implementation. For example: https://github.com/dotnet/runtime/blob/84fd85932fd3b371cad69cc2412404ea50158c63/src/libraries/Common/tests/System/Diagnostics/Tracing/TestEventListener.cs#L84C1-L93C14
2. The locks on `preInitEventSources` were added in this [PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/703) to address the issue -
"As OnEventSourceCreated can run before the constructor is done, you can run into a situation where you fail to enable an EventSource instance."
@MihaZupan: Instead of the current locking mechanism, Is it possible in `OnEventSourceCreated`, we use Spinlock to wait until `this.options` was set to ensure that `OnEventSourceCreated` will be triggered only after the `ctor` got executed? Too hacky?

### DIFF
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -46,6 +46,7 @@ internal sealed class EventCountersMetrics : EventListener
     /// <param name="options">The options to define the metrics.</param>
     public EventCountersMetrics(EventCountersInstrumentationOptions options)
     {
+        _ = EventCountersInstrumentationEventSource.Log;
         lock (this.preInitEventSources)
         {
             this.options = options;
--- a/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
@@ -40,13 +40,20 @@ internal sealed class EventCountersMetrics : EventListener
     private readonly ConcurrentDictionary<(string, string), double> values = new();
     private bool isDisposed;
 
+    static EventCountersMetrics()
+    {
+        // Ensure EventCountersInstrumentationEventSource got initialized when the class was accessed for the first time
+        // to prevent potential deadlock:
+        // https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1024.
+        _ = EventCountersInstrumentationEventSource.Log;
+    }
+
     /// <summary>
     /// Initializes a new instance of the <see cref="EventCountersMetrics"/> class.
     /// </summary>
     /// <param name="options">The options to define the metrics.</param>
     public EventCountersMetrics(EventCountersInstrumentationOptions options)
     {
-        _ = EventCountersInstrumentationEventSource.Log;
         lock (this.preInitEventSources)
         {
             this.options = options;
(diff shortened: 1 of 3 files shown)


besseb05 (approved):
lgtm - although I am not familiar with the codebase

src/OpenTelemetry.Instrumentation.EventCounters/EventCountersMetrics.cs
```diff
@@ -46,6 +46,7 @@ internal sealed class EventCountersMetrics : EventListener
     /// <param name="options">The options to define the metrics.</param>
     public EventCountersMetrics(EventCountersInstrumentationOptions options)
     {
+        _ = EventCountersInstrumentationEventSource.Log;
```
utpilla, line 49: Need to add comments describing why this is required.

reyang: +1, also consider making it static (e.g. in cctor).

reyang (approved):
LGTM with a changelog entry.

reyang (approved):

utpilla (approved):

PR #1261, opened 2023-07-10
[OpenTelemetry.Instrumentation.AWSLambda]: consider two sources when setting http.method and http.target: original request and request context.

rypdal:
**Problem**

For AWS Lambda functions (V1 only) invoked via HTTP depending on setup there might be different sources for setting http.method and http.target: original request (APIGatewayProxyRequest) and request context (ProxyRequestContext) defined inside of the request.

**Fix**

Consider Path and HttpMethod properties of APIGatewayProxyRequest and APIGatewayProxyRequest.ProxyRequestContext when setting http.target and http.method. Request context should be prioritised over request while setting http.target and http.method

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
@@ -44,7 +44,7 @@ internal class AWSLambdaHttpUtils
         {
             case APIGatewayProxyRequest request:
                 httpScheme = AWSLambdaUtils.GetHeaderValues(request, HeaderXForwardedProto)?.LastOrDefault();
-                httpTarget = string.Concat(request.RequestContext?.Path ?? string.Empty, GetQueryString(request));
+                httpTarget = string.Concat(request.RequestContext?.Path ?? request.Path ?? string.Empty, GetQueryString(request));
                 httpMethod = request.HttpMethod;
                 var hostHeader = AWSLambdaUtils.GetHeaderValues(request, HeaderHost)?.LastOrDefault();
                 (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeader);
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -44,8 +44,9 @@ internal class AWSLambdaHttpUtils
         {
             case APIGatewayProxyRequest request:
                 httpScheme = AWSLambdaUtils.GetHeaderValues(request, HeaderXForwardedProto)?.LastOrDefault();
-                httpTarget = string.Concat(request.RequestContext?.Path ?? string.Empty, GetQueryString(request));
-                httpMethod = request.HttpMethod;
+                var path = request.Path ?? request.RequestContext?.Path ?? string.Empty;
+                httpTarget = string.Concat(path, GetQueryString(request));
+                httpMethod = request.HttpMethod ?? request.RequestContext?.HttpMethod;
```
Oberon00, line 49: Should we also consider the case where the method is null/empty? I guess in that case it does not make sense to set any (other) http attributes

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
rypdal: I would consider it more general: should we set all required tags of one (or more) of required are missing ? But then it makes code more complex. I think we still should be more tolerant and set as more as we can.

Oberon00 (changes requested):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -44,8 +44,9 @@ internal class AWSLambdaHttpUtils
         {
             case APIGatewayProxyRequest request:
                 httpScheme = AWSLambdaUtils.GetHeaderValues(request, HeaderXForwardedProto)?.LastOrDefault();
-                httpTarget = string.Concat(request.RequestContext?.Path ?? string.Empty, GetQueryString(request));
-                httpMethod = request.HttpMethod;
+                var path = request.Path ?? request.RequestContext?.Path ?? string.Empty;
```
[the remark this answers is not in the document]
rypdal: I swapped setting and mention prioritisation in the PR description.

Oberon00 (approved):

utpilla (approved):
LGTM overall. Resorting to the component owner's approval for the specific changes made in the PR.

PR #1111, opened 2023-03-28
[Exporter.Geneva] Add support for exporting metrics to different combinations of account and namespace

utpilla:
## Changes
- Add support for exporting metrics to different combinations of account and namespace

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #1269, opened 2023-07-17
[Exporter.Geneva] Can configure metrics namespace per Meter

specialforest:
Fixes #.

## Changes

Add support in metrics for namespace configuration per meter.

Example scenario: two services report metrics into their own namespaces, but common metrics, like metrics from HTTP client and ASP.NET Core, to a shared namespace.

### DIFF
--- a/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
@@ -24,9 +24,9 @@ using System.Linq;
 using System.Net.Sockets;
 using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Threading.Tasks;
 using Kaitai;
 using Microsoft.Extensions.DependencyInjection;
+using Moq;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Trace;
 using Xunit;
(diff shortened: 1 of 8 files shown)


test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
```diff
@@ -1264,125 +1026,71 @@ private static void AssertExemplarFilteredTagSerialization(Exemplar expectedExem
...
+    {
+        int headerSize;
+        unsafe
+        {
+            headerSize = sizeof(BinaryHeader);
+        }
 
-        if (metricType == MetricType.LongSum)
+        var result = new byte[size + headerSize];
+        Array.Copy(src, result, size + headerSize);
```
specialforest, line 1046: I wonder if it's a bug when the exporter calls IMetricDataTransport.Send the _size_ parameter doesn't include binary header size.

utpilla:
There is already a feature to send metrics to different namespaces by adding the tag `_microsoft_metrics_namespace` with the required namespace name. This was added in #1111 and has been available starting `1.5.1` version of the package.

You should be able to use that for your requirement.

specialforest:
> @specialforest There is already a feature to send metrics to different namespaces by adding the tag `_microsoft_metrics_namespace` with the required namespace name. This was added in #1111 and has been available starting `1.5.1` version of the package.
> 
> You should be able to use that for your requirement.

I cannot do that for metrics I don't own, e.g., HttpClient or ASP.NET Core metrics. Yes technically, I can setup the default namespace to be the shared namespace and all _standard_ metrics will be reported there and for my own metrics I will need to use `_microsoft_metrics_namespace` every time to report to service specific namespace.

specialforest:
I cannot edit code of HttpClient to add _microsoft_metrics_namespace ;)

utpilla:
> I cannot edit code of HttpClient to add _microsoft_metrics_namespace ;)

@specialforest  You have a few ways to achieve what you're trying to do here:
- One of them is what you already mentioned in your previous comment about using the namespace for the instrumentation librariess' metrics in the ConnectionString and override the namespace for the metrics that you control by adding `_microsoft_metrics_namespace` dimension.
- Create two instances of `MeterProvider` along with their respective `GenevaExporter`s. You could have one configured with namespace of instrumentation libraries and the other one with the namespace for your custom metrics

PR #1270, opened 2023-07-18
Instrumenation.AWSLambda: Drop netstandard2, lose Newtonsoft (non-)dependency

Oberon00:
Closes #1273 (alternative to that PR)

Related to a similar change in the XRay dependency: #1092

## Changes

Quoting the CHANGELOG:

* The minimum required .NET version is now .NET Core 3.1 (previously any netstandard2.0-compliant).
  As the minimum .NET version in AWS Lambda is already 6.0, no compatibility problems are expected.

  This resolves a warning that some dependency analyzers may produce where this package would transitively
  depend on a vulnerable version of `Newtonsoft.Json` through
  [Amazon.Lambda.APIGatewayEvents](https://www.nuget.org/packages/Amazon.Lambda.APIGatewayEvents/2.4.1#dependencies-body-tab).

  This also avoids a potential issue where the instrumentation would try to call a Newtonsoft.Json function
  when no other package nor the app itself depends on Newtonsoft.Json, since the transitive dependency would be
  ignored unless using application were compiled against a TargetFramework older than Core 3.1.

For significant contributions please make sure you have completed the following items:

* [X] Appropriate `CHANGELOG.md` updated for non-trivial changes
* ~~[ ] Design discussion issue #~~ N/A
* ~~[ ] Changes in public API reviewed~~ API remains identical

PR #1273, opened 2023-07-20
Instrumentation.AWSLambda: Upgrade & explicitly depend on Newtonsoft.Json

Oberon00:
Closes #1270 (alternative to that PR).

Related to a similar change in the XRay dependency: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1092

## Changes

Quoting the CHANGELOG:

* Add explicit dependency on Newtonsoft.Json, upgrading the mimimum version.

  This resolves a warning that some dependency analyzers may produce where this
  package would transitively depend on a vulnerable version of Newtonsoft.Json
  through [Amazon.Lambda.APIGatewayEvents][].

  This also avoids a potential issue where the instrumentation would try to call
  a Newtonsoft.Json function when no other package nor the app itself depends on
  Newtonsoft.Json, since the transitive dependency would be ignored unless using
  application were compiled against a TargetFramework older than Core 3.1.

[Amazon.Lambda.APIGatewayEvents]: https://www.nuget.org/packages/Amazon.Lambda.APIGatewayEvents/2.4.1#dependencies-body-tab

* ~~[ ] Design discussion issue #~~ N/A
* ~~[ ] Changes in public API reviewed~~ N/A

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Unreleased
 
+* Add explicit dependency on Newtonsoft.Json, upgrading the minimum version.
+
+  This resolves a warning that some dependency analyzers may produce where this
+  package would transitively depend on a vulnerable version of Newtonsoft.Json
+  through [Amazon.Lambda.APIGatewayEvents][].
+
+  This also avoids a potential issue where the instrumentation would try to call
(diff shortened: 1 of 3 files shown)


utpilla:
Could you update the branch with the latest changes? It looks like maintainers wouldn't have access to do that for this branch.

utpilla (approved):

Oberon00:
I updated the branch and also changed permissions so that repository maintainers should now be able to push/update here.

Kielek (approved):

PR #1275, opened 2023-07-21
Rename package: [OpenTelemetry.Contrib.Instrumentation.AWS] to [OpenTelemetry.Instrumentation.AWS]

rypdal:
Name of the package _OpenTelemetry.Contrib.Instrumentation.AWS_ is not consistent with current naming conventions and `Contrib `suffix should be removed.

Package project and all related artefacts like namespaces, etc should be changed to OpenTelemetry.Instrumentation.AWS 

This PR is the result of discussions in the other PR: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1224#issuecomment-1581162972


rypdal:
FYI: @Oberon00 and @Kielek

Oberon00:
I just noticed that the AWS packages are also the only ones that break the capitalization rule at https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers which states that for acronyms longer than two letters, only the first one should be uppercased. So the namespace & package name ought to be Aws instead of AWS.
However, I wonder if this is worth the disruption (if we also rename the AwsLambda/AWSLambda package) or inconsistency (if we don't).

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
Oberon00, line 1: This is shown as added instead of moved, and I can't see a corresponding deletion. Did you forget to delete / commit the deletion of the old csproj?

src/OpenTelemetry.Instrumentation.AWS/TracerProviderBuilderExtensions.cs
```diff
@@ -15,11 +15,11 @@
...
 
 using System;
-using OpenTelemetry.Contrib.Instrumentation.AWS;
-using OpenTelemetry.Contrib.Instrumentation.AWS.Implementation;
+using OpenTelemetry.Instrumentation.AWS.Implementation;
 using OpenTelemetry.Internal;
+using OpenTelemetry.Trace;
 
-namespace OpenTelemetry.Trace;
+namespace OpenTelemetry.Instrumentation.AWS;
```
Kielek, line 22: Old namespace was here for a reason.
Easy way to call `AddAWSInstrumentation` without adding new using at the top of the file.

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
Kielek, line 1: Please add information about changes at the top of Unreleased section.

rypdal:
> I just noticed that the AWS packages are also the only ones that break the capitalization rule at https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers which states that for acronyms longer than two letters, only the first one should be uppercased. So the namespace & package name ought to be Aws instead of AWS. However, I wonder if this is worth the disruption (if we also rename the AwsLambda/AWSLambda package) or inconsistency (if we don't).

I think it's a subject of different PR where we should include AWSLambda package too. The current renaming follows initial rules introduced in AWS projects.

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
rypdal: I forgot to remove the old project file from the OpenTelemetry.Contrib.Instrumentation.AWS folder. It's removed now.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs
```diff
@@ -115,7 +115,7 @@ private static void ProcessEndRequest(IExecutionContext executionContext, Activi
             var httpResponse = responseContext.HttpResponse;
             if (httpResponse != null)
             {
-                int statusCode = (int)httpResponse.StatusCode;
+                var statusCode = (int)httpResponse.StatusCode;
```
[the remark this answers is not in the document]
rypdal: Seems I applied  some offered by VS auto-refactoring. Although some changes are reasonable, but since we would like to keep changes related only to renaming of package it was reverted.

test/OpenTelemetry.Instrumentation.AWS.Tests/OpenTelemetry.Instrumentation.AWS.Tests.csproj
```diff
@@ -20,13 +20,13 @@
...
-    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Contrib.Instrumentation.AWS\OpenTelemetry.Contrib.Instrumentation.AWS.csproj" />
-  </ItemGroup>
-
   <ItemGroup Condition="'$(TargetFramework)' == '$(NetFrameworkMinimumSupportedVersion)'">
     <Compile Remove="Tools\CustomWebResponse.cs" />
     <Compile Remove="Tools\HttpResponseMessageBody.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\OpenTelemetry.Instrumentation.AWS\OpenTelemetry.Instrumentation.AWS.csproj" />
```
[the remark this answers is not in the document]
rypdal: Thanks. Added the `RepoRoot`.

src/OpenTelemetry.Instrumentation.AWS/TracerProviderBuilderExtensions.cs
rypdal: I see. Reverted to the original namespace.

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
rypdal: Added.

Oberon00 (approved):

Kielek (approved):

rypdal:
Seems failed tests are introduced after merge.

rypdal:
@srprash , @Kielek , @Oberon00: could this PR be merged ?

Oberon00 (approved):

.github/component_owners.yml
```diff
@@ -6,7 +6,7 @@ components:
   src/OpenTelemetry.Contrib.Extensions.AWSXRay/:
     - srprash
     - atshaw43
-  src/OpenTelemetry.Contrib.Instrumentation.AWS/:
+  src/OpenTelemetry.Instrumentation.AWS/:
```
utpilla, line 9: Please sort them alphabetically.

utpilla (approved):

Kielek:
Merging per comment https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/875#issuecomment-1452338635

PR #1277, opened 2023-07-24
[ResourceDetectors.Azure] update changelog and readme

TimothyMothra:
## Changes
- update changelog
- update readme for VM detector.

For significant contributions please make sure you have completed the following items:

* [ ] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #1280, opened 2023-07-26
[ResourceDetectors.Azure] update readme to use `ConfigureResource`

TimothyMothra:
Follow up: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1277#discussion_r1272943541


rajkumar-rangaraj:
Should we also include meter provider example as a part of this update. @utpilla has called this out in earlier PR.

utpilla:
> Should we also include meter provider example as a part of this update. @utpilla has called this out in earlier PR.

The earlier PR had added a comment saying "The following example shows how to add AzureVMResourceDetector to TracerProvider configuration, **but this can be added to logs and metrics as well**."

utpilla (approved):

Issue #1268, opened 2023-07-14
[OpenTelemetry.Contrib.Instrumentation.AWS]: issue for next -beta package release.


rypdal:
# Issue with OpenTelemetry.Contrib.Instrumentation.AWS

Request: next beta release of the OpenTelemetry.Contrib.Instrumentation.AWS package. 

Recently implemented in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1034  functionality is distributed between 2 nuget packages:

**OpenTelemetry.Instrumentation.AWSLambda**: released -> https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1224
**OpenTelemetry.Contrib.Instrumentation.AWS**: not released

We are not owners of the corresponding project and can't create PR for the release of the package. That's why the issue is created instead.
We would like to ask for release of the next -beta version of the **OpenTelemetry.Contrib.Instrumentation.AWS** package.

rypdal:
FYI: @Oberon00

rypdal:
Hello @srprash  and @atatshaw43. 
We would like to ask for the next beta release of the OpenTelemetry.Contrib.Instrumentation.AWS package. Please, see the issue description.

Megasware128:
I noticed the same for OpenTelemetry.ResourceDetectors.Azure. #1182 has been merged for quite a while now but the feature has not been published to NuGet yet

rypdal:
Hi @srprash , @Kielek , @Oberon00. The title and description of this PR has been adapted according to the target package rename. Can we release the renamed package OpenTelemetry.Instrumentation.AWS ?

Kielek:
First part:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/releases/tag/Extensions.AWS-1.3.0-beta.1
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/releases/tag/ResourceDetectors.AWS-1.3.0-beta.1

Kielek:
@rypdal, technical changes are done. Could you please create PR similar to
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1285?

I suppose that 1.1.0-beta.1 should be fine for Instrumentation.AWS.
Probably 1.2.0-beta.1 for Instrumentation.AWSLambda.

rypdal:
> 1.2.0-beta.1 for Instrumentation.AWSLambda

[...]

PR #1285, opened 2023-07-28
[Extensions.AWS, ResourceDetectors.AWS] Release 1.3.0-beta.1

Kielek:
Towards #1268

## Changes

Prerequisite to release another AWS packages.
Please check changes in changelog for details.

The question is if it should be released as 1.3.0-beta.1 (as we changed the package name) or it has to be major release (2.0.0-beta.1) for Extension package.
My recommendation is to use the same version for ResourceDetectors as it was extracted from the old package.
@srprash, @Oberon00, @rypdal, please let me know what do you think about versioning.

### DIFF
--- a/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.3.0-beta.1
+
+Released 2023-Aug-01
+
 * Rename package from `OpenTelemetry.Contrib.Extensions.AWSXRay`
   to `OpenTelemetry.Extensions.AWS`
   ([#1232](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1232))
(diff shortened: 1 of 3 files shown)


rypdal:
We already released the OpenTelemetry.Instrumentation.AWSLambda with changed package name as [1.1.0-beta.2](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/Instrumentation.AWSLambda-1.1.0-beta.3/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md#110-beta2). I think we should stick with the same versioning for the other AWS packages.

srprash:
I agree that we don't need a major version bump.

srprash:
After the `Extensions.AWS` is released, we need to update the references of the old package name in rest of the packages.
https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-dotnet-contrib%20OpenTelemetry.Contrib.Extensions.AWSXRay&type=code

But I think there could be an issue. Stable packages cannot have unstable dependencies. The build for the packages depending on `Extensions.AWS-1.3.0-beta.1` will likely start to fail.

Kielek:
@srprash, if it is ok for your, I would go first with beta release. Then, when needed and confirmed that we do no make any mistake, we can make follow up release as stable package.

srprash:
@Kielek That sounds good to me.

srprash (approved):

Issue #1201, opened 2023-05-23
Component owners - OpenTelemetry membership status


Kielek:
From https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md

> The owner should at least be an [OpenTelemetry Member](https://github.com/open-telemetry/community/blob/main/community-membership.md#member) to be eligible to assigned as component owner. This is required to ensure that reviews can be automatically requested from the owners.

For now, our auto-assign PR job is usually not working because of lack of OTel membership.

If you are mentioned below, please consider to create requests https://github.com/open-telemetry/community/issues/new?assignees=&labels=area%2Fgithub-membership&projects=&template=membership.md&title=REQUEST%3A+New+membership+for+%3Cyour-GH-handle%3E.

I will be happy to support you as a OTel member.

| name              | GH link                              | OTel Member |
|-------------------|--------------------------------------|-------------|
| ~~@atshaw43~~          | https://github.com/atshaw43          | No - temporarry removed from component owner files         |
| cijothomas        | https://github.com/cijothomas        | Yes         |
| codeblanch        | https://github.com/codeblanch        | Yes         |
| ejsmith           | https://github.com/ejsmith           | ~~No~~ Yes          |
| fred2u            | https://github.com/fred2u            | ~~No~~ Yes          |
| hananiel          | https://github.com/hananiel          | Yes         |
| havret            | https://github.com/havret            | ~~No~~ Yes          |
| iskiselev         | https://github.com/iskiselev         | ~~No~~ Yes        |
| maldago           | https://github.com/maldago           | ~~No~~ Yes          |
| mic-max           | https://github.com/mic-max           | Yes         |
| moonheart         | https://github.com/moonheart         | ~~No~~ Yes          |
| Oberon00          | https://github.com/Oberon00          | Yes         |
| pcwiese           | https://github.com/pcwiese           | ~~No~~ Yes          |
| rajkumar-rangaraj | https://github.com/rajkumar-rangaraj | Yes         |
| reyang            | https://github.com/reyang            | Yes         |
| rypdal            | https://github.com/rypdal            | ~~No~~Yes          |
| SergeyKanzhelev   | https://github.com/SergeyKanzhelev   | Yes         |
| srprash           | https://github.com/srprash           | Yes         |
| twenzel           | https://github.com/twenzel           | ~~No~~ Yes       |
| utpilla           | https://github.com/utpilla           | Yes         |
[...]

PR #1316, opened 2023-08-21
[Instrumentation.Quartz] Nullable

Kielek:
Towards #894
## Changes


Kielek:
@maldago, could you please review?

maldago (approved):
I don't see any issue with this. Thanks @Kielek

src/OpenTelemetry.Instrumentation.Quartz/QuartzInstrumentationOptions.cs
```diff
@@ -43,7 +43,7 @@ public class QuartzInstrumentationOptions
     /// <para>object: the raw object from which additional information can be extracted to enrich the activity.
     /// The type of this object depends on the event, which is given by the above parameter.</para>
     /// </remarks>
-    public Action<Activity, string, object> Enrich { get; set; }
+    public Action<Activity, string, object?>? Enrich { get; set; }
```
utpilla, line 46: I think the better thing to do would be to not mark `object` nullable and update the code to call `Enrich` only when the `object` passed to it is not null.

Kielek:
@maldago, could you please also check #1201?

(sorry for mixing topics, but I do not found any better way to send you a message.

src/OpenTelemetry.Instrumentation.Quartz/QuartzInstrumentationOptions.cs
Kielek: Please check 8b32c4b3837e37eeb2bfef7edc5e16d414a58e57

maldago: I think the update is fair.

PR #1336, opened 2023-08-30
Nullable for ASP.NET instrumentation projects

Kielek:
Towards #894

## Changes

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/AspNetTelemetryEventSource.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/AspNetTelemetryEventSource.cs
@@ -33,47 +33,47 @@ internal sealed class AspNetTelemetryEventSource : EventSource
     public static readonly AspNetTelemetryEventSource Log = new();
 
     [NonEvent]
-    public void ActivityStarted(Activity? activity)
+    public void ActivityStarted(Activity activity)
     {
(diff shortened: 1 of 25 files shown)


pjanotti (approved):

src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/AspNetTelemetryEventSource.cs
```diff
@@ -33,7 +33,7 @@ internal sealed class AspNetTelemetryEventSource : EventSource
     public static readonly AspNetTelemetryEventSource Log = new();
 
     [NonEvent]
-    public void ActivityStarted(Activity activity)
+    public void ActivityStarted(Activity? activity)
```
utpilla, line 36: These methods would not be called when the activity is `null`. You should update the usage of activity instead to this: `this.ActivityStarted(activity!.Id);`

Kielek: Fixed in 698292581464ae3971a4cce874da1202245f2319

This kind of methods accepts string nullable as Id technically can returns null.
```csharp
Event(2, Message = "Activity started, Id='{0}'", Level = EventLevel.Verbose)]
public void ActivityStarted(string? id)
```

utpilla (approved):

PR #1338, opened 2023-08-30
[AWSLambda] - Replace Newtonsoft dependency with System.Text.Json

jamiewinder:
## Changes

Removes the Newtonsoft.Json dependency and replaces it with System.Text.Json.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
@@ -20,7 +20,6 @@ using System.Diagnostics;
 using System.Linq;
 using Amazon.Lambda.SNSEvents;
 using Amazon.Lambda.SQSEvents;
-using Newtonsoft.Json;
 using OpenTelemetry.Context.Propagation;
 
 namespace OpenTelemetry.Instrumentation.AWSLambda.Implementation;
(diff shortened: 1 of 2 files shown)


Oberon00:
This is basically #1270 and I think we might be ready to do this now. But we probably should do the full #1270 and also upgrade the target framework version. At the moment you are probably using System.Text.Json through an "accidental" transitive dependency? Can you check where it is coming from?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -145,7 +144,7 @@ internal static PropagationContext ExtractParentContext(SNSEvent.SNSMessage? mes
         {
             try
             {
-                snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
+                snsMessage = JsonSerializer.Deserialize<SNSEvent.SNSMessage>(body);
```
rypdal, line 147: You need `System.Text.Json` namespace.

rypdal (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: Ah, OK this does not even compile. Please see my above comment. I'm marking this as "Changes requested"

Oberon00 (changes requested):

PR #1352, opened 2023-09-15
Nullable Hangfire

Kielek:
Towards #894

## Changes

Nullable Hangfire.
Tests project already covered.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
@@ -1,9 +1,9 @@
 OpenTelemetry.Trace.HangfireInstrumentationOptions
-OpenTelemetry.Trace.HangfireInstrumentationOptions.DisplayNameFunc.get -> System.Func<Hangfire.BackgroundJob, string>
+OpenTelemetry.Trace.HangfireInstrumentationOptions.DisplayNameFunc.get -> System.Func<Hangfire.BackgroundJob!, string!>!
 OpenTelemetry.Trace.HangfireInstrumentationOptions.DisplayNameFunc.set -> void
 OpenTelemetry.Trace.HangfireInstrumentationOptions.HangfireInstrumentationOptions() -> void
 OpenTelemetry.Trace.HangfireInstrumentationOptions.RecordException.get -> bool
 OpenTelemetry.Trace.HangfireInstrumentationOptions.RecordException.set -> void
 OpenTelemetry.Trace.TracerProviderBuilderExtensions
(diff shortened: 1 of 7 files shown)


fred2u (approved):

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
```diff
@@ -32,13 +32,13 @@ internal sealed class HangfireInstrumentationJobFilterAttribute : JobFilterAttri
 {
     private readonly HangfireInstrumentationOptions options;
 
+#pragma warning disable CA1019 // Define accessors for attribute arguments
```
utpilla, line 35: Is this just about making `options` a property instead of a field? I think we could simply fix that if that's the case.

src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
```diff
@@ -1,9 +1,9 @@
 OpenTelemetry.Trace.HangfireInstrumentationOptions
-OpenTelemetry.Trace.HangfireInstrumentationOptions.DisplayNameFunc.get -> System.Func<Hangfire.BackgroundJob, string>
+OpenTelemetry.Trace.HangfireInstrumentationOptions.DisplayNameFunc.get -> System.Func<Hangfire.BackgroundJob!, string!>!
```
utpilla, line 2: Simplify this line: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/b3c7ae771f77a8661db996a3e2b8e91e3242b281/src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs#L64-L65

to this since `DisplayNameFunc` is non-nullable now?
```csharp
Func<BackgroundJob, string> displayNameFunc = this.options.DisplayNameFunc;
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireInstrumentationJobFilterAttribute.cs
Kielek: This construction `private HangfireInstrumentationOptions Options { get; }` leads to: `CA1019	If Options is the property accessor for positional argument options, make it public	OpenTelemetry.Instrumentation.Hangfire`.

Which is nob better. I do not think that we should make it public.

src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/netstandard2.0/PublicAPI.Unshipped.txt
Kielek: Fixed 462c6fdd6ba42efa280218085e7bdda737d70616

utpilla (approved):

PR #1366, opened 2023-09-21
Nullable Instrumentation.EntityFrameworkCore

Kielek:
Towards #894 .

## Changes

Nullable Instrumentation.EntityFrameworkCore.
Introduce `IncludeSharedSpanHelper` and `ActivityInstrumentationHelper` to include files into compilation only when needed.


ejsmith (approved):
LGTM

src/Shared/ListenerHandler.cs
```diff
@@ -26,9 +29,11 @@ internal abstract class ListenerHandler
     /// Initializes a new instance of the <see cref="ListenerHandler"/> class.
     /// </summary>
     /// <param name="sourceName">The name of the <see cref="ListenerHandler"/>.</param>
-    public ListenerHandler(string sourceName)
+    /// <param name="supportsNullActivity">Indicates whether the  <see cref="ListenerHandler"/> supports NULL <see cref="Activity"/>.</param>
+    protected ListenerHandler(string sourceName, bool supportsNullActivity)
```
utpilla, line 33: Why is this new `bool` parameter being added? I don't think we need this.

Kielek: Previously there to return true by `SupportsNullActivity` we need to override this value in child class. For now, we have to inject by constructor. It was more intuitive for me.

The property is used under DiagnosticSourceListener -> OnNext method

utpilla: Any particular benefit to updating the shared file?

I think we could achieve this by simply using `#nullable enable` in the individual classes and annotating the code accordingly (by using `?` and `!`).

Kielek: Changes related to SupportNullActivity reverted in 7a455f055c158cb374308e738ee5c06c554477a9

utpilla (approved):

PR #1370, opened 2023-09-22
Nullable for tests-projects with already handled projects

Kielek:
Towards #894.

## Changes

Nullable for tests-projects with already handled projects.

While reviewing consider to check each commit separately.


test/OpenTelemetry.Extensions.Enrichment.Tests/OpenTelemetryEnrichmentProviderBuilderExtensions.cs
```diff
@@ -41,6 +41,7 @@ public void AddTraceEnricherOfTRegistersEnricher()
         using var source = new ActivitySource(SourceName);
         using (var activity = source.StartActivity(SourceName))
         {
+            Assert.NotNull(activity);
```
utpilla, line 44: We don't really need these `Assert.NotNull(activity);` statements. The unit tests would fail anyway if the activity ends up being `null`.

utpilla (approved):

test/OpenTelemetry.Extensions.Enrichment.Tests/OpenTelemetryEnrichmentProviderBuilderExtensions.cs
Kielek: I can use either use `!` as null-forgiving operator or add this assertion. Assertion looks better IMO.

PR #1388, opened 2023-10-10
[Instrumentation.AspNet] Fix activity being closed before processing completes

### LINKED ISSUE #1138: AspNet HttpInListener OnStopActivity is called after Activity is stopped by TelemetryHttpModule
# Bug Report

* OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule 1.0.0-rc9.3
* OpenTelemetry.Instrumentation.AspNet 1.0.0-rc9.3

Runtime version:

* net48

## Symptom

The "OpenTelemetry.Instrumentation.AspNet.Telemetry" span is processed and exported before it can be enriched by the the `HttpInListener` and any custom `Enrich` callbacks.

For example, the DisplayName of the Activity should be set to the MVC Route instead of the raw URL.

The DisplayName of the Activity remains the raw URL

## Additional Context

In https://github.com/open-telemetry/opentelemetry-dotnet/blob/1.0.0-rc9.3/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs#L141, we can see that `aspNetActivity.Stop();` is called before invoking the `onRequestStoppedCallback`. That callback is responsible for calling into `OnStopActivity` on `HttpInListener` which can update the name based on the route template can call the `Enrich` callback method.

I think that calling `Stop()` on the activity automatically triggers the ActivityListeners, including processors and exporters, to handle when the Activity is stopped.

Is it as simple as moving `Stop()` to after the callbacks?

qhris:
Fixes #1138.

## Findings

This PR implements the fix suggested by @jdaigle.

I have tested this on an a .NET Framework 4.8 ASP MVC application I've been maintaining.

I ran into this issue trying to modify the span name to be prefixed by the http method as [recommended by the spec](<https://github.com/open-telemetry/semantic-conventions/blob/4040095eda0159e38edfe7084ed32d3077a6ffb0/docs/http/http-spans.md?plain=1#L70>). The problem I ran into was the `HttpInListener` overriding whatever changes the processor made to the span name.

It also seems odd to me that the activity is closed and *then* updated (through `HttpInListener.OnStopActivity`). I think there is a race condition with exporters here. I was able to trigger this by setting a breakpoint before `onRequestStoppedCallback` is called; sometimes the exported span would have the pre-callback display name and sometimes it would have the post-callback display name. This is due to the exporters being processors, and processors are triggered on stopping the activity (?). This of course depends on you're using the simple exporter or a batched one.

This is how I think it works (pre-patch):

- `aspNetActivity.Stop()` is called.
- Processors run (including exporters)
- `HttpInListener.OnStopActivity` is called (through `onRequestStoppedCallback`.

I'm still pretty new to OTEL in dotnet, so I might be missing something here, but this seems like a problem to me.

## Questions

- [x] Should the `content.Items[ContextKey] = null`  be moved as well, as the context is cleared out now before `asnNetActivity.Stop()` is called?

## Changes

This changes the point at which the activity is stopped to after the `onRequestStoppedCallback` function is called.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs
@@ -131,8 +131,6 @@ internal static class ActivityHelper
         Debug.Assert(context.Items[ContextKey] is ContextHolder, "Context item is not an ContextHolder instance.");
 
         var currentActivity = Activity.Current;
-
-        aspNetActivity.Stop();
         context.Items[ContextKey] = null;
 
         try
@@ -144,6 +142,7 @@ internal static class ActivityHelper
             AspNetTelemetryEventSource.Log.CallbackException(aspNetActivity, "OnStopped", callbackEx);
         }
 
+        aspNetActivity.Stop();
         AspNetTelemetryEventSource.Log.ActivityStopped(aspNetActivity);
 
         if (textMapPropagator is not TraceContextPropagator)
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Fixed an issue where activities were stopped incorrectly before calling Enrich callbacks.
+* Fixed an issue where activities were stopped incorrectly before processing completed.
   Activity processor's `OnEnd` will now happen after `AspNetInstrumentationOptions.Enrich`.
   ([#1388](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1388))
 
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Fixed an issue where activities were closed incorrectly before processing completed.
+* Fixed an issue where activities were stopped incorrectly before calling Enrich callbacks.
   Activity processor's `OnEnd` will now happen after `AspNetInstrumentationOptions.Enrich`.
   ([#1388](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1388))
 
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fixed an issue where activities were closed incorrectly before processing completed.
+  Activity processor's `OnEnd` will now happen after `AspNetInstrumentationOptions.Enrich`.
+  ([#1388](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1388))
+
 * Update `OpenTelemetry.Api` to `1.6.0`.
   ([#1344](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1344))
 
--- a/src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Release together with `OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule`.
+  Fixed an issue where activities were closed incorrectly before processing completed.
+  Activity processor's `OnEnd` will now happen after `AspNetInstrumentationOptions.Enrich`.
+  ([#1388](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1388))
+
 ## 1.0.0-rc9.9
 
 Released 2023-Jun-09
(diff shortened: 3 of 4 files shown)


Kielek:
@qhris, please sign CLA, it is hard requirement to proceed with PR.

Please let us know if you have any issue with the procedure.

qhris:
@Kielek I had missed the CLA requirement as I submitted the PR. I have no idea how that works so I'll look it over tomorrow since I wasn't prepared having to sign something with my physical address...

In regards to the changes:

Something that struck me is that this will be a breaking change as it changes the order of the processor's `OnEnd` and `AspNetInstrumentationOptions.Enrich` delegate. Not sure if the current ordering is on purpose or not.

qhris:
Did some additional testing this morning comparing the flow to a .NET 7 application.

I used `OpenTelemetry.Instrumentation.AspNetCore` `1.5.1-beta.1`

In the .NET 7 application we have the following flow:

1. Incoming http request.
2. Activity is started.
3. Processors run their `OnStart`.
4. `EnrichWithHttpRequest` from `HttpInListener` is called.
5. Request processing (middlewares/controllers).
6. `EnrichWithHttpResponse` from `HttpInListener` is called.
7. Processors run their `OnEnd`.

The .NET Framework application (pre-patch) has the following behavior:

1. Incoming http request.
2. Activity is started.
3. Processors run their `OnStart`.
4. `Enrich` with `OnStartActivity` is called.
5. Request processing (middlewares/controllers).
6. Processors run their `OnEnd`
7. `Enrich` with `OnStopActivity` is called.

This PR would bring the implementations in line between .NET and .NET Framework by swapping items 6 and 7 for .NET Framework.

cijothomas:
> Did some additional testing this morning comparing the flow to a .NET 7 application.
> 
> I used `OpenTelemetry.Instrumentation.AspNetCore` `1.5.1-beta.1`
> 
> In the .NET 7 application we have the following flow:
> 
> 1. Incoming http request.
> 2. Activity is started.
> 3. Processors run their `OnStart`.
> 4. `EnrichWithHttpRequest` from `HttpInListener` is called.
> 5. Request processing (middlewares/controllers).
> 6. `EnrichWithHttpResponse` from `HttpInListener` is called.
> 7. Processors run their `OnEnd`.
> 
> The .NET Framework application (pre-patch) has the following behavior:
> 
> 1. Incoming http request.
> 2. Activity is started.
> 3. Processors run their `OnStart`.
> 4. `Enrich` with `OnStartActivity` is called.
> 5. Request processing (middlewares/controllers).
> 6. Processors run their `OnEnd`
> 7. `Enrich` with `OnStopActivity` is called.
> 
> This PR would bring the implementations in line between .NET and .NET Framework by swapping items 6 and 7 for .NET Framework.

The ASP.NET Core behavior is correct and this fix for asp.net is correct! Thanks.

src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs
```diff
@@ -144,6 +142,7 @@ public static void StopAspNetActivity(TextMapPropagator textMapPropagator, Activ
             AspNetTelemetryEventSource.Log.CallbackException(aspNetActivity, "OnStopped", callbackEx);
         }
 
+        aspNetActivity.Stop();
```
cijothomas, line 145: please add a changelog entry with this bugfix link.
It'd be great to have a test covering this, but not a blocker.

cijothomas (approved):
LGTM with a changelog entry!
Thanks for your contribution!

src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs
qhris: I've added both tests and changelog updates, tell me if there's any wording I need to change.

src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fixed an issue where activities were closed incorrectly before processing completed.
```
cijothomas, line 5:
```suggestion
* Fixed an issue where activities were stopped incorrectly before processing completed.
```

cijothomas, line 5:
```suggestion
* Fixed an issue where activities were stopped incorrectly before calling Enrich callbacks.
```

cijothomas (approved):

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Release together with `OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule`.
```
utpilla, line 5: You can remove this line. These two packages are always released together.

utpilla (approved):

utpilla:
@qhris Thanks for fixing the issue.

You can use the `1.6.0-beta.1` version of the instrumentation library which has this fix in it: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/releases/tag/Instrumentation.AspNet-1.6.0-beta.1

vishweshbankwar:
@qhris This change has a side effect of broken metrics instrumentation.

Now that the activity is stopped after the onStopCallBack is fired, the [activity.Duration](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/a372bc065452a7c13bcc69a021e842be440cce13/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs#L51) would always result in `0`. In order to fix this, you would need to set the duration of activity before the end callback like below:

```
if (aspNetActivity.Duration == TimeSpan.Zero)
{
    aspNetActivity.SetEndTime(DateTime.UtcNow);
}
```

qhris:
@vishweshbankwar thanks for letting me know, I'll take a look to reproduce and I'll submit a fix for it.
I guess this is another downside of having the metrics being tied to traces...

PR #1407, opened 2023-10-22
[Instrumentation.AspNet] Add metric enrich functionality

### LINKED ISSUE #1226: AspNet Metrics Instrumentation Enrich & Filter

Using Open Telemetry's [AspNetCore instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/src/OpenTelemetry.Instrumentation.AspNetCore#metrics), we can automatically generate, enrich, and filter a default `http.server.duration` metric using [AspNetCoreMetricsInstrumentationOptions](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreMetricsInstrumentationOptions.cs) in our .NET Core services. 

Through the similar [AspNet instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AspNet#metrics) used in .NET Framework services it is possible to emit the `http.server.duration` metric, but the same filtering and enrichment options are not available. Does any mechanism currently exist to attach additional dimensions or apply filtering logic to the auto-instrumented metric in .NET Framework? 

For example, we're able to use the below logic in .NET Core and are aiming to do the same in .NET Framework.
image

* [x] Feature Request
* [ ] Bug

Parity between .NET Core and .NET Framework metric configuration options.

Currently, .NET Framework services do not have the same ability to enrich and filter the default metrics generated by Open Telemetry instrumentation libraries. .NET Framework services would be limited to the default dimensions and have no ability to filter based on custom logic.

## Additional Context

Pointed here by @cijothomas after initially posting in the opentelemetry-dotnet Discussions board [here](https://github.com/open-telemetry/opentelemetry-dotnet/discussions/4567).

qhris:
Fixes #1226 by adding adding functionality.

## Changes

- Adds enrich functionality for metrics


src/OpenTelemetry.Instrumentation.AspNet/.publicApi/net462/PublicAPI.Unshipped.txt
```diff
@@ -6,8 +6,17 @@ OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentationOptions.Filter.get ->
...
 OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentationOptions.RecordException.get -> bool
 OpenTelemetry.Instrumentation.AspNet.AspNetInstrumentationOptions.RecordException.set -> void
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.AspNetMetricsInstrumentationOptions() -> void
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.Enrich.get -> OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.EnrichFunc?
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.Enrich.set -> void
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.EnrichFunc
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.Filter.get -> OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.FilterFunc?
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.Filter.set -> void
+OpenTelemetry.Instrumentation.AspNet.AspNetMetricsInstrumentationOptions.FilterFunc
```
Kielek, line 16: It is opposed to https://github.com/open-telemetry/opentelemetry-dotnet/pull/4981 for ASP.NET Core metrics.

@vishweshbankwar, I suppose that we should be able to release also ASP.NET as stable together/after ASP.NET. It is potential blocker for this.

qhris: From what I can gather, these features are for .NET 8 and will not be available for .NET Framework (as per @cijothomas comment) which this library targets? What would be the way forward for enrich filter/functionailty here then?

vishweshbankwar: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4981 does not apply here. Regarding the stable release of ASP.NET, there is no ETA for it.

qhris:
Is there anything you want me to do with this?

Ideally I'd build upon this to introduce the new `http.request.server.duration` metric. I can do that in this PR or do it as a follow up. If you want me to introduce the new metric; maybe the old one should be removed? We can't exactly do a switch like .NET did between 7 and 8 where the `OTEL_SEMCONV_STABILITY_OPT_IN` became irrelevant.

vishweshbankwar:
@qhris Do you need filter as well? If yes, could you explain the use case. ASP.NET Core has removed the filter capability https://github.com/open-telemetry/opentelemetry-dotnet/pull/4981

qhris:
@vishweshbankwar No particular reason, I was just matching the API's to the now removed functionality in .NET 8.
I'll update this to remove filtering.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+* Added enrich functionality to metric instrumentation
+  ([#1407](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407)).
+
+  * New overload of `AddAspnetInstrumentation` now accepts a configuration delegate.
+  * The `Enrich` can be used to add additional metric tags.
```
Kielek, line 9: I think that we should change tags -> attributes
It is common name in OTel. Tag is the only implementation in .NET.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
+
+* Additional metric tags will now be emitted
+  ([#1407](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407)).
```
Kielek, line 12: Any chance

Kielek (commented):
Sorry for late feedback. Any chance to split the PR into 2 parts - exactly as you have in CHANGE:OG:
1. Enriching functionality
2. Adding tags.

2 reasons behind it.
1. Changelog will have better links to the changes, so the end-user can easily check potential changes.
2. It is easier to verify smaller functionalities while reviewing.

jdom:
If I read the specs, it does look like the `http.server.request.duration` metric in seconds is stable: https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#metric-httpserverrequestduration
(I'm not an expert in OTel, so perhaps I'm not looking in the right spot)

Kielek:
@jdom, you are looking correctly, but we should switch whole package to the new convention at once. IMO the worst option is mixing them.

I am fine with merging this as is, then make one more PR to switch convention.

qhris:
@Kielek that sounds good to me, since I'd have to undo a lot of work just to add it back in with new values.
I like the idea of just removing the old conventions for metrics, if we are clear about it in the release notes.

So for now I'll just make the suggested changes to the changelog and fix the conflicts.
There's also probably no need to release a new version of the package until the 2nd PR is made.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
qhris: Listed the new attributes as requested.

ZRich97:
> @qhris Do you need filter as well? If yes, could you explain the use case. ASP.NET Core has removed the filter capability [open-telemetry/opentelemetry-dotnet#4981](https://github.com/open-telemetry/opentelemetry-dotnet/pull/4981)

@vishweshbankwar - our team has used the filter logic in ASP.NET Core's instrumentation to avoid emitting metrics/traces for healthchecks and other unmonitored paths. We'd like to have the same functionality from the ASP.NET instrumentation (though I realize this has now been removed in the newer Core packages). Is this a valid use case that would merit adding filtering, or is there a separate OTel mechanism we should use? I can open a separate PR for filtering if it would be more appropriate to move the discussion there.

Kielek (approved):
Approving. Still I would prefer to have it as 2 PRs.

@vishweshbankwar, could you please check this?

qhris:
@Kielek I can still split it if necessary but that would be later tonight. Either way works for me. If it doesn't get merged until then the plan is:

1. Make this PR only be the enrich functionality (with additional changes for tests and such).
2. Prepare a new PR with the new `http.server.request.duration` metrics that @jdom mentioned are stable.

qhris:
@Kielek @vishweshbankwar 

Sorry for the inconvenience.
I have retargeted this PR to *only* add the `Enrich` functionality now!

vishweshbankwar (approved):
LGTM - Thanks @qhris!

Kielek:
Thanks for contribution!

PR #1420, opened 2023-10-31
Fix up Redis trimming annotations

eerhardt:
#1415 had an issue in it due to how the DynamicDependency attribute is written.

1. Referring to a nested type with `+` isn't correct in DynamicDependency attributes. You need to just use a normal `.`. The current code is producing a warning when using the .NET 8.0 SDK to publish for AOT. "warning IL2036: OpenTelemetry.Instrumentation.StackExchangeRedis.Implementation.RedisProfilerEntryToActivityConverter.<.cctor>g__GetCommandAndKey|5_3(PropertyFetcher`1<String>,Object,String&): Unresolved type 'StackExchange.Redis.RedisDatabase+ScriptEvalMessage' in 'DynamicDependencyAttribute'."

This wasn't caught in our CI because we use the .NET 7.0 SDK to publish, which doesn't catch this bug.

2. The "CommandAndKey" property isn't on the nested ScriptEvalMessage type, but instead a virtual on the base Message class.

The fix here is to use the base Message class in the DynamicDependency. I've verified this works with a .NET 8.0 SDK.

cc @Yun-Ting @utpilla @vitek-karas

### DIFF
--- a/src/OpenTelemetry.Instrumentation.StackExchangeRedis/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.StackExchangeRedis/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fix an issue in the trimming annotations to refer to the correct Type
+  ([#1420](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1420))
+
 ## 1.0.0-rc9.11
 
 Released 2023-Oct-31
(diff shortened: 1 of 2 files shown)


Yun-Ting:
maybe add a unit test to cover this case?

eerhardt:
> maybe add a unit test to cover this case?

Do we have unit tests that actually publish the trimmed/AOT'd app and run it?

This bug would have been caught if we used the .NET 8.0 SDK in the verify-aot-compat leg instead of the 7.0 SDK. Do you know when that will happen? After 8.0 RTM ships?

utpilla (approved):

PR #1426, opened 2023-11-04
[Instrumentation.AspNet] Add Filter Support For ASP.NET Instrumented Metrics

### LINKED ISSUE #1226: AspNet Metrics Instrumentation Enrich & Filter

ZRich97:
Fixes #1226

## Changes

This PR adds AspNetMetricsInstrumentationOptions with the ability to filter and enrich emitted metrics to the OpenTelemetry.Instrumentation.AspNet package. These changes are meant to match those already available in the AspNetCore instrumentation library and based on this PR: https://github.com/open-telemetry/opentelemetry-dotnet/pull/3948

Our main use case for this functionality is that services can rely on the out-of-the-box instrumentation to emit the "http.server.duration" metric while adding custom tags (in our case these may be "Region", "Cluster", etc.). This would be preferred to building custom handlers and middleware to publish a separate metric with custom tags.

TODO: Documentation and changelog updates. 


Kielek:
@ZRich97, could you please review #1407? @ghris created PR few days ago with similar functionality (Enrich).
Based on the discussion Filtering probably will not be needed.

Thanksfor the contribution!

ZRich97:
> @ZRich97, could you please review #1407? @ghris created PR few days ago with similar functionality (Enrich). Based on the discussion Filtering probably will not be needed.
> 
> Thanksfor the contribution!

@Kielek / @vishweshbankwar - Given https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407 was merged as Enrich-only, would it make sense for me to update and publish this PR to enable filtering? I believe there's a valid use case for filtering (mentioned [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407#issuecomment-1800841111)), but let me know your thoughts. Thanks!

vishweshbankwar:
> > @ZRich97, could you please review #1407? @ghris created PR few days ago with similar functionality (Enrich). Based on the discussion Filtering probably will not be needed.
> > Thanksfor the contribution!
> 
> @Kielek / @vishweshbankwar - Given #1407 was merged as Enrich-only, would it make sense for me to update and publish this PR to enable filtering? I believe there's a valid use case for filtering (mentioned [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407#issuecomment-1800841111)), but let me know your thoughts. Thanks!

@ZRich97 What issues do you face if you dont filter out the endpoints you mentioned in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407#issuecomment-1800841111

ZRich97:
> > > @ZRich97, could you please review #1407? @ghris created PR few days ago with similar functionality (Enrich). Based on the discussion Filtering probably will not be needed.
> > > Thanksfor the contribution!
> > 
> > 
> > @Kielek / @vishweshbankwar - Given #1407 was merged as Enrich-only, would it make sense for me to update and publish this PR to enable filtering? I believe there's a valid use case for filtering (mentioned [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407#issuecomment-1800841111)), but let me know your thoughts. Thanks!
> 
> @ZRich97 What issues do you face if you dont filter out the endpoints you mentioned in [#1407 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1407#issuecomment-1800841111)

@vishweshbankwar - The main use case in my mind is filtering out unused metrics (noise). Some of our supported services don't need to collect metrics for every route in the application (and they currently don't in their non-OTel monitoring solutions). This filtering could also be done at the alerting/monitoring stage (metrics we don't care about could be identified by the dimensions collected/enriched by this instrumentation and filtered out of any dashboards/alerts), but why not avoid collecting unused metrics altogether?

cijothomas:
Metrics Filter was recently removed from AspNetCore instrumentation, so it may be better to hold off from this, and gather all feedback here: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4982

@ZRich97 Could you add your use-case to above as well?

ZRich97:
> Metrics Filter was recently removed from AspNetCore instrumentation, so it may be better to hold off from this, and gather all feedback here: [open-telemetry/opentelemetry-dotnet#4982](https://github.com/open-telemetry/opentelemetry-dotnet/issues/4982)
> 
> @ZRich97 Could you add your use-case to above as well?

@cijothomas - sure, I'll add our use-case there as well. From what I understand filtering is present for traces, is there a reason it shouldn't also be added for metrics? I would imagine the use-cases would be similar (reduce noise, only monitor certain paths or types of requests, etc.) - let me know if we are missing something. Thanks!

cijothomas:
> > Metrics Filter was recently removed from AspNetCore instrumentation, so it may be better to hold off from this, and gather all feedback here: [open-telemetry/opentelemetry-dotnet#4982](https://github.com/open-telemetry/opentelemetry-dotnet/issues/4982)
> > @ZRich97 Could you add your use-case to above as well?
> 
> @cijothomas - sure, I'll add our use-case there as well. From what I understand filtering is present for traces, is there a reason it shouldn't also be added for metrics? I would imagine the use-cases would be similar (reduce noise, only monitor certain paths or types of requests, etc.) - let me know if we are missing something. Thanks!

For traces,logs -every telemetry adds cost - memory cost, serialization costs, network costs, and Vendor billing (typically based on volume).
For metrics, due to the pre-aggregation - the cost increase is less significant (I am not saying it is not there, but relatively low), so it may not be worthwhile to execute filter conditions in metric hot path! Happy to hear more feedback on this topic, which will also ultimately influence if Asp.Net Core want to offer some native filtering ability!

ZRich97:
> For traces,logs -every telemetry adds cost - memory cost, serialization costs, network costs, and Vendor billing (typically based on volume). For metrics, due to the pre-aggregation - the cost increase is less significant (I am not saying it is not there, but relatively low), so it may not be worthwhile to execute filter conditions in metric hot path! Happy to hear more feedback on this topic, which will also ultimately influence if Asp.Net Core want to offer some native filtering ability!

Yep, it makes sense that this would be more important for traces/logs. I don't see a downside to including filtering for metrics as well - is there one from your perspective (or some general guidance against it)?

PR #1429, opened 2023-11-08
[Instrumentation.AspNet] Update semantic conventions for metrics

qhris:
Updates metrics to match [semantic conventions for 1.23](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#http-server).

This was previously discussed in #1407.

## Changes

Replaces `http.server.duration` with the new `http.request.server.duration` metric and adds missing attributes.

## Remaining issues

* [x] ~~Unable to get `Application_Error` to fire in my sample application in order to verify `error.type` for exceptions. It works fine for status codes.~~ Excluded for now.
* [x] Unable to write a test for `network.protocol.version` which uses `request.ServerVariables["SERVER_PROTOCOL"]` because that property/class is heavily locked down. I have verified this to work on a sample application.
* [x] The `HttpRequestRouteHelper` works fine for fetching the route template but it is marginally useful because the route easily ends up like `/{controller}/{action}` which doesn't say much.


qhris:
Hey,

I'm leaving this up as a draft until I can finish up the remaining tasks and run some more tests.
I could however use some input from @Kielek and @vishweshbankwar for the direction of this PR and the remaining tasks?

vishweshbankwar:
> Hey,
> 
> I'm leaving this up as a draft until I can finish up the remaining tasks and run some more tests. I could however use some input from @Kielek and @vishweshbankwar for the direction of this PR and the remaining tasks?

@qhris 
1) Let's handle `error.type` outside of this PR. It would be tricky to get that value on metrics side. Take a look at https://github.com/open-telemetry/opentelemetry-dotnet/pull/4986 if you can do something similar here.
2) I see the test creates `HttpContext.Current`. Is it possible to do  `HttpContext.Current.Request.ServerVariables.Add` and validate that you get the `network.protocol.version`? 
3) For `http.route`, this is a known issue. Please open a separate issue for this if there isn't one out there already. For now, you can copy how tracing is doing it.

qhris:
@vishweshbankwar thanks for the input!

1. Thanks for the link, I can look into this a bit more. Are you suggesting to remove it entirely for this PR if not?
2. A tried a variety of things to get it working.
    - `ServerVariables.Add()` throw with `System.NotSupportedException`.
    - `ServerVariables` only has a getter so I can't assign to it.
    - I also tried setting the backing field `_serverVariables` through reflection but ended up running into an issue since the type `HttpServerVarsCollection` is internal to `System.Web`. I could try to instantiate that type through reflection as well but I didn't feel comfortable with that as it would get very messy. Might be the only way to go though.
    - I tried making `HttpRequest` a mock but that throws: `System.NotSupportedException : Type to mock (HttpRequest) must be an interface, a delegate, or a non-sealed, non-static class`.
    - It seems using `HttpContextBase` is the recommended way to make the code more testable but that would require making quite a bigger change to the codebase (i.e. rewriting `HttpTelemetryModule`, `ActivityHelper`, et al.).
3. Sounds good 👍

vishweshbankwar:
> @vishweshbankwar thanks for the input!
> 
> 1. Thanks for the link, I can look into this a bit more. Are you suggesting to remove it entirely for this PR if not?
> 2. A tried a variety of things to get it working.
>    
>    * `ServerVariables.Add()` throw with `System.NotSupportedException`.
>    * `ServerVariables` only has a getter so I can't assign to it.
>    * I also tried setting the backing field `_serverVariables` through reflection but ended up running into an issue since the type `HttpServerVarsCollection` is internal to `System.Web`. I could try to instantiate that type through reflection as well but I didn't feel comfortable with that as it would get very messy. Might be the only way to go though.
>    * I tried making `HttpRequest` a mock but that throws: `System.NotSupportedException : Type to mock (HttpRequest) must be an interface, a delegate, or a non-sealed, non-static class`.
>    * It seems using `HttpContextBase` is the recommended way to make the code more testable but that would require making quite a bigger change to the codebase (i.e. rewriting `HttpTelemetryModule`, `ActivityHelper`, et al.).
> 3. Sounds good 👍

Yes - A separate PR for `error.type`.

For `network.protocol.version`, You could remove this one too for now in favor of making progress. I will try to take a look at it when I get chance.

qhris:
Hey, I'm made some additional findings here.

The reason why I wasn't able to get `Application_Error` to fire was because I was testing a `ASP Web API` application. This type of app has a completely different error flow as desribed in [this Microsoft documentation](<https://learn.microsoft.com/en-us/aspnet/web-api/overview/error-handling/web-api-global-error-handling>). 

This means that traces do not behave as expected either, since the error status of the span will never be set [here](<https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInListener.cs#L184>).

For now, I went ahead and removed the `error.type` as suggested.

Would it be a blocker to leave in the protocol version as is? It is a very useful metric and it would be a shame to leave it out.

vishweshbankwar:
> Would it be a blocker to leave in the protocol version as is

No, I think it is ok.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs
```diff
@@ -39,17 +44,44 @@ public void Dispose()
...
+        var request = context.Request;
+        var url = request.Url;
         var tags = new TagList
         {
-            { SemanticConventions.AttributeHttpMethod, context.Request.HttpMethod },
-            { SemanticConventions.AttributeHttpScheme, context.Request.Url.Scheme },
-            { SemanticConventions.AttributeHttpStatusCode, context.Response.StatusCode },
+            { SemanticConventions.AttributeServerAddress, url.Host },
+            { SemanticConventions.AttributeServerPort, url.Port },
+            { SemanticConventions.AttributeNetworkProtocolName, NetworkProtocolName },
```
vishweshbankwar, line 67: Remove `network.protocol.name` -- it is conditionally required if not defaults to `http`.
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-metrics.md#metric-httpserverrequestduration

qhris: Thanks, I missed that 👍

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs
```diff
@@ -39,17 +43,43 @@ public void Dispose()
...
+        var request = context.Request;
+        var url = request.Url;
         var tags = new TagList
         {
-            { SemanticConventions.AttributeHttpMethod, context.Request.HttpMethod },
-            { SemanticConventions.AttributeHttpScheme, context.Request.Url.Scheme },
-            { SemanticConventions.AttributeHttpStatusCode, context.Response.StatusCode },
+            { SemanticConventions.AttributeServerAddress, url.Host },
+            { SemanticConventions.AttributeServerPort, url.Port },
+            { SemanticConventions.AttributeHttpRequestMethod, request.HttpMethod },
```
vishweshbankwar, line 66: Needs special handling for httpMethod. see https://github.com/open-telemetry/opentelemetry-dotnet/pull/5001

src/Shared/SemanticConventions.cs
```diff
@@ -114,5 +114,12 @@ internal static class SemanticConventions
     public const string AttributeHttpRequestMethod = "http.request.method"; // replaces: "http.method" (AttributeHttpMethod)
     public const string AttributeHttpResponseStatusCode = "http.response.status_code"; // replaces: "http.status_code" (AttributeHttpStatusCode)
     public const string AttributeUrlScheme = "url.scheme"; // replaces: "http.scheme" (AttributeHttpScheme)
+
+    // v1.23.0
+    // https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#http-server
+    public const string AttributeNetworkProtocolName = "network.protocol.name"; // replaces: "net.transport" (AttributeNetTransport)
```
vishweshbankwar, line 120: remove this.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpRequestRouteHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Web;
+using System.Web.Routing;
+
+namespace OpenTelemetry.Instrumentation.AspNet.Implementation;
+
+/// <summary>
+/// Helper class for processing http requests.
+/// </summary>
+internal class HttpRequestRouteHelper
```
vishweshbankwar, line 26:
```suggestion
internal sealed class HttpRequestRouteHelper
```

test/OpenTelemetry.Instrumentation.AspNet.Tests/HttpInMetricsListenerTests.cs
```diff
@@ -130,9 +143,16 @@ public class HttpInMetricsListenerTests
             ExpectTag("true", "enriched");
         }
 
-        ExpectTag("GET", SemanticConventions.AttributeHttpMethod);
-        ExpectTag(200, SemanticConventions.AttributeHttpStatusCode);
-        ExpectTag(expectedScheme, SemanticConventions.AttributeHttpScheme);
+        // Do not use constants from SemanticConventions here in order to detect mistakes.
+        // https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#http-server
+        // Unable to check for "network.protocol.version" because we can't set server variables due to the accessability
```
vishweshbankwar, line 148:
```suggestion
        // Unable to check for "network.protocol.version" because we can't set server variables due to the accessibility
```

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -8,6 +8,23 @@
   * New overload of `AddAspNetInstrumentation` now accepts a configuration delegate.
   * The `Enrich` can be used to add additional metric attributes.
 
+* HTTP server metrics now follow stable
```
vishweshbankwar, line 11: Mention that this is a **breaking change**

Kielek:
@qhris, I think that. semantic convention also for spans/traces has to be adjusted. It can be done in separate PR, but it should be done before the release.

carlos-neto-maersk:
Can you guys please check https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1431 ?  
  
It is related to this, because due to the changes done to telemetry module to have the activity stopped on this `OnActivityStoppedCallback`, under load we are experiencing that weird pattern (and I think that it is because it because it is closing activities that aren't to be closed in other `OnActivityStoppedCallback` handlers.  
  
I'm working on not closing the activity after the handlers, and using the current timestamp to figure it out the duration instead of using `Activity.Duration`, but the code isn't great.  
  

**And thank you so much for your great work.**

qhris:
@carlos-neto-maersk I have seen the issue but I haven't had the time to check it out yet. I'll keep the discussion to that issue/PR as it is not exactly related to this one.

qhris:
> @qhris, I think that. semantic convention also for spans/traces has to be adjusted. It can be done in separate PR, but it should be done before the release to make the transition from unstable to stable world only once.

@Kielek It should go in a different PR agreed. This one should probably be worked out before though.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs
qhris: I went ahead and implemented something similar, while also handling `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS` properly.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs
```diff
@@ -39,17 +45,51 @@ public void Dispose()
...
-            { SemanticConventions.AttributeHttpMethod, context.Request.HttpMethod },
-            { SemanticConventions.AttributeHttpScheme, context.Request.Url.Scheme },
-            { SemanticConventions.AttributeHttpStatusCode, context.Response.StatusCode },
+            { SemanticConventions.AttributeServerAddress, url.Host },
+            { SemanticConventions.AttributeServerPort, url.Port },
+            { SemanticConventions.AttributeUrlScheme, url.Scheme },
+            { SemanticConventions.AttributeHttpResponseStatusCode, context.Response.StatusCode },
         };
 
+        if (this.requestMethodHelper.TryGetMethod(request.HttpMethod, out var method))
```
vishweshbankwar, line 72: nit: Take a look at this PR to refactor requestMethodHelper.
https://github.com/open-telemetry/opentelemetry-dotnet/pull/5026

src/OpenTelemetry.Instrumentation.AspNet/Implementation/RequestMethodHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+using System;
+
+namespace OpenTelemetry.Instrumentation.AspNet.Implementation;
+
+internal sealed class RequestMethodHelper
+{
+    private const string KnownHttpMethodsEnvironmentVariable = "OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS";
+
+    // List of known HTTP methods in order of expected frequency.
+    private readonly string[] knownHttpMethods = new[]
```
vishweshbankwar, line 26: why not use dictionary?

src/OpenTelemetry.Instrumentation.AspNet/Implementation/HttpInMetricsListener.cs
qhris: Yeah that looks cleaner, I'll rewrite it as that.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/RequestMethodHelper.cs
qhris: Since we're on a hot code path I figured looping over a small set of items should be faster than using a dictionary, but I can change this back if necessary.

Since you asked I decided to do a small performance test and found this:
image

I can change it to a dictionary to match the other implementation you linked.

qhris:
This is a getting a bit long at this point, but I noticed that `SemanticConventions.cs` now has a lot of deprecated and unused properties such as `AttributeNetTransport`. Should I remove the relevant ones or leave that for future cleanup?

src/OpenTelemetry.Instrumentation.AspNet/Implementation/RequestMethodHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        var suppliedKnownMethods = Environment.GetEnvironmentVariable(KnownHttpMethodsEnvironmentVariable)
+            ?.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+        if (suppliedKnownMethods?.Length > 0)
+        {
+            this.knownHttpMethods = suppliedKnownMethods.ToDictionary(x => x, x => x, StringComparer.OrdinalIgnoreCase);
+        }
+    }
+
+    public string GetNormalizedHttpMethod(string method)
```
utpilla, line 55: Make this static? You could make the whole class static.

qhris: That would not be possible in this case.
The way it is written it allows the unit tests to re-instatiate the class which will re-read the environment variables in order to test `OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS`.

Making it all static will make it untestable.

I think it is fine in this case because the class will only be instantiated twice (once in `HttpInMetricsListener` and in the future `HttpInListener`.

Kielek:
> This is a getting a bit long at this point, but I noticed that `SemanticConventions.cs` now has a lot of deprecated and unused properties such as `AttributeNetTransport`. Should I remove the relevant ones or leave that for future cleanup?

As it is in Shared folder, my vote is for separate PR.

src/OpenTelemetry.Instrumentation.AspNet/Implementation/RequestMethodHelper.cs
```diff
@@ -0,0 +1,61 @@
...
+internal sealed class RequestMethodHelper
+{
+    private const string KnownHttpMethodsEnvironmentVariable = "OTEL_INSTRUMENTATION_HTTP_KNOWN_METHODS";
+
+    // The value "_OTHER" is used for non-standard HTTP methods.
+    // https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-spans.md#common-attributes
+    private const string OtherHttpMethod = "_OTHER";
+
+    // List of known HTTP methods as per spec.
+    private readonly Dictionary<string, string> knownHttpMethods = new(StringComparer.OrdinalIgnoreCase)
```
vishweshbankwar, line 32: nit: instantiate in the ctor itself based on the value of env variable?

vishweshbankwar (approved):
LGTM: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1429/files#r1396542874 should be addressed before merging.

vishweshbankwar:
@qhris Could you please open an issue for following up on network.protocol.version testing?

src/OpenTelemetry.Instrumentation.AspNet/Implementation/RequestMethodHelper.cs
qhris: Yeah that makes sense.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -8,6 +8,23 @@
   * New overload of `AddAspNetInstrumentation` now accepts a configuration delegate.
   * The `Enrich` can be used to add additional metric attributes.
 
+* BREAKING: HTTP server metrics now follow stable
+  [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#http-server).
+
+  The `http.request.duration` metric is replaced with `http.server.request.duration`.
+  Note that the unit changes from milliseconds to seconds.
```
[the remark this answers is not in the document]
qhris: I'm so glad you posted this, I just realized I didn't check this properly when testing.

The buckets have not been updated, the implementation [here](<https://github.com/samlam/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Metrics/Metric.cs#L32>) does not actually take `OpenTelemetry.Instrumentation.AspNet` into consideration.

I updated the `CHANGELOG`, fixed it, and added tests for it locally now. But if we want it fixed upstream this PR will have to wait.
I also had to pull in `OpenTelemetry` as a dependency as that's the only way to access `AddView`.

qhris:
Please do not merge this until [this comment](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1429#discussion_r1397501852) is resolved.

qhris:
> @qhris Could you please open an issue for following up on network.protocol.version testing?

@vishweshbankwar I created an issue for making things more testable here: #1438.

src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
```diff
@@ -23,4 +23,8 @@
     <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule\OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
```
cijothomas, line 27: this is really against OpenTelemetry principles, and should not be done.
Please do either of the below:
1. Just documenting this in readme and ask users to add views themselves.
2. Request to make OTel sdk give special treatment to this one also. (I am supportive of this idea).

cijothomas (changes requested):
See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1429/files#r1397902272

Requesting to change the dependency on SDK before merge.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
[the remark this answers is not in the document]
vishweshbankwar: @qhris Good catch. We should not be adding SDK adding dependency to instrumentation as Cijo mentioned. Could you remove that? and send a PR upstream to include ASPNET in the list?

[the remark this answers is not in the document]
qhris: I made the suggested changes, and I also updated the buckets to remove the zero bucket as was done in https://github.com/open-telemetry/opentelemetry-dotnet/pull/5021.
I posted a link to the upstream PR in response to Cijo.

src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
qhris: Yes, I figured that was the case, I just wanted to make the tests pass earlier, I completely agree it's not a good idea.

I removed it and went with option 2 for now, I opened at issue at https://github.com/open-telemetry/opentelemetry-dotnet/pull/5063 for now.

I have removed the dependency again but leaving the tests to fail until we decide whether to wait on the upstream PR.

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
```diff
@@ -8,6 +8,29 @@
   * New overload of `AddAspNetInstrumentation` now accepts a configuration delegate.
   * The `Enrich` can be used to add additional metric attributes.
 
+* BREAKING: HTTP server metrics now follow stable
+  [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.0/docs/http/http-metrics.md#http-server)
+  ([#1429](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1429)).
+
+  * New metric: `http.server.request.duration`
+    * Unit: `s` (seconds)
+    * Histogram Buckets: `0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5,
```
cijothomas, line 17: Need to explicitly mention that this buckets are coming because OTel Sdk is hardwired to specially treat this metrics, and mention the starting version of OTel SDK as well. (once we know the version)

qhris: Sounds good to me, but I'll hold off until we know.

Kielek (approved):
@cijothomas, can we merge it with current wording?

cijothomas:
> @cijothomas, can we merge it with current wording?

Since we know the exact version of the SDK which is required to get correct histogram buckets, I'd suggest to call it out in the changelog.
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/CHANGELOG.md#170-rc1

src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
qhris: I've went ahead with the suggested changes.

qhris:
Fixed all the remaining issues but the tests won't pass until the SDK is updated.

utpilla:
@qhris Thanks for your patience!

PR #1435, opened 2023-11-16
Updated AWSSDK, Amazon packages to support newer versions

jakenuts:
The Instrumentation.AWS SDK packages are quite old (3 years) and using this package in most projects creates alot of warnings. I think the "[legacy, next_ver)" format will fix this (not certain). If not, perhaps they could just be updated the 3.7 generation.

## Changes

In OpenTelemetry.Instrumentation.AWS:
Updated AWSSDK.Core to accept versions 3.5.1.24 up to 4.x
Updated AWSSDK.SimpleNotificationService to accept versions 3.5.0.27 up to 4.x
Updated AWSSDK.SQS to accept versions 3.5.0.26 up to 4.x

In OpenTelemetry.Instrumentation.AWSLambda:
Updated Amazon.Lambda.APIGatewayEvents to accept versions 2.4.1 up to 3.x
Updated Amazon.Lambda.Core to accept versions 1.2.0 up to 3.x
Updated Amazon.Lambda.SNSEvents to accept versions 2.0.0 up to 3.x
Updated Amazon.Lambda.SQSEvents to accept versions 2.1..0 up to 3.x

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
+++ b/src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
@@ -9,9 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.5.1.24" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.5.0.27" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.5.0.26" />
+    <PackageReference Include="AWSSDK.Core" Version="[3.5.1.24,4.0)" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[3.5.0.27,4.0)" />
+    <PackageReference Include="AWSSDK.SQS" Version="[3.5.0.26, 4.0)" />
     <PackageReference Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
   </ItemGroup>
 
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
@@ -8,10 +8,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.1" />
-    <PackageReference Include="Amazon.Lambda.Core" Version="1.2.0" />
-    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.0.0" />
-    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="2.1.0" />
+    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="[2.4.1, 3.0)" />
+    <PackageReference Include="Amazon.Lambda.Core" Version="[1.2.0,3.0)" />
+    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="[2.0.0, 3.0)" />
+    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="[2.1.0, 3.0)" />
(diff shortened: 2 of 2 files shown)


Kielek:
@jakenuts, please sign EasyCLA agreement. We cannot proceed without this.

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
```diff
@@ -9,9 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.5.1.24" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.5.0.27" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.5.0.26" />
+    <PackageReference Include="AWSSDK.Core" Version="[3.5.1.24,4.0)" />
```
cijothomas, line 12: without this change also, these versions could have been used right? The title of the PR is misleading, as this is not adding support for newer versions, instead putting upper bound on the versions supported?

jakenuts, line 12: Yes, sorrry, I've only just discovered ranges (as opposed to wildcards) in PackageReference. My intent is to allow the use of this package with apps that are using the current 3.7.x AWS packages without warnings about 3.5 dependencies. Whatever achieves that goal while allowing the broadest usage is fine by me.

jakenuts: Correct, they can be used, but the fixed dependance on the outdated versions causes warnings in all of our projects so my intent was to open up the accepted versions to the current 3.7.x versions and above. I'm struggling a bit with both the package boundary notation and pr/codespace/traveling so please feel free to ditch this whole thing if it seems untenable. Just hoping to use 3.7 without endless notifications about 3.5.

cijothomas: For that goal, this PR should bump the minimum supported version from 3.5 to 3.7. I'll let the owners of this package decide if they are okay with that change.

utpilla:
@srprash @atshaw43 @rypdal @Oberon00 Please review.

jakenuts:
For reference, this is what I'm hoping to resolve.. minor but not great.
image

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
jakenuts: Sorry, hopelesly lost in codespaces/pr/screen glare while traveling. Would it be possible to open up the accepted range of sdk's to allow current ones (thereby letting the end user decide which version they'll use) rather than moving to the current version and fixing it there? Moving to 3.7 will resolve my issue but I worry that someone stuck on 3.5 will lose their access to updates. Could be I'm out of my depth on how those ranges affect things.

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
```diff
@@ -9,10 +9,10 @@
...
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="3.5.1.24" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="3.5.0.27" />
-    <PackageReference Include="AWSSDK.SQS" Version="3.5.0.26" />
-    <PackageReference Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
+    <PackageReference Include="AWSSDK.Core" Version="[3.5.1.24,)" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[3.5.0.27,)" />
+    <PackageReference Include="AWSSDK.SQS" Version="[3.5.0.26,)" />
+    <PackageReference Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-*" />
```
Oberon00, line 15: Did I miss something? AFAIK we never use `*` here

PR #1454, opened 2023-11-30
[Instrumentation.AWS] Update AWS SDK references

normj:
## Changes
Update the AWS .NET SDK dependencies for the OpenTelemetry.Instrumentation.AWS package. This is needed because the version that was previously had a max range cap of `3.6` but the SDK has been on version `3.7` since 2021. Without this change users are getting restore warnings whenever they are using an SDK versions released in the last 2 years.

For background purposes I'm a developer of the AWS .NET SDK.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Updated dependency on AWS .NET SDK to version 3.7.100.
+  ([#1454](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1454))
+
 ## 1.1.0-beta.1
 
 Released 2023-Aug-07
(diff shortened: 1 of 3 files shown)


Kielek:
@normj, please add entry to the changelog.

Oberon00:
>  had a max range cap of 3.6

Where does this max range cap come from? The reference here should be only a lower bound, see https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

normj:
> > had a max range cap of 3.6
> 
> Where does this max range cap come from? The reference here should be only a lower bound, see https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges
> 
> > ```
> > 
> > <PackageReference Include="ExamplePackage" Version="6.1" />
> > ```

The library is currently referencing version 3.5.0.26 of AWSSDK.SQS. If you look at the dependencies for that version it has a dependency on AWSSDK.Core set to `>= 3.5.1.24 && < 3.6.0`.  In more recent versions the service packages like AWSSDK.SQS have a much wider range set for the dependency on AWSSDK.Core. `>= 3.7.300 && < 4.0.0`.

https://www.nuget.org/packages/AWSSDK.SQS/3.5.0.26#dependencies-body-tab

normj:
> @normj, please add entry to the changelog.

What should I set the release date to for the changeling?

Oberon00 (approved):

cijothomas (approved):

normj:
@Kielek change log has been updated.

PR #1486, opened 2023-12-11
Update OTel SDK version to 1.7.0

utpilla:
## Changes
- Update OTel SDK version to 1.7.0

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #1488, opened 2023-12-11
Fix build errors

utpilla:
## Changes
- Fix errors introduced in #1486

### DIFF
--- a/src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj
+++ b/src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
-    
+
     <!--We need the package reference to `System.Net.Http` to use `HttpClientHandler.ServerCertificateCustomValidationCallback` property
     on net462 target. Starting net471, we could just add an assembly `Reference` instead of `PackageReference`.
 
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj
```diff
@@ -12,10 +12,16 @@
...
+    on net462 target. Starting net471, we could just add an assembly `Reference` instead of `PackageReference`.
 
-  <ItemGroup>
-    <Reference Include="System.Net.Http" Condition="'$(TargetFramework)' == '$(NetFrameworkMinimumSupportedVersion)'" />
+    For example:
+    <ItemGroup>
+      <Reference Include="System.Net.Http" Condition="'$(TargetFramework)' == '$(NetFrameworkMinimumSupportedVersion)'" />
+    </ItemGroup>
+    -->
+    <PackageReference Include="System.Net.Http" Version="4.3.4" Condition="'$(TargetFramework)' == 'net462'" />
```
utpilla, line 24: With `1.7.0` version of the SDK, this package now pulls in `8.0.0` version of `Microsoft.Extensions.Logging.Configuration` instead of `3.1.0` version that used to come with the previous dependency.

`Microsoft.Extensions.Logging.Configuration` has added a dedicated net462 target in `8.0.0` which does not have the `System.Net.Http` assembly included. That's why we need an explicit package dependency for `net462` targets now. With `3.1.0` version, there was no dedicated `net462` target, and the package ended up using `netstandard2.0` bits which includes the `System.Net.Http` assembly.

Follow this discussion for more details: https://github.com/open-telemetry/opentelemetry-dotnet/pull/5020#discussion_r1380902962

CodeBlanch (approved):
LGTM

I have the same fixes on #1487 but I can merge these in.

PR #1537, opened 2024-01-18
[Exporter.Geneva] Skip dispose for ETW singleton of GenevaMetricExporter

### LINKED ISSUE #1520: Geneva metric exporter disposes shared transport object

`OpenTelemetry.Exporter.Geneva` - 1.7.0 (the same issue happens with 1.6.0)

`net8.0`

* [ ] Feature Request
* [x] Bug

Consider this repro:
```csharp
using OpenTelemetry;
using OpenTelemetry.Exporter.Geneva;
using OpenTelemetry.Metrics;
using System.Diagnostics.Metrics;

await EmitMetrics("one");
await EmitMetrics("two");

static async Task EmitMetrics(string attempt)
{
		.AddMeter("*")
		.AddGenevaMetricExporter(x =>
		{
			x.MetricExportIntervalMilliseconds = 1000;
			x.ConnectionString = "Account=OTelGeneva;Namespace=MeteringSample";
		})
		.Build();

	using var meter = new Meter("GenevaMetricExporterBug");

	meter.CreateCounter<long>("counter_" + attempt)
		.Add(1);

	// Give the exporter some time to send the metrics
	await Task.Delay(2000).ConfigureAwait(false);
}
```

This is expected to emit metrics in both calls.

Second time metrics aren't emitted via Geneva exporter (works just fine with Console exporter though).

What did you see instead? If you are reporting a bug, create a self-contained
project using the template of your choice and apply the minimum required code to
result in the issue you're observing. We will close this issue if:

* The repro project you share with us is complex. We can't investigate custom
  projects, so don't point us to such, please.
* If we can not reproduce the behavior you're reporting.

## Additional Context

Geneva metric exporter [uses](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/Exporter.Geneva-1.6.0/src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs#L90) `Shared` instance for the transport and later [disposes](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/Exporter.Geneva-1.6.0/src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs#L284) its value in `metricDataTransport` field.

If this issue isn't fixed, at least would be great to have an option to control the behavior (e.g. a flag to disable shared instance).

xiang17:
Fixes #1520. The current code disposes the `MetricEtwDataTransport` singleton when an instance of `GenevaMetricExporter` is disposed. However, the singleton shouldn't be disposed. Future instances of `GenevaMetricExporter` will still need to reference it to send data to ETW.

This only happens on Windows, because ETW only allows one instance of EventSource with the same Guid (#1225). On Linux, such limit doesn't exist: `GenevaMetricExporter` uses Unix Domain Socket to send Metrics data.

## Changes

Fix a bug in `GenevaMetricExporter` where the `MetricEtwDataTransport` singleton is disposed.

Tested manually with the program in the issue. Didn't add tests in this PR since it's a bit tricky. We don't have any test of checking whether data is actually sent into ETW. Might have to use a [`EventListener`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventlistener?view=net-8.0) to read and parse data from the ETW.


src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -268,7 +268,11 @@ protected override void Dispose(bool disposing)
         {
             try
             {
-                this.metricDataTransport?.Dispose();
+                // MetricEtwDataTransport is a singleton and should not be disposed
+                if (this.metricDataTransport is not MetricEtwDataTransport)
```
reyang, line 272: Does this line of code reflecting the actual intention?

rajkumar-rangaraj:
Could you explain in the PR description how the current changes address the issue?

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -268,7 +268,11 @@ protected override void Dispose(bool disposing)
...
+                {
+                    this.metricDataTransport?.Dispose();
```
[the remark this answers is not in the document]
xiang17: My plan was to test with EventListener, however, I got some unexpected data (error message instead of the binary data sent into ETW) in the payload.

Sample code:
```
private class TestMetricEtwDataListener : EventListener
{
    private readonly Action<EventWrittenEventArgs> eventHandler;

    public TestMetricEtwDataListener(Action<EventWrittenEventArgs> eventHandler)
    {
        this.eventHandler = eventHandler;
    }

    protected override void OnEventSourceCreated(EventSource eventSource)
    {
        if (eventSource.Name == "OpenTelemetryGenevaMetricExporter")
        {
            this.EnableEvents(eventSource, EventLevel.Verbose);
        }

        base.OnEventSourceCreated(eventSource);
    }

    protected override void OnEventWritten(EventWrittenEventArgs eventData)
    {
        if (eventData.EventSource.Name != "OpenTelemetryGenevaMetricExporter")
        {
            return;
        }

        for (int i = 0; i < eventData.Payload.Count; ++i)
        {
            object obj = eventData.Payload[i];
            this.testOutputHelper.WriteLine($"Type is: `{obj.GetType()}`.");
            this.testOutputHelper.WriteLine($"EventData.Payload[{i}] is: `{obj}`.");

            Debug.WriteLine($"EventData.Payload[{i}] is {obj}");
        }
    }
}
```

The output is:
```
Type is: `System.String`.
EventData.Payload[0] is: `Event 70 was called with 1 argument(s), but it is defined with 0 parameter(s).`.
Type is: `System.String`.
EventData.Payload[0] is: `Event 70 was called with 1 argument(s), but it is defined with 0 parameter(s).`.
```

Looks like it's due to the way we define the `[Event]` attribute. (I found the definition of the error message and where the error is thrown. The definition of the error message is [here](https://github.com/microsoft/referencesource/blob/51cf7850defa8a17d815b4700b67116e3fa283c2/mscorlib/mscorlib.txt#L1023C1-L1023C37). And it's used [here](https://source.dot.net/#System.Private.CoreLib/artifacts/obj/coreclr/System.Private.CoreLib/linux.x64.Debug/System.SR.cs,98cb18dbb6afb4c7,references).)

The metrics exporter use TLV based serialization to send data to ETW. However, EventSource would send error message to EventListener's callback.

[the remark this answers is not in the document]
xiang17: I'll look for other ways to directly read the binary bytes of the ETW event without EventListener.

xiang17:
> Could you explain in the PR description how the current changes address the issue?

I've updated the description with details.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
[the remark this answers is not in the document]
xiang17: Considering this involves the external components (ETW mechanisms), it is more of an integration test to validate with an ETW consumer tool like `EventListener`.

I've added a unit test which focuses on that the `MetricEtwDataTransport` singleton instance is shared and NOT disposed among multiple instances of `GenevaMetricExporter`.

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
```diff
@@ -57,4 +58,22 @@ private void ExternallyAggregatedDoubleDistributionMetric()
     private void TLVMetricEvent()
     {
     }
+
+    /// <summary>
+    /// Disposes the object. This is only for Test purposes.
+    /// This should NOT be called during the lifetime of the thread because it
```
rajkumar-rangaraj, line 64:
```suggestion
    /// This method should NOT be called as there should only be one instance of this class throughout the application's lifecycle.
```

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -268,7 +268,12 @@ protected override void Dispose(bool disposing)
         {
             try
             {
-                this.metricDataTransport?.Dispose();
+                // The ETW data transport singleton on Windows should NOT be disposed.
```
vishweshbankwar, line 271:
```suggestion
                // The singleton instance used by the ETW data transport on Windows should NOT be disposed.
```

vishweshbankwar (approved):

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
```diff
@@ -57,4 +58,22 @@ private void ExternallyAggregatedDoubleDistributionMetric()
...
     }
+
+    /// <summary>
+    /// Disposes the object. This is only for Test purposes.
+    /// This method should NOT be called as there should only be one instance of
+    /// this class throughout the application's lifecycle.
+    /// Unit tests will check the value of IsDisposed to verify it's not called
+    /// by mistake.
+    /// </summary>
+    public new void Dispose()
```
reyang, line 69: This is anti-pattern.

reyang (changes requested):
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1537/files#r1467981314

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
reyang: Read https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
xiang17: It's not an instance used by "ETW data transport", the "ETW data transport" object here itself is the singleton object.

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
```diff
@@ -9,14 +9,17 @@ namespace OpenTelemetry.Exporter.Geneva;
...
+    private bool isDisposed;
 
     static MetricEtwDataTransport()
     {
-        Shared = new();
+        singleton = new();
     }
 
-    public static readonly MetricEtwDataTransport Shared;
+    public static MetricEtwDataTransport Shared { get => singleton; }
```
reyang, line 22: Need to change the variable names as "singleton" and "shared" are too broad/general.

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
xiang17: I've updated it according to the article: using `protected override void Dispose(bool)` to perform the actual cleanup of the derived class.

Now the code sets the reference to null so that later calls can fail fast if a bug had caused it to dispose early. The unit test checks this scenario and verifies that it's not disposed when an instance of `GenevaMetricExporter` is disposed.

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
xiang17: I can change the name to `MetricEtwDataTransportSingleton` or `MetricEtwDataTransportInstance`, but the the usage will be `this.metricDataTransport = MetricEtwDataTransport.MetricEtwDataTransportSingleton;`, which seems too redundant.

I can rename it to `Instance` as it is used many times as the name for a singleton in the opentelemetry-dotnet repo: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-dotnet+instance&type=code.

reyang: "Instance" is fine IMHO.

xiang17: I've updated it.

reyang: Do we need to have the field? I think just having the attribute is good enough (e.g. follow https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-dotnet+instance&type=code)

xiang17: I need to set it to null, so I need to add a setter, which is only needed as `private`.

`get; private set;` can do exactly that...

reyang (approved):

cijothomas (approved):

PR #1544, opened 2024-01-20
[Instrumentation.AWSLambda] Fix AoT warnings

### LINKED ISSUE #1523: Getting rid of Newtonsoft.Json dependency in OpenTelemetry.Instrumentation.AWSLambda
This is not a bug but future direction for all AWS libraries those are part of Opentelemetry. We would like to make them AOT compatible

While investigating, OpenTelemetry.Instrumentation.AWSLambda takes dependency on Newtonsoft.Json. Are there any restriction on using System.Text.Json/JsonSerializer.Deserialize ? If not, let me know happy to submit PR.

```C#
                // AWSMessagingUtils , GetSnsMessage
                //snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
                snsMessage = JsonSerializer.Deserialize<SNSEvent.SNSMessage>(body);
```

martincostello:
- Fixes #1523.
- Depends on #1545.

## Changes

Make OpenTelemetry.Instrumentation.AWSLambda AoT compatible by:
- Using System.Text.Json to deserialize SNS messages
- Updating to Amazon.Lambda packages that support native AoT.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -11,6 +11,10 @@
   ([#1295](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1295))
 * BREAKING: Target `net6.0` instead of `netstandard2.0`
   ([#1545](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1545))
+* Add support for native AoT.
+  `Amazon.Lambda.*` NuGet package dependencies have been upgraded, see package
+  dependencies for details.
+  ([#1544](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1544))
 
 ## 1.2.0-beta.1
 
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
@@ -5,9 +5,9 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Text.Json;
 using Amazon.Lambda.SNSEvents;
 using Amazon.Lambda.SQSEvents;
-using Newtonsoft.Json;
 using OpenTelemetry.Context.Propagation;
 
 namespace OpenTelemetry.Instrumentation.AWSLambda.Implementation;
@@ -132,7 +132,7 @@ internal class AWSMessagingUtils
         {
             try
             {
-                snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
+                snsMessage = JsonSerializer.Deserialize(body, SourceGenerationContext.Default.SNSMessage);
             }
             catch (Exception)
             {
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
@@ -8,12 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.1" />
-    <PackageReference Include="Amazon.Lambda.Core" Version="1.2.0" />
-    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.0.0" />
-    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="2.1.0" />
+    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.7.0" />
+    <PackageReference Include="Amazon.Lambda.Core" Version="2.2.0" />
+    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.1.0" />
+    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="2.2.0" />
     <PackageReference Include="OpenTelemetry.Extensions.AWS" Version="1.3.0-beta.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/SourceGenerationContext.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/SourceGenerationContext.cs
@@ -1,0 +1,20 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Text.Json.Serialization;
+using Amazon.Lambda.SNSEvents;
+
+namespace OpenTelemetry.Instrumentation.AWSLambda;
+
+/// <summary>
+/// "Source Generation" is feature added to System.Text.Json in .NET 6.0.
+/// This is a performance optimization that avoids runtime reflection when performing serialization.
+/// Serialization metadata will be computed at compile-time and included in the assembly.
+/// <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/source-generation-modes" />.
+/// <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/source-generation" />.
+/// <see href="https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/" />.
+/// </summary>
+[JsonSerializable(typeof(SNSEvent.SNSMessage))]
+internal sealed partial class SourceGenerationContext : JsonSerializerContext
+{
+}
--- a/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj
+++ b/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj
@@ -14,7 +14,6 @@
       .github\workflows\ci.yml so that it runs for the project being added.
     -->
     <TrimmerRootAssembly Include="OpenTelemetry.Extensions" />
-    <TrimmerRootAssembly Include="OpenTelemetry.Extensions.AWS" />
     <TrimmerRootAssembly Include="OpenTelemetry.Extensions.Enrichment" />
     <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.AWSLambda" />
     <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.EventCounters" />
--- a/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj
+++ b/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj
@@ -16,8 +16,9 @@
     <TrimmerRootAssembly Include="OpenTelemetry.Extensions" />
     <TrimmerRootAssembly Include="OpenTelemetry.Extensions.AWS" />
     <TrimmerRootAssembly Include="OpenTelemetry.Extensions.Enrichment" />
-    <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.Runtime" />
+    <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.AWSLambda" />
     <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.EventCounters" />
+    <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.Runtime" />
     <TrimmerRootAssembly Include="OpenTelemetry.Instrumentation.StackExchangeRedis" />
     <TrimmerRootAssembly Include="OpenTelemetry.ResourceDetectors.AWS" />
     <TrimmerRootAssembly Include="OpenTelemetry.ResourceDetectors.Azure" />


src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
```diff
@@ -127,12 +139,22 @@ internal static PropagationContext ExtractParentContext(SNSEvent.SNSMessage? mes
...
             body.TrimStart().StartsWith("{", StringComparison.Ordinal) &&
             body.Contains(SnsMessageAttributes))
+#endif
         {
             try
             {
-                snsMessage = JsonConvert.DeserializeObject<SNSEvent.SNSMessage>(body);
+#if NET6_0_OR_GREATER
+                using var stream = new MemoryStream(System.Text.Encoding.UTF8.GetBytes(body));
+                snsMessage = LambdaSerializer.Deserialize<SNSEvent.SNSMessage>(stream);
```
martincostello, line 154: I went this route as the Lambda serializer [has custom options to deal with the serialization](https://github.com/aws/aws-lambda-dotnet/blob/a2421013def1e938d24b3e9f9c3453868d77a360/Libraries/src/Amazon.Lambda.Serialization.SystemTextJson/AbstractLambdaJsonSerializer.cs#L129-L150), so I figured it was better to use that than just the vanilla source generator support. That requires the string to be turned into a `Stream` to use though.

src/OpenTelemetry.Instrumentation.AWSLambda/SourceGenerationContext.cs
```diff
@@ -0,0 +1,22 @@
...
+
+/// <summary>
+/// "Source Generation" is feature added to System.Text.Json in .NET 6.0.
+/// This is a performance optimization that avoids runtime reflection when performing serialization.
+/// Serialization metadata will be computed at compile-time and included in the assembly.
+/// <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/source-generation-modes" />.
+/// <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/source-generation" />.
+/// <see href="https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/" />.
+/// </summary>
+[JsonSerializable(typeof(SNSEvent.SNSMessage))]
```
martincostello, line 18: The tests pass, but there's no `[JsonPropertyName]` attributes on this class, so I'm not 100% sure if the serialization is right or not.

martincostello:
@normj Do you have any insight into the deserialization of the payload here being right in the context of the changes here to use the source generator?

Oberon00:
@martincostello There is a unit test with an example payload here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/8a89a69445a1f89917a6a159e895b9267c7cc925/test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSMessagingUtilsTests.cs#L62-L101
If that test passes, the deserialization should actually be good, but feel free to double-check

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -23,9 +23,9 @@ public static class AWSLambdaWrapper
     private static readonly AssemblyName AssemblyName = typeof(AWSLambdaWrapper).Assembly.GetName();
 
     [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1202:ElementsMustBeOrderedByAccess", Justification = "Initialization order.")]
-    internal static readonly string ActivitySourceName = AssemblyName.Name;
+    internal static readonly string ActivitySourceName = AssemblyName.Name!;
```
Oberon00, line 26: Are you sure about this? Isn't this actually `null` in AoT mode? CC @rypdal

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -88,17 +88,17 @@ internal static string GetCloudProvider()
 
     internal static string GetAWSRegion()
     {
-        return Environment.GetEnvironmentVariable(AWSRegion);
+        return Environment.GetEnvironmentVariable(AWSRegion) ?? string.Empty;
```
Oberon00, line 91: Are you sure this is safe? Wouldn't it be safer to use `string?` as return value? Or did you check all callers?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
Oberon00: That seems overly complex. The custom options are not needed for the two properties we fetch, so maybe go with the simpler change like in #1270?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
normj: In Lambda the AWS_REGION environment variable will never not be set but in a test environment it could not be set. I'm with @Oberon00 that changing the return type to `string?` is the safer choice and more in the spirit of nullability.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
```diff
@@ -1,19 +1,26 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>net6.0;netstandard2.0</TargetFrameworks>
```
normj, line 4: Given .NET Core 3.1 is well out of support is there a reason .NET Standard 2.0 still needs to be targeted? That would allow you to get rid of some if the conditional compilation blocks and reduce the complexity.

src/OpenTelemetry.Instrumentation.AWSLambda/SourceGenerationContext.cs
normj: You generally only need to use `JsonPropertyName` on response objects where the casing on the outgoing JSON matters. When going from JSON to POCO we enable case insensitive matches in the serializer.

normj:
> @normj Do you have any insight into the deserialization of the payload here being right in the context of the changes here to use the source generator?

Sorry I'm not understanding the question. Are you asking if the `SourceGenerationContext` is setup correctly for deserializing the `SNSEvent.SNSMessage` type correctly? If so yes that looks right for using source generator serializer.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
Oberon00: Actually a good point. We did want to get rid of netstandard2.0 already, but delayed it because we wanted to have a particular bugfix in a backwards compatible release. But by now I think we could upgrade to at least `netcoreapp3.1` as done in #1270, although I also see no reason not to upgrade to 6.0 (AWS already has that as minimum).

Would be great if we could get rid of all the `#ifdef`s and have only a single target framework specifier.

martincostello:
Yep, that was the question.

Thanks for the feedback both, I'll address these today.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
martincostello: I'd prefer to not do that in the same PR.

Would you rather:
1. I do another PR to change `netstandard2.0` to `net6.0` and then rebase this one on top of it.
2. Merge this (once feedback addressed) as-is, then I do a follow-up to remove `netstandard2.0`.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSMessagingUtils.cs
martincostello: I originally added this for fear of breaking the serialization after doing it the simpler way at first, but double-checking with the test, the easy way works fine, so I'll revert it back.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -23,10 +23,10 @@ public static class AWSLambdaWrapper
     private static readonly AssemblyName AssemblyName = typeof(AWSLambdaWrapper).Assembly.GetName();
 
     [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1202:ElementsMustBeOrderedByAccess", Justification = "Initialization order.")]
-    internal static readonly string ActivitySourceName = AssemblyName.Name;
+    internal static readonly string? ActivitySourceName = AssemblyName.Name;
 
-    private static readonly Version Version = AssemblyName.Version;
-    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version.ToString());
+    private static readonly Version? Version = AssemblyName.Version;
+    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName ?? "OpenTelemetry.Instrumentation.AWSLambda", Version?.ToString());
```
martincostello, line 29: `ActivitySource.ctor()` [allows the second parameter to be `null`](https://learn.microsoft.com/dotnet/api/system.diagnostics.activitysource.-ctor), so we should be able to safely ignore if we can't get a version.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
```diff
@@ -14,12 +14,12 @@ internal sealed class AWSLambdaResourceDetector : IResourceDetector
     /// <returns>Detected resource.</returns>
     public Resource Detect()
     {
-        var resourceAttributes = new List<KeyValuePair<string, object>>
+        var resourceAttributes = new List<KeyValuePair<string, object>>(4)
```
martincostello, line 17: Resource requires `<string, object>` and not `<string, object?>`, so I went with setting the values to `string.Empty` though they would be `null` in TFMs without NRT I believe.

Alternatively, I could change this to only add the values to the list if they aren't null, but that might trip people up if they didn't use a guard like `TryGetValue()`.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
Oberon00: OK, it won't throw, but we still lose the version info

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
Oberon00: Sending empty strings is not a convention that is usually applied in OTel, so I would go with only adding the values if they are not null.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
martincostello: Conversation should move to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1545#discussion_r1461683945

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
martincostello: Addressed in #1545.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
martincostello: #1545

martincostello:
Have rebased onto #1545 and will rebase onto main once that is merged.

Just the native AoT changes can be reviewed here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1544/commits/f7220fd75fdacc124b30ccc9f4c8ec0980be4c36

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
   ([#1295](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1295))
 * BREAKING: Target `net6.0` instead of `netstandard2.0`
   ([#1545](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1545))
+* Fix native AoT warnings
+  ([#1543](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1543))
```
martincostello, line 15:
```suggestion
  ([#1544](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1544))
```

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
```diff
@@ -8,12 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.4.1" />
-    <PackageReference Include="Amazon.Lambda.Core" Version="1.2.0" />
-    <PackageReference Include="Amazon.Lambda.SNSEvents" Version="2.0.0" />
-    <PackageReference Include="Amazon.Lambda.SQSEvents" Version="2.1.0" />
+    <PackageReference Include="Amazon.Lambda.APIGatewayEvents" Version="2.7.0" />
+    <PackageReference Include="Amazon.Lambda.Core" Version="2.2.0" />
```
Oberon00, line 12: Is raising the lower bound of these dependencies necessary? In particular this one from major version 1 to 2?

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
   ([#1295](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1295))
 * BREAKING: Target `net6.0` instead of `netstandard2.0`
   ([#1545](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1545))
+* Fix native AoT warnings
```
Oberon00, line 14: Warning fixes are not relevant for end users, but if the warnings were relevant, would it be true to say "Add support for AoT compilation"?

martincostello:
```suggestion
* Add support for native AoT.
```

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
martincostello: These versions all correlate to the versions that first introduced support for native AoT: [CHANGELOG](https://github.com/aws/aws-lambda-dotnet/blob/master/RELEASE.CHANGELOG.md#release-2023-10-26).

Version 2.1.0 of Amazon.Lambda.Core is the first version that added explicit support for `net6.0`, which was required by the changes asked for in #1545, and the only release since then was the 2.2.0 release for `net8.0` support that resolves the native AoT warnings.

Based on that, I think these are reasonable increases to require of users while also making this assembly usable without trim warnings in native AoT applications. There were also already breaking changes listed in the CHANGELOG for this library before this PR or #1545.

Thoughts @normj?

Oberon00: > while also making this assembly usable without trim warnings in native AoT applications

I wonder, because I assume the following:

1. Users of the instrumentation _should_ still have their own dependencies on Lambda packages and not rely on transitive dependencies here (unfortunately no implementation/api distinction exists in Nuget like in Java Gradle modules)
2. If users want AoT, they can & should upgrade their own dependencies as needed. The library we release only specifies a lower bound.

Based on what you write with the .NET 6 support, an upgrade to 2.1 is totally reasonable without question, but I wonder about 2.2.

martincostello: I can downgrade it a bit in my native AoT Lambda using .NET 8 and see if I get any IL warnings (https://github.com/martincostello/alexa-london-travel/pull/1020) or not.

My library does already specify its own dependencies for the Lambda assemblies [it needs](https://github.com/martincostello/alexa-london-travel/blob/d82a420645cfc52e0c9f3615bf3a7665e52ae5b8/Directory.Packages.props#L6-L8), _but_ I don't use API Gateway, SNS or SQS, but I get all of those dependencies via this library transiently. Needing the end-user to manually upgrade the transient dependencies to get it to work properly for native AoT isn't ideal.

martincostello:
AoT failure will be resolved when #1540 is merged.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
Oberon00: >  but I don't use API Gateway, SNS or SQS, but I get all of those dependencies via this library transiently

Good point. This is not ideal anyway. I don't think there is a way to have these as compile-only dependencies and then check at runtime if they are actually present?

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
martincostello: Norm might know some tricks for dealing with that as the core AWS SDKs have to do similar things for things like STS authentication via the built-in fallback credential providers, but yeah that's a bit of a spiral out of the original scope here 😅.

normj: Sorry I can't think of any special tricks here. The only difference between 2.1 and 2.2 for Amazon.Lambda.Core is updates to the project file to have the .NET 8 target and mark the package trimmable for .NET 8 if that makes you feel better taking the updated dependency.

normj (approved):
As a non-component owner the changes look good.

Kielek (approved):

birojnayak:
good to see this is closed ...

PR #1545, opened 2024-01-22
[Instrumentation.AWSLambda] Update to net6.0

martincostello:
Contributes to #1523.

## Changes

Spun off from #1544.

Target `net6.0` instead of `netstandard2.0` as AWS Lambda no longer supports versions of .NET less than .NET 6.


src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -23,10 +23,10 @@ public static class AWSLambdaWrapper
     private static readonly AssemblyName AssemblyName = typeof(AWSLambdaWrapper).Assembly.GetName();
 
     [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1202:ElementsMustBeOrderedByAccess", Justification = "Initialization order.")]
-    internal static readonly string ActivitySourceName = AssemblyName.Name;
+    internal static readonly string? ActivitySourceName = AssemblyName.Name;
 
-    private static readonly Version Version = AssemblyName.Version;
-    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version.ToString());
+    private static readonly Version? Version = AssemblyName.Version;
+    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName ?? "OpenTelemetry.Instrumentation.AWSLambda", Version?.ToString());
```
martincostello, line 29: Open question on what to do if `Version` is `null`: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1544#discussion_r1461645540

Oberon00: What do other components do here?
If AoT does not provide this version info, should we auto-generate a C# file with msbuild during build to hold the version info?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
```diff
@@ -14,14 +14,26 @@ internal sealed class AWSLambdaResourceDetector : IResourceDetector
...
-        var resourceAttributes = new List<KeyValuePair<string, object>>
+        var resourceAttributes = new List<KeyValuePair<string, object>>(4)
         {
             new(AWSLambdaSemanticConventions.AttributeCloudProvider, AWSLambdaUtils.GetCloudProvider()),
-            new(AWSLambdaSemanticConventions.AttributeCloudRegion, AWSLambdaUtils.GetAWSRegion()),
-            new(AWSLambdaSemanticConventions.AttributeFaasName, AWSLambdaUtils.GetFunctionName()),
-            new(AWSLambdaSemanticConventions.AttributeFaasVersion, AWSLambdaUtils.GetFunctionVersion()),
         };
 
+        if (AWSLambdaUtils.GetAWSRegion() is { } region)
```
Oberon00, line 22: This looks arcane, would this work too?

```suggestion
        if (AWSLambdaUtils.GetAWSRegion() is not null region)
```

martincostello: Yes it would, they mean the same thing. I'll change them.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
martincostello: I don't know - I'll have to look into whether any other of the libraries here have already hit the same thing.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
martincostello: Actually, no, that isn't valid C#.

Options are:

1. What was already there;
2. The commit I just pushed to do `{ Length: > 0 }` instead;
3. Explicitly declare the variables and check them (more verbose IMHO).

```csharp
var region = AWSLambdaUtils.GetAWSRegion();
if (region is not null)
{
    resourceAttributes.Add(new(AWSLambdaSemanticConventions.AttributeCloudRegion, region));
}

// etc
```

Oberon00: Sorry, if `is not null` does not work, I'm totally fine with the `{ }`, it's probably an idiom people who write more modern C# than me are familiar with 😃

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
martincostello: Ignoring tests, the libraries fall into two camps:

1. Don't specify a version at all.
2. Pass through the assembly version, which won't work for native AoT as noted.

image

I can test it out separately, but I think if the version wants to be kept, the better solution is to investigate whether native AoT retains [`AssemblyFileVersion]`/ `[AssemblyInformationalVersion]` to use instead.

This is already present in the assembly according to [nuget.info](https://nuget.info/packages/OpenTelemetry.Instrumentation.AWSLambda/1.2.0-beta.1):

image

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
martincostello: OK, in that case I'll change it back as it's terser.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -20,13 +19,10 @@ namespace OpenTelemetry.Instrumentation.AWSLambda;
...
-    private static readonly AssemblyName AssemblyName = typeof(AWSLambdaWrapper).Assembly.GetName();
+    internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.AWSLambda";
 
-    [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1202:ElementsMustBeOrderedByAccess", Justification = "Initialization order.")]
-    internal static readonly string ActivitySourceName = AssemblyName.Name;
-
-    private static readonly Version Version = AssemblyName.Version;
-    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version.ToString());
+    private static readonly string Version = typeof(AWSLambdaWrapper).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()!.Version;
+    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version);
```
martincostello, line 25: I'll double-check in an application of mine that uses native AoT this won't get trimmed away now, but I think this is the best way to resolve this question.

`Assembly.GetName()` looks like mainly a vehicle to get the name and version, but that won't work for native AoT.

If `[AssemblyFileInfo]` is not trimmed, we can use that for the version, and changing the name of the assembly is going to be a breaking change (and probably won't change again?) so should be safe enough to hard-code. It also ensures that this cod never gets a `null`, which would break stuff if it did:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/8a89a69445a1f89917a6a159e895b9267c7cc925/src/OpenTelemetry.Instrumentation.AWSLambda/TracerProviderBuilderExtensions.cs#L42

martincostello: Yep, getting the assembly attribute appears to work.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -20,13 +19,10 @@ namespace OpenTelemetry.Instrumentation.AWSLambda;
...
 {
-    private static readonly AssemblyName AssemblyName = typeof(AWSLambdaWrapper).Assembly.GetName();
+    internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.AWSLambda";
 
-    [SuppressMessage("StyleCop.CSharp.OrderingRules", "SA1202:ElementsMustBeOrderedByAccess", Justification = "Initialization order.")]
-    internal static readonly string ActivitySourceName = AssemblyName.Name;
-
-    private static readonly Version Version = AssemblyName.Version;
-    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version.ToString());
+    private static readonly string Version = typeof(AWSLambdaWrapper).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()!.Version;
```
Kielek, line 24: Consider `AssemblyInformationalVersion.Split('+')[0]`.

eg. if you have versions 1.2.5-beta.1 this attribute should be equal to `1.2.5.beta.1+commit_hash`

Ref: https://github.com/adamralph/minver?tab=readme-ov-file#version-numbers and https://github.com/dotnet/sdk/issues/34568

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
rypdal: I wonder if the call `Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()!.Version `still works and what it returns for native AoT. Did you test it just with a simple AoT application ? Hopefully this list is complete: https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/overview?tabs=cli#api-incompatibility

martincostello: I have a native AoT published ASP.NET Core application which returns metadata about its execution context, and this works correctly:

```csharp
        endpoints.MapGet("/configuration/version", () =>
        {
            return new JsonObject()
            {
                ["frameworkDescription"] = JsonValue.Create(RuntimeInformation.FrameworkDescription),
                ["operatingSystem"] = new JsonObject()
                {
                    ["description"] = JsonValue.Create(RuntimeInformation.OSDescription),
                    ["architecture"] = JsonValue.Create(RuntimeInformation.OSArchitecture.ToString()),
                    ["version"] = JsonValue.Create(Environment.OSVersion.VersionString),
                    ["is64Bit"] = JsonValue.Create(Environment.Is64BitOperatingSystem),
                },
                ["process"] = new JsonObject()
                {
                    ["architecture"] = JsonValue.Create(RuntimeInformation.ProcessArchitecture.ToString()),
                    ["is64BitProcess"] = JsonValue.Create(Environment.Is64BitProcess),
                    ["isNativeAoT"] = JsonValue.Create(!RuntimeFeature.IsDynamicCodeSupported),
                    ["isPrivilegedProcess"] = JsonValue.Create(Environment.IsPrivilegedProcess),
                },
                ["dotnetVersions"] = new JsonObject()
                {
                    ["runtime"] = JsonValue.Create(GetVersion<object>()),
                    ["aspNetCore"] = JsonValue.Create(GetVersion<HttpContext>()),
                },
            };
        }).ExcludeFromDescription();

        static string GetVersion<T>()
            => typeof(T).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()!.InformationalVersion;
```

I adapted the code to check `[AssemblyFileVersion]` earlier - can test it again if necessary with `AssemblyInformationalVersionAttribute` specifically from this assembly.

martincostello: The response:

```json
{
  "frameworkDescription": ".NET 8.0.1",
  "operatingSystem": {
    "description": "Ubuntu 22.04.3 LTS",
    "architecture": "X64",
    "version": "Unix 5.10.184.175",
    "is64Bit": true
  },
  "process": {
    "architecture": "X64",
    "is64BitProcess": true,
    "isNativeAoT": true,
    "isPrivilegedProcess": false
  },
  "dotnetVersions": {
    "runtime": "8.0.1+bf5e279d9239bfef5bb1b8d6212f1b971c434606",
    "aspNetCore": "8.0.1+8e941eb42f819adb116b881195158b3887a70a1c"
  }
}
```

rypdal: If it works for `AssemblyInformationalVersionAttribute` then should be fine for `AssemblyFileVersion`. No need for double check.

rypdal (approved):

cijothomas (approved):

PR #1572, opened 2024-02-06
bugfix: AWS SQS MD5 Hash Mismatch

### LINKED ISSUE #1492: SendMessage with MessageAttributes on latest version of AWS SDK throws "Attribute MD5 hash mismatch" exception
# Issue with OpenTelemetry.Instrumentation.AWS

* OpenTelemetry.Instrumentation.AWS 1.1.0-beta.2
* AWSSDK.SQS 3.7.300.11

* net7.0 and net48

* [ ] Feature Request
* [X] Bug

Messages are sent without exceptions.

When we use `.AddAWSInstrumentation()` and then call `IAmazonSQS.SendMessageAsync` with a message with custom MessageAttributes, an `AmazonSQSException` exception is thrown with the message "Attribute MD5 hash mismatch".

## Additional Context

I think that problem lies in `SqsRequestContextHelper.AddAttribute`, where the attributes are added to the `IRequestContext` parameters that is not compatible with the JSON protocol.

In the most recent versions of "AWSSDK.SQS", the SDK changed from the old XML protocol to the new JSON protocol (see https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-sqs-support-json-protocol/).

It would seem the attributes are not properly added to the request, so SQS produces an MD5 hash that does not include the attributes added by OTEL. This doesn't match the MD5 hash generated by the client.

We can workaround by reverting to an older version of "AWSSDK.SQS", but that is not desirable.

ppittle:
Fixes #1492 

## Changes

A recent change to the AWS SDK implementation of SQS client changes wireline protocol from XML to Json.  This causes the SDK's internal request pipeline to behave differently, causing `AWSTracingPipelineHandler` to no longer be able to manipulate `SendMessageRequest.MessageAttributes` after the Marshalling step has been completed.  

This change moves `AWSTracingPipelineHandler` to run before Marshalling (ie Serialization) so that Open Telemetry headers can be correctly injected.

Injecting propagation context into outgoing requests have been moved to a new handler, `AWSPropagatorPipelineHandler`, which runs later in the SDK request pipeline.

### Service Name Mapping

Moving `AWSTracingPipelineHandler` ahead in the AWS SDK Pipeline made it necessary to also tweak `AWSServiceHelper` to use `requestContext.ServiceMetaData.ServiceId` instead of `requestContext.Request.ServiceName` as the `Request` has not yet been populated at this point.  

ServiceId and ServiceName can differ for some AWS Services, requiring an update to [`AWSServiceType`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1572/files/aaf6fbabedc770a2cc6848ec5bf25851cc98a87b#diff-e1335da399a2c5c58f572a860866987f5235e287da0e3730a0fb6c9c500dbe56).  I have manually verified all 3 Service Names in `AWSServiceType` for correctness.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
@@ -21,7 +21,7 @@ internal class AWSServiceHelper
     };
 
     internal static string GetAWSServiceName(IRequestContext requestContext)
-        => Utils.RemoveAmazonPrefixFromServiceName(requestContext.Request.ServiceName);
+        => Utils.RemoveAmazonPrefixFromServiceName(requestContext.ServiceMetaData.ServiceId);
 
     internal static string GetAWSOperationName(IRequestContext requestContext)
     {
--- a/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceType.cs
+++ b/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceType.cs
@@ -7,9 +7,9 @@ namespace OpenTelemetry.Instrumentation.AWS.Implementation;
 
 internal class AWSServiceType
 {
-    internal const string DynamoDbService = "DynamoDBv2";
+    internal const string DynamoDbService = "DynamoDB";
     internal const string SQSService = "SQS";
-    internal const string SNSService = "SimpleNotificationService"; // SNS
+    internal const string SNSService = "SNS";
 
     internal static bool IsDynamoDbService(string service)
         => DynamoDbService.Equals(service, StringComparison.OrdinalIgnoreCase);
--- a/test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
+++ b/test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
@@ -174,6 +174,7 @@ public class TestAWSClientInstrumentation
             var send_msg_req = new SendMessageRequest();
             send_msg_req.QueueUrl = "https://sqs.us-east-1.amazonaws.com/123456789/MyTestQueue";
             send_msg_req.MessageBody = "Hello from OT";
+            send_msg_req.MessageAttributes.Add("Custom", new MessageAttributeValue { StringValue = "Value", DataType = "String" });
 #if NETFRAMEWORK
             sqs.SendMessage(send_msg_req);
 #else
@@ -198,8 +199,8 @@ public class TestAWSClientInstrumentation
 
     private void ValidateDynamoActivityTags(Activity ddb_activity)
     {
-        Assert.Equal("DynamoDBv2.Scan", ddb_activity.DisplayName);
-        Assert.Equal("DynamoDBv2", Utils.GetTagValue(ddb_activity, "aws.service"));
+        Assert.Equal("DynamoDB.Scan", ddb_activity.DisplayName);
+        Assert.Equal("DynamoDB", Utils.GetTagValue(ddb_activity, "aws.service"));
         Assert.Equal("Scan", Utils.GetTagValue(ddb_activity, "aws.operation"));
         Assert.Equal("us-east-1", Utils.GetTagValue(ddb_activity, "aws.region"));
         Assert.Equal("SampleProduct", Utils.GetTagValue(ddb_activity, "aws.table_name"));
--- a/test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
+++ b/test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
@@ -16,6 +16,7 @@ namespace OpenTelemetry.Instrumentation.AWS.Tests;
 internal class TestRequest : IRequest
 {
     private readonly ParameterCollection parameters;
+
     public TestRequest(ParameterCollection? parameters = null)
     {
         this.parameters = parameters ?? new ParameterCollection();
--- a/test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
+++ b/test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
@@ -13,9 +13,13 @@ using Amazon.Runtime.Internal.Util;
 
 namespace OpenTelemetry.Instrumentation.AWS.Tests;
 
-internal class TestRequest(ParameterCollection parameters) : IRequest
+internal class TestRequest : IRequest
 {
-    private readonly ParameterCollection parameters = parameters;
+    private readonly ParameterCollection parameters;
+    public TestRequest(ParameterCollection? parameters = null)
+    {
+        this.parameters = parameters ?? new ParameterCollection();
+    }
 
     public string RequestName => throw new NotImplementedException();
 
--- a/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* BREAKING: Switched AWSServiceName tag to use ServiceId instead of ServiceName
+  ([#1572](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1572))
+
 ## 1.1.0-beta.3
 
 Released 2024-Jan-26
--- a/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSPropagatorPipelineHandler.cs
+++ b/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSPropagatorPipelineHandler.cs
@@ -1,0 +1,68 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Amazon.Runtime;
+using Amazon.Runtime.Internal;
+using OpenTelemetry.Context.Propagation;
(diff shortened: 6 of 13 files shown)


ppittle:
Would it be possible to add @normj or @birojnayak as optional reviewers?

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
```diff
@@ -21,7 +21,7 @@ internal class AWSServiceHelper
     };
 
     internal static string GetAWSServiceName(IRequestContext requestContext)
-        => Utils.RemoveAmazonPrefixFromServiceName(requestContext.Request.ServiceName);
+        => Utils.RemoveAmazonPrefixFromServiceName(requestContext.ServiceMetaData.ServiceId);
```
normj, line 24: I would suggest giving context for this change in the PR description so people not familiar with AWS SDK details understand this is a safe change.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceType.cs
```diff
@@ -7,9 +7,9 @@ namespace OpenTelemetry.Instrumentation.AWS.Implementation;
 
 internal class AWSServiceType
 {
-    internal const string DynamoDbService = "DynamoDBv2";
+    internal const string DynamoDbService = "DynamoDB";
```
normj, line 10: Same comment as above, give context in the PR description that these changes are due to the switch in using `ServiceMetaData.ServiceId` and by changing these constants it makes the `IsDynamoDbService` and `IsSnsService` work the same when using `ServiceName`.

normj (approved):
Looks good, I recommend adding more context to the PR description for other reviewers that are not as familiar with the internals of the AWS .NET SDK.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
ppittle: 👍 
update PR description above.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceType.cs
ppittle: 👍 
update PR description above.

birojnayak:
@ppittle let's try to add a test if it's possible ? Customer has reported this as bug, so see if we can add a test.

ppittle:
> @ppittle let's try to add a test if it's possible ? Customer has reported this as bug, so see if we can add a test.

I have reworked the unit tests to reflect the change in implementation.

I also looked at adding a proper regression test - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1572/files#diff-e5d83ef47b6e8c2bac87983d6e1ede500c9c42506c262afb183d4ff7bb557db2R177

However, this doesn't fully cover the issue raised.  The exception is caused because the older version didn't send all Attributes to the SQS Service.  When SQS replied with the hash of what it had received, this did not match the hash calculated by the SDK and an exception was raised by the user.

The unit tests intercept the SDK Pipeline to [Set a Mock Web Response](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/89ff089bf07fe8576e7cf2027a020a305537ad91/test/OpenTelemetry.Instrumentation.AWS.Tests/Tools/CustomResponses.cs#L72) rather than making a live call to a SQS instance.  Covering this issue in a unit test would require the test to precalculate the expected hash and inject that value into the mock response.  In my opinion, the maintenance cost for that test code is high and would not provide a high degree of certainty of being able to detect future changes to the SDK Pipeline's internals.  

It would be more appropriate to have an integration test, but without an AWS Account to test against, this is also not a feasible option.

I recommend we leave the test coverage as it is, but I welcome a second opinion

test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
```diff
@@ -198,8 +199,8 @@ private void ValidateAWSActivity(Activity aws_activity, Activity parent)
 
     private void ValidateDynamoActivityTags(Activity ddb_activity)
     {
-        Assert.Equal("DynamoDBv2.Scan", ddb_activity.DisplayName);
-        Assert.Equal("DynamoDBv2", Utils.GetTagValue(ddb_activity, "aws.service"));
+        Assert.Equal("DynamoDB.Scan", ddb_activity.DisplayName);
```
birojnayak, line 202: why this assert changed ?

ppittle: The [AWSServiceName tag is set](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/89ff089bf07fe8576e7cf2027a020a305537ad91/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs#L232) based on the result from calling [`AWSServiceHelper.GetAWSServiceName`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/89ff089bf07fe8576e7cf2027a020a305537ad91/src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs#L215).  The [implementation of `GetAWServiceName` was changed in this PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1572/files/31344ee2876fb471c3de5719afd3cad383c7a078#diff-f63d829e66f547c60534775fb2216f1fe15ea5e74d04b3233d540926ef29fc31L24) to use `ServiceMetaData.ServiceId` instead of `requestContext.Request.ServiceName`; moving the handler code earlier in the SDK Pipeline means the `requestContext.Request` object isn't yet available.

`ServiceId` and `ServiceName` are sometimes subtly different, as is the case with dynamo: **DynamoDB** vs **DynamoDBv2**.

I could see an argument that this is a breaking change - existing users would see the new tag value after upgrading.  To prevent it, I'd need to hardcode a mapping table for dynamodb and sns.  When I considered this, my opinion was the name change was acceptable and the maintenance overhead of the mapping table was not worth it.

But I'd very much welcome additional opinions, especially if CNCF has guidance around this issue.

birojnayak: As you said, the name shouldn't change, while we can change the logic... But I see  [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/Instrumentation.AWS-1.1.0-beta.3/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md#110-beta1) similar changes. Approving and let's mark this as breaking changes. @Kielek FYI..

birojnayak:
@Kielek or anyone , could you add tag to mark this change as breaking change ? or let us know the procedure...

birojnayak (approved):

Kielek:
> @Kielek or anyone , could you add tag to mark this change as breaking change ? or let us know the procedure...

Typically, you should add entry to the Changelog.
Consider following format

```md
* **Breaking Change**: Your message here
  ([#PR_ID](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/PR_ID))
```

ppittle:
> > @Kielek or anyone , could you add tag to mark this change as breaking change ? or let us know the procedure...
> 
> Typically, you should add entry to the Changelog. Consider following format
> 
> ```
> * **Breaking Change**: Your message here
>   ([#PR_ID](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/PR_ID))
> ```

Updated Changelog

srprash (approved):

ppittle:
@Kielek - looks like this PR has a pending reviewer:
>open-telemetry/dotnet-contrib-approvers was requested for review as a code owner

Are you able to approve, or point me towards someone who can?

utpilla:
@ppittle Could you please fix the CI errors? We can then merge the PR.

ppittle:
> @ppittle Could you please fix the CI errors? We can then merge the PR.

I've pushed an update.  @utpilla, can you approve the CI/CD run?

test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
```diff
@@ -13,9 +13,13 @@
 
 namespace OpenTelemetry.Instrumentation.AWS.Tests;
 
-internal class TestRequest(ParameterCollection parameters) : IRequest
+internal class TestRequest : IRequest
 {
-    private readonly ParameterCollection parameters = parameters;
+    private readonly ParameterCollection parameters;
```
utpilla, line 18: @ppittle Could you add a newline after this? This is the [error](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1516.md): https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/7999444791/job/21847990924?pr=1572#step:4:239

test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
```diff
@@ -13,9 +13,13 @@
...
+    public TestRequest(ParameterCollection? parameters = null)
```
Kielek, line 19:
```suggestion
    private readonly ParameterCollection parameters;

    public TestRequest(ParameterCollection? parameters = null)
```

test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequest.cs
ppittle: Looks like @Kielek fixed it for me :)

utpilla (approved):

Issue #1585, opened 2024-02-20
Support otlp protocol for metric exporter.


vishweshbankwar:
Add support for sending otlp formatted metric data over etw and user_events.

vishweshbankwar:
This is now available with https://www.nuget.org/packages/OpenTelemetry.Exporter.Geneva/1.8.0 as an experimental feature. See [instructions](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Exporter.Geneva#:~:text=Set%20PrivatePreviewEnableOtlpProtobufEncoding%3Dtrue%20to%20opt%2Din%20to%20the%20experimental%20feature%20for%20changing%20the%20underlying%20serialization%20format%20to%20binary%20protobuf%20following%20the%20schema%20defined%20in%20OTLP%20specification.) to onboard.

martincostello:
Is there a reason this was re-opened? What else is left to do?

martincostello:
Re-closing due to lack of details on what is remaining to do.

PR #1596, opened 2024-02-29
[Exporter.Geneva] Add otlp protobuf metric exporter

vishweshbankwar:
Towards #1585

## Changes

* Adds metric exporter which follows the otlp protocol to format metric data in protobuf encoding. The exporter will only be used when `PrivatePreviewOtlpProtobufMetricExporter` is set to `true` in connection string and on windows environment. For Linux based environments, the support will be added later via user_events.

## Implementation Details

* A single MetricPoint is exported per etw event. This mimics the current behavior of TlvExporter. 
* While Serializing, a fixed size of `4` bytes is left for writing the length of sub messages. [Protobuf encoding](https://protobuf.dev/programming-guides/encoding/) specifies a varint should be used instead which only takes the minimum number of bytes required to specify the length. `4` bytes is used to avoid the size calculation in advance. In some cases calculating the size in advance could be more efficient (Added TODOs).
*  This PR only covers serializing Counters with Long type.

## Future considerations and TODOs

* Add remaining metric types.
* Add more unit test cases
* Add benchmarks
* Apply optimizations in serialization. Noted few TODOs in the code (not a comprehensive list).
* Writing single MetricPoint per etw event is not a very efficient use of max available buffer size. An alternate approach could be to write as much data as we can in single event and if the buffer overflows, send the remaining data in a separate event.


vishweshbankwar:
// * Summary *

BenchmarkDotNet v0.13.12, Windows 11 (10.0.22631.3155/23H2/2023Update/SunValley3)
13th Gen Intel Core i9-13900H, 1 CPU, 20 logical and 14 physical cores
.NET SDK 8.0.200
  [Host]     : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.2 (8.0.224.6711), X64 RyuJIT AVX2


| Method                                         | Mean      | Error    | StdDev   | Allocated |
|----------------------------------------------- |----------:|---------:|---------:|----------:|
| SerializeCounterMetricItemWith3Dimensions      |  90.26 ns | 0.584 ns | 0.517 ns |         - |
| SerializeCounterMetricItemWith4Dimensions      | 104.05 ns | 0.519 ns | 0.434 ns |         - |
| SerializeCounterMetricItemWith3Dimensions_otlp | 254.96 ns | 4.065 ns | 3.802 ns |         - |
| SerializeCounterMetricItemWith4Dimensions_otlp | 288.26 ns | 3.449 ns | 3.226 ns |         - |

// * Hints *
Outliers
  MetricExporterBenchmarks.SerializeCounterMetricItemWith3Dimensions: Default -> 1 outlier  was  removed (93.04 ns)
  MetricExporterBenchmarks.SerializeCounterMetricItemWith4Dimensions: Default -> 2 outliers were removed (107.99 ns, 108.21 ns)

// * Legends *
  Mean      : Arithmetic mean of all measurements
  Error     : Half of 99.9% confidence interval
  StdDev    : Standard deviation of all measurements
  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  1 ns      : 1 Nanosecond (0.000000001 sec)

// * Diagnostic Output - MemoryDiagnoser *


// ***** BenchmarkRunner: End *****

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -22,31 +23,46 @@ public class GenevaMetricExporter : BaseExporter<Metric>
...
-    private delegate ExportResult ExportMetricsFunc(in Batch<Metric> batch);
+    private delegate ExportResult ExportMetricsFunc(in Batch<Metric> batch, Resource resource);
 
     private readonly ExportMetricsFunc exportMetrics;
 
     private bool isDisposed;
 
+    private Resource resource;
+
+    internal Resource Resource => this.resource ??= this.ParentProvider.GetResource();
```
utpilla, line 36: Make this an `internal readonly` field and initialize it in the ctor?

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -22,31 +23,46 @@ public class GenevaMetricExporter : BaseExporter<Metric>
...
         Guard.ThrowIfNull(options);
         Guard.ThrowIfNullOrWhitespace(options.ConnectionString);
 
-        // TODO: parse connection string to check if otlp protobuf format is enabled.
-        // and then enable either TLV Exporter or Protobuf based exporter.
-        var tlvMetricsExporter = new TlvMetricExporter(options);
+        var connectionStringBuilder = new ConnectionStringBuilder(options.ConnectionString);
 
-        this.exportMetrics = tlvMetricsExporter.Export;
+        if (connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter != null && connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter.Equals("true", StringComparison.OrdinalIgnoreCase))
```
utpilla, line 45:
```suggestion
        if (connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter != null && connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase))
```

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: Since this is only required by the new exporter, could you just pass the resource to its ctor instead of updating the Export method signature for TlvMetricExporter as well.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/FieldNumberConstants.cs
```diff
@@ -0,0 +1,132 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Runtime.CompilerServices;
+using OpenTelemetry.Metrics;
+
+namespace OpenTelemetry.Exporter.Geneva;
+
+internal class FieldNumberConstants
```
utpilla, line 9:
```suggestion
internal static class FieldNumberConstants
```

utpilla, line 9: Add the comment to mention that these constants are based on the v1.1.0 proto file and add a link to the proto file: https://github.com/open-telemetry/opentelemetry-proto/blob/v1.1.0/opentelemetry/proto/metrics/v1/metrics.proto

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
```diff
@@ -0,0 +1,51 @@
...
+    private readonly byte[] buffer = new byte[BufferSize];
+
+    private readonly OtlpProtobufSerializer otlpProtobufSerializer;
+
+    public OtlpProtobufMetricExporter()
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            // Temporary until we add support for user_events.
+            throw new NotSupportedException("Unix domain socket should not be used on Windows.");
```
utpilla, line 24:
```suggestion
            throw new NotSupportedException("Exporting data in protobuf format is not supported on Linux.");
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/WireFormat.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.Exporter.Geneva;
+
+/// <summary>
+/// This class contains constants and helper functions useful for dealing with
```
utpilla, line 7: Add a link to the protobuf doc that mentions this encoding: https://protobuf.dev/programming-guides/encoding/#structure

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/WireFormat.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// <summary>
+    /// Makes a tag value given a field number and wire type.
+    /// </summary>
+    /// <param name="fieldNumber">Field number specified in proto definition.</param>
+    /// <param name="wireType">WireType for the field.</param>
+    /// <returns>Tag value.</returns>
+    internal static uint MakeTag(int fieldNumber, WireType wireType)
+    {
+        return (uint)(fieldNumber << TagTypeBits) | (uint)wireType;
```
utpilla, line 58: I'm not sure if the operator precedence would be an issue here but it would be simpler to just add an extra pair of parenthesis to make our intent clear.

```suggestion
        return ((uint)(fieldNumber << TagTypeBits)) | (uint)wireType;
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/TimestampHelpers.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+
+namespace OpenTelemetry.Exporter.Geneva;
+
+/// <summary>
+/// Helpers to convert .NET time related types to the timestamp used in OTLP.
+/// </summary>
+internal static class TimestampHelpers
+{
+    private const long NanosecondsPerTicks = 100;
```
utpilla, line 13:
```suggestion
    private const long NanosecondsPerTick = 100;
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/TimestampHelpers.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        return (dt.Ticks - UnixEpochTicks) * NanosecondsPerTicks;
+    }
+
+    internal static long ToUnixTimeNanoseconds(this DateTimeOffset dto)
+    {
+        return (dto.Ticks - UnixEpochTicks) * NanosecondsPerTicks;
+    }
+
+    internal static long ToNanoseconds(this TimeSpan duration)
```
utpilla, line 26: Remove the methods that are not used.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System;
+using System.Buffers.Binary;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace OpenTelemetry.Exporter.Geneva;
+
+internal static class OtlpProtobufSerializerHelper
```
utpilla, line 11: `ProtobufSerializer` would be a better name for this class as it does not have an Otlp specific code.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        int currentPosition = 0;
+
+        try
+        {
+            this.otlpProtobufSerializer.SerializeAndSendMetrics(this.buffer, ref currentPosition, resource, batch);
+        }
+        catch (Exception ex)
+        {
+            ExporterEventSource.Log.ExporterException("Failed to export metrics batch", ex);
```
utpilla, line 42: Set `result` to `ExportResult.Failure` in the catch block.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    public ExportResult Export(in Batch<Metric> batch, Resource resource = null)
+    {
+        var result = ExportResult.Success;
+
+        int currentPosition = 0;
+
+        try
+        {
+            this.otlpProtobufSerializer.SerializeAndSendMetrics(this.buffer, ref currentPosition, resource, batch);
```
utpilla, line 38: Could you split the responsibilities here? `OtlpProtobufSerializer` class and its methods should only be responsible for serializing the data and getting the `byte[]` buffer ready. You could then call `MetricEtwDataTransport.Send` separately after you have serialized the data.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
vishweshbankwar: resource is not available before the .Build(). it gets set only after the exporter is initialized.

utpilla: > Since this is only required by the new exporter, could you just pass the resource to its ctor instead of updating the Export method signature for TlvMetricExporter as well.

Resource is only needed by `OtlpProtobufMetricExporter` and need not be shared with `TlvMetricExporter`.

vishweshbankwar: `ParentProvider` is also set after the processor is created so this won't work.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
vishweshbankwar: @utpilla Just adding the notes based on offline discussion. With the current state it is possible to do so as we would be serializing 1 metricPoint at a time. However, with the follow up I am planning to do (sending as much data as possible in one etw event), it would get coupled again. There isn't a way to completely decouple the serializer with sender.

There was one concern I had with regards to testing/benchmarking, but that can be managed by setting the transport to no-op. LMK what you think.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
utpilla: There are a few ways to address this issue. One would be to let the `OtlpProtobufExporter` have a field that pointes to `GenevaMetricExporter` or you could provide `OtlpProtobufExporter` a `Func` that would return the Resource. How about this?

```csharp
public class GenevaMetricExporter : BaseExporter<Metric>
{
    public GenevaMetricExporter(GenevaMetricExporterOptions options)
    {
        ...

        var otlpProtobufExporter = new OtlpProtobufMetricExporter(this.GetResource);

        ...
    }

    private Resource GetResource(
    {
        return this.ParentProvider.GetResource();
    }
}




internal sealed class OtlpProtobufMetricExporter : IDisposable
{
    private readonly Func<Resource> getResource;
    private Resource resource;
    private Resource Resource => this.resource ??= this.getResource();

    public OtlpProtobufMetricExporter(Func<Resource> getResource)
    {
        ...

        this.getResource = getResource;

        ...
    }

    public ExportResult Export(in Batch<Metric> batch)
    {
        ...

        try
        {
            this.otlpProtobufSerializer.SerializeAndSendMetrics(this.buffer, ref currentPosition, this.Resource, batch);
        }
        catch (Exception ex)
        {
            ...
        }

        return result;
    }
}
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public ExportResult Export(in Batch<Metric> batch)
+    {
+        var result = ExportResult.Success;
+
+        int currentPosition = 0;
+
+        try
+        {
+            this.otlpProtobufSerializer.SerializeAndSendMetrics(this.buffer, ref currentPosition, this.getResource(), batch);
```
utpilla, line 41: This method should always start writing from the beginning of the buffer. There is no need of `currentPosition` parameter for this method.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+                {
+                    newList = new List<Metric>();
+                    newList.Add(metric);
+                    this.scopeMetrics[metric.MeterName] = newList;
+                }
+            }
+        }
+
+        // Serialize
+        this.SerializeResourceMetrics(buffer, ref currentPosition, resource);
```
utpilla, line 63: Same thing as this comment: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1596/files#r1529583054

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+    {
+        foreach (var metric in metricBatch)
+        {
+            if (this.scopeMetrics.TryGetValue(metric.MeterName, out var metricList))
+            {
+                metricList.Add(metric);
+            }
+            else
+            {
+                if (MetricListPool.TryTake(out var newList))
```
utpilla, line 48: Why do we need to pool the lists? We can simply create a dedicated list for each meter and clear each of them at the end of export.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+
+    internal void ReturnScopeMetrics()
+    {
+        foreach (var list in this.scopeMetrics)
+        {
+            list.Value.Clear();
+            MetricListPool.Add(list.Value);
+        }
+
+        this.scopeMetrics.Clear();
```
utpilla, line 76: Why is this needed?

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+                {
+                    SerializeTag(buffer, ref currentPosition, tag.Key, tag.Value, fieldNumber);
+                }
+            }
+        }
+    }
+
+    private static void SerializeResource(byte[] buffer, ref int currentPosition, Resource resource)
+    {
+        if (resource != null && resource.Attributes != null)
```
utpilla, line 302: We want to serialize Resources only if it has some attributes otherwise we would waste some space to serialize zero as the length of the resources.
```suggestion
        if (resource != Resource.Empty)
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+    internal void SerializeScopeMetrics(byte[] buffer, ref int currentPosition, string scopeName, List<Metric> metrics)
+    {
+        bool scopeInformationIncluded = false;
+
+        foreach (Metric metric in metrics)
+        {
+            if (!scopeInformationIncluded)
+            {
+                // Serialize scope information
+                SerializeInstrumentationScope(buffer, ref currentPosition, scopeName, metric.MeterTags);
```
utpilla, line 115: Could you move this outside of the foreach loop? This method would only be called for metrics of the same `Meter`.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            ProtobufSerializerHelper.WriteStringTag(buffer, ref currentPosition, metric.Unit, FieldNumberConstants.Metric_unit);
+        }
+    }
+
+    internal static void SerializeInstrumentationScope(byte[] buffer, ref int currentPosition, string name, IEnumerable<KeyValuePair<string, object>> meterTags)
+    {
+        int previousPosition = currentPosition;
+
+        currentPosition += LengthAndTagSize;
```
utpilla, line 275: Same issue as [this](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1596/files#r1529602334). Check if there are any meter tags first.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla, line 115: Add a TODO to avoid serializing this on every export.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+
+            this.previousMetricStartIndex = currentPosition;
+
+            currentPosition += LengthAndTagSize;
+
+            // Serialize metrics for the meter/scope
+            this.SerializeMetric(buffer, ref currentPosition, metric);
+
+            // Reset the current position to start writing again from where we started the last metric.
+            currentPosition = this.previousMetricStartIndex;
```
utpilla, line 128: This would override the previously serialized metric data, right?

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,373 @@
...
+        {
+            case MetricType.LongSum:
+            case MetricType.LongSumNonMonotonic:
+                {
+                    this.previousIntrumentStartIndex = currentPosition;
+
+                    currentPosition += LengthAndTagSize;
+
+                    // Write isMonotonic tag
+                    ProtobufSerializerHelper.WriteBoolWithTag(buffer, ref currentPosition, FieldNumberConstants.Sum_is_monotonic, metric.MetricType == MetricType.LongSum);
```
utpilla, line 148: Why are we serializing this in the reverse order?

From the proto:

```proto
message Sum {
  repeated NumberDataPoint data_points = 1;

  // aggregation_temporality describes if the aggregator reports delta changes
  // since last report time, or cumulative changes since a fixed start time.
  AggregationTemporality aggregation_temporality = 2;

  // If "true" means that the sum is monotonic.
  bool is_monotonic = 3;
}
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla: Ah we are only serializing one MetricPoint at a time! That essentially means that this method should always start writing at the same index. It would be more readable to have the variable `previousMetricStartIndex` to `metricStartIndex` and have that populated in the caller of this method instead of doing this everytime:

```csharp
this.previousScopeMetricsStartIndex = currentPosition;
...
currentPosition = this.previousScopeMetricsStartIndex;
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: This is slightly different. Instrumentation scope will always have the name field serialized even if there are no meter tags.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: It's just easier this way. We need to serialize field # 2 and 3 only once so we serialize it and then keep writing metricPoint one by one. Sequence does not matter here.

Alternate way would be to serialize those fields every time as we do not know in advance how much space a metricPoint would take.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: refactored.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: Please see updated, is this what you were expecting?

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -22,26 +23,41 @@ public class GenevaMetricExporter : BaseExporter<Metric>
...
         Guard.ThrowIfNull(options);
         Guard.ThrowIfNullOrWhitespace(options.ConnectionString);
 
-        // TODO: parse connection string to check if otlp protobuf format is enabled.
-        // and then enable either TLV Exporter or Protobuf based exporter.
-        var tlvMetricsExporter = new TlvMetricExporter(options);
+        var connectionStringBuilder = new ConnectionStringBuilder(options.ConnectionString);
 
-        this.exportMetrics = tlvMetricsExporter.Export;
+        if (connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter != null && connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase))
```
rajkumar-rangaraj, line 45: nit:
```suggestion
        if (connectionStringBuilder.PrivatePreviewOtlpProtobufMetricExporter?.Equals(bool.TrueString, StringComparison.OrdinalIgnoreCase))
```

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
```diff
@@ -22,26 +23,41 @@ public class GenevaMetricExporter : BaseExporter<Metric>
...
+        {
+            var otlpProtobufExporter = new OtlpProtobufMetricExporter(() => { return this.Resource; });
```
rajkumar-rangaraj, line 47: What is the benefit of passing the resource as delegate here?

If I understand correctly, the Resource should only be utilized for a protobuf-based exporter. However, we have a class-level variable declared for it.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufMetricExporter.cs
```diff
@@ -0,0 +1,53 @@
...
+        catch (Exception ex)
+        {
+            ExporterEventSource.Log.ExporterException("Failed to export metrics batch", ex);
+            result = ExportResult.Failure;
+        }
+
+        return result;
+    }
+
+    public void Dispose()
```
rajkumar-rangaraj, line 50: If we are not using Dispose, why to mark the class as IDisposable?

utpilla: I think we can still make good use of making this type an `IDisposable` by setting the byte array and `OtlpProtobufSerializer` as `null` to expedite their collection by GC.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+            else
+            {
+                var newList = new List<Metric>();
+                newList.Add(metric);
+                this.scopeMetrics[metric.MeterName] = newList;
+            }
+        }
+
+        // Serialize
+        this.SerializeResourceMetrics(buffer, resource);
```
rajkumar-rangaraj, line 52: Split this into SerializeResource and SerializeAllScopeMetrics. SerializeResource could be called only once and not in every iteration. You don’t need to free the entire buffer. Instead, provide the buffer and the length utilized earlier by resource to SerializeAllScopeMetrics

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+
+        this.ClearScopeMetrics();
+    }
+
+    internal void ClearScopeMetrics()
+    {
+        foreach (var list in this.scopeMetrics)
```
rajkumar-rangaraj, line 59: It's not list, rather KeyValuePair, either use the correct type instead of var or rename list to something like entry.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
rajkumar-rangaraj: I see, SerializeResourceMetrics already does the split and a TODO.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteStringTag(byte[] buffer, ref int currentPosition, string value, int fieldNumber)
+    {
+        int stringSize = Utf8Encoding.GetByteCount(value);
+
+        WriteTag(buffer, ref currentPosition, fieldNumber, WireFormat.WireType.LengthDelimited);
+
+        WriteLength(buffer, ref currentPosition, stringSize);
+
+        _ = Encoding.UTF8.GetBytes(value, 0, value.Length, buffer, currentPosition);
```
rajkumar-rangaraj, line 26: This returns the size of string, is there a way to update the buffer with value and then go back and update the length?
There were some todo's not sure if it was to sort this out.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    private static void SerializeResource(byte[] buffer, ref int currentPosition, Resource resource)
+    {
+        if (resource != null && resource != Resource.Empty)
+        {
+            var previousPosition = currentPosition;
+            currentPosition += LengthAndTagSize;
+            SerializeTags(buffer, ref currentPosition, resource.Attributes, FieldNumberConstants.Resource_attributes);
+            ProtobufSerializerHelper.WriteTagAndLengthPrefix(buffer, ref previousPosition, currentPosition - previousPosition - LengthAndTagSize, FieldNumberConstants.ResourceMetrics_resource, WireFormat.WireType.LengthDelimited);
```
rajkumar-rangaraj, line 287: Perhaps you could note down the previous position after incrementing LengthAndTagSize. This could simplify the length calculation here.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using OpenTelemetry.Metrics;
+using OpenTelemetry.Resources;
+
+namespace OpenTelemetry.Exporter.Geneva;
+
+internal class OtlpProtobufSerializer
```
utpilla, line 12:
```suggestion
internal sealed class OtlpProtobufSerializer
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla: We don't need the `null` check here. The public API [documentation](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/ProviderExtensions.cs#L21) states that it would return `Resource.Empty` when no resource is found.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+    {
+        foreach (var metric in metricBatch)
+        {
+            if (this.scopeMetrics.TryGetValue(metric.MeterName, out var metricList))
+            {
+                metricList.Add(metric);
+            }
+            else
+            {
+                var newList = new List<Metric>();
```
utpilla, line 45: nit: Simplify the list initialization

```csharp
var newList = new List<Metric>
{
    metric,
};
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla: > Instrumentation scope will always have the name field serialized even if there are no meter tags.

Why is that? The proto [file](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L69-L71) says:

"Semantically when InstrumentationScope isn't set, it is equivalent with an empty instrumentation scope name (unknown)."

It looks like it should be fine to not set the instrumentation scope.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla: I was trying to avoid repeated calculations of the same thing. We are doing this for each scope metrics instance:
```csharp
currentPosition = this.scopeMetricsStartIndex;
currentPosition += LengthAndTagSize;
```

Something like this:

```csharp
this.scopeMetricsLengthAndTagIndex = currentPosition;
this.scopeMetricsValueIndex = currentPosition + LengthAndTagSize;

foreach (KeyValuePair<string, List<Metric>> metric in this.scopeMetrics)
{
    if (metric.Value.Count > 0)
    {
        // Serialize this meter/scope
        this.SerializeScopeMetrics(buffer, ref this.scopeMetricsValueIndex, metric.Key, metric.Value);
    }
}
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
rajkumar-rangaraj: It's not an optional parameter in proto, we may need to set unknown when InstrumentationScope is not set

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+
+        // Serialize Resource
+        // TODO: Avoid serializing it multiple times.
+        SerializeResource(buffer, ref currentPosition, resource);
+
+        // TODO: Serialize schema_url field
+
+        // Serialize ScopeMetrics field
+        this.scopeMetricsStartIndex = currentPosition;
+        foreach (KeyValuePair<string, List<Metric>> metric in this.scopeMetrics)
```
utpilla, line 81: `metric` is not the best name for current element in the enumeration. Also, rename the dictionary `scopeMetrics` to something like `scopeToMetricsMapping` 

```suggestion
        foreach (KeyValuePair<string, List<Metric>> entry in this.scopeToMetricsMapping)
```

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+    internal void SerializeScopeMetrics(byte[] buffer, ref int currentPosition, string scopeName, List<Metric> metrics)
+    {
+        // Serialize scope information
+        // TODO: Avoid serializing for each export.
+        SerializeInstrumentationScope(buffer, ref currentPosition, scopeName, metrics[0].MeterTags);
+
+        this.metricStartIndex = currentPosition;
+        foreach (Metric metric in metrics)
+        {
+            currentPosition = this.metricStartIndex;
```
utpilla, line 104: Same comment as https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1596#discussion_r1531443362

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: InstrumentationScope has multiple fields. https://github.com/open-telemetry/opentelemetry-proto/blob/9d139c87b52669a3e2825b835dd828b57a455a55/opentelemetry/proto/common/v1/common.proto#L71-L80

So even if a scope(Meter) does not have attributes, we need to set the name/version fields.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,353 @@
...
+                    ProtobufSerializerHelper.WriteEnumWithTag(buffer, ref currentPosition, FieldNumberConstants.Sum_aggregation_temporality, metric.Temporality == AggregationTemporality.Cumulative ? 2 : 1);
+
+                    this.metricPointStartIndex = currentPosition;
+                    foreach (var metricPoint in metric.GetMetricPoints())
+                    {
+                        currentPosition = this.metricPointStartIndex;
+
+                        currentPosition += LengthAndTagSize;
+
+                        var sum = (ulong)metricPoint.GetSumLong();
```
utpilla, line 142: Why do we have to cast this to `ulong`?

utpilla: The proto defines this as `sfixed64` which should be a `long`: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L679

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
vishweshbankwar: This is different than other cases. Here we want to optimize on the number on bytes used to encode length. By writing the length first we ensure that we used the minimum # of required bytes.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+namespace OpenTelemetry.Exporter.Geneva;
+
+internal static class ProtobufSerializerHelper
+{
+    private const int Fixed64Size = 8;
+
+    internal static Encoding Utf8Encoding => Encoding.UTF8;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteStringTag(byte[] buffer, ref int currentPosition, string value, int fieldNumber)
```
utpilla, line 18: Rearrange the order of the parameters to be consistent with the other methods.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteFixed64WithTag(byte[] buffer, ref int currentPosition, int fieldNumber, ulong value)
+    {
+        WriteTag(buffer, ref currentPosition, fieldNumber, WireFormat.WireType.Fixed64);
+        WriteRawLittleEndian64(buffer, ref currentPosition, value);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteInt64WithTag(byte[] buffer, ref int currentPosition, int fieldNumber, ulong value)
```
utpilla, line 56: Use the same convention of having the proto type name in the method name.

```suggestion
    internal static void WriteFixed64WithTag(byte[] buffer, ref int currentPosition, int fieldNumber, ulong value)
```

test/OpenTelemetry.Exporter.Geneva.Tests/OpenTelemetry.Exporter.Geneva.Tests.csproj
```diff
@@ -1,16 +1,19 @@
...
-  <Import Project="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\Common.GenevaExporter.props"/>
+  <Import Project="$(RepoRoot)\src\OpenTelemetry.Exporter.Geneva\Common.GenevaExporter.props" />
 
   <PropertyGroup>
     <Description>Unit test project for Geneva Exporters for OpenTelemetry</Description>
     <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
     <TargetFrameworks>$(SupportedNetTargets)</TargetFrameworks>
     <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net48;net472;net471;net47;net462</TargetFrameworks>
-    <NoWarn>$(NoWarn),SA1311,SA1312,SA1313,SA1123,SA1202</NoWarn>
+    <NoWarn>$(NoWarn),SA1311,SA1312,SA1313,SA1123,SA1202,SA1518</NoWarn>
```
utpilla, line 9: Is this just to avoid the check for proto files? Could you check if you could update the editorconfig instead to skip the proto files for this check?

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
utpilla: Also, rename those variables to `TagAndLength` instead of `LengthAndTag`, if the we first encode Tag and then the length 😄

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+            value >>= 7;
+        }
+
+        WriteRawByte(buffer, ref currentPosition, (byte)value);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteRawByte(byte[] buffer, ref int currentPosition, byte value)
+    {
+        if (currentPosition < 0)
```
rajkumar-rangaraj: When we are incrementing the currentPosition, how would be end up with value less than zero.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: long and ulong both are encoded the same way. The bit representation does not change, the change would be on the decoding side on how it is interpreted (in terms of signed versus unsigned).

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
vishweshbankwar: This is different than fixed64. It used varint https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/9865323c25f0a55f52fae2221faef4b4338d990e/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs#L59

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
vishweshbankwar: good catch. Fixed.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
vishweshbankwar: Because resource is only available after the exporter is built. Note that `GeneveMetricExporter` would be the one configured with MeterProvider so the resource can only be accessed using the `ParentProvider` of `GenevaMetricExporter`.

src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs
vishweshbankwar: This won't work for if

test/OpenTelemetry.Exporter.Geneva.Tests/OpenTelemetry.Exporter.Geneva.Tests.csproj
vishweshbankwar: added

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+    internal static void WriteDoubleWithTag(byte[] buffer, ref int cursor, int fieldNumber, double value)
+    {
+        WriteTag(buffer, ref cursor, fieldNumber, WireFormat.WireType.Fixed64);
+        WriteRawLittleEndian64(buffer, ref cursor, (ulong)BitConverter.DoubleToInt64Bits(value));
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteLengthCustom(byte[] buffer, ref int cursor, int length)
+    {
+        WriteRawVarintCustom(buffer, ref cursor, (uint)length);
```
rajkumar-rangaraj, line 72: Should we avoid calling WriteRawVarintCustom and have the implementation here, as this is custom?

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/ProtobufSerializerHelper.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        while (index < 3)
+        {
+            buffer[cursor++] = 0x00;
+            index++;
+        }
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void WriteRawVarint32(byte[] buffer, ref int cursor, uint value)
```
rajkumar-rangaraj, line 117: I'm concerned about copying only the methods from Protobuf. The best approach would be to obtain the file from Protobuf and then move your custom implementations to a different file. Before copying the content, have you verified the license of the Protobuf repository? Here are a few approaches to address this issue

1. Review the license information and collaborate with the maintainer on how to obtain the code from another repository.
1. Customize and develop our own solution.

vishweshbankwar: @rajkumar-rangaraj - I have not copied the methods as is. The implementations may look same but that's because of the encoding.

> Customize and develop our own solution.

I think that is what I am doing. Is there something specific you see that could be an issue?

rajkumar-rangaraj: It's only a couple of methods that have an exact match, and I agree that, apart from those, it's all your implementation. If you're customizing, It would be better to change the method names instead of following the naming conventions from another repository.

utpilla (changes requested):
Need to ensure that if at all any code is being used from a different repository, we do it in accordance with their license requirements.

Please confirm which one of the below two statements hold true:
1. Code from some other repo has been used in accordance with their license requirements
2. No code has been used from any other repo (or it has been removed if it was previously used)

vishweshbankwar:
> Need to ensure that if at all any code is being used from a different repository, we do it in accordance with their license requirements.
> 
> Please confirm which one of the below two statements hold true:
> 
> 1. Code from some other repo has been used in accordance with their license requirements
> 2. No code has been used from any other repo (or it has been removed if it was previously used)

@utpilla @rajkumar-rangaraj - I have revised the code to ensure that any unintentional similarities to external code have been eliminated. The code in this pull request is written from scratch and does not borrow from any external libraries. Please review. Thanks!

utpilla (approved):

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,388 @@
...
+                    this.metricPointTagAndLengthIndex = cursor;
+                    this.metricPointValueIndex = cursor + TagAndLengthSize;
+                    foreach (var metricPoint in metric.GetMetricPoints())
+                    {
+                        try
+                        {
+                            // Reset cursor to write new metricPoint
+                            cursor = this.metricPointValueIndex;
+
+                            var sum = (ulong)metricPoint.GetSumLong();
```
utpilla, line 168: Write a comment here saying why the casting should be okay.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,388 @@
...
+                case bool b:
+                    ProtobufSerializerHelper.WriteBoolWithTag(buffer, ref cursor, FieldNumberConstants.AnyValue_bool_value, (bool)value);
+                    break;
+                case byte:
+                case sbyte:
+                case short:
+                case ushort:
+                case int:
+                case uint:
+                case long:
```
utpilla, line 366: Add a case for `ulong`?

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            TemporalityPreference = MetricReaderTemporalityPreference.Delta,
+        };
+
+        using var meterProvider = Sdk.CreateMeterProviderBuilder()
+            .AddMeter(nameof(this.LongCounterSerializationSingleMetricPoint))
+            .AddReader(inMemoryReader)
+            .Build();
+
+        longCounter.Add(123, new("tag1", "value1"), new("tag2", "value2"));
```
utpilla, line 37: Add a test case for testing negative long numbers and ulong numbers as well.

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+        otlpProtobufSerializer.SerializeAndSendMetrics(buffer, Resource.Empty, new Batch<Metric>(exportedItems.ToArray(), exportedItems.Count));
+
+        Assert.Single(testTransport.ExportedItems);
+
+        var request = new OtlpCollector.ExportMetricsServiceRequest();
+
+        request.MergeFrom(testTransport.ExportedItems[0]);
+
+        Assert.Single(request.ResourceMetrics);
```
utpilla, line 54: Add some resource attributes and test if the resources provided are contained in the serialized data.

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            Assert.Equal(1, (int)metric.Sum.AggregationTemporality);
+
+            Assert.True(metric.Sum.IsMonotonic);
+
+            Assert.Single(metric.Sum.DataPoints);
+
+            var dataPoint = metric.Sum.DataPoints[0];
+
+            Assert.Equal(124, dataPoint.AsInt);
```
utpilla, line 164: Add asserts for StartTime and EndTime as well.

src/OpenTelemetry.Exporter.Geneva/Internal/ConnectionStringBuilder.cs
```diff
@@ -75,6 +75,12 @@ public string PrivatePreviewEnableTraceLoggingDynamic
         set => this._parts[nameof(this.PrivatePreviewEnableTraceLoggingDynamic)] = value;
     }
 
+    public string PrivatePreviewOtlpProtobufMetricExporter
+    {
+        get => this._parts.TryGetValue(nameof(this.PrivatePreviewOtlpProtobufMetricExporter), out var value) ? value : null;
```
utpilla, line 80: We probably would have to change the name to something else such as `EnableOtlpProtobufEncoding`.

utpilla (approved):
Left a few comments (non-blocking). These should be worked upon in future PRs.

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -0,0 +1,391 @@
...
+                    this.metricPointValueIndex = cursor + TagAndLengthSize;
+                    foreach (var metricPoint in metric.GetMetricPoints())
+                    {
+                        try
+                        {
+                            // Reset cursor to write new metricPoint
+                            cursor = this.metricPointValueIndex;
+
+                            // Casting to ulong is ok here as the bit representation for long versus ulong will be the same
+                            // The difference would in the way the bit representation is interprested on decoding side (signed versus unsigned)
```
utpilla, line 169:
```suggestion
                            // The difference would in the way the bit representation is interpreted on decoding side (signed versus unsigned)
```

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
vishweshbankwar: Added for negative number. Planning to do ulong and negative attribute value in follow up.

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        var dataPoint = metric.Sum.DataPoints[0];
+
+        Assert.Equal(value, dataPoint.AsInt);
+
+        // Assert time
+        var metricPointsEnumerator = exportedItems[0].GetMetricPoints().GetEnumerator();
+        metricPointsEnumerator.MoveNext();
+        var metricPoint = metricPointsEnumerator.Current;
+
+        Assert.Equal((ulong)TimestampHelpers.ToUnixTimeNanoseconds(metricPoint.StartTime), dataPoint.StartTimeUnixNano);
```
utpilla, line 94: nit: You can also check this without relying on the helper method in `src/`.

```suggestion
            Assert.Equal((ulong)((DateTimeOffSet)metricPoint.StartTime.ToUnixTimeSeconds() * 1000000000), dataPoint.StartTimeUnixNano);
```

Issue #1416, opened 2023-10-31
Update to AOT-compatible PropertyFetcher and remove older version


eerhardt:
# Issue that does not fit into other categories

**What are you trying to achieve?**
With #1415 we are copying a new version of PropertyFetcher from open-telemetry/opentelemetry-dotnet that is AOT-compatible. This was copied as a separate version in order to reduce the risk of changing other libraries.

What do you expect to see?

We should update all the references of the old PropertyFetcher to the new AOT-compatible version and then remove the old version.

**Additional context.**

See also https://github.com/open-telemetry/opentelemetry-dotnet/pull/4675

Kielek:
Done in #2558

PR #1600, opened 2024-03-05
[Instrumentation.AspNet, Instrumentation.Quartz] Use new PropertyFetcher

Kielek:
Towards #1416.

## Changes

[Instrumentation.AspNet, Instrumentation.Quartz] Use new PropertyFetcher

Leftwovers: WCF and EFCore packages. New version does not contain Fetch method, only TryFetch.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
+++ b/src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
@@ -15,7 +15,7 @@
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\Shared\ExceptionExtensions.cs" Link="Includes\ExceptionExtensions.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
-    <Compile Include="$(RepoRoot)\src\Shared\PropertyFetcher.cs" Link="Includes\PropertyFetcher.cs" />
+    <Compile Include="$(RepoRoot)\src\Shared\PropertyFetcher.AOT.cs" Link="Includes\PropertyFetcher.AOT.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\SemanticConventions.cs" Link="Includes\SemanticConventions.cs" />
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
```diff
@@ -15,7 +15,7 @@
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\Shared\ExceptionExtensions.cs" Link="Includes\ExceptionExtensions.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
-    <Compile Include="$(RepoRoot)\src\Shared\PropertyFetcher.cs" Link="Includes\PropertyFetcher.cs" />
+    <Compile Include="$(RepoRoot)\src\Shared\PropertyFetcher.AOT.cs" Link="Includes\PropertyFetcher.AOT.cs" />
```
vishweshbankwar, line 18: @Kielek - Do we need `PropertyFetcher.cs`? Can we delete that and rename the AOT one to keep it consistent with main repo?

src/Shared/PropertyFetcher.AOT.cs
```diff
@@ -9,10 +9,14 @@
...
 #nullable enable
 
+#pragma warning disable IDE0005 // Using directive is unnecessary.
 #if NETSTANDARD2_1_0_OR_GREATER || NET6_0_OR_GREATER
 using System.Diagnostics.CodeAnalysis;
 #endif
+using System;
+using System.Linq;
 using System.Reflection;
+#pragma warning restore IDE0005 // Using directive is unnecessary.
```
vishweshbankwar, line 19: Curious why do we need to do this?

Kielek: Part of the projects are using implicit usings. It is not enabled globally.

src/OpenTelemetry.Instrumentation.AspNet/OpenTelemetry.Instrumentation.AspNet.csproj
Kielek: When we fix WCF and Efcore.
It requires code modification to use Try get method.

vishweshbankwar (approved):

PR #1624, opened 2024-03-25
Instrumentation spans and metrcis - set ActivitySource/Meter.Version to NuGet package version

Kielek:
Fixes: issue when it is hard to determine what is the instrumentation library version used in end-user project.
Most of packages just reporte 1.0.0.0 (Assembly version).
MinVer is setting more user friendly value in `AssemblyInformationalVersionAttribute` attribute in following format: `1.7.2-alpha.0.39+69cc79aed00a37c103b9e46488e2001db959a1ab`.

It was previously set correctly only in one package.

## Changes

Instrumentation spans - set ActivitySource.Version and Meter.Version to NuGet package version


Kielek:
I will be happy to propagate to main repository if we have agree on the naming and values.

utpilla:
Why does the ActivitySource version need to match the package version?

utpilla:
> I will be happy to propagate to main repository if we have agree on the naming and values.

This would probably be considered a breaking change for the stable packages.

cijothomas:
reporting nuget version looks better than assembly version which is 1.0.0.0 always, but i wonder if there was any guidance on this in terms of using nuget version.. Spec says library version as an example, so it is probably good to use nuget version.

> I will be happy to propagate to main repository if we have agree on the naming and values.
> 
> This would probably be considered a breaking change for the stable packages.

I am not sure how would this be a breaking change? It is not changing the semantics on the telemetry itself. Also there is no way to enable/disable based on Version today in the SDK. Ofcourse it should be possible in collector possibly, but still this does not look like breaking change.

vishweshbankwar:
Just the heads up. This will work well for libraries that create their own ActivitySource. It won't work for libraries that are simply enriching the activity created by respective frameworks. e.g. `OpenTelemetry.Instrumentation.AspNetCore` and `OpenTelemetry.Instrumentation.HttpClient`

So there will still be some inconsistency.

utpilla:
> I am not sure how would this be a breaking change? It is not changing the semantics on the telemetry itself. Also there is no way to enable/disable based on Version today in the SDK. Ofcourse it should be possible in collector possibly, but still this does not look like breaking change.

I was thinking about its implications for someone using a custom `ActivityListener` but I guess we don't need to worry about that since the `ActvitySource` instances are not marked `public` for both the stable instrumentation packages.

What problem are we trying to solve though? Who is using ActivitySource version and for what purpose? Also, as @vishweshbankwar pointed out we do not have any control over the libraries that natively support `ActivitySource` based instrumentation, so we wouldn't have any kind of consistency either.

cijothomas:
> I am not sure how would this be a breaking change? It is not changing the semantics on the telemetry itself. Also there is no way to enable/disable based on Version today in the SDK. Ofcourse it should be possible in collector possibly, but still this does not look like breaking change.
> 
> I was thinking about its implications for someone using a custom ActivityListener but I guess we don't need to worry about that since the ActvitySource instances are not marked public for both the stable instrumentation packages.

Got it!

cijothomas:
> Just the heads up. This will work well for libraries that create their own ActivitySource. It won't work for libraries that are simply enriching the activity created by respective frameworks. e.g. `OpenTelemetry.Instrumentation.AspNetCore` and `OpenTelemetry.Instrumentation.HttpClient`
> 
> So there will still be some inconsistency/gaps.

Good point. Same challenges for "schema_url" too (not yet supported in .NET)!

Kielek:
> Just the heads up. This will work well for libraries that create their own ActivitySource. It won't work for libraries that are simply enriching the activity created by respective frameworks. e.g. `OpenTelemetry.Instrumentation.AspNetCore` and `OpenTelemetry.Instrumentation.HttpClient`
> 
> So there will still be some inconsistency/gaps.

I would not say that inconsistency. As a follow up to this (and main reposiroty) we should start recommending what to put as a version. For most cases it should be not assembly version, as it typically is `nonzero.0.0.0`. Putting such versions is more or less useless if we speaking about telemetry/debugging purposes.

Kielek:
Updated PR to include the same version for the meters.

Oberon00:
> For most cases it should be not assembly version, as it typically is nonzero.0.0.0

Shouldn't we fix the assembly version then?

Kielek:
> > For most cases it should be not assembly version, as it typically is nonzero.0.0.0
> 
> Shouldn't we fix the assembly version then?

It is against this recommendation: https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/versioning#assembly-version

src/Shared/SignalVersionHelper.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Reflection;
+
+namespace OpenTelemetry.Instrumentation;
+
+internal static class SignalVersionHelper
+{
+    public static string GetVersion<T>()
+    {
+        return typeof(T).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()!.InformationalVersion.Split('+')[0];
```
Oberon00, line 12: Since we are using `!.` here, would it be better to use any fallback, or maybe a `?.` chain?

Oberon00 (approved):

fred2u (approved):

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -21,8 +20,7 @@ public static class AWSLambdaWrapper
 {
     internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.AWSLambda";
 
-    private static readonly string Version = typeof(AWSLambdaWrapper).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()!.InformationalVersion.Split('+')[0];
-    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, Version);
+    private static readonly ActivitySource AWSLambdaActivitySource = new(ActivitySourceName, SignalVersionHelper.GetVersion<AWSLambdaInstrumentationOptions>());
```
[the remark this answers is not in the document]
Kielek: I have started with this :)

```
`OpenTelemetry.Instrumentation.AWSLambda.AWSLambdaWrapper': static class type cannot be used as type arguments
```

src/Shared/SignalVersionHelper.cs
Kielek: I do not see any case when the null value can happen in any of library.
As long as we are using MinVer'sioning it should be safe to use this construction.

cijothomas:
@Kielek Would it be possible for you to take this to the OTel community meeting? If you cannot make it to the meeting, @vishweshbankwar can you help?

utpilla:
> As a follow up to this (and main reposiroty) **we should start recommending what to put as a version**.

Users are not required to provide a version. The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer) already says that version is optional. I don't think there is any need for us to recommend what value to use for it.

> Putting such versions is more or less useless if we speaking about telemetry/debugging purposes.

How are users using ActivitySource version for debugging purposes?

Kielek:
> > Putting such versions is more or less useless if we speaking about telemetry/debugging purposes.
> 
> How are users using ActivitySource version for debugging purposes?

From observability vendors perspective, if this value is associated with spans, we can easily check what version of instrumentation is used. It is valid especially if customers are using library instrumentations. Based in such data, we can recommend some steps, such us upgrade if fixes are in newer versions.

src/Shared/SignalVersionHelper.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Reflection;
```
ppittle, line 4: naïve question: Does this have any implication for AOT?  Does AOT support getting assembly info?

Kielek: I do not see any compilation warnings. Some instrumentation packages are included into AOT app.
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/e39306fc82c2e8f0ef9cde54b3d2701828add459/test/OpenTelemetry.AotCompatibility.TestApp/OpenTelemetry.AotCompatibility.TestApp.csproj#L18-L22

cijothomas:
> > > Putting such versions is more or less useless if we speaking about telemetry/debugging purposes.
> > 
> > 
> > How are users using ActivitySource version for debugging purposes?
> 
> From observability vendors perspective, if this value is associated with spans, we can easily check what version of instrumentation is used. It is valid especially if customers are using library instrumentations. Based in such data, we can recommend some steps, such us upgrade if fixes are in newer versions.

Got it!
Another example: a vendor has some special handling for v1 of semantic conventions, and want to get rid of that special handling, as v2 is out now, and v2 is considered stable. But the vendor need to figure out what percentage of its users is producing v1 vs v2.
This can be understood by looking at instrumentationscope.version.

Agree that assembly version won't be able to give such breakdowns.! (the special set of instrumentations where scope is owned by one component, but actual filling of data is done by another won't work as already pointed out, but that is also a temporary state, assuming these libraries will eventually natively produce telemetry without the need of a instrumentation library, like already done for metrics in some cases.)

Oberon00:
@cijothomas 
>  But the vendor need to figure out what percentage of its users is producing v1 vs v2.
This can be understood by looking at instrumentationscope.version.

Not really. The same library may produce different versions, e.g. for HTTP depending on settings. Also, this use case is perfectly suited for the schema URL that is reported with each scope.

vishweshbankwar (approved):
LGTM: There is still open question on how to handle the version when instrumentations do not create the ActivitySource for the activities they enrich. None of the libraries here falls under that category. 

@Kielek - We could open issues for the [special cases](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1624#issuecomment-2019085828) in their respective repos.

ppittle (approved):

Kielek:
@vishweshbankwar, I created: #5497. Will merge this one shortly (when the PR will finished).

PR #1629, opened 2024-03-28
[Exporter.Geneva][otlp protobuf] Add histogram support

vishweshbankwar:
Towards #1585

## Changes

Adds support for exporting histogram instrument type.

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
@@ -269,7 +269,69 @@ internal sealed class OtlpProtobufSerializer
 
             case MetricType.Histogram:
                 {
-                    // TODO
+                    cursor = this.instrumentValueIndex;
+
+                    // Write aggregationTemporality tag
+                    ProtobufSerializerHelper.WriteEnumWithTag(buffer, ref cursor, FieldNumberConstants.Histogram_aggregation_temporality, metric.Temporality == AggregationTemporality.Cumulative ? 2 : 1);
+
+                    this.metricPointTagAndLengthIndex = cursor;
+                    this.metricPointValueIndex = cursor + TagAndLengthSize;
+                    foreach (var metricPoint in metric.GetMetricPoints())
+                    {
+                        try
+                        {
+                            // Reset cursor to write new metricPoint
+                            cursor = this.metricPointValueIndex;
+
+                            var startTime = (ulong)metricPoint.StartTime.ToUnixTimeNanoseconds();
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -269,7 +269,69 @@ private void SerializeMetric(byte[] buffer, ref int cursor, Metric metric)
...
+
+                            if (metricPoint.TryGetHistogramMinMaxValues(out double min, out double max))
+                            {
+                                ProtobufSerializerHelper.WriteDoubleWithTag(buffer, ref cursor, FieldNumberConstants.HistogramDataPoint_min, min);
+                                ProtobufSerializerHelper.WriteDoubleWithTag(buffer, ref cursor, FieldNumberConstants.HistogramDataPoint_max, max);
+                            }
+
+                            foreach (var histogramMeasurement in metricPoint.GetHistogramBuckets())
+                            {
+                                var bucketCount = (ulong)histogramMeasurement.BucketCount;
```
utpilla, line 308: The proto file has bucket counts and explicit bounds as two different repeated fields. Doesn't that mean that we should first serialize all the bucket counts together and then all the explicit bounds together?

vishweshbankwar: We could do that but it does not matter from serialization perspective. Take a look at this: https://protobuf.dev/programming-guides/encoding/#optional:~:text=However%2C%20records%20for%20e%20do%20not%20need%20to%20appear%20consecutively%2C%20and%20can%20be%20interleaved%20with%20other%20fields%3B%20only%20the%20order%20of%20records%20for%20the%20same%20field%20with%20respect%20to%20each%20other%20is%20preserved.%20Thus%2C%20this%20could%20also%20have%20been%20encoded%20as

Serializing it together is more relevant in packed repeated fields which is a different type: https://protobuf.dev/programming-guides/encoding/#packed

utpilla: Thanks!

utpilla (approved):

utpilla:
Towards #1585 

It would be nice to link all of these PRs to the main issue. You could then just add one entry in the CHANGELOG for the issue which would have all the relevant PRs linked.

vishweshbankwar:
> Towards #1585
> 
> It would be nice to link all of these PRs to the main issue. You could then just add one entry in the CHANGELOG for the issue which would have all the relevant PRs linked.

Good catch. I have been doing that but missed on this and the last one. Added.

PR #1640, opened 2024-04-08
Move existing ServerCertificateValidationHandler to Shared

joegoldman2:
Related to #1550.
Test classes probably need to be moved somewhere else but I don't know exactly where.


iskiselev:
@joegoldman2 , please take a look at my commit https://github.com/open-telemetry/opentelemetry-dotnet-contrib/commit/3a112b03f00514110c8426e8bbcb9e4a7350dc1e
It does some cleanup to certificate validation logic - it is what was desired in original PR. Probably you can just grab this commit and push it in your PR.

joegoldman2:
> @joegoldman2 , please take a look at my commit [3a112b0](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/commit/3a112b03f00514110c8426e8bbcb9e4a7350dc1e) It does some cleanup to certificate validation logic - it is what was desired in original PR. Probably you can just grab this commit and push it in your PR.

Thank you for your review. I cherry-picked your commit 👍

iskiselev:
We need this code to be merged, to implement what have been started in #1529. There we also will need to establish HTTPS connection using certificate provided in a file.
I participated in original implementation and review of that certificate-based validation logic in AWS.
It should not change anything except fixes of error messages for AWS. @ppittle and @srprash, would you be able to review this change?
From my side, it looks good.

src/OpenTelemetry.ResourceDetectors.AWS/AWSResourcesEventSource.cs
```diff
@@ -8,10 +8,16 @@
 namespace OpenTelemetry.ResourceDetectors.AWS;
 
 [EventSource(Name = "OpenTelemetry-ResourceDetectors-AWS")]
-internal sealed class AWSResourcesEventSource : EventSource
+internal sealed class AWSResourcesEventSource : EventSource, IServerCertificateValidationEventSource
 {
     public static AWSResourcesEventSource Log = new();
 
+    private const int EVENTIDFAILEDTOEXTRACTATTRIBUTES = 1;
```
Kielek, line 15: Consider (same for other consts). I think that we follow Pascale Case for most of the cosnts in this repo.

```suggestion
    private const int EventIdFailedToExtractAttributes = 1;
```

src/OpenTelemetry.ResourceDetectors.AWS/OpenTelemetry.ResourceDetectors.AWS.csproj
```diff
@@ -23,6 +23,9 @@
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\Shared\ExceptionExtensions.cs" Link="Includes\ExceptionExtensions.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
+    <Compile Include="$(RepoRoot)\src\Shared\IServerCertificateValidationEventSource.cs" Link="Includes\IServerCertificateValidationEventSource.cs" />
```
Kielek, line 26: Could you please apply alphabetical order?

src/Shared/IServerCertificateValidationEventSource.cs
Kielek, line 1: Please add new Shared files into sln https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/113e5b82b06b4cb9da232f3f27547591d98d38f6/opentelemetry-dotnet-contrib.sln#L271-L290

Kielek (commented):
Some minor things, when fixed IMO it can be merged.

src/Shared/IServerCertificateValidationEventSource.cs
joegoldman2: It's already the case: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/fe3807301b5fcdee9dc3d618f5d16b8f8ac81607/opentelemetry-dotnet-contrib.sln.

src/OpenTelemetry.ResourceDetectors.AWS/OpenTelemetry.ResourceDetectors.AWS.csproj
joegoldman2: Not sure to understand this one. It already seems to be in alphabetical order.

src/Shared/IServerCertificateValidationEventSource.cs
Kielek: Sorry, I missed that file somehow.

src/OpenTelemetry.ResourceDetectors.AWS/OpenTelemetry.ResourceDetectors.AWS.csproj
Kielek: Stardted reviewing this to early my time. Sorry :]

Kielek (approved):

src/Shared/ServerCertificateValidationHandler.cs
```diff
@@ -6,35 +6,36 @@
...
         {
-            ServerCertificateValidationProvider? serverCertificateValidationProvider =
-                ServerCertificateValidationProvider.FromCertificateFile(certificateFile);
+            ServerCertificateValidationProvider? serverCertificateValidationProvider = ServerCertificateValidationProvider.FromCertificateFile(certificateFile, log);
 
             if (serverCertificateValidationProvider == null)
             {
-                AWSResourcesEventSource.Log.FailedToValidateCertificate(nameof(Handler), "Failed to Load the certificate file into trusted collection");
-                return null;
+                return new HttpClientHandler();
```
ppittle, line 21: This seems like a behavior change - why remove the log statement, and change the `return null` to `return new`?

iskiselev: It is, but should not really affect anything. All required log information would be logged inside call `ServerCertificateValidationProvider.FromCertificateFile`. If we were unable to create handler with specific policy to trust the certificate, we still create default handler that will use system-wide SSL trust to establish connection. It is OK, as we usually want to trust additional certificate, not limit anything (at least it is how custom ServerCertificateValidationProvider is implemented).
Most probably the connection would fail to be established, but it would also be caught and logged later - with reasonable information that SSL connection was not established. With prior messages about failure to create validation provider, entire picture would be clear.

src/OpenTelemetry.ResourceDetectors.AWS/AWSResourcesEventSource.cs
```diff
@@ -21,15 +27,33 @@ public void ResourceAttributesExtractException(string format, Exception ex)
...
     public void FailedToExtractResourceAttributes(string format, string exception)
     {
-        this.WriteEvent(1, format, exception);
+        this.WriteEvent(EventIdFailedToExtractAttributes, format, exception);
+    }
+
+    [Event(EventIdFailedToValidateCertificate, Message = "Failed to validate certificate. Details: '{0}'", Level = EventLevel.Warning)]
+    public void FailedToValidateCertificate(string error)
+    {
+        this.WriteEvent(EventIdFailedToValidateCertificate, error);
```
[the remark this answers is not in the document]
iskiselev: The log implementation previously was bad. It called `void FailedToValidateCertificate(string format, string error)`, that have logged message `Failed to validate certificate in format: '${format}', error: '{error}'.` It was called with `"ServerCertificateValidationProvider"` as a format argument always. Now log messages should have more sense.

ppittle (approved):

ppittle:
Looks good to me.  @iskiselev - thanks for driving this!

PR #1644, opened 2024-04-10
[Exporter.Geneva][Otlp protobuf] Add PrepopulatedMetricDimensions and Account/Namespace support

vishweshbankwar:
Towards #1585 

## Changes

Adds support for PrepopulatedMetricDimensions options and specifying account and namespace via connection string. 


src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
```diff
@@ -37,9 +49,40 @@ internal sealed class OtlpProtobufSerializer
...
+            cursor = 0;
+            SerializeTags(temp, ref cursor, prepopulatedMetricDimensions, FieldNumberConstants.HistogramDataPoint_attributes);
+            this.prepopulatedHistogramDataPointAttributes = new byte[cursor];
+            Array.Copy(temp, this.prepopulatedHistogramDataPointAttributes, cursor);
+            this.prepopulatedHistogramDataPointAttributesLength = cursor;
+
+            // TODO: exponential histogram.
+        }
+
+        if (connectionStringBuilder != null && connectionStringBuilder.Namespace != null)
```
utpilla, line 77: Add a try/catch block as accessing `Namespace` and `Account` from `ConnectionStringBuilder` could throw an exception when connectionString does not contain them.

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
```diff
@@ -115,7 +150,7 @@ public void CounterSerializationSingleMetricPoint(string instrumentName, long? l
 
         Assert.NotNull(request.ResourceMetrics[0].Resource);
 
-        AssertOtlpAttributes([new KeyValuePair<string, object>("TestResourceKey", "TestResourceValue")], request.ResourceMetrics[0].Resource.Attributes);
+        AssertOtlpAttributes(addAccountAndNamespace ? resourceAttributes.Concat(accountAndNamespace) : resourceAttributes, request.ResourceMetrics[0].Resource.Attributes);
```
utpilla, line 153: How is it testing if the account and namespace provided in the ConnectionString are also added to the resource?

Also, `AssertOtlpAttributes` should have failed. It looks like `AssertOtlpAttributes` should be updated to verify that the count of expected values matches the count of actual values.

vishweshbankwar: Take a look at this e48ff46359568badac25662a1ccc932d393c6c36. I refactored it a bit to make it more clear.

> AssertOtlpAttributes should have failed. It looks like AssertOtlpAttributes should be updated to verify that the count of expected values matches the count of actual values

It does here - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/113e5b82b06b4cb9da232f3f27547591d98d38f6/test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs#L955

src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
vishweshbankwar: added

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
utpilla: The tests should also check for the case where a different set of `Account` and `Namespace` are provided in the Resource by the user. Then check whether the serialized resource attributes contain both the sets of `Account` and `Namespace`.

utpilla: Not related to this PR but consider updating the `AssertOtlpAttributes` method:
- There is a redundant check for attribute key.
- There is no need to count `expectedSize` in the loop. `expectedAttributesCount` should have the expected value.

utpilla (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricExporterTests.cs
vishweshbankwar: Done - 001a1fb5a4bd8792906abf0820630d3f141f646b

CodeBlanch (approved):
LGTM

PR #1646, opened 2024-04-11
Sync Shared/AssemblyVersionExtensions with main repository

Kielek:
## Changes

Propagate changes from https://github.com/open-telemetry/opentelemetry-dotnet/pull/5528 and https://github.com/open-telemetry/opentelemetry-dotnet/pull/5498

Additional compilation fix in cc4cca59395381722439feafb3df0437b8fd8d83


src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -22,12 +24,10 @@ internal sealed class EntityFrameworkDiagnosticListener : ListenerHandler
     internal const string AttributeDbName = "db.name";
     internal const string AttributeDbStatement = "db.statement";
 
-    internal static readonly string ActivitySourceName = typeof(EntityFrameworkDiagnosticListener).Assembly.GetName().Name;
+    internal static readonly Assembly Assembly = typeof(EntityFrameworkDiagnosticListener).Assembly;
```
Oberon00, line 27: Create a new static attribute that is only used as a "local variable" during initialization could be avoided by using a static initializer like in src/OpenTelemetry.Instrumentation.Cassandra/CassandraMeter.cs, or just calling `typeof` + `.Assembly` twice.

Oberon00 (approved):

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
Kielek: Agree, but there is a problem with static fields initialization by constructor.
Problem does not occur when value are set to property.

I will keep it as in this PR.

```log
Error CA1810 : Initialize all static fields in 'AspNetMetrics' when those fields are declared and remove the explicit static constructor.
```

fred2u (approved):

rajkumar-rangaraj (approved):

vishweshbankwar (approved):

PR #1648, opened 2024-04-12
[Instrumentation.GrpcCore] Record exceptions as activity events flag

IliaBrahinets:
## Changes

`RecordException` flag has been added in both `ServerTracingInterceptorOptions` and `ClientTracingInterceptorOptions`. When set to `true` exceptions are recored as activity events.


src/OpenTelemetry.Instrumentation.GrpcCore/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   ([#1635](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1635))
 * `ActivitySource.Version` is set to NuGet package version.
   ([#1624](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1624))
+* RecordException option for both client and server interceptors
```
Kielek, line 13:
```suggestion
* `RecordException` option for both client and server interceptors
```

test/OpenTelemetry.Instrumentation.GrpcCore.Tests/GrpcCoreClientInterceptorTests.cs
```diff
@@ -355,6 +357,12 @@ static void ValidateCommonEventAttributes(ActivityEvent activityEvent)
             Assert.Contains(responseMessage.Tags, t => t.Key == SemanticConventions.AttributeMessageType && (string)t.Value == "RECEIVED");
             Assert.Contains(requestMessage.Tags, t => t.Key == SemanticConventions.AttributeMessageCompressedSize && (int)t.Value == FoobarService.DefaultResponseMessageSize);
         }
+
+        if (recordedExceptions)
+        {
+            Assert.NotNull(activity.Events);
+            Assert.Single(activity.Events, e => e.Name == SemanticConventions.AttributeExceptionEventName);
```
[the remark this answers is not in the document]
IliaBrahinets: Sounds good. The only thing which bothers me is that basically it is a validation of a result which is produced by an external library (`Activity.RecordException`), but on the other hand it is a part of the specifcation. Added validation for "exception.type", "exception.message", "exception.stacktrace". Acording to the specification https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md

IliaBrahinets:
@cijothomas , @Kielek I've made suggested changes. Ready for another review

src/OpenTelemetry.Instrumentation.GrpcCore/SemanticConventions.cs
```diff
@@ -22,5 +22,11 @@ internal static class SemanticConventions
 
     // Used for unit testing only.
     internal const string AttributeActivityIdentifier = "activityidentifier";
+
+    // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md
+    internal const string AttributeExceptionEventName = "exception";
+    internal const string AttributeExceptionType = "exception.type";
+    internal const string AttributeExceptionMessage = "exception.message";
+    internal const string AttributeExceptionStacktrace = "exception.stacktrace";
```
Kielek, line 30: All this const are used only in tests file. I am not sure if it worth to put it into production code.

src/OpenTelemetry.Instrumentation.GrpcCore/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   ([#1624](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1624))
 * Update `OpenTelemetry.Api` to `1.8.1`.
   ([#1668](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1668))
+* `RecordException` option for both client and server interceptors
```
Kielek, line 13:
```suggestion
* `RecordException` option for both client and server interceptors.
```

Kielek (commented):
@cijothomas, based on https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md

I would say that when possible exception should be always recorded as events. Do we really need the option to enable it? I would consider adding possibility to opt-out. To met the spec, it should be enabled by default.

@IliaBrahinets, please do not change the code based on this comment. We need to find agreement here, it should probably impact more packages than only this one.

cijothomas:
> @cijothomas, based on https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md
> 
> I would say that when possible exception should be always recorded as events. Do we really need the option to enable it? I would consider adding possibility to opt-out. To met the spec, it should be enabled by default.
> 
> @IliaBrahinets, please do not change the code based on this comment. We need to find agreement here, it should probably impact more packages than only this one.

I think it should be off by default, and controlled via an option. 
1. Exceptions are expensive (string allocations, and the cost of obtaining stack etc.).
2. Exceptions could contain sensitive/pii things. (in the message usually, that is included when ToString() is invoked)
3. It is possible that the exception is already reported using other mechanism (like `ILogger`), so this could be just undesired duplicate information.
4. All existing instrumentation in the main repo/this repo offer this option already and they are all opt-in.
5. We have seen folks asking for on/off flag for `ActivityEvent` in general before. One example is [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/416#issuecomment-1164728734).

src/OpenTelemetry.Instrumentation.GrpcCore/SemanticConventions.cs
IliaBrahinets: Then existing `AttributeActivityIdentifier` const should be also taken into account. I think it is simpler to manage all semantic convetions from one place within a library (including tests). Libraries which use `Shared/SemanticConventions.cs` include many unused consts. BTW probably it is worth to migrate `OpenTelemetry.Instrumentation.GrpcCore` to `Shared/SemanticConventions.cs`. though it lacks "otel.status_code" and "otel.status_description" attributes.

Kielek (approved):
LGTM.

@IliaBrahinets, please resolve conflicts. It is related to moving publicAPI file to the parent folder. Ref: #1711

@pcwiese, it will be great if you can review

IliaBrahinets:
> LGTM.
> 
> @IliaBrahinets, please resolve conflicts. It is related to moving publicAPI file to the parent folder. Ref: #1711
> 
> @pcwiese, it will be great if you can review

The conflicts have been resolved.

src/OpenTelemetry.Instrumentation.GrpcCore/SemanticConventions.cs
IliaBrahinets: Turns out that otel.status_code and otel.status_descrription are used in tests only. As the result of the discussion:
- otel.status_code and otel.status_description should be replaced by `Activity.Status` and `Activity.StatusDescription` in tests
- Use `Shared/SemanticConventions.cs` in `OpenTelemetry.Instrumentation.GrpcCore`

I would like to submit another PR for the suggested changes, @Kielek what do you think ?

Kielek: PRs are more than welcomed :)

pcwiese (approved):

PR #1666, opened 2024-04-19
[Exporter.Geneva] Resolve AOT warnings in OpenTelemetry.Exporter.Geneva

eerhardt:
## Changes

Make OpenTelemetry.Exporter.Geneva compatible with native AOT.

- Also use Regex source generator for faster startup, less memory usage, and the ability to trim the Regex engine.


reyang (approved):

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
```diff
@@ -24,6 +27,9 @@ private MetricEtwDataTransport()
     }
 
     [NonEvent]
+#if NET6_0_OR_GREATER
```
vishweshbankwar, line 30: @eerhardt - will need suppress on `SendOtlpProtobufEvent` below on line 45 as well.

vishweshbankwar (approved):

src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/MetricEtwDataTransport.cs
eerhardt: This is a good catch. It looks like we need to update the Aot test app to catch these. Will update.

eerhardt: Fixed.

utpilla:
@eerhardt GenevaExporter also uses `Expression.Lambda` from [System.Linq.Expressions](https://learn.microsoft.com/en-us/dotnet/api/system.linq.expressions?view=net-8.0). Did we not have a Trim/AOT issue when using that?

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/ab7bdabacb8907b1745ecdc2ee7544a681989e6c/src/OpenTelemetry.Exporter.Geneva/Internal/ReentrantExportProcessor.cs#L18-L25

eerhardt:
> Did we not have a Trim/AOT issue when using that?

There wasn't an AOT warning for it, which means that it will work correctly when AOT'd.

Linq Expressions do work with trimming and AOT, as long as the tooling can statically tell what methods/properties/etc you are referencing. In this case, it can since the types are all static - `Batch<T>` and `typeof(T)`.

The drawback to using Linq Expressions on native AOT'd apps is that the "compiled" expression is just compiled into an expression tree and the tree is interpreted every time it is executed. So it is a slower than it would be on a normal CoreCLR process where it can be compiled to IL and JIT'd to machine code.

If this is a problem, I would recommend making the constructor of `Batch<T>(T item)` public so this code can call it. Or the comment says this class should be gotten rid of once OpenTelemetry .NET officially supports not having synchronization. If neither of them can happen, the code can be changed to use pure Reflection when on native AOT, which performs better than interpreted Linq Expressions.

utpilla:
> If this is a problem, I would recommend making the constructor of `Batch<T>(T item)` public so this code can call it. Or the comment says this class should be gotten rid of once OpenTelemetry .NET officially supports not having synchronization. If neither of them can happen, the code can be changed to use pure Reflection when on native AOT, which performs better than interpreted Linq Expressions.

Thanks @eerhardt for the explanation! We would have to switch to using pure Reflection on Native AOT but that can be address later. For now, this PR looks good.

utpilla (approved):

eerhardt:
> We would have to switch to using pure Reflection on Native AOT

"have to" is too strong of a phrase. The code as is will work on Native AOT.

My recommendation would be to not use private/internal Reflection at all. If this API is supposed to be called outside of the assembly, it should be public. Or if this code is supposed to be removed, we should do the work to remove it.

utpilla:
> "have to" is too strong of a phrase. The code as is will work on Native AOT.

Ah yes pardon my English 😀. What I actually wanted to say was we could "consider" switching to pure reflection. We don't have to unless there is a valid complaint/ user ask for it. 

> My recommendation would be to not use private/internal Reflection at all. If this API is supposed to be called outside of the assembly, it should be public. Or if this code is supposed to be removed, we should do the work to remove it.

Unfortunately, both of those alternatives are probably going to be hard (time-taking) as the private API is coming from the main package which has stricter requirements for making things public. Maybe it's time to revisit this issue in the main repo.

eerhardt:
One other reason to use pure Reflection (besides potential throughput gains) would be for size. If this is the only code in the app pulling in Linq Expressions, there is potential size savings with trimming Linq Expressions from this code. I can look to see if that's the case for the case I care about.

PR #1669, opened 2024-04-19
Disable metric name validation for otlp format

vishweshbankwar:
Towards #1585 

## Changes

Moving the disable logic to common place so that it can be used for both tlv and otlp protobuf format.

### DIFF
--- a/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
@@ -282,9 +282,11 @@ public class GenevaMetricExporterTests
     }
 
     [Theory]
-    [InlineData(true)]
-    [InlineData(false)]
-    public void DisableMetricNameValidationTest(bool disableMetricNameValidation)
+    [InlineData(true, false)]
+    [InlineData(false, false)]
+    [InlineData(true, true)]
+    [InlineData(false, true)]
+    public void DisableMetricNameValidationTest(bool disableMetricNameValidation, bool enableOtlpProtobufexporter)
     {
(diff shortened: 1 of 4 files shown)


cijothomas (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaMetricExporterTests.cs
```diff
@@ -301,12 +303,12 @@ public void DisableMetricNameValidationTest(bool disableMetricNameValidation)
                 {
                     if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                     {
-                        options.ConnectionString = $"Account=OTelMonitoringAccount;Namespace=OTelMetricNamespace;DisableMetricNameValidation={disableMetricNameValidation}";
+                        options.ConnectionString = $"Account=OTelMonitoringAccount;Namespace=OTelMetricNamespace;DisableMetricNameValidation={disableMetricNameValidation};PrivatePreviewOtlpProtobufMetricExporter={enableOtlpProtobufexporter}";
```
utpilla, line 306: Could you conditionally append the string with the opt-in key instead of always having it in the connection string? This test was originally written to test ConnectionStrings that don't have this new opt-in key.

```csharp
var connectionString = $"Account=OTelMonitoringAccount;Namespace=OTelMetricNamespace;DisableMetricNameValidation={disableMetricNameValidation}";

if (enableOtlpProtobufexporter)
{
 connectionString = connectionString + "PrivatePreviewOtlpProtobufMetricExporter=true";
}
```

vishweshbankwar: sure, I can do that.

Issue #1219, opened 2023-06-06
[Sampler.AWS] Unit tests are consistently failing in VS


Kielek:
# Issue with OpenTelemetry.Sampler.AWS

List of [all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.3.2`):

* OpenTelemetry 1.4.0, same for 1.5.0

Runtime version (e.g. `net462`, `net48`, `net6.0`, `net7.0` etc. You can
find this information from the `*.csproj` file):

* all tests targets

**Is this a feature request or a bug?**

* [ ] Feature Request
* [x] Bug

**What is the expected behavior?**

Unit tests passed.

**What is the actual behavior?**

Execute tests from current main
![image](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/assets/5972917/8fba09e5-ca28-4932-8904-545acb08dcff)

martincostello:
`TestAWSXRayRemoteSampler.TestSamplerUpdateAndSample` is still failing, at least on `net462`.

PR #1682, opened 2024-04-23
[Sampler.AWS]  Use DateTimeOffset to beter handle timezones.

ppittle:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1219

## Changes

Changes Sampler.AWS to use `DateTimeOffset` instead of `DateTime` as it has better time zone support.

## Note

Do a second pass on change set to evaluate any breaking changes.


ppittle:
@Kielek - you were previously able to reproduce the test failure in #1219.  

Can you confirm this fix.

Kielek (approved):
All tests passed.
I do not think that it could harm anything.

Please let me know if you want to include any entry in the CHANGELOG. or we are ready to merge.

PR #1684, opened 2024-04-23
[Instrumentation.ElasticsearchClient] Fix sensitive data in url.full (former db.url) tag

### LINKED ISSUE #1194: Elasticsearch db.statement and db.url tags contain authentication details.
# Issue with OpenTelemetry.Instrumentation.ElasticsearchClient

`db.statement` and `db.url` contains an http link that is authenticated `https://user:pass@host/....`

* 1.4.0 RTM all other packages.
* OpenTelemetry.Instrumentation.ElasticsearchClient 1.0.0-beta4

* net6.0

* [ ] Feature Request
* [x] Bug

I'd expect user and passwords to be redacted.

I see the http url contain both user name and password.

soalexmn:
Fixes #1194

## Changes

This PR should partially close https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1194

Remove sensitive information (username and password) in `url.full` ( former `db.url`) as described in https://opentelemetry.io/docs/specs/semconv/database/elasticsearch/

Plus, adjust semantic conventions, use url.full instead of db.url

I'm not sure about `db.statement` tag as soon as library redact password itself (checked with my traces) and cleaning such data in a large query may affect the query data itself in case of false positives.


soalexmn:
> @soalexmn, 2 important thigs
> 
>     1. Please sign CLA. Without this we cannot proceed with PR.
> 
>     2. `db.url` seems to be kind of custom attribute. Could you please adjust semantic convention first, then make appropriate sanitisation?

Thanks, I've corrected all the points

ejsmith (approved):

src/Shared/SemanticConventions.cs
```diff
@@ -53,7 +53,7 @@ internal static class SemanticConventions
     public const string AttributeDbStatement = "db.statement";
     public const string AttributeDbOperation = "db.operation";
     public const string AttributeDbInstance = "db.instance";
-    public const string AttributeDbUrl = "db.url";
+    public const string AttributeDbUrl = "url.full";
```
Kielek, line 56: This attribute is used only by Elasticsearch package. It should be fine to rename it, but please replace also the C# const name. Without this, you are breaking convention used in the whole file.

```suggestion
    public const string AttributeUrlFull = "url.full";
```

Kielek (commented):
Please add also CHANGELOG.md entry. Inform about 2 things
1. Replacing old attribute by the new one (include names)
2. Redacting values.

src/OpenTelemetry.Instrumentation.ElasticsearchClient/CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   ([#1624](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1624))
 * Update OpenTelemetry SDK version to `1.8.1`.
   ([#1668](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1668))
+* Fix sensitive data in `url.full` (former `db.url`) tag
+  ([#1684](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1684))
+  * Replace `db.url` attribute with `url.full` to comply with conventions.
+  * Fix instrumentation to replace username and password in `url.full` with `REDACTED:REDACTED`.
```
Kielek, line 14:
```suggestion
* Replace `db.url` attribute with `url.full` to comply with [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.25.0/docs/database/elasticsearch.md#attributes).
  Redact `username` and `password` part of the `url.full`.
  ([#1684](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1684))
```

soalexmn:
I apologize for the slight mix up, the build should be going well now. Should I squash the commits?

Kielek (approved):

Kielek:
> @Kielek I apologize for the slight mix up, the build should be going well now. Should I squash the commits?

There is no need. We are squashing while merging to the main.

PR #1687, opened 2024-04-24
[Instrumentation.SqlClient] nullable enable

TimothyMothra:
Towards #894

## Changes
- remove `nullable disable` from SqlClient project and test project.


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -94,14 +94,20 @@ public override void OnEventWritten(string name, object payload)
...
+                        if (database != null)
+                        {
+                            activity.DisplayName = (string)database;
+                        }
 
                         _ = this.dataSourceFetcher.TryFetch(connection, out var dataSource);
                         _ = this.commandTextFetcher.TryFetch(command, out var commandText);
 
-                        activity.SetTag(SemanticConventions.AttributeDbName, (string)database);
+                        activity.SetTag(SemanticConventions.AttributeDbName, database);
```
Kielek, line 105: Not directly related to this PR.

Shouldn't it be in databased != statement? I think that we do not need empty/null attribute,

src/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs
```diff
@@ -132,7 +132,7 @@ public class SqlClientTraceInstrumentationOptions
     /// cref="Activity"/>.</item>
     /// </list>
     /// </remarks>
-    public Action<Activity, string, object> Enrich { get; set; }
+    public Action<Activity, string, object>? Enrich { get; set; }
```
Kielek, line 135: Just double checking if object (SqlCommand) is always non null here.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
TimothyMothra: good catch, I missed that.
I'm fine making this change in this PR, but let me know if this should be done in a follow up instead.

Kielek: I would put it in scope of this pr

src/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs
TimothyMothra: I'm not sure on the requirements, but reviewing how Enrich is invoked:

`SqlClientDiagnosticListener` first checks if command is null before invoking.


https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c34b5cd08ca9b81f04a85a0503984d7e479c4e67/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L67-L72


https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c34b5cd08ca9b81f04a85a0503984d7e479c4e67/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L133

Kielek: Thanks for confirmation

Kielek (approved):

vishweshbankwar (approved):

PR #1699, opened 2024-04-27
[Resources.Container] Add Kubernetes support

### LINKED ISSUE #1562: Incorrect container id reported
# Issue with OpenTelemetry.ResourceDetectors.Container

List of [all OpenTelemetry NuGet packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are using (e.g. `OpenTelemetry 1.3.2`):

- OpenTelemetry.Exporter.OpenTelemetryProtocol 1.7.0
- OpenTelemetry.Extensions.Hosting 1.7.0
- OpenTelemetry.Instrumentation.AspNetCore 1.7.0
- OpenTelemetry.ResourceDetectors.Container 1.0.0-beta.5

* [ ] Feature Request
* [x] Bug

The correct container id is reported in the Resource attributes.

The container id doesn't seem to be correct.

Otel Collector with debug exporter:
```
2024-02-02T09:32:08.834Z        info    ResourceSpans #0
Resource SchemaURL:
Resource attributes:
     -> container.id: Str(b7f67556f672a21eef1344ce6670797fe54e1f49bcd4fdaf8a746d2cd24acf6e)
```

But the output of `kubectl describe pods` is not reporting the same value:
```
> kubectl describe pods -n api
Name:             basket-api-765214ace9-6frsx
....
Containers:
  basket-api:
    Container ID:   containerd://b6e87264844654c35996d2e67b1bd7904b1620310bce307bddd92bd724124361
// rest of the output omitted for brevity
```

## Additional Context

The application is using .NET 8 and is running on AKS.

joegoldman2:
Fixes #1562.
Only tested manually at the moment.


src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -132,45 +152,129 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
+
+                using (var streamReader = ResourceDetectorUtils.GetStreamReader(path))
+                {
+                    while (!streamReader.EndOfStream)
+                    {
+                        stringBuilder.Append(streamReader.ReadLine()?.Trim());
+                    }
+                }
+
+                stringBuilder.Insert(0, "Bearer ");
```
matt-hensley, line 210: Instead of appending the "Bearer " prefix after the fact, could the StringBuilder instance be initialized with the value?

For example:
`var stringBuilder = new StringBuilder("Bearer ");`

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -132,45 +152,129 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
+            var credentials = GetK8sCredentials(K8sCredentialPath);
+            if (string.IsNullOrEmpty(credentials))
+            {
+                return string.Empty;
+            }
+
+            using var httpClientHandler = ServerCertificateValidationHandler.Create(K8sCertificatePath, ContainerResourceEventSource.Log);
+            var response = ResourceDetectorUtils.SendOutRequest(url, "GET", new KeyValuePair<string, string>("Authorization", credentials), httpClientHandler).GetAwaiter().GetResult();
+            var pod = DeserializeK8sResponse(response);
+            if (pod == null || pod.Status == null || pod.Status.ContainerStatuses == null)
```
matt-hensley, line 174: nit: could be simplified to `if (pod?.Status?.ContainerStatuses == null)`

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: I reused the logic from AWSEKSResourceDetector
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/9d4b5f102269f883f15974c35e0676aa2186779d/src/OpenTelemetry.ResourceDetectors.AWS/AWSEKSResourceDetector.cs#L80
I changed it in both places. Thank you!

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: Done

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -18,6 +21,13 @@ public class ContainerResourceDetector : IResourceDetector
     private const string Filepath = "/proc/self/cgroup";
     private const string FilepathV2 = "/proc/self/mountinfo";
     private const string Hostname = "hostname";
+    private const string K8sServiceHostKey = "KUBERNETES_SERVICE_HOST";
+    private const string K8sServicePortKey = "KUBERNETES_SERVICE_PORT_HTTPS";
+    private const string K8sNamespaceKey = "KUBERNETES_NAMESPACE";
```
rajkumar-rangaraj, line 26: Is this environment variable and `KUBERNETES_CONTAINER_NAME` available in all Kubernetes environments?

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -132,45 +152,127 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
+    private static string? ExtractK8sContainerId()
+    {
+        try
+        {
+            var host = Environment.GetEnvironmentVariable(K8sServiceHostKey);
+            var port = Environment.GetEnvironmentVariable(K8sServicePortKey);
+            var @namespace = Environment.GetEnvironmentVariable(K8sNamespaceKey);
+            var hostname = Environment.GetEnvironmentVariable(K8sHostnameKey);
+            var containerName = Environment.GetEnvironmentVariable(K8sContainerNameKey);
+            var url = $"https://{host}:{port}/api/v1/namespaces/{@namespace}/pods/{hostname}";
```
rajkumar-rangaraj, line 164: What is the guarantee that this will be a well-formed URL?

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: From what I know, `KUBERNETES_NAMESPACE` and `KUBERNETES_CONTAINER_NAME` are not available by default in Kubernetes. The user will have to provide them using the [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: Nothing I think but in this case an exception will be thrown, catched and logged using the EventSource.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -18,6 +21,13 @@ public class ContainerResourceDetector : IResourceDetector
     private const string Filepath = "/proc/self/cgroup";
     private const string FilepathV2 = "/proc/self/mountinfo";
     private const string Hostname = "hostname";
+    private const string K8sServiceHostKey = "KUBERNETES_SERVICE_HOST";
+    private const string K8sServicePortKey = "KUBERNETES_SERVICE_PORT_HTTPS";
+    private const string K8sNamespaceKey = "KUBERNETES_NAMESPACE";
+    private const string K8sHostnameKey = "HOSTNAME";
+    private const string K8sContainerNameKey = "KUBERNETES_CONTAINER_NAME";
```
iskiselev, line 28: The tricky part here - k8s doesn't have any standard variable to provide container name. It is entirely based on the user. I don't see any issues that extractor would expect it to be named "KUBERNETES_CONTAINER_NAME" - but probably it may be better to be passed by user to detector as constructor argument.
If it is passed by user, we can also report it as additional attribute.

joegoldman2: I think it's the same for `KUBERNETES_NAMESPACE` in this case, which is not standard as well. Perhaps we can let the user supply the values via the constructor but also keep automatic detection via the environment variables if they are not supplied. What do you think?

> If it is passed by user, we can also report it as additional attribute.

I'm not sure that's the purpose of this detector. From my point of view, the user can already use the env variable `OTEL_RESOURCE_ATTRIBUTES` or implement his own detector to add this resource attribute.

iskiselev:
Not stale. I'll try to finish review this week.

ppittle (approved):
Approve for the changes in OpenTelemetry.ResourceDetectors.AWS

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
iskiselev, line 26: Namespace should be published in `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. Probably the best way to read it from file instead of asking user to configure passing it in environment variable though downwards api.

iskiselev:
Similar work happens in opentelemtry-js: open-telemetry/opentelemetry-js-contrib#1962
There are still few open questions:
- there is no standard way to get container name. It should be provided by customer - most probably on constructor. We may try to use some environment variable as fallback option. Javascript implementation currently suggest usage of `CONTAINER_NAME` environment variable. This PR use KUBERNETES_CONTAINER_NAME environment varaible. ContainerName is not available with downwards API, but can be provided using yaml anchors or manually
- there is no standard way to pod name. It should be provided by customer - most probably on constructor. We may try to use  `HOSTNAME` environment variable as fallback option. Javascript implementation currently suggest usage of `POD_NAME` environment variable. This PR uses `HOSTNAME` environment variable for it. It may not work on some environments of if user override hostname in PodSpec. PodName can be available through environment variable with downwards API.
- Namespace is available be default in `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file. It can be also provided by downwards API, though reading it from file I like more

It would be good if we synchronize fallback environment variable with JS implementation

To make it work, customer should provide get permission on pods object. We should extend readme showing how to make it working. I was able to make it functional with next sample:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: pod-reader-account
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  namespace: default
  name: pod-reader
rules:
- apiGroups: [""] 
  resources: ["pods"]
  verbs: ["get"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: read-pods
  namespace: default
subjects:
- kind: ServiceAccount
  name: pod-reader-account
  namespace: default
roleRef:
  kind: Role
  name: pod-reader
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: v1
kind: Pod
metadata:
  name: container-resolve-demo
spec:
  serviceAccountName: pod-reader-account
  volumes:
  - name: shared-data
    emptyDir: {}
  containers:
  - name: &container_name my_container_name
    image: ubuntu:latest
    command: [ "/bin/bash", "-c", "--" ]
    args: [ "while true; do sleep 30; done;" ]
    env:
    - name: KUBERNETES_CONTAINER_NAME
      value: *container_name
    - name: KUBERNETES_POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
```

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -132,45 +152,127 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
+            var containerName = Environment.GetEnvironmentVariable(K8sContainerNameKey);
+            var url = $"https://{host}:{port}/api/v1/namespaces/{@namespace}/pods/{hostname}";
+            var credentials = GetK8sCredentials(K8sCredentialPath);
+            if (string.IsNullOrEmpty(credentials))
+            {
+                return string.Empty;
+            }
+
+            using var httpClientHandler = ServerCertificateValidationHandler.Create(K8sCertificatePath, ContainerResourceEventSource.Log);
+            var response = ResourceDetectorUtils.SendOutRequest(url, "GET", new KeyValuePair<string, string>("Authorization", credentials), httpClientHandler).GetAwaiter().GetResult();
```
iskiselev, line 172: I'm very nervous with using sync-over-async and potential dead-lock if some `ConfigureAwait(false)` is not done. Can we use [Send ](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.send) instead? Or would it create problems with default client handler or if `AllowSynchronousIO = false` for ASP.NET Core.

Otherwise, it may be safe to use:
```
    private static readonly TaskFactory _taskFactory = new
        TaskFactory(CancellationToken.None,
            TaskCreationOptions.None,
            TaskContinuationOptions.None,
            TaskScheduler.Default);

    public static T RunSync<T>(Func<Task<T>> task)
        => _taskFactory
            .StartNew(task)
            .Unwrap()
            .GetAwaiter()
            .GetResult();
```

iskiselev:
And the hardest part would be a unit test / integration for it. We can move reading environment variables / files to separate class, so that we would be able to mock it. After it we can start ASP.NET web service and use it to validate that request correctly send / parsed. It may be much more work to make it use custom certificate, so it can probably be skipped. Do you have other ideas on easier tests?

joegoldman2:
> * there is no standard way to get container name. It should be provided by customer - most probably on constructor. We may try to use some environment variable as fallback option. Javascript implementation currently suggest usage of `CONTAINER_NAME` environment variable. This PR use KUBERNETES_CONTAINER_NAME environment varaible. ContainerName is not available with downwards API, but can be provided using yaml anchors or manually

The PR in opentelemetry-js has not yet been approved. I think it's better to prefix with `KUBERNETES_` to follow the KUBERNETES naming convention. What do you think?

> * there is no standard way to pod name. It should be provided by customer - most probably on constructor. We may try to use  `HOSTNAME` environment variable as fallback option. Javascript implementation currently suggest usage of `POD_NAME` environment variable. This PR uses `HOSTNAME` environment variable for it. It may not work on some environments of if user override hostname in PodSpec. PodName can be available through environment variable with downwards API.

I added `KUBERNETES_POD_NAME` and a fallback to `HOSTNAME` if not provided.

> * Namespace is available be default in `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file. It can be also provided by downwards API, though reading it from file I like more

I removed `KUBERNETES_NAMESPACE` and I'm now reading the namespace from this file. Thank you for the suggestion!

> To make it work, customer should provide get permission on pods object. We should extend readme showing how to make it working. I was able to make it functional with next sample:

I added some details in the README explaining what permission is needed to use this detector in a Kubernetes environment.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: We have the same issue in all AWS resource detectors as they are using the same `ResourceDetectorUtils.SendOutRequest`

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c81572ee8d5dafb20c91486f967c0935c5290adf/src/OpenTelemetry.ResourceDetectors.AWS/AWSEC2ResourceDetector.cs#L95

Is it ok for you if I fix this issue for all detectors in a follow-up PR?

joegoldman2:
> And the hardest part would be a unit test / integration for it. We can move reading environment variables / files to separate class, so that we would be able to mock it. After it we can start ASP.NET web service and use it to validate that request correctly send / parsed. It may be much more work to make it use custom certificate, so it can probably be skipped. Do you have other ideas on easier tests?

I can move the reading environment variables / files to a separate class but I don't know how to inject the dependency then. Detectors always have a parameter-less constructor until now. Any suggestion?

matt-hensley:
> Detectors always have a parameter-less constructor until now. Any suggestion?

`HostDetector` [has an internal constructor to facilitate testing](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Resources.Host/HostDetector.cs#L44)

src/OpenTelemetry.ResourceDetectors.Container/README.md
```diff
@@ -34,6 +34,66 @@ your application is running:
...
+    emptyDir: {}
+  containers:
+  - name: &container_name my_container_name
+    image: ubuntu:latest
+    command: [ "/bin/bash", "-c", "--" ]
+    args: [ "while true; do sleep 30; done;" ]
+    env:
+    - name: KUBERNETES_CONTAINER_NAME
+      value: *container_name
+```
```
iskiselev, line 95: Should we add to readme how to define KUBERNETES_POD_NAME with downwards API?
```
    - name: KUBERNETES_POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
```

src/OpenTelemetry.ResourceDetectors.Container/README.md
```diff
@@ -34,6 +34,66 @@ your application is running:
 
 - **ContainerResourceDetector**: container.id.
 
+## Kubernetes
+
```
iskiselev, line 38:
```
To make container ID resolution work, container and pod name should be provided through KUBERNETES_CONTAINER_NAME and KUBERNETES_POD_NAME environment variable respectively and pod should have at least get permission to kubernetes resource pods.
It can be done by utilizing YAML anchoring, downwards API and RBAC (Role-Based Access Control).

If KUBERNETES_POD_NAME is not provided, detector will use HOSTNAME as a fallback, but it may not work in some environments or if hostname was overridden in pod spec.

Below is an example of how to configure sample pod to make container ID resolution working:

File: src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs [hunk 9]
```diff
@@ -132,45 +143,102 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
         return input.Substring(startIndex, endIndex - startIndex);
     }
 
+    private string? ExtractK8sContainerId()
+    {
+        try
+        {
+            var baseUrl = this.k8sMetadataFetcher.GetServiceBaseUrl();
```
iskiselev, line 150: We try to extract K8s result first - which is correct - but it utilize HTTP request which can be relatively slow. Can we make extract code fail early (with no warning, as it is perfectly fine - we don't want any warning on non-K8s environments) if any required environment variable is not set or any file is not available?

iskiselev (changes requested):
Looks nearly done, thank you @joegoldman2. I have suggested minor tweaks for readme and early fallback if we are non on K8s environment (or on K8s environment, but any required environment variable / file is not available).

I will be on trip next week and may not be able to approve it before Friday.

src/OpenTelemetry.ResourceDetectors.Container/README.md
joegoldman2: Done.

src/OpenTelemetry.ResourceDetectors.Container/README.md
joegoldman2: Thank you for the suggestion. Done.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: Done, please tell me what you think.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
```diff
@@ -132,45 +141,107 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
+            if (string.IsNullOrEmpty(baseUrl))
+            {
+                return null;
+            }
+
+            var @namespace = this.k8sMetadataFetcher.GetNamespace();
+            var hostname = this.k8sMetadataFetcher.GetPodName() ?? this.k8sMetadataFetcher.GetHostname();
+            var containerName = this.k8sMetadataFetcher.GetContainerName();
```
iskiselev, line 156: ContainerName is only parameter that should be provided by user manually. Should we check it first and exit if it was not provided? If it is provided, most probably user would like detector to resolve container ID through K8s API.
I thought whether we would like to have errors in log if token/namespace/certificate file are not available or should we check at least one with File.Exists before. Probably current implementation would be good, as we already would check some existence of environment variable and could reasonable think that we identified K8s environment.

src/OpenTelemetry.ResourceDetectors.Container/K8sMetadataFetcher.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace OpenTelemetry.ResourceDetectors.Container;
+
+internal sealed class K8sMetadataFetcher : IK8sMetadataFetcher
+{
+    private const string KubernetesServiceHostKey = "KUBERNETES_SERVICE_HOST";
+    private const string KubernetesServicePortKey = "KUBERNETES_SERVICE_PORT_HTTPS";
+    private const string KubernetesHostnameKey = "HOSTNAME";
+    private const string KubernetesPodNameKey = "KUBERNETES_POD_NAME";
+    private const string KubernetesContainerNameKey = "KUBERNETES_CONTAINER_NAME";
+    private const string KubernetesNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace";
```
iskiselev, line 17: One more small edge case (we can address it in follow-up PR, as it would require some preparation for testing.
CGroup based resolution was Linux-specific, but K8s mechanics should work in Windows also.
We should test, where K8s places namespace/token/certificate on Windows pods. Most probably, it would be the same path on drive `C:`. In that case, we should use C-drive prefix on Windows, otherwise it would not work when current working directory is on another drive.

iskiselev (changes requested):
Please validate that container name was defined by user and fail early if it was not.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: > If it is provided, most probably user would like detector to resolve container ID through K8s API.

I'm not sure to understand this part. Do you mean that if the user doesn't provide the container name, it means he wants to resolve the container ID via CGroup?

joegoldman2:
> Please validate that container name was defined by user and fail early if it was not.

Instead of `KUBERNETES_SERVICE_HOST`, right?

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
iskiselev: If user have not provided  container name, we will not be able to resolve it with K8s API. We have several options to treat it:
- (1.a) probably we are not on K8s environment and we should not do anything
- (1.b) probably user don't want us to resolve container id with K8s API, for example he also have not configured rights
- (2) we make sure that we are in K8s environment b other means (Kubernetes server URL environment variable + expected files exist) and we believe it is user error that container name was not provided to us. In that case we should still special handle it with proper error message and advise to look in README.

It is up to you, which option (1 or 2) you like more.

iskiselev: In option 2, we still should not do HTTP request, as it take time but we still will not be able to resolve container id.

iskiselev:
> > Please validate that container name was defined by user and fail early if it was not.
> 
> Instead of `KUBERNETES_SERVICE_HOST`, right?

You can do it either instead of or in addition. Both have some sense.

src/OpenTelemetry.ResourceDetectors.Container/ContainerResourceDetector.cs
joegoldman2: I did something: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1699/commits/2977bc7e66919e4c7652a2a60015f7338b5c3f9d. Please tell me what you think.

iskiselev (approved):
Thank you, looks good and ready to be merged.

joegoldman2:
Anything missing for this PR to be approved?

joegoldman2:
Anything missing for this PR to be merged?

joegoldman2:
There are too much issues due to conflicts between PRs. I'll wait for #1849 to be completed before fixing this one.

joegoldman2:
@iskiselev I fixed all merge conflicts. Can you please have a look? Thank you.

src/OpenTelemetry.Resources.Container/CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ from `OpenTelemetry-ResourceDetectors-Container`
 to `OpenTelemetry-Resources-Container`.
   ([#1849](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1849))
 
+* Add Kubernetes support.
```
iskiselev, line 23: That should go to ## Unreleased section

joegoldman2: Done

src/OpenTelemetry.Resources.Container/ContainerResourceEventSource.cs
```diff
@@ -0,0 +1,59 @@
...
+        if (this.IsEnabled(EventLevel.Error, (EventKeywords)(-1)))
+        {
+            this.FailedToExtractResourceAttributes(format, ex.ToInvariantString());
+        }
+    }
+
+    [Event(EventIdFailedToExtractResourceAttributes, Message = "Failed to extract resource attributes in '{0}'.", Level = EventLevel.Error)]
+    public void FailedToExtractResourceAttributes(string format, string exception)
+    {
+        this.WriteEvent(1, format, exception);
```
iskiselev, line 33: Can EventIdFailedToExtractResourceAttributes be used here instead of hard-coded 1?

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
```diff
@@ -131,45 +140,107 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
         return input.Substring(startIndex, endIndex - startIndex);
     }
 
+    private string? ExtractK8sContainerId()
+    {
+        try
+        {
+            var baseUrl = this.k8sMetadataFetcher.GetServiceBaseUrl();
+            var containerName = this.k8sMetadataFetcher.GetContainerName();
+            if (string.IsNullOrEmpty(baseUrl) && string.IsNullOrEmpty(containerName))
```
iskiselev, line 149: I think we agreed that we fail fast if either baseUrl or containerName is null.
So, `&&` should be replaced with `||`

iskiselev (changes requested):
Please fix fail-fast condition for K8s (if either url or container name not set, we should skip k8s resolution)

src/OpenTelemetry.Resources.Container/ContainerResourceEventSource.cs
joegoldman2: Yes, done.

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
joegoldman2: Fixed.

src/Shared/ResourceDetectorUtils.cs
```diff
@@ -12,7 +12,7 @@
 #endif
 using System.Threading.Tasks;
 
-namespace OpenTelemetry.Resources.AWS;
```
iskiselev, line 15: In `ResourceDetectorUtils.SendOutRequest` one-time used `HttpClient` can be created and not disposed.
It would be great to fix it in future to something like:
```
using var httpClient = handler == null ? new HttpClient() : new HttpClient(handler, false);
```

iskiselev (approved):

src/Shared/ResourceDetectorUtils.cs
joegoldman2: Yes, I've already planned to rework `ResourceDetectorUtils` in a future PR as agreed above (in particular to avoid sync-over-async).

test/OpenTelemetry.Resources.Container.Tests/OpenTelemetry.Resources.Container.Tests.csproj
```diff
@@ -7,12 +7,14 @@
     <Nullabe>enable</Nullabe>
   </PropertyGroup>
 
-  <ItemGroup>
-    <ProjectReference Include="$(RepoRoot)\src\OpenTelemetry.Resources.Container\OpenTelemetry.Resources.Container.csproj" />
+  <ItemGroup Condition="'$(TargetFramework)' != '$(NetFrameworkMinimumSupportedVersion)'">
```
Kielek, line 10: Is this condition really needed? You are targeting only .NET 6+ here.

src/OpenTelemetry.Resources.Container/SourceGenerationContext.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NET6_0_OR_GREATER
```
Kielek, line 4: Remove conditional compilation. Whole project is targeting .NET6+.

src/OpenTelemetry.Resources.Container/README.md
```diff
@@ -35,6 +35,75 @@ your application is running:
 
 - **ContainerDetector**: container.id.
 
+## Kubernetes
+
+To make container ID resolution work, container and pod name should be provided
+through KUBERNETES_CONTAINER_NAME and KUBERNETES_POD_NAME environment variable
```
Kielek, line 41:
```suggestion
through `KUBERNETES_CONTAINER_NAME` and `KUBERNETES_POD_NAME` environment variable
```

src/OpenTelemetry.Resources.Container/README.md
```diff
@@ -35,6 +35,75 @@ your application is running:
...
+respectively and pod should have at least
+get permission to kubernetes resource pods.
+It can be done by utilizing YAML anchoring, downwards API
+and RBAC (Role-Based Access Control).
+
+If KUBERNETES_POD_NAME is not provided, detector will use HOSTNAME
```
Kielek, line 47:
```suggestion
If `KUBERNETES_POD_NAME` is not provided, detector will use `HOSTNAME`
```

src/OpenTelemetry.Resources.Container/K8sMetadataFetcher.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace OpenTelemetry.Resources.Container;
+
+internal sealed class K8sMetadataFetcher : IK8sMetadataFetcher
+{
+    private const string KubernetesServiceHostKey = "KUBERNETES_SERVICE_HOST";
+    private const string KubernetesServicePortKey = "KUBERNETES_SERVICE_PORT_HTTPS";
+    private const string KubernetesHostnameKey = "HOSTNAME";
+    private const string KubernetesPodNameKey = "KUBERNETES_POD_NAME";
+    private const string KubernetesContainerNameKey = "KUBERNETES_CONTAINER_NAME";
```
Kielek, line 16: Consider: KubernetesServiceHostKey -> KubernetesServiceHostEnvironmentVariable/ KubernetesServiceHostEnvVar

Same for others.

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
```diff
@@ -40,25 +46,29 @@ internal enum ParseMode
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
-        var cGroupBuild = this.BuildResource(Filepath, ParseMode.V1);
-        if (cGroupBuild == Resource.Empty)
+        var resource = this.BuildResource(Filepath, ParseMode.K8s);
```
Kielek, line 49: Call directly, ExtractK8sContainerId.
Passing unused file path can be misleading. You can fully remove ParseMode.K8s.

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
```diff
@@ -131,45 +140,107 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
 
+    private string? ExtractK8sContainerId()
+    {
+        try
+        {
+            var baseUrl = this.k8sMetadataFetcher.GetServiceBaseUrl();
+            var containerName = this.k8sMetadataFetcher.GetContainerName();
+            if (string.IsNullOrEmpty(baseUrl) || string.IsNullOrEmpty(containerName))
+            {
+                return null;
```
Kielek, line 151: In the method, you are returning null or string.Empty. I think you should decide common convention.

Kielek (commented):
As I understand, K8s environmental variables are not standard. Is there any chance to define then on semantic convention level?
https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/container.md
or https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/k8s.md
?

If not possible, it should be agreement at least with go-lang community as you working on the same feature.


Do you know if other languages implements similar feature?

test/OpenTelemetry.Resources.Container.Tests/OpenTelemetry.Resources.Container.Tests.csproj
joegoldman2: Not needed. Removed.

src/OpenTelemetry.Resources.Container/SourceGenerationContext.cs
joegoldman2: Removed.

src/OpenTelemetry.Resources.Container/README.md
joegoldman2: Done.

src/OpenTelemetry.Resources.Container/README.md
joegoldman2: Done.

src/OpenTelemetry.Resources.Container/K8sMetadataFetcher.cs
joegoldman2: Renamed.

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
joegoldman2: Thank you for the suggestion. I did something, can you please tell me if it is better?

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
joegoldman2: Fixed.

joegoldman2:
> Do you know if other languages implements similar feature?

JS is currently implementing the same feature. @iskiselev asked them to align with us: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1962#issuecomment-2127903740.

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
```diff
@@ -131,47 +177,58 @@ private static string RemovePrefixAndSuffixIfNeeded(string input, int startIndex
...
         {
-            ContainerExtensionsEventSource.Log.ExtractResourceAttributesException($"{nameof(ContainerDetector)} : Failed to extract Container id from path", ex);
+            ContainerResourceEventSource.Log.ExtractResourceAttributesException($"{nameof(ContainerDetector)}: Failed to extract container id", ex);
         }
 
         return null;
+
+        static K8sPod? DeserializeK8sResponse(string response)
+        {
+#if NET6_0_OR_GREATER
```
Kielek, line 227: Can this part be also simplified? All targets to .NET6+

src/OpenTelemetry.Resources.Container/ContainerDetector.cs
Kielek: LGTM, thanks.

iskiselev:
Logic / code wise it looks good to me. But I fully support concerns of @Kielek about establishing and following standards across OTEL implementation. And it is tricky for that logic.

We are working on container resource detector package. Talking in terms defined in [container](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/container.md) and [k8s](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/k8s.md) resources, in this PR logic to extract `container.id` in Kuberentes environment was implemented. To do it, we need to know: `k8s.container.name`, `k8s.namespace.name`, and `k8s.pod.name`.  There is well defined way to extract `k8s.namespace.name`, semi-working way to resolve `k8s.pod.name` and no way to resolve `k8s.container.name`. To overcome that, we suggest use user-defined environment variables `KUBERNETES_CONTAINER_NAME` and `KUBERNETES_POD_NAME` (with advices how to configure it in README.md.

There are attributes that "MAY be configurable via a dedicated environment variable" per [resources](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/README.md#semantic-attributes-with-dedicated-environment-variable) and [variables list](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md), but all OTEL used variables use `OTEL_` prefix.

Should that package be responsible for both cgroup containers and k8s container? Or should it be separate packages? Is it ok for detector to identify resource in another category? How detectors / detector package can reuse data collected by another resource detector?

[SKD spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) suggests:
> Custom resource detectors related to generic platforms (e.g. Docker, Kubernetes) or vendor specific environments (e.g. EKS, AKS, GKE) MUST be implemented as packages separate from the SDK.
> Resource detector packages MAY detect resource information from multiple possible sources and merge the result using the Merge operation described above.

With all of that: @Kielek, do you think we can merge it in current state (we still have pre-release of the package), and open issues here and semantic-conventions repo to discuss it? If not, I can suggest to create special user-provided configuration to enable K8s container resolution and do not resolve it from environment variable internally. Something that would be used like
```
var tracerProvider = Sdk.CreateTracerProviderBuilder()
                        // other configurations
                        .ConfigureResource(resource => resource
                            .AddContainerDetector(
                               settings => settings.EnableK8sResolution(
                                 Environment.GetVaraible("KUBERNETES_CONTAINER_NAME"),
                                 Environment.GetVaraible("KUBERNETES_POD_NAME")))
                        .Build();
```

iskiselev (approved):

lachmatt:
As this is something that can already be done by collector, as noted in this [comment](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1962#issuecomment-2075526877), I think it'd be better to have consensus on this approach in broader opentelemetry community, as suggested by @Kielek, before merging this PR.
I assume scenario with applications running in k8s cluster sending telemetry directly to the backend (e.g without collector deployed) is not that common.

joegoldman2:
> As this is something that can already be done by collector

From what I understood from [the doc](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md) and tested, the Kubernetes attributes processor doesn't allow to set container.id but if provided the processor can the attributes k8s.container.name, container.image.name and container.image.tag.

iskiselev:
If I read k8s processor code correct, it in fact [set container id](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/4700e4b262fed6676a6af6191ee3765e64c3c3b1/processor/k8sattributesprocessor/processor.go#L250C68-L250C79) if it was not set. For it, AttributeK8SContainerName should be set. Same is [documented](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/5f53e33dd7ad4f0ccc797d5eb9145f4fc6749fdf/processor/k8sattributesprocessor/README.md?plain=1#L83).

iskiselev:
I have an idea how we can finish this PR and still be in full compliance with already approved semantic conventions and idea behind k8s processor implementation. Though it will have one small hack (that can be improved later once proper API will be provided).

K8s processor imply that user will provide `k8s.container.name` to be able to resolve container id. And user can do it already, using OTEL_RESOURCE_ATTRIBUTES environment variable. All it need to do is to add proper pairs there. Something like:
```
apiVersion: v1
kind: Pod
metadata:
  name: container-resolve-demo
spec:
  serviceAccountName: pod-reader-account
  volumes:
  - name: shared-data
    emptyDir: {}
  containers:
  - name: &container_name my-container-name
    image: ubuntu:latest
    command: [ "/bin/bash", "-c", "--" ]
    args: [ "while true; do sleep 30; done;" ]
    env:
    - name: OTEL_RESOURCE_ATTRIBUTES_K8S_CONTAINER_NAME
      value: *container_name
    - name: OTEL_RESOURCE_ATTRIBUTES_K8S_POD_NAME
      valueFrom:
        fieldRef:
          fieldPath: metadata.name
    - name: OTEL_RESOURCE_ATTRIBUTES
      value: k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_K8S_POD_NAME),k8s.container.name=$(OTEL_RESOURCE_ATTRIBUTES_K8S_CONTAINER_NAME)
```

Now, instead of accessing some environment variables, we just need to get attributes from `OTEL_RESOURCE_ATTRIBUTES` - effectively use one resource detector in another one. I have not seen it anywhere, but we can do something like:
```
internal static class K8sSemanticConventions
{
    public const string AttributePodName = "k8s.pod.name";
    public const string AttributeContainerName = "k8s.container.name";
}

private string? ExtractK8sContainerId()
{
  // Is there any way to do it better?
  var environmentVariableAttributes = ResourceBuilder.CreateEmpty()
          .AddEnvironmentVariableDetector().Build().Attributes;
  var podName = environmentVariableAttributes
      .FirstOrDefault(atr => atr.Key == K8sSemanticConventions.AttributeContainerName)
      .Value?.ToString();
  var containerName = environmentVariableAttributes
      .FirstOrDefault(atr => atr.Key == K8sSemanticConventions.AttributePodName)
      .Value?.ToString();
  this.k8sMetadataFetcher = k8sMetadataFetcher;
  ...
}

```

joegoldman2:
> And user can do it already, using OTEL_RESOURCE_ATTRIBUTES environment variable.

With this approach, the pod name and container name will be part of the Resource (as it's quite common to enable the env variable detector) and will be included in all traces/metrics, etc 😕

iskiselev:
@joegoldman2, currently I see several options:
- start new discussion in opentelemetry-specification about new environment variable 
- make customer pass container name manually as parameter
- use container name as already specified resource
- move K8s logic to separate detector that would require container name on constructor (and somehow make container detector never override container-id detected by controller)
- do nothing, suggesting to use processor feature on collector (not everyone use collector!)

Or some combination of that approaches. I don't like last option (do nothing).

joegoldman2:
@iskiselev I'm not really sure what to do next? Do you want to initiate a new discussion in opentelemetry-specification about new environment variables?

joegoldman2:
Not stale. Waiting for @iskiselev's feedback.

iskiselev:
Not stale. open-telemetry/opentelemetry-specification#4140 opened to discuss next step.
@joegoldman2, if new config environment variables will be approved, we will still need to make it configurable though factory method parameters per:
https://github.com/open-telemetry/opentelemetry-specification/blob/27bc9ffad2b386626367dd2f98d0ba2ddf35a18f/specification/configuration/sdk-environment-variables.md?plain=1#L54

joegoldman2:
Ok, so I guess the next step is to define the desired API to allow users to set the values programmatically, right?

iskiselev:
@joegoldman2,  I already suggested API at https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1699#issuecomment-2164160848, but any would works once it allows to specify values.
I'm not sure, whether we will see any move open-telemetry/opentelemetry-specification#4140. With program API we should be able to merge this PR even without environment values support if consensus would not be reached.
Btw, do you see an option to merge it already (with @Kielek approval) or is it requires anything else? I still fully agree with @Kielek , that we should either get common agreement on environment values usage and document them - or do not use them at all.

iskiselev:
And looks like new environment variable would not be added any time soon: https://github.com/open-telemetry/opentelemetry-specification/issues/2891#issuecomment-1289241503. So, using API-based values may be the only option.

joegoldman2:
Thank you for the API proposal https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1699#issuecomment-2164160848. What is the procedure to review/validate a new API?

iskiselev:
@joegoldman2, I actually don't know what should be the procedure to review/validate a new API. From my point of view (I'm curently codeowner for src/OpenTelemetry.Resources.Container), anything that will give a way to provide required values from code should work, as it will be in compliance with specification. As it will be also isolated as arumnets of AddContainerDetector method, we probably don't need any wider audience review. Alternatvely we can try to visit Tuesday meeting and seek opinions / advices there.

iskiselev:
BTW, we just finished validation of using k8sattributes processor on collector, and it perfectly solves the problem of resovling container.id for k8s cluster (once `k8s.container.name` attibute added).

src/OpenTelemetry.Resources.Container/ContainerResourceEventSource.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics.Tracing;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.Resources.Container;
+
+[EventSource(Name = "OpenTelemetry-Resources-Container")]
+internal class ContainerResourceEventSource : EventSource, IServerCertificateValidationEventSource
```
WeihanLi, line 10:
```suggestion
internal sealed class ContainerResourceEventSource : EventSource, IServerCertificateValidationEventSource
```

joegoldman2:
> BTW, we just finished validation of using k8sattributes processor on collector, and it perfectly solves the problem of resovling container.id for k8s cluster (once `k8s.container.name` attibute added).

Do you have a link to the related issue?

iskiselev:
@joegoldman2, we published instruction how to use K8 attribute processor for resolveing container id in OTEL collector: https://docs.appdynamics.com/observability/cisco-cloud-observability/en/application-performance-monitoring/monitor-applications-in-kubernetes/infrastructure-correlation#InfrastructureCorrelation-KubernetesAttributeProcessorConfiguredontheOpenTelemetryCollector

PR #1718, opened 2024-05-13
[Instrumentation.AspNetCore] Nullable enabled

Kielek:
Towards #894

## Changes

[Instrumentation.AspNetCore] Nullable enabled
Move _shared_ files to Shared folder


test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -1208,22 +1213,27 @@ public override void Inject<T>(PropagationContext context, T carrier, Action<T,
...
         private readonly SamplingDecision samplingDecision = samplingDecision;
-        private readonly IEnumerable<KeyValuePair<string, object>> attributes = attributes;
+        private readonly IEnumerable<KeyValuePair<string, object>>? attributes = attributes;
 
         public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
         {
-            return new SamplingResult(this.samplingDecision, this.attributes);
+            if (this.attributes != null)
+            {
+                return new SamplingResult(this.samplingDecision, this.attributes);
```
Kielek, line 1225: @open-telemetry/dotnet-approvers, I have a question about accepted parameters here.

Attributes is forced to be not null https://github.com/open-telemetry/opentelemetry-dotnet/blob/core-1.8.1/src/OpenTelemetry/Trace/SamplingResult.cs#L35, but the constructor only with SamplingDecision as parameters, internally [is calling this one with a `null` as a value](https://github.com/open-telemetry/opentelemetry-dotnet/blob/core-1.8.1/src/OpenTelemetry/Trace/SamplingResult.cs#L15-L16).

Is it intentional? Or it can be changed?  Similar case is for `traceStateString` parameter.

vishweshbankwar: Does not look intentional. @CodeBlanch - do you know?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -61,33 +58,40 @@ public HttpInListener(AspNetCoreTraceInstrumentationOptions options)
         this.options = options;
     }
 
-    public override void OnEventWritten(string name, object payload)
+    public override void OnEventWritten(string name, object? payload)
     {
+        var activity = Activity.Current;
+
+        if (activity == null)
```
vishweshbankwar, line 65: Do we need this check? Could we `Activity.Current!`? IIRC, this is already checked at caller side.

Kielek: Fixed: 1aaea5295b0b1d1c41c588bb94fe1e1989845d65

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
Kielek: Created https://github.com/open-telemetry/opentelemetry-dotnet/pull/5614 to discuss it/fix it.

vishweshbankwar (approved):

PR #1831, opened 2024-05-22
[Instrumentation.Http] Set error description and error type for Task Cancellation on HTTPClient

### LINKED ISSUE #1812: HttpClient not setting error message when using CancellationToken
### Component

### Package Version

### Runtime Version

net8.0

### Description

HttpClient supports using the CancellationToken to cancel requests in-flight. The current implementation of the HttpHandlerDiagnosticListener will set an error status in OnStopActivity if the request was canceled and the comments specify that it relies on the OnException handler to set the span status and description. However, when using CancellationToken the OnException handler will not be raised (presumably because the TaskCanceledException that is triggered by a cancellation is already handled and does not bubble up to the OnException handler as you would expect). This means that telemetry exported will have an error status set, but there will be no error message or exception available on the span.

### Steps to Reproduce

- Execute a long-running HTTP call via HttpClient
- Use the CancellationToken to cancel the request before it completes

### Expected Result

I would expect to see some information regarding the nature of the error. It might not be needed to include the TaskCanceledException's full stacktrace here as that exception is 'by design', but there should at least be some feedback available on the span to indicate why the status was set to error.

### Actual Result

There is no message or exception available

### Additional Context

_No response_

jdt:
When HTTPClient Task Cancellation occurs, set the status message and assign error.type

Fixes #1812.

## Changes

* Set an error description
* Set error.type


src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -181,7 +181,9 @@ public void OnStopActivity(Activity activity, object payload)
                 {
                     if (currentStatusCode == ActivityStatusCode.Unset)
                     {
-                        activity.SetStatus(ActivityStatusCode.Error);
+                        // Task cancellation won't trigger the OnException so set the span error information here
+                        activity.SetStatus(ActivityStatusCode.Error, "Operation Canceled");
```
vishweshbankwar, line 185: @jdt - We also need to fix https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerMetricsDiagnosticListener.cs. `error.type` should be added to metric also.

vishweshbankwar, line 185: @jdt - I left a comment on the [issue](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1812#issuecomment-2135825058). I think we should do this for metric also. Wanted to see if @antonfirsov has any additional thoughts on this.

jdt: Makes sense to keep this the same, even if it would somewhat change the current implementation. I'll wait for feedback and add it to my PR if we reach an agreement on it.

vishweshbankwar: Thanks @antonfirsov. I think Task Cancelled would be a better choice then. @jdt - does that sounds good to you?

PR #1850, opened 2024-05-30
[Exporter.Geneva] Add TraceState support

xiang17:
Add TraceState support in Geneva exporter.

## Changes


rajkumar-rangaraj (approved):
LGTM

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
```diff
@@ -243,6 +243,13 @@ internal int SerializeActivity(Activity activity)
             cntFields += 1;
         }
 
+        if (!string.IsNullOrEmpty(activity.TraceStateString))
+        {
+            cursor = MessagePackSerializer.SerializeAsciiString(buffer, cursor, "traceState");
+            cursor = MessagePackSerializer.SerializeAsciiString(buffer, cursor, activity.TraceStateString);
```
reyang, line 249: Could this be a non-ASCII string, if that's the case what should be the expected behavior?

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
```diff
@@ -243,6 +243,13 @@ internal int SerializeActivity(Activity activity)
             cntFields += 1;
         }
 
+        if (!string.IsNullOrEmpty(activity.TraceStateString))
```
cijothomas, line 246: could we make this opt-in? `Activity` does some special mechanisms (walk the parent chain...), to find tracestatestring, so it may not be free! (That was one reason why it was removed in OTel LogRecord).

I'd suggest to do benchmarks and see if this affects perf. (make sure to test with A1->A2->A3 scenario with tracestate set on A1 only/similar)

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
reyang: I understand the spec part (I wrote part of that spec). My question was about the .NET implementation https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activity.tracestatestring.

joegoldman2: The .NET implementation accepts non-ASCII characters: [sharplab.io](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAIwCwAUPgAwAE+xALANwXV3EB0AIlBAOYwA9gGcEUAMbCW5VrQBuEZDQA2UUQFMY6pQF4aWgO40AguLFyoCAJ4AZNQk3aKAbwo13NAMoALQamUYdhowACqCNHoA+hEAfDQIaOpwbh6eEGAADsrqETQAFMjqAGYmZlAW1gDChRBigjAA8hl1MMIAPKbmllaV9Q4AHghxkQCUsaVd1mmZqjB8AErqwv4IHMbKyly1EMYwGIvigsgY6hgUAL7SneXdnn7I4uprGBjXFbb2jsh5qsHaI9JWMQqABtAC6NAw21yIIARFQqABaBjAIoATkIRQAzABWADseOAWIYUIAbBAsY8MWiMFR1HSGHisaTEQiilRSXiIBAqMAIGiqIRgHjWcRYXAaLCioJBDoIMBxABLwC8G4AAPdhYMB5B4omQUGAqAcGF6MAGCAACshBBl+LUjjQMlabXw7boaDqEvrDacTWbLdbbQgjhxqupauouMUICt/c7XXkARQKI6Ay6g8gOABRQbICBmEK5x4ASVee08CHDeShFeS5A87FJ+UEwAAVuozAB+GjiRR67QS+i0IpQdQBABy6SSND8CHYVC7w9HGAAahBlKgpzP2Fi2oOYguRwFV+ulmMdDEUu5XHX6x5F0e1xvhLkTkVo8oENJbx4FEpqxBcjyQdwRGHtkD7ZAAEIv2/Gh7xXR8cj0eCJzAHJhAMSxxG8S9b2vWDb1hBI83UG1ClNWFxn/EEqDBWsCPrIjCyWCsHEo89IW2EFiDo3CCOiDjX3fBB6Ngy5cPOUT3C3X94mY81FE0ETcJkxQ5JI8tw0TGRKGINE8mIx5NIcbSKFkvNJisO5UAeJD9HUIw3lue5HjyWEAFkrCcqYXPUWFtPMsp3l9dRBlybyej6ULVgU5BhHUAz5MU01IIlQz1GM9QArUiybmsXJcveazbI4TTkAQCK3M8iLxQmPKrAAaVgDBSu0OR+xoQruhCwZTN0/SuusSCOALDTWIyz05m0oA).

xiang17: I found opentelemetry doing validation for the tracestate's key and value: https://github.com/open-telemetry/opentelemetry-dotnet/blob/4d02fe1a0803daaf6e5db1cf6f45f5d8d89c99a4/src/OpenTelemetry.Api/Context/Propagation/TraceStateUtilsNew.cs#L166

I'll add something similar to validate here as well.

Regarding the expected behavior, [Section 4.3.3](https://www.w3.org/TR/trace-context/#a-traceparent-is-received) of the spec says:
> The vendor MAY validate the tracestate header. If the tracestate header cannot be parsed the vendor MAY discard the entire header. Invalid tracestate entries MAY also be discarded.

Looks like I can either drop invalid entries or the entire header. I'll keep aligned with the core library to drop the entire header.

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
xiang17: It's already made into Part B. I'm not sure if it's right to make it opt-in. If anything I feel like it's should be more of an opt-out if a customer is concerned of the performance impact.

I'll run the benchmarks and test its effects on performance.

Can you link me to more details regarding the removal in OTel LogRecord?

cijothomas: > Can you link me to more details regarding the removal in OTel LogRecord?

https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Logs/ILogger/OpenTelemetryLoggerOptions.cs#L78

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
reyang: @xiang17 I don't understand, why making this complex instead of just serializing the Unicode string directly? The exporter is not supposed to validate the tracestate content IMHO.

```suggestion
            cursor = MessagePackSerializer.SerializeUnicodeString(buffer, cursor, activity.TraceStateString);
```

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
xiang17: Thanks, Cijo. I have found the original discussion on it: https://github.com/open-telemetry/opentelemetry-dotnet/pull/4334#discussion_r1160873655

But the difference is that the use of tracestate on log record is "probably rare" and more importantly,  [TraceState is not on log data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#log-and-event-record-definition).

However, traceState is in Span of Part B.

xiang17: I did some benchmark tests (see code in PR). Here is also some data with those tests for with/without the code change in src files.

# Without code change in MsgPackTraceExporter
| Method                                       | Mean     | Error   | StdDev  | Gen0   | Allocated |
|--------------------------------------------- |---------:|--------:|--------:|-------:|----------:|
| ExportActivity                               | 276.2 ns | 2.20 ns | 2.06 ns |      - |         - |
| SerializeActivity                            | 240.1 ns | 4.41 ns | 6.74 ns |      - |         - |
| SerializeActivityWithoutTraceState           | 253.9 ns | 2.16 ns | 2.02 ns |      - |         - |
| SerializeActivityWithTraceState              | 257.7 ns | 2.30 ns | 2.15 ns |      - |         - |
| SerializeActivityWithTraceStateInGrandparent | 258.0 ns | 3.10 ns | 2.59 ns |      - |         - |
| CreateActivityWithGenevaExporter             | 441.7 ns | 8.79 ns | 7.79 ns | 0.0396 |     416 B |


# With code change in MsgPackTraceExporter
| Method                                       | Mean     | Error   | StdDev  | Gen0   | Allocated |
|--------------------------------------------- |---------:|--------:|--------:|-------:|----------:|
| ExportActivity                               | 266.2 ns | 0.34 ns | 0.28 ns |      - |         - |
| SerializeActivity                            | 243.1 ns | 1.42 ns | 1.26 ns |      - |         - |
| SerializeActivityWithoutTraceState           | 248.7 ns | 0.78 ns | 0.65 ns |      - |         - |
| SerializeActivityWithTraceState              | 273.8 ns | 3.75 ns | 3.32 ns |      - |         - |
| SerializeActivityWithTraceStateInGrandparent | 280.2 ns | 5.54 ns | 5.69 ns |      - |         - |
| CreateActivityWithGenevaExporter             | 452.1 ns | 4.50 ns | 3.99 ns | 0.0396 |     416 B |




# Without code change in TldTraceExporter
| Method                                           | Mean       | Error     | StdDev    | Allocated |
|------------------------------------------------- |-----------:|----------:|----------:|----------:|
| MsgPack_SerializeActivity                        | 239.387 ns | 0.6162 ns | 0.4811 ns |         - |
| TLD_SerializeActivity                            | 283.528 ns | 1.9925 ns | 1.8638 ns |         - |
| TLD_SerializeActivityWithoutTraceState           | 293.012 ns | 3.6684 ns | 3.2519 ns |         - |
| TLD_SerializeActivityWithTraceState              | 293.466 ns | 1.1386 ns | 1.0093 ns |         - |
| TLD_SerializeActivityWithTraceStateInGrandparent | 285.946 ns | 0.9807 ns | 0.9174 ns |         - |
| MsgPack_ExportActivity                           | 267.778 ns | 1.0824 ns | 1.0125 ns |         - |
| TLD_ExportActivity                               |   1.050 ns | 0.0124 ns | 0.0110 ns |         - |

# With code change in TldTraceExporter
| Method                                           | Mean       | Error     | StdDev    | Allocated |
|------------------------------------------------- |-----------:|----------:|----------:|----------:|
| MsgPack_SerializeActivity                        | 239.467 ns | 1.1202 ns | 0.9930 ns |         - |
| TLD_SerializeActivity                            | 279.810 ns | 1.2504 ns | 0.9763 ns |         - |
| TLD_SerializeActivityWithoutTraceState           | 294.145 ns | 1.2952 ns | 1.1481 ns |         - |
| TLD_SerializeActivityWithTraceState              | 319.608 ns | 5.9573 ns | 5.5724 ns |         - |
| TLD_SerializeActivityWithTraceStateInGrandparent | 314.195 ns | 1.1738 ns | 0.9164 ns |         - |
| MsgPack_ExportActivity                           | 265.354 ns | 1.0098 ns | 0.9445 ns |         - |
| TLD_ExportActivity                               |   1.102 ns | 0.0469 ns | 0.0482 ns |         - |

The weird thing is that if tracestate is in grandparent activity, sometimes it's faster than if it's in the grandchild activity.

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
xiang17: I had a misunderstanding on it. Updated it to use Unicode.

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
reyang: > could we make this opt-in?

+1

test/OpenTelemetry.Exporter.Geneva.Benchmarks/Exporter/TraceExporterBenchmarks.cs
```diff
@@ -99,6 +150,24 @@ public void SerializeActivity()
         this.exporter.SerializeActivity(this.activity);
     }
 
+    [Benchmark]
+    public void SerializeActivityWithoutTraceState()
```
utpilla, line 154: `TraceState` is too specific to have dedicated benchmarks for it. I would be in favor of keeping the benchmarks simple and targeted towards the primary use case.

cijothomas: we only need it for discussion in this PR, no need to merge this changes.

xiang17: Sure, I'm temporarily using it to test the impact. I'll remove it from the Benchmark.

src/OpenTelemetry.Exporter.Geneva/MsgPackExporter/MsgPackTraceExporter.cs
cijothomas: 285.946 ns to 314.195 ns is ~10% perf hit. (Most likely, the deeper the chain is, the more the perfhit is)

> But the difference is that the use of tracestate on log record is "probably rare"
I am not sure if that is the case or not.. It could be!

> However, traceState is in Span of Part B.
That does not prevent us from making it opt-in.

xiang17: > That does not prevent us from making it opt-in.

I see. I'll add an option.

Meanwhile, just for the sake of discussion, I agree with your claim that deeper chain could cause more perfhit because the implementation of `Activity.TraceStateString`. However, there are two strange things contrary to my understanding:
1. in many cases, `TLD_SerializeActivityWithTraceStateInGrandparent` is faster than `TLD_SerializeActivityWithTraceState` in the same run. 285.946 ns < 293.466 ns is an example and I've got other examples after running `TraceExporterBenchmarks` multiple times.
2. I step debugged the unit test to `var activity = source.StartActivity("SayHello", ActivityKind.Internal, parentActivity.Context, null, links)` line, and found that right after this line, VS shows that `activity.Parent` is `null` and `TraceStateString` got `parentActivity`'s value `"some=state"`. Instead, I was expecting at least `activity.Parent` points to `parentActivity`. (`activity.TraceStateString` might have got its value while it's being displayed.)

I'm not sure what I'm missing.

reyang: I'm more coming from security/privacy side - if users bump the exporter version and start to see tracestate leaking EUII and there is no easy way to turn it off, they might be very upset.

test/OpenTelemetry.Exporter.Geneva.Benchmarks/Exporter/TraceExporterBenchmarks.cs
xiang17: Reverted all changes in Benchmark.

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -25,6 +25,8 @@ public class GenevaExporterOptions
 
     public EventNameExportMode EventNameExportMode { get; set; }
 
+    public bool IncludeTraceState { get; set; }
```
cijothomas, line 28:
```suggestion
    public bool IncludeTraceStateForSpan { get; set; }
```

something to indicate this is only for Spans, not LogRecords..

xiang17: Would `IncludeTraceStateForTrace` be a better name?

xiang17: "TraceStateForTrace" might be a little tongue twisting...

xiang17: Renamed to `IncludeTraceStateForSpan`.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Update GenevaTraceExporter to export `activity.TraceStateString` as the value
```
cijothomas, line 5: nit: 
1. make it clear that this is opt-in feature. i.e not enabled by default.
2. make it clearer that this is only for Spans.

cijothomas (approved):

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
xiang17: I've updated the CHANGELOG and also added a section about this option in the README file.

reyang (approved):
LGTM.

PR #1854, opened 2024-06-01
[Instrumentation.AspNetCore, Instrumentation.Http] Cache activity names

### LINKED ISSUE #1759: [Instrumentation.AspNetCore, Instrumentation.Http] Cache activity names
> Q: Should we maintain a cache of these `$"{namePrefix} {httpRoute}"` names somewhere?

_Originally posted by @CodeBlanch in https://github.com/open-telemetry/opentelemetry-dotnet/pull/5484#discussion_r1543724380_

joegoldman2:
Fixes #1759.

## Changes

Introduce a static `ConcurrentDictionary` to cache the activity display name. I don't think it's worth adding an entry in the changelog for this change.


src/Shared/RequestDataHelper.cs
```diff
@@ -26,6 +27,7 @@ internal sealed class RequestDataHelper
     private const string OtherHttpMethod = "_OTHER";
 
     private static readonly char[] SplitChars = new[] { ',' };
+    private static readonly ConcurrentDictionary<string, string> DisplayNameCache = new ConcurrentDictionary<string, string>();
```
Kielek, line 30: Should be the size limit for this collection?

joegoldman2: I don't expect this collection to grow to more than a few hundred items (which is already a lot).

joegoldman2: You're right, the dictionary will grow without limit. Let me try to thing about a solution.

joegoldman2: My understanding is that there's no built-in "bounded" collection in .NET. While it's possible to implement a custom one, probably by wrapping `ConcurrentDictionary` with additional checks, I'm not entirely enthusiastic about it. I'm open to suggestions.

vishweshbankwar: @joegoldman2 - One possible option is to check the Count of items within the dictionary before adding a new item. The Count could be some pre-defined number. If the Count exceeds that number, then we could simply skip adding items to the Dictionary.

src/Shared/RequestDataHelper.cs
```diff
@@ -84,7 +87,27 @@ public void SetActivityDisplayName(Activity activity, string originalHttpMethod,
...
+            }
+
+            if (DisplayNameCache.TryGetValue(httpRoute!, out var displayName))
+            {
+                return displayName;
+            }
+
+            if (DisplayNameCache.Count < DisplayNameCacheSize)
+            {
+                return DisplayNameCache.GetOrAdd(httpRoute!, $"{namePrefix} {httpRoute}");
```
vishweshbankwar, line 106: Re-checking - Can two http methods have the same route values. if that happens, we will incorrectly set the display name here.

e.g. 
Put /api/document/{id}
Get api/document/{id}

src/Shared/RequestDataHelper.cs
```diff
@@ -84,7 +87,27 @@ public void SetActivityDisplayName(Activity activity, string originalHttpMethod,
         var normalizedHttpMethod = this.GetNormalizedHttpMethod(originalHttpMethod);
         var namePrefix = normalizedHttpMethod == "_OTHER" ? "HTTP" : normalizedHttpMethod;
 
-        activity.DisplayName = string.IsNullOrEmpty(httpRoute) ? namePrefix : $"{namePrefix} {httpRoute}";
+        activity.DisplayName = GetDisplayName();
```
xiang17, line 90: How much performance would it gain? The new method seems to have enough if checks to counter the benefit. Plus the additional memory usage.

joegoldman2:
> Re-checking - Can two http methods have the same route values?. if that happens, we will incorrectly set the display name here.

Right, so the cache key should take into account the http method.

> How much performance would it gain? The new method seems to have enough if checks to counter the benefit. Plus the additional memory usage.


I did a benchmark and here is the result:

| Method             | HttpRoute       | HttpMethod | Mean       | Error     | StdDev    | Gen0   | Allocated |
|------------------- |---------------- |----------- |-----------:|----------:|----------:|-------:|----------:|
| Original           |                 | _OTHER     |  0.9941 ns | 0.1721 ns | 0.0094 ns |      - |         - |
| New_CacheKeyString |                 | _OTHER     |  1.2971 ns | 0.8686 ns | 0.0476 ns |      - |         - |
| New_CacheKeyTuple  |                 | _OTHER     |  1.3056 ns | 0.8290 ns | 0.0454 ns |      - |         - |
| Original           |                 | GET        |  0.9614 ns | 0.5957 ns | 0.0327 ns |      - |         - |
| New_CacheKeyString |                 | GET        |  0.8245 ns | 1.0305 ns | 0.0565 ns |      - |         - |
| New_CacheKeyTuple  |                 | GET        |  0.9863 ns | 0.0562 ns | 0.0031 ns |      - |         - |
| Original           | /resources/{id} | _OTHER     |  8.8834 ns | 4.1025 ns | 0.2249 ns | 0.0068 |      64 B |
| New_CacheKeyString | /resources/{id} | _OTHER     | 26.0081 ns | 3.1322 ns | 0.1717 ns | 0.0076 |      72 B |
| New_CacheKeyTuple  | /resources/{id} | _OTHER     | 19.5532 ns | 7.1305 ns | 0.3908 ns |      - |         - |
| Original           | /resources/{id} | GET        |  9.0777 ns | 4.2050 ns | 0.2305 ns | 0.0068 |      64 B |
| New_CacheKeyString | /resources/{id} | GET        | 25.6794 ns | 3.3270 ns | 0.1824 ns | 0.0068 |      64 B |
| New_CacheKeyTuple  | /resources/{id} | GET        | 17.8230 ns | 4.8392 ns | 0.2653 ns |      - |         - |

<details>
<summary>Benchmark code</summary>

```cs
[MemoryDiagnoser]
[ShortRunJob]
public class Tests
{
    private static readonly ConcurrentDictionary<string, string> DisplayNameCache = new ConcurrentDictionary<string, string>();
    private static readonly ConcurrentDictionary<(string, string), string> DisplayNameCache2 = new ConcurrentDictionary<(string, string), string>();
    private const int DisplayNameCacheSize = 1000;

    [Params("", "/resources/{id}")]
    public string? HttpRoute { get; set; }

    [Params("GET", "_OTHER")]
    public string HttpMethod { get; set; }

    [Benchmark]
    public string Original()
    {
        var namePrefix = HttpMethod == "_OTHER" ? "HTTP" : HttpMethod;
        return string.IsNullOrEmpty(HttpRoute) ? namePrefix : $"{namePrefix} {HttpRoute}";
    }

    [Benchmark]
    public string New_CacheKeyString()
    {
        var namePrefix = HttpMethod == "_OTHER" ? "HTTP" : HttpMethod;
        return GetDisplayName();

        string GetDisplayName()
        {
            if (string.IsNullOrEmpty(HttpRoute))
            {
                return namePrefix;
            }

            if (DisplayNameCache.TryGetValue($"{HttpMethod} {HttpRoute!}", out var displayName))
            {
                return displayName;
            }

            if (DisplayNameCache.Count < DisplayNameCacheSize)
            {
                return DisplayNameCache.GetOrAdd($"{HttpMethod} {HttpRoute!}", $"{namePrefix} {HttpRoute}");
            }

            return $"{namePrefix} {HttpRoute}";
        }
    }

    [Benchmark]
    public string New_CacheKeyTuple()
    {
        var namePrefix = HttpMethod == "_OTHER" ? "HTTP" : HttpMethod;
        return GetDisplayName();

        string GetDisplayName()
        {
            if (string.IsNullOrEmpty(HttpRoute))
            {
                return namePrefix;
            }

            if (DisplayNameCache2.TryGetValue((HttpMethod, HttpRoute!), out var displayName))
            {
                return displayName;
            }

            if (DisplayNameCache2.Count < DisplayNameCacheSize)
            {
                return DisplayNameCache2.GetOrAdd((HttpMethod, HttpRoute!), $"{namePrefix} {HttpRoute}");
            }

            return $"{namePrefix} {HttpRoute}";
        }
    }
}
```

</details>

The original version is better in all cases. Either my approach isn't right, or this optimization isn't worth it.

joegoldman2:
I'm not sure what to do with this PR considering the benchmark result. Any suggestion?

src/Shared/RequestDataHelper.cs
```diff
@@ -78,7 +81,27 @@ public void SetActivityDisplayName(Activity activity, string originalHttpMethod,
...
+            {
+                return namePrefix;
+            }
+
+            if (DisplayNameCache.TryGetValue(httpRoute!, out var displayName))
+            {
+                return displayName;
+            }
+
+            if (DisplayNameCache.Count < DisplayNameCacheSize)
```
AndrewP-GH, line 98: `.Count` gets a lock on all buckets, so this won't be efficient on the hot path

joegoldman2: Yes that's what the benchmark shows above.

joegoldman2: Yes, that's what the benchmark above shows.

vishweshbankwar:
@joegoldman2 - Sorry for the delay on this. I will take a look at benchmark and get back.

joegoldman2:
Not stale.

PR #1856, opened 2024-06-02
[Instrumentation.AspNet, Instrumentation.AspNetCore] EnrichWithException - improve example

### LINKED ISSUE #1828: [Instrumentation.AspNet, Instrumentaion.AspNetCore] EnrichWithException - improve example (tag)
> would be nice to change this tag to something else, as exception.type would be already covered by the instrumentation library, following the otel semantic convention!
Not a blocker, as we need to do it (if we chose to), for other instrumentation too.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1824#discussion_r1608534788_

joegoldman2:
Fixes #1828.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet/README.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet/README.md
@@ -189,12 +189,9 @@ this.tracerProvider = Sdk.CreateTracerProviderBuilder()
         };
         o.EnrichWithException = (activity, exception) =>
         {
-            // If there is an inner exception, capture its details as well
-            if (exception.InnerException != null)
+            if (exception.Source != null)
             {
-                activity.SetTag("exception.inner.type", exception.InnerException.GetType().FullName);
-                activity.SetTag("exception.inner.message", exception.InnerException.Message);
-                activity.SetTag("exception.inner.stacktrace", exception.InnerException.StackTrace);
+                activity.SetTag("exception.source", exception.Source);
             }
         };
     })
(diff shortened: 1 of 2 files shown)


Kielek (approved):

src/OpenTelemetry.Instrumentation.AspNet/README.md
```diff
@@ -189,7 +189,13 @@ this.tracerProvider = Sdk.CreateTracerProviderBuilder()
         };
         o.EnrichWithException = (activity, exception) =>
         {
-            activity.SetTag("exceptionType", exception.GetType().ToString());
+            // If there is an inner exception, capture its details as well
+            if (exception.InnerException != null)
+            {
+                activity.SetTag("exception.inner.type", exception.InnerException.GetType().FullName);
+                activity.SetTag("exception.inner.message", exception.InnerException.Message);
+                activity.SetTag("exception.inner.stacktrace", exception.InnerException.StackTrace);
```
cijothomas, line 197: not sure if this is potentially confusing.. given `stacktrace` is filled with `ToString()` by the otel core api (and in .net runtime too due to recent addition)..
Also, the inner could have further inner?

joegoldman2: I guess one level of depth is sufficient for the example. Will try to find another example if this one is confusing for you.

cijothomas: It is not confusing to me, really. What I meant to share is that - quite often, people just copy the code as-is, ignoring the fact that this is merely an example of what can be done using the Enrich mechanism!

Maybe use https://learn.microsoft.com/en-us/dotnet/api/system.exception.source?view=net-8.0#system-exception-source instead?

joegoldman2: I'm sorry, I didn't mean that it was confusing for you, but that you had the feeling that it could be confusing, in general. Thank you for the suggestion, I used it.

cijothomas (approved):

PR #1936, opened 2024-07-01
[Instrumentation.GrpcCore] Nullable support

IliaBrahinets:
Related to #894 

## Changes

- Nullable support for both `Instrumentation.GrpcCore` and `Instrumentation.GrpcCore.Tests`
- Minimal supported version update of `Grpc.Core.Api` to `2.43.0` as it is the minimal version with nullable support
- Update of `Grpc.Tools` to `2.38.0` for `Instrumentation.GrpcCore.Tests`


src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>.NET gRPC Core based client and server interceptors for OpenTelemetry.</Description>
     <PackageTags>$(PackageTags);gRPC Core;interceptors</PackageTags>
     <MinVerTagPrefix>Instrumentation.GrpcCore-</MinVerTagPrefix>
-    <Nullable>disable</Nullable>
+    <Nullable>enable</Nullable>
```
Kielek, line 7:
```suggestion
```

You can safely remove this entry. It is inherited from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/4c69afa2fbb3b9d84b45bd73dcfad43331516e69/build/Common.props#L14

test/OpenTelemetry.Instrumentation.GrpcCore.Tests/OpenTelemetry.Instrumentation.GrpcCore.Tests.csproj
```diff
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
-    <Nullable>disable</Nullable>
+    <Nullable>enable</Nullable>
```
Kielek, line 5:
```suggestion
```

src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
```diff
@@ -15,7 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="[3.15.0,4.0)" />
-    <PackageReference Include="Grpc.Core.Api" Version="[2.23.0,3.0)" />
+    <PackageReference Include="Grpc.Core.Api" Version="[2.43.0,3.0)" />
```
Kielek, line 18: I would consider even 2.46.6.
It is the latest (security fixed) version of `Grpc.Core` package. Nobody should use any older version. Can be done separately. Ref: https://groups.google.com/g/grpc-io/c/OTj5mb1qzb0 It can stay as you proposed.

test/OpenTelemetry.Instrumentation.GrpcCore.Tests/GrpcCoreServerInterceptorTests.cs
```diff
@@ -161,6 +161,6 @@ private static async Task TestHandlerFailure(Func<Foobar.FoobarClient, Metadata,
 
         var activity = activityListener.Activity;
         GrpcCoreClientInterceptorTests.ValidateCommonActivityTags(activity, StatusCode.ResourceExhausted, interceptorOptions.RecordMessageEvents, interceptorOptions.RecordException);
-        Assert.Equal(FoobarService.DefaultParentFromTraceparentHeader.SpanId, activity.ParentSpanId);
+        Assert.Equal(FoobarService.DefaultParentFromTraceparentHeader.SpanId, activity!.ParentSpanId);
```
Kielek, line 164: Consider, it applies in other places where you put `!` is tests.

```suggestion
        Assert.NotNull(activity);
        Assert.Equal(FoobarService.DefaultParentFromTraceparentHeader.SpanId, activity.ParentSpanId);
```

src/OpenTelemetry.Instrumentation.GrpcCore/ServerTracingInterceptor.cs
```diff
@@ -207,7 +207,10 @@ public ServerRpcScope(ServerCallContext context, ServerTracingInterceptorOptions
                 currentContext ?? default,
                 tags: options.AdditionalTags);
 
-            this.SetActivity(activity);
+            if (activity != null)
```
Kielek, line 210: Is it be better to change SetActivity contract to accpet null?
The check is inside already.

src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
IliaBrahinets: Makes sense, will create another PR for this

test/OpenTelemetry.Instrumentation.GrpcCore.Tests/GrpcCoreServerInterceptorTests.cs
IliaBrahinets: The idea is that `ValidateCommonActivityTags` ensures that `activity` is not null. Added `[NotNull]` attribute for the activity argument to the method. So, `!` is no longer needed

src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
```diff
@@ -1,10 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;netstandard2.1;netstandard2.0</TargetFrameworks>
```
Kielek, line 3: What is the reason for adding new TFMs?

What is more, tests are executed only against net6.0. NET Standard 2.1/2.0 will be not covered.

IliaBrahinets: To use nullable annotated BCL and use nullable attribute `[NotNull]` in unit-tests (currently there is the conflict with the one defined in `Guard.cs`), but i agree it is not the best idea due to compatibility with not annotated netstandard2.0 BCL and the point you mentioned. Reverted this change and replaced `[NotNull]` attribute with `Assert.NotNull` checks.

IliaBrahinets:
@Kielek , I've made changes according to the comments. Ready for another review.

Kielek (approved):

PR #1943, opened 2024-07-09
[Resources.OperatingSystem] initial implementation

### LINKED ISSUE #1844: Provide an operating system resource package
### Component

None

I want information about my os to be visible in my open Telemetry data

The following properties are set unless a user opts out

- os.build_id
- os.description
- os.name
- os.type

Manually adding them

### Additional context

https://opentelemetry.io/docs/specs/semconv/resource/os/

ysolomchenko:
Fixes #1844

## Changes

[Resources.OperatingSystem] initial implementation


src/OpenTelemetry.Resources.OperatingSystem/OperatingSystemSemanticConventions.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.Resources.OperatingSystem;
+
+internal static class OperatingSystemSemanticConventions
+{
+    public const string AttributeOperatingSystemType = "os.type";
+
+    public static readonly string[] OperatingSystems =
```
cijothomas, line 10: this does not look like is needed. If needed for tests, its best to declare it in tests project.

Kielek: I would say that you should define here separate values of the resource values.

You can see example here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/49a01d3a8b72cbdba499d3bac302e055209eb89a/src/OpenTelemetry.SemanticConventions/Attributes/CloudAttributes.cs#L73-L81

test/OpenTelemetry.Resources.OperatingSystem.Tests/OperatingSystemDetectorTests.cs
```diff
@@ -0,0 +1,23 @@
...
+using Xunit;
+
+namespace OpenTelemetry.Resources.OperatingSystem.Test;
+
+public class OperatingSystemDetectorTests
+{
+    [Fact]
+    public void TestOperatingSystemAttributes()
+    {
+        var resource = ResourceBuilder.CreateEmpty().AddDetector(new OperatingSystemDetector()).Build();
```
cijothomas, line 13: I'd suggest to use the extension method instead of using the ctor, so as to mimic the user code more closely, and also this can help remove the InternalsVisible trick as well. There is no need for Tests to know any internals.

cijothomas: Won't be a bad idea at all to copy the semantic convention OR add ref to it from test project.

Kielek: @cijothomas, in general I agree, but current approach is in sync with all other projects.
If we want to change lets do it in separate PR/issue.

cijothomas: yes agree, no need to block this PR at all.

Kielek (approved):
@open-telemetry/dotnet-contrib-maintainers, if you are fine, I wuld like to merge this and make `v0.1.0-alpha.1` release tomorrow.

src/OpenTelemetry.Resources.OperatingSystem/README.md
```diff
@@ -0,0 +1,44 @@
...
+Operating System Resource Detectors.
+
+```shell
+dotnet add package OpenTelemetry.Resources.OperatingSystem --prerelease
+```
+
+## Usage
+
+You can configure Operating System resource detector to
+the `TracerProvider` with the following example below.
```
cijothomas, line 24: nit: Resource can be used with all signals, not just TracerProvider.

cijothomas (approved):

PR #1955, opened 2024-07-15
[Instrumentation.Http] Set http cancellation error description and type

### LINKED ISSUE #1812: HttpClient not setting error message when using CancellationToken

jdt:
When HTTPClient Task Cancellation occurs, set the status message and assign error.type

Fixes #1812.

## Changes

* Set an error description
* Set error.type

* [X] Changes in public API reviewed (if applicable)


Kielek (approved):

Kielek:
@jdt, could you please resolve compilation issues?

jdt:
Done! Looks ready to merge now

vishweshbankwar (approved):
LGTM. Just an additional note: This does not fix the metrics part for `.NET6.0`/`.NET7.0`, I think that is ok as `.NET7.0` has already reached eol and `.NET6.0` will be in Nov. The `error.type` is assigned correctly in `.NET8.0` for metrics for this case by the runtime itself. https://learn.microsoft.com/en-us/dotnet/core/diagnostics/built-in-metrics-system-net#metric-httpclientrequestduration

PR #1962, opened 2024-07-16
Update instrumentation for AWS services: S3, Kinesis & SQS

zhenweig:
## Changes
* Add AWS resource name attributes for S3 bucket and Kinesis stream
  * S3 bucket: use attribute name ```aws.s3.bucket```,  following [Semantic Conventions for AWS S3](https://opentelemetry.io/docs/specs/semconv/object-stores/s3/)
  * Kinesis stream: use attribute name ```aws.kinesis.stream_name```. There is no official Semantic Conventions for Kinesis, thus pattern of ```aws.$SERVICE.$RESOURCE_NAME``` is applied
* For SQS, add queue name attribute when queue URL is not present in the request
   * [SQS APIs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html) fact: For all SQS APIs on a specific queue, either ```QueueURL``` or ```QueueName``` param is in the request
   * Change also enables other AWS services to populate multiple attributes from request

* [X] Changes in public API reviewed (if applicable)


AsakerMohd (approved):

srprash:
Can we change the attributes to the following aligning with [what we have for ADOT Python](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_attribute_keys.py)?
- `aws.sqs.queue.name`
- `aws.sqs.queue.url`
- `aws.kinesis.stream.name`

Though the current sem conv so not align with the above names, I think we plan to propose them to be updated.

srprash (approved):

srprash:
> Can we change the attributes to the following aligning with [what we have for ADOT Python](https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_attribute_keys.py)?
> 
> * `aws.sqs.queue.name`
> * `aws.sqs.queue.url`
> * `aws.kinesis.stream.name`
> 
> Though the current sem conv so not align with the above names, I think we plan to propose them to be updated.

Sync internally and discussed this is not needed at the moment. Can ignore this comment.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSSemanticConventions.cs
```diff
@@ -12,6 +12,9 @@ internal static class AWSSemanticConventions
 
     public const string AttributeAWSDynamoTableName = "aws.table_name";
     public const string AttributeAWSSQSQueueUrl = "aws.queue_url";
+    public const string AttributeAWSSQSQueueName = "aws.sqs.queue_name";
```
Kielek, line 15: Do you really need custom attribute here? If it is related to the messaging, you could probably use some of the generic values https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/messaging/messaging-spans.md#messaging-attributes

Maybe even it is worth to implement more of them?

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSSemanticConventions.cs
```diff
@@ -12,6 +12,9 @@ internal static class AWSSemanticConventions
...
+    public const string AttributeAWSS3BucketName = "aws.s3.bucket";
+    public const string AttributeAWSKinesisStreamName = "aws.kinesis.stream_name";
```
Kielek, line 17: If there is no more generic concept, please define it fist on semantic convention level.
Without this, it can be easily out of sync with other languages.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSSemanticConventions.cs
zhenweig: Yes, this attribute name is specific to AWS SQS queues and not applicable to other messaging services.
The name:

* Aligns with the AWS SDK instrumentation in other languages:
For example, the [Python agent](https://github.com/aws-observability/aws-otel-dotnet-instrumentation/blob/74902e85d03c0a8b6e66d27a4646e623acb03e0c/src/AWS.Distro.OpenTelemetry.AutoInstrumentation/AwsAttributeKeys.cs#L36) uses the same attribute name
* Follows the naming convention of ```aws.$SERVICE.$RESOURCE_NAME```
* AWS SDK team is currently working on defining all AWS resource attributes at the semantic convention level

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSSemanticConventions.cs
zhenweig: Same as above, these attribute names:
* Aligns with the AWS SDK instrumentation in other languages
* Follows the naming convention of aws.$SERVICE.$RESOURCE_NAME
* Aligns with OTel semantic conventions, if already available:
  -  For example,[ S3 bucket semantics](https://opentelemetry.io/docs/specs/semconv/object-stores/s3/)
* AWS SDK team is currently working on defining all AWS resource attributes at the semantic convention level

zhenweig:
Synced with repo maintainer Piotr offline, and also spoke with AWS CloudWatch team. 

CloudWatch team tech lead explained that defining AWS resource attributes at the instrumentation agent level is currently necessary because the [OTel semantic convention package](https://github.com/open-telemetry/semantic-conventions) is still in its incubating phase. Actually recently, CloudWatch team had to [remove the dependency on this package for the Java Instrumentation Agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11324) to prevent breaking changes.

In the long term, the CloudWatch team plans to collaborate with the AWS SDK team to incorporate AWS semantic conventions upstream and then migrate the instrumentation agents' dependency to it.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSSemanticConventions.cs
Kielek: Could you please put information in comment which attributes are not part of sem conv?

PR #1974, opened 2024-07-24
[Instrumentation.AWS] Update the instrumentation logic to use AWS TracerProvider.

muhammad-othman:
## Changes

This PR focuses on leveraging the newly introduced tracing APIs within the AWS SDK. The main changes include:

* Integration of New Tracing APIs:
Implemented the new AWS SDK tracing APIs and updated the `AddAWSInstrumentation` method to register this implementation using `AWSTracerProvider`.

* Simplification of Tracing Logic:
Removed `AWSTracingPipelineHandler` and migrated its logic to either the .NET SDK or `AWSPropagatorPipelineHandler`.

* SDK Version Update:
Updated the SDK version to `3.7.400` to incorporate the new tracing APIs.


Kielek:
@muhammad-othman, please sign EasyCLA. Without this we are not allowed to proceed with the PR.

ppittle (approved):

normj:
With the previous implementation we had the following AWS attributes added on the span 

image

And with the new implementation we are only getting queue url and request id.

image

Can we get operation, region and service added on the spans in the new attribute?

muhammad-othman:
> With the previous implementation we had the following AWS attributes added on the span
> 
> image
> 
> And with the new implementation we are only getting queue url and request id.
> 
> image
> 
> Can we get operation, region and service added on the spans in the new attribute?

@normj 
Yep, the following attributes was removed since they aren't part of OTel specs

image


I believe they were added them preemptively before they were added to the specs, however now we have  [rpc.method](https://opentelemetry.io/docs/specs/semconv/attributes-registry/rpc/) and [rpc.service](https://opentelemetry.io/docs/specs/semconv/attributes-registry/rpc/) to replace `aws.service` and `aws.operation` attributes 
https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/#common-attributes

The only attribute that I removed is `aws.region` which doesn't have any replacement in the specs yet, not sure if we should keep it knowing that it may be added in a different name to specs, and if we will keep it, shouldn't it get added to the other SDKs too?

normj:
> I believe they were added them preemptively before they were added to the specs, however now we have [rpc.method](https://opentelemetry.io/docs/specs/semconv/attributes-registry/rpc/) and [rpc.service](https://opentelemetry.io/docs/specs/semconv/attributes-registry/rpc/) to replace aws.service and aws.operation attributes
https://opentelemetry.io/docs/specs/semconv/cloud-providers/aws-sdk/#common-attributes

I missed that the service and operation were moved to OTel specific attributes. Region probably doesn't make sense at the OTel spec level since that is more of an AWS attribute. We should discuss with the rest of the SDK if we should add an `aws.region` attribute to our standardization. That isn't really tied to this PR so we don't need to hold this PR back for that.

test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
```diff
@@ -207,9 +207,6 @@ private void ValidateAWSActivity(Activity aws_activity, Activity parent)
     private void ValidateDynamoActivityTags(Activity ddb_activity)
     {
         Assert.Equal("DynamoDB.Scan", ddb_activity.DisplayName);
-        Assert.Equal("DynamoDB", Utils.GetTagValue(ddb_activity, "aws.service"));
```
normj, line 210: Instead of removing these checks for `aws.service` and `aws.operation` shouldn't they updated to check `rpc.service` and `rpc.method`?

test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
```diff
@@ -220,9 +217,6 @@ private void ValidateDynamoActivityTags(Activity ddb_activity)
     private void ValidateSqsActivityTags(Activity sqs_activity)
     {
         Assert.Equal("SQS.SendMessage", sqs_activity.DisplayName);
-        Assert.Equal("SQS", Utils.GetTagValue(sqs_activity, "aws.service"));
```
normj, line 223: Instead of removing these checks for `aws.service` and `aws.operation` shouldn't they updated to check `rpc.service` and `rpc.method`?

normj (changes requested):

test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
muhammad-othman: Yep, they do exist few lines below
image

test/OpenTelemetry.Instrumentation.AWS.Tests/TestAWSClientInstrumentation.cs
muhammad-othman: Yep, they do exist few lines below
image

normj (approved):

cijothomas:
@srprash Could you review?

src/OpenTelemetry.Instrumentation.AWS/Implementation/Tracing/AWSTraceSpan.cs
```diff
@@ -0,0 +1,91 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics;
+using Amazon.Runtime.Telemetry;
+using Amazon.Runtime.Telemetry.Tracing;
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.Instrumentation.AWS.Implementation.Tracing;
+
+internal class AWSTraceSpan : TraceSpan
```
martincostello, line 11: Make `AWSTraceSpan`, `AWSTracer` and `AWSTracerProvider` `sealed` to aid with JIT optimisations?

birojnayak:
`aws.region` is missing as pointed out earlier, while others have replacement.  See if you can have, because existing customers/users might be relying on it to build custom dashboard or extensions.

muhammad-othman:
@birojnayak Unfortunately  `aws.region` attribute doesn't currently exist in the other AWS SDKs and there is no replacement for it in the current specifications. My concern is that if we add this attribute now, we may later reach an agreement with other SDKs on a different name, similar to how `aws.service` and `aws.operation` were replaced by `rpc.service and `rpc.method`. 

Since that we're still in beta, I believe it's better to remove `aws.region` for now and we can add it later with a name that has been agreed upon across all SDKs. 
Also this change will not be part of this PR since this package isn't producing the shared attributes anymore, it will be added to the Core SDK similar to https://github.com/aws/aws-sdk-net/blob/2cd501b0108ea1c0136094c04c3e1785ff0f0aac/sdk/src/Core/Amazon.Runtime/Telemetry/Tracing/TracingUtilities.cs#L52

srprash (approved):

muhammad-othman:
Hi @cijothomas,
I've addressed the comments on this PR and the two component owners approved it, please let me know if there are any additional steps needed from my side to merge it.

cijothomas:
There is a CI failure: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/10498004987/job/29082060927?pr=1974 Could you check and fix? I can merge once CI is passing.

muhammad-othman:
Thank you @cijothomas for the quick response.
Looks like exported items were mutated when other tests run at the same time, which caused `Collection was modified` exception.
I moved the exportItems asserts outside the using clause, this way we are sure the tracer provider was disposed before enumerating over the exportedItems and they shouldn't be modified at this stage.

cijothomas:
@vishweshbankwar Need your help. One of the merge requirement is not met, maybe a CI is missing? I am not able to tell quickly..

muhammad-othman:
@cijothomas I think it needs code owner approver first, someone from ` open-telemetry/dotnet-contrib-approvers`. I believe [the previous PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1980) was approved by you then `Code owner review required` disappeared.

cijothomas (approved):

PR #1980, opened 2024-07-26
[Instrumentation.AWS] Add AWS metrics instrumentation

muhammad-othman:
## Changes

* Integration of New Metrics APIs:
Implemented the new AWS SDK tracing APIs, starting with `AWSMeterProvider`.

* Added Instrumentation Method:
Introduced `MeterProviderBuilderExtensions.AddAWSInstrumentation` to enable metrics instrumentation.

* SDK Version Update:
Updated the SDK version to `3.7.400` to incorporate the new tracing APIs.
This update conflicts with the changes in PR [#1974](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1974). The last merged PR will rebase accordingly.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
+++ b/src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
@@ -20,6 +20,7 @@ internal class AWSHistogram<T> : Histogram<T>
     {
         if (attributes != null)
         {
+            // TODO: remove ToArray call and use when AttributesAsSpan expected to be added at AWS SDK v4.
             this.histogram.Record(value, attributes.AllAttributes.ToArray());
         }
         else
--- a/src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
+++ b/src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
@@ -1,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using Amazon.Runtime.Telemetry;
+using Amazon.Runtime.Telemetry.Metrics;
+
+namespace OpenTelemetry.Instrumentation.AWS.Implementation.Metrics;
+
+internal class AWSHistogram<T> : Histogram<T>
+    where T : struct
+{
+    private readonly System.Diagnostics.Metrics.Histogram<T> histogram;
+
+    public AWSHistogram(System.Diagnostics.Metrics.Histogram<T> histogram)
+    {
+        this.histogram = histogram;
+    }
+
+    public override void Record(T value, Attributes? attributes = null)
+    {
+        if (attributes != null)
+        {
+            this.histogram.Record(value, attributes.AllAttributes.ToArray());
+        }
+        else
+        {
+            this.histogram.Record(value);
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


ppittle (approved):

normj (approved):

normj:
Getting metrics added for the AWS SDK looks really cool in the Aspire Dashboard.

image

vishweshbankwar:
@muhammad-othman - Could you please take a look at failing tests?

src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using Amazon.Runtime.Telemetry;
+using Amazon.Runtime.Telemetry.Metrics;
+
+namespace OpenTelemetry.Instrumentation.AWS.Implementation.Metrics;
+
+internal class AWSHistogram<T> : Histogram<T>
```
cijothomas, line 9: what is the need of these wrappers over existing Histogram and others? Metrics API are designed to support high performance, allocation-free scenarios, and this wrappers is negating some of the perf optimizations... Wondering if this is really needed?

src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
```diff
@@ -0,0 +1,30 @@
...
+    public AWSHistogram(System.Diagnostics.Metrics.Histogram<T> histogram)
+    {
+        this.histogram = histogram;
+    }
+
+    public override void Record(T value, Attributes? attributes = null)
+    {
+        if (attributes != null)
+        {
+            this.histogram.Record(value, attributes.AllAttributes.ToArray());
```
cijothomas, line 23: The array allocation of heap is not desirable.

src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
muhammad-othman: We needed these wrappers to allow the AWS SDK to call these methods directly, and unfortunately the AWS SDK cannot take on new dependencies, and we supported .NET Standard and .NET Core 3.1, which lack `System.Diagnostics.DiagnosticSource` that is available in newer versions.

src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
muhammad-othman: unfortunately I couldn't avoid that with the current APIs, `Histogram.Record` accepts params as an array, and the API that the AWS SDK exposes uses `IEnumerable` since the attributes can be mutable in other places, maybe at some point  `Histogram.Record` may accepts params as an  `IEnumerable`  which would be more flexible.

cijothomas: `Histogram.Record` has many overloads to accepts attributes(Tags) without forcing allocation. 
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics#metrics-api

It is unfortunate that the AWS wrapper is forcing heap allocation for metrics in hot path.

cijothomas (approved):
Glad to see more metrics instrumentation.
I am concerned about the heap allocated in the hot path, see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1980#discussion_r1704419693 
but given the owners have approved, will proceed to merge.

cijothomas:
@srprash Could you review?

src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
normj: @muhammad-othman When we move the SDK's observability changes into the V4 development branch we can add an `AttributesAsSpan` property and use it here to remove the allocation.

muhammad-othman: Will create a v4 task for this, also this isn't the only place that we use `.ToArray()` so it should be useful to use it in other places including the tracing changes.

cijothomas: can you add a TODO comment here that future aws sdk will offer api's to avoid allocation.?

srprash (approved):

PR #1989, opened 2024-08-01
[Instrumentation.SqlClient] Database metrics `db.client.operation.duration` prototype

matt-hensley:
## Changes

Implements proposed `db.client.operation.duration` metric for SqlClient with required attributes.

Only implemented against .NET 6+ for now.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientMetricDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientMetricDiagnosticListener.cs
@@ -1,0 +1,71 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+using System.Reflection;
+
+namespace OpenTelemetry.Instrumentation.SqlClient.Implementation;
+
+#if NET6_0_OR_GREATER
+internal class SqlClientMetricDiagnosticListener : ListenerHandler
+{
+    internal static readonly Assembly Assembly = typeof(SqlClientMetricDiagnosticListener).Assembly;
+    internal static readonly AssemblyName AssemblyName = Assembly.GetName();
+    internal static readonly string MeterName = AssemblyName.Name!;
+    internal static readonly string MeterVersion = AssemblyName.Version!.ToString();
+    internal static readonly Meter Meter = new(MeterName, MeterVersion);
+    private static readonly Histogram<double> DbClientOperationDuration = Meter.CreateHistogram<double>(
+        "db.client.operation.duration",
+        "s",
+        "Duration of database client operations.");
+
+    private readonly AsyncLocal<long> beginTimestamp = new AsyncLocal<long>();
+
+    public SqlClientMetricDiagnosticListener(string sourceName)
+        : base(sourceName)
+    {
+    }
+
+    public override void OnEventWritten(string name, object? payload)
+    {
+        var activity = Activity.Current;
+        switch (name)
+        {
+            case SqlClientDiagnosticListener.SqlDataBeforeExecuteCommand:
+            case SqlClientDiagnosticListener.SqlMicrosoftBeforeExecuteCommand:
+                {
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientMetricDiagnosticListener.cs
```diff
@@ -0,0 +1,88 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+using System.Reflection;
+
+namespace OpenTelemetry.Instrumentation.SqlClient.Implementation;
+
+#if NET6_0_OR_GREATER
+internal class SqlClientMetricDiagnosticListener : ListenerHandler
```
gregkalapos, line 11: Since both this new `SqlClientMetricDiagnosticListener` and the existing `SqlClientDiagnosticListener` rely on the same events, can they really work well together?

E.g. if I change the order in the `SqlClientCallsAreCollectedSuccessfully` and first call `meterProviderBuilder.AddSqlClientInstrumentation();` and after that `traceProviderBuilder.AddSqlClientInstrumentation` (so first this new class is registered and then the existing `SqlClientDiagnosticListener`), then the test fail, because `SqlClientDiagnosticListener.SqlMicrosoftAfterExecuteCommand` never fires in this class. 

I believe users can register span and metric creation in any order and nothing would prevent them from doing so, right? If that's the case I think this design should be reconsidered.

I totally agree in general with the separation of the class that takes care of the activity and this one that takes care of metrics, but I think in this form, it doesn't really work.

Maybe an option would be to keep a single class handling the events, but then the activity creation and the metric creation could still be in 2 separate classes and the one handling the events can just call both.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientMetricDiagnosticListener.cs
```diff
@@ -0,0 +1,88 @@
...
+                        if (this.beginTimestamp.Value != 0)
+                        {
+                            var endTimestamp = Stopwatch.GetTimestamp();
+                            var duration = GetElapsedTime(this.beginTimestamp.Value, endTimestamp);
+                            DbClientOperationDuration.Record(duration.TotalSeconds, tags);
+                        }
+                    }
+                    else
+                    {
+                        activity.Stop();
```
gregkalapos, line 69: Related to the other comment: I think stopping this activity here is a bit brutal: this activity is not really started by this class - it's actually "owned" by `SqlClientDiagnosticListener` and it expects this to also stop. In the current form, it'll trigger this log: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L153

Now.. I guess it's stopped here, because otherwise there is no `activity.Duration`, right? I think this again could be solved by handling the events in a common class and not subscribing twice to the same event.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientMetricDiagnosticListener.cs
```diff
@@ -0,0 +1,88 @@
...
+                        DbClientOperationDuration.Record(activity.Duration.TotalSeconds, tags);
```
gregkalapos, line 70: There are some other non-required attributes on this metric: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#database-operation

If the metric is generated based on the `activity`, would it be maybe possible to look for those attributes on the `activity` and move the ones that are present there to the attributes of the metric that is generated here?

gregkalapos (commented):
Hey @matt-hensley,

nice PR, great to see this being worked on 👏 . I attended most of the DB stabilization working group meetings and @alanwest pointed me to this PR and told me you work on some of these things. So I looked at it.

I understand this is a prototype - what's the plan here? Would you want this to merge at some point?

I left some comments - basically all those are only relevant if this really gets merged and maybe you already know those things. Just hope maybe those comments help to make this into a final PR.

PR #1967, opened 2024-07-17
Adds a RateLimiting sampler based on the Jaeger sampler for Java.

samsp-msft:
https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java

Fixes #
Design discussion issue #

## Changes

Please provide a brief description of the changes here.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #1996, opened 2024-08-06
Rate Limiting sampler for .NET

samsp-msft:
Port of https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jaeger-remote-sampler/src/main/java/io/opentelemetry/sdk/extension/trace/jaeger/sampler/RateLimitingSampler.java to .NET

The rate limiting sampler is a sampler that will limit the number of traces to
the specified rate per second. It is typically used in conjunction with the ParentBasedSampler
to ensure that the rate limiting sampler is only applied to the root spans. If
a request comes in without a sampling decision, the rate limiting sampler will
make a decision based on the rate limit. The sampling decision is stored in the
activity context, via the Recorded property, and is passed along with calls to
other services called with HttpClient so the same decision is used for the entire
trace.

Example of RateLimitingSampler usage:

```cs
builder.Services.AddOpenTelemetry()
    .WithTracing(tracing =>
    {
        tracing.AddAspNetCoreInstrumentation()
            .AddHttpClientInstrumentation()
            // Add the rate limiting sampler with a limit of 3 traces per second
            .SetSampler(new ParentBasedSampler(new RateLimitingSampler(3)))
    });
```

## Changes

### DIFF
--- a/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
+++ b/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 using System.Diagnostics;
 
-namespace OpenTelemetry;
+namespace OpenTelemetry.Extensions.Internal;
 
 internal sealed class RateLimiter
 {
--- a/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
+++ b/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
@@ -7,28 +7,28 @@ namespace OpenTelemetry.Extensions.Internal;
 internal sealed class RateLimiter
 {
     private readonly Stopwatch stopwatch = Stopwatch.StartNew();
-    private readonly double creditsPerNanosecond;
+    private readonly double creditsPerTick;
     private readonly long maxBalance; // max balance in ticks
     private long currentBalance; // last op ticks less remaining balance, using long directly with Interlocked for thread safety
 
     public RateLimiter(double creditsPerSecond, double maxBalance)
     {
-        this.creditsPerNanosecond = creditsPerSecond / Stopwatch.Frequency;
-        this.maxBalance = (long)(maxBalance / this.creditsPerNanosecond);
+        this.creditsPerTick = creditsPerSecond / Stopwatch.Frequency;
+        this.maxBalance = (long)(maxBalance / this.creditsPerTick);
         this.currentBalance = this.stopwatch.ElapsedTicks - this.maxBalance;
     }
 
     public bool TrySpend(double itemCost)
     {
-        long cost = (long)(itemCost / this.creditsPerNanosecond);
-        long currentNanos;
-        long currentBalanceNanos;
+        long cost = (long)(itemCost / this.creditsPerTick);
+        long currentTicks;
+        long currentBalanceTicks;
         long availableBalanceAfterWithdrawal;
         do
         {
-            currentBalanceNanos = Interlocked.Read(ref this.currentBalance);
-            currentNanos = this.stopwatch.ElapsedTicks;
-            long currentAvailableBalance = currentNanos - currentBalanceNanos;
+            currentBalanceTicks = Interlocked.Read(ref this.currentBalance);
+            currentTicks = this.stopwatch.ElapsedTicks;
+            long currentAvailableBalance = currentTicks - currentBalanceTicks;
             if (currentAvailableBalance > this.maxBalance)
             {
                 currentAvailableBalance = this.maxBalance;
@@ -40,7 +40,9 @@ internal sealed class RateLimiter
                 return false;
             }
         }
-        while (Interlocked.CompareExchange(ref this.currentBalance, currentNanos - availableBalanceAfterWithdrawal, currentBalanceNanos) != currentBalanceNanos);
+
+        // CompareExchange will fail if currentBalance has changed since the last read, implying another thread has updated the balance
+        while (Interlocked.CompareExchange(ref this.currentBalance, currentTicks - availableBalanceAfterWithdrawal, currentBalanceTicks) != currentBalanceTicks);
         return true;
     }
 }
--- a/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
+++ b/src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
@@ -1,0 +1,46 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+using System.Diagnostics;
+
+namespace OpenTelemetry;
+
+internal sealed class RateLimiter
+{
+    private readonly Stopwatch stopwatch = Stopwatch.StartNew();
+    private readonly double creditsPerNanosecond;
+    private readonly long maxBalance; // max balance in ticks
+    private long currentBalance; // last op ticks less remaining balance, using long directly with Interlocked for thread safety
+
+    public RateLimiter(double creditsPerSecond, double maxBalance)
+    {
+        this.creditsPerNanosecond = creditsPerSecond / Stopwatch.Frequency;
+        this.maxBalance = (long)(maxBalance / this.creditsPerNanosecond);
+        this.currentBalance = this.stopwatch.ElapsedTicks - this.maxBalance;
+    }
+
+    public bool TrySpend(double itemCost)
+    {
+        long cost = (long)(itemCost / this.creditsPerNanosecond);
+        long currentNanos;
+        long currentBalanceNanos;
+        long availableBalanceAfterWithdrawal;
+        do
+        {
+            currentBalanceNanos = Interlocked.Read(ref this.currentBalance);
+            currentNanos = this.stopwatch.ElapsedTicks;
+            long currentAvailableBalance = currentNanos - currentBalanceNanos;
+            if (currentAvailableBalance > this.maxBalance)
+            {
+                currentAvailableBalance = this.maxBalance;
+            }
+
+            availableBalanceAfterWithdrawal = currentAvailableBalance - cost;
+            if (availableBalanceAfterWithdrawal < 0)
+            {
+                return false;
+            }
+        }
+        while (Interlocked.CompareExchange(ref this.currentBalance, currentNanos - availableBalanceAfterWithdrawal, currentBalanceNanos) != currentBalanceNanos);
+        return true;
+    }
+}
--- a/src/OpenTelemetry.Extensions/README.md
+++ b/src/OpenTelemetry.Extensions/README.md
@@ -79,3 +79,27 @@ var tracerProvider = Sdk.CreateTracerProviderBuilder()
 Warning: The baggage key predicate is executed for every baggage entry for each
 started activity.
 Do not use slow or intensive operations.
+
+### RateLimitingSampler
+
+The rate limiting sampler is a sampler that will limit the number of traces to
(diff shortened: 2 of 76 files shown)


samsp-msft:
redo of #1967

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// The OpenTelemetry.Trace.SamplingParameters used by the OpenTelemetry.Trace.Sampler
+    /// to decide if the System.Diagnostics.Activity to be created is going to be sampled
+    /// or not.
+    /// </param>
+    /// <returns>
+    /// Sampling decision on whether activity needs to be sampled or not.
+    /// </returns>
+    public override SamplingResult ShouldSample(in SamplingParameters samplingParameters)
+    {
+        return this.rateLimiter.TrySpend(1.0) ? this.onSamplingResult : this.offSamplingResult;
```
cijothomas, line 56: Not sure I follow the algorithm.. Why pass 1.0 always here?

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
```diff
@@ -0,0 +1,68 @@
...
+    private readonly SamplingResult onSamplingResult;
+    private readonly SamplingResult offSamplingResult;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RateLimitingSampler"/> class.
+    /// </summary>
+    /// <param name="maxTracesPerSecond">The maximum number of traces that will be emitted each second.</param>
+    public RateLimitingSampler(int maxTracesPerSecond)
+    {
+        double maxBalance = maxTracesPerSecond < 1.0 ? 1.0 : maxTracesPerSecond;
```
cijothomas, line 31: maxTracesPerSecond < 1.0 -- when would this happen ever?

MikeGoldsmith: As the input arg is an int, it's possible someone could pass in a value < 1. It's not gonna work though, so defaulting to 1 makes sense here.

In this scenario, I think we should log a warning to indicate the passed in value has been overriden.

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
```diff
@@ -0,0 +1,68 @@
...
+        this.rateLimiter = new RateLimiter(maxTracesPerSecond, maxBalance);
+        var attributes = new Dictionary<string, object>()
+        {
+            { SAMPLERTYPEKEY, SAMPLERTYPE },
+            { SAMPLERPARAMKEY, (double)maxTracesPerSecond },
```
MikeGoldsmith, line 36: Is there a reason why you're converting from int to double here?

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
```diff
@@ -0,0 +1,68 @@
...
+    private readonly SamplingResult offSamplingResult;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RateLimitingSampler"/> class.
+    /// </summary>
+    /// <param name="maxTracesPerSecond">The maximum number of traces that will be emitted each second.</param>
+    public RateLimitingSampler(int maxTracesPerSecond)
+    {
+        double maxBalance = maxTracesPerSecond < 1.0 ? 1.0 : maxTracesPerSecond;
+        this.rateLimiter = new RateLimiter(maxTracesPerSecond, maxBalance);
```
MikeGoldsmith, line 32: I don't see how this works, given a maxTracesPerSecond of >= 1, both maxTracesPerSeconf and maxBalance will always be the same. Is that intended?

MikeGoldsmith (commented):
I can see this is a direct copy of what's in Java and the rudimentary tests show it to work (on windows), but I don't understand the logic to know how this is meant to work.

Also, we'll need to figure out why the tests are failing on ubuntu.

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
samsp-msft: The underlying algorithm in the RateLimiter has an item cost for the relative cost of each item. We treat traces as all having the same cost, hence 1.0 (as the rate limiter takes doubles).

The algorithm is to essentially convert the cost into ticks (cost/max_samples_sec) and then see if there is a balance available to spend. If so the value is deducted and the decision is to keep that trace, if not it rejects it and the balance continues to grow. 
The max_balance is to account for bursty traffic. If you have 3s of no traffic, the balance would build up, it gets capped at the allowance for 1s worth of items.

samsp-msft:
> I can see this is a direct copy of what's in Java and the rudimentary tests show it to work (on windows), but I don't understand the logic to know how this is meant to work.
> 
> Also, we'll need to figure out why the tests are failing on ubuntu.

I did some more experimentation, including running in WSL (Ubuntu). I had not accounted for the initial balance in the math in the test - with that, the first N asks will be sampled in as they chew through that balance. With this accounted for the expected value and actual are exact in my personal experiments, but I want to keep a small fudge factor as tests that rely on timing don't always work out the way you expect them to.

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
samsp-msft: If the maxBalance is less than 1, and we use a weight of 1, there will never be enough of a balance to pass anything. Using the sample rate as the maxBalance provides a good default - the first N items will always be passed, after that they will be sampled out until enough balance is available. The maxBalance provides a cap for the case that there are pauses in the data - if there is no data for multiple seconds, it doesn't enable a large balance to build up.

src/OpenTelemetry.Extensions/Trace/RateLimitingSampler.cs
samsp-msft: It's what the original Java does :-)

The sampling algorithm supports a sample rate of a double, but the wrapper is exposing only an integer rate. It could change to a double, and I want this to be consistent across languages.

samsp-msft:
@cijothomas the ubuntu run should pass tests now.

MikeGoldsmith (approved):
Looks good to me. Thanks for fixing linux tests 👍🏻

samsp-msft:
@cijothomas - poke 😸

src/OpenTelemetry.Extensions/README.md
```diff
@@ -84,3 +84,27 @@ var tracerProvider = Sdk.CreateTracerProviderBuilder()
...
 started activity.
 Do not use slow or intensive operations.
+
+### RateLimitingSampler
+
+The rate limiting sampler is a sampler that will limit the number of traces to
+the specified rate per second. It is typically used in conjunction with the ParentBasedSampler
+to ensure that the rate limiting sampler is only applied to the root spans. If
+a request comes in without a sampling decision, the rate limiting sampler will
+make a decision based on the rate limit. The sampling decision is stored in the
```
cijothomas, line 94:
```suggestion
make a decision based on the rate limit.
```

The last sentence is not really required here. That is something OTel SDK does, irrespective of the Sampler

src/OpenTelemetry.Extensions/README.md
```diff
@@ -84,3 +84,27 @@ var tracerProvider = Sdk.CreateTracerProviderBuilder()
 Warning: The baggage key predicate is executed for every baggage entry for each
 started activity.
 Do not use slow or intensive operations.
+
+### RateLimitingSampler
+
+The rate limiting sampler is a sampler that will limit the number of traces to
+the specified rate per second. It is typically used in conjunction with the ParentBasedSampler
+to ensure that the rate limiting sampler is only applied to the root spans. If
+a request comes in without a sampling decision, the rate limiting sampler will
```
cijothomas, line 93:
```suggestion
an `Activity` creation request comes in without a sampling decision, the rate limiting sampler will
```

cijothomas: this statements feels incorrect. RateLimitingSampler does not check whether there is a parent decision or not. that is upto end user to ensure by wrapping this inside ParentBased.

src/OpenTelemetry.Extensions/README.md
```diff
@@ -84,3 +84,27 @@ var tracerProvider = Sdk.CreateTracerProviderBuilder()
 Warning: The baggage key predicate is executed for every baggage entry for each
 started activity.
 Do not use slow or intensive operations.
+
+### RateLimitingSampler
+
+The rate limiting sampler is a sampler that will limit the number of traces to
```
cijothomas, line 90: is it counting the number of traces or number of spans (Activity)?

samsp-msft: It works on activities, but when used in conjunction with ParentBasedSampler, it becomes the number of traces. As its only called for the root spans, it does not cost or account for the number of spans in each trace.

src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+internal sealed class RateLimiter
+{
+    private readonly Stopwatch stopwatch = Stopwatch.StartNew();
+    private readonly double creditsPerNanosecond;
+    private readonly long maxBalance; // max balance in ticks
+    private long currentBalance; // last op ticks less remaining balance, using long directly with Interlocked for thread safety
+
+    public RateLimiter(double creditsPerSecond, double maxBalance)
+    {
+        this.creditsPerNanosecond = creditsPerSecond / Stopwatch.Frequency;
```
vishweshbankwar, line 16: Should this be simply dividing it by 10^9?. Looking at java implementation --> https://github.com/open-telemetry/opentelemetry-java/blob/ad120a5bff0887dffedb9c73af8e8e0aeb63659a/sdk/common/src/main/java/io/opentelemetry/sdk/internal/RateLimiter.java#L36

samsp-msft: The .NET stopwatch class does not have a fixed frequency for its ticks - it depends on the hardware/OS that it is running on, and provides the `Frequency` property to indicate how many ticks there are a second.
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.elapsedticks?view=net-8.0#remarks

vishweshbankwar: got it- it's the name `creditsPerNanosecond` that confused me. Cannot think of a better name atm.

src/OpenTelemetry.Extensions/Internal/RateLimiter.cs
```diff
@@ -0,0 +1,46 @@
...
+        this.maxBalance = (long)(maxBalance / this.creditsPerNanosecond);
+        this.currentBalance = this.stopwatch.ElapsedTicks - this.maxBalance;
+    }
+
+    public bool TrySpend(double itemCost)
+    {
+        long cost = (long)(itemCost / this.creditsPerNanosecond);
+        long currentNanos;
+        long currentBalanceNanos;
```
vishweshbankwar, line 25:
```suggestion
        long currentBalanceInTicks;
```

vishweshbankwar: @samsp-msft - would this be better name? Same for other `*Nanos`.

vishweshbankwar (approved):
LGTM - left some non-blocking comments.

PR #2001, opened 2024-08-08
[Instrumentation.StackExchangeRedis] Use different source names for connections

### LINKED ISSUE #1997: StackExchangeRedis support for uniquely instrumenting multiple Redis connections
### Component

OpenTelemetry.Instrumentation.StackExchangeRedis

We have an application that uses a single Redis server for multiple purposes (e.g. caching and SignalR). If we try to call `.AddRedisInstrumentation` for each of our different Redis connections, we end up with duplicate spans in our traces.

The `ActivityName` and `ActivitySource` in `StackExchangeRedisConnectionInstrumentation` are the same for all Redis connections that are instrumented, rather than being able to support unique names/sources per Redis connection.

Being able to support "unique" instrumentation per Redis connection would prevent the duplicate spans we're currently experiencing. Further, it would allow us to specify different options (e.g. `Enrich` actions) for each connection as needed.

Ideally, each call to `.AddRedisInstrumentation` could effectively configure instrumentation that is unique per Redis connection if the consumer of the `OpenTelemetry.Instrumentation.StackExchangeRedis` library chooses to go this route.

The default behavior (all Redis connections are instrumented the same) would still be an option. This per-connection instrumentation I'm requesting would simply be optional.

I tried different instrumentation approaches using the library, but was unable to get the desired result. I either ended up with traces that seemed to be missing spans for calls related to certain Redis connections/commands, or traces with duplicate spans for a given Redis command.

### Additional context

I've messed around with the code a bit to try to get a working PoC. I do have it working, but there may be a more elegant way to provide this support.

I'm happy to submit a PR for this and have folks much more familiar with the code to have a look.

YayBurritos:
Fixes #1997 

## Changes

Use existing `name` parameter of `TracerProviderBuilderExtensions -> AddRedisInstrumentation` to create a unique `ActivitySource` for the Redis connection. 

### Instrumenting multiple Redis connections

```csharp
using OpenTelemetry.Trace;

public class Program
{
    public static void Main(string[] args)
    {
        // Connect to the first server.
        var firstConnectionName = "myRedisConnection1";
        using var connection1 = ConnectionMultiplexer.Connect("server1:6379");

        // Connect to the second server.
        var secondConnectionName = "myRedisConnection2";
        using var connection2 = ConnectionMultiplexer.Connect("server2:6379");

            .AddRedisInstrumentation(firstConnectionName, connection1, null)
            .AddRedisInstrumentation(secondConnectionName, connection2, null)
            .AddConsoleExporter()
            .Build();
    }
}
```

If `name` isn't provided, the prior (default) behavior will be respected.

### Default Behavior

```csharp
using OpenTelemetry.Trace;

public class Program
{
    public static void Main(string[] args)
    {
        // Connect to the server.
        using var connection = ConnectionMultiplexer.Connect("localhost:6379");

            .AddRedisInstrumentation(connection)
            .AddConsoleExporter()
            .Build();
    }
}
```


vishweshbankwar:
@YayBurritos - Please sign the CLA so that the PR can be reviewed.

YayBurritos:
@vishweshbankwar, @matt-hensley: FYI - CLA has been signed. Thanks for your patience!

Kielek (changes requested):
In general, I do not like the idea of this PR. The source name should be constant for the given library.
You can distinguish/decorate attributes using `Enrich` on the options:

Consider something like:

```csharp
x.Enrich = (activity, command) => activity.AddTag("your-custom-connection-attribute-name", "your-custom-connection-name")
```

matt-hensley:
Agree with @Kielek, dynamic ActivitySource naming is unusual. I believe the use case can be handled by named configurations /connections.

YayBurritos:
@Kielek , @matt-hensley : Totally understand your concerns on this PR. The issue I'm trying to address is duplicate activities/spans being created when you attempt to instrument different Redis connection multiplexers. Enriching doesn't solve this problem. 

In our testing, if we have Connection A and Connection B, and there's an HMGET call for Connection A, we see **two** spans for the HMGET (since there are two multiplexers that are instrumented) rather than one. This is the behavior I'm trying to address here.

This solution (while it may be "unusual") simply provides an _option_ to instrument the connections independently - with the main goal of avoiding duplicate span creation. Users can continue to do things the old/standard way if they prefer. This is simply a workaround.

If there's a better solution, I'm open to that so feel free to provide recommendations. But the library as-is seems to have an issue for our particular scenario.

Kielek:
@YayBurritos,
Did you have tested construction from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1193#issue-1712724215 - Usage section?

You should not call AddRedisInstrumentation twice. There is a dedicated way to add additional connection to the instrumentation.

YayBurritos:
> @YayBurritos, Did you have tested construction from [#1193 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1193#issue-1712724215) - Usage section?
> 
> You should not call AddRedisInstrumentation twice. There is a dedicated way to add additional connection to the instrumentation.
> 
> EDIT: If it does not help, could you please provide [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example). There is a chance that we will be able to advice better solution for managing the connections.

Hi @Kielek - Thanks for the reply! I did try the approach shown in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1193#issue-1712724215, but in my testing, that resulted in no Redis-related spans being produced. I was quite surprised by that result.

I can look to provide an MRE. That will help me better understand if there's something odd/unusual with our application, or if we still see the same behavior with the library.

Kielek:
@YayBurritos, any updated about MRE?

YayBurritos:
> @YayBurritos, any updated about MRE?

HI @Kielek! I don't yet have an MRE to share, but I have been working on some reorg within our code and I may (still need to do more testing) have resolved our duplicate span issue without requiring any changes to the `Instrumentation.StackExchangeRedis` package. Our issue may be the result of chaining an Exporter each time we call `AddRedisInstrumentation`.

Again, still need to test a bit to prove that theory, but that's the path I'm heading down at the moment. Will let you know when I have some results.

YayBurritos:
@Kielek : Upon further testing, it does appear that our duplicate span issue for Redis-related activity was the result of our overzealous use of exporters. I updated our application to only use one exporter (chained after the first instrumented Redis connection), but continue to call `AddRedisInstrumentation` for each of the other connections we use. Results show what appear to be the "correct" spans (spans associated with a single Redis connection, rather than duplicates). As such, I think I'm good if we close this PR as well as the related issue. Thanks for the review and feedback on this!

Kielek:
Great to hear that. Feel free to create issues or prs if you see any place for improvements. Especially PRs are.more than welcomed :)

PR #2021, opened 2024-08-22
[OneCollector] Introduce EventFullNameMappings on options

flaviocdc:
## Changes

When using OneCollector exporter, there's no way to override the how `EventNameManager` generates event names.
Currently the name generated will always have the following format: `Logger Category.Logger Name`
So say you do:
```csharp
var logger = LoggerFactory.CreateLogger("MyNamespace.MyClass");
logger.LogInfo(new EventId(1, "MyLogName"), "My log message");
```
The event name generated would be `MyNamespace.MyClass.MyLogName`.
Since Event name is what used to generate the name of the table that will be used to store data in Aria, the table name will then be `MyNamespace_MyClass_MyLogName`, which some times is not desirable.

OpenTelemetry.Exporter.Geneva supports what they call `TableNameMappings` (more details [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Exporter.Geneva#log-table-name-mappings))

I'm pretty much mimicking most of what's implemented in Geneva package, except for a few things that I'm leaving out for the sake of simplicity (for my use case, I don't need full parity).

### Event full name mappings

EventFullNameMappings allow users to control the event full name which is generated for a given log `CategoryName` and/or `EventId.Name`.

* To change the default event full name for all Logs add an entry with the key `*` and set the value to the desired custom event full name.

* To change the event full name for a specific log [category](https://docs.microsoft.com/dotnet/core/extensions/logging#log-category) add an entry with the key set to the full "category" of the log messages or a prefix that will match the starting portion of the log message "category". Set the value of the key to either the desired custom event full name

  For example, given the configuration...

  ```csharp
  logging.AddOneCollectorExporter(
    "InstrumentationKey=instrumentation-key-here",
    builder => builder.SetEventFullNameMappings(
        new Dictionary<string, string>()
        {
            { "*", "DefaultLogs" },
            { "MyCompany", "InternalLogs" },
            { "MyCompany.Product1", "InternalProduct1Logs" },
            { "MyCompany.Product2", "InternalProduct2Logs" },
            { "MyCompany.Product2.Security", "InternalSecurityLogs" },
        });
  ```

  ...log category mapping would be performed as such:

  * `ILogger<ThirdParty.Thing>`: This would go to "DefaultLogs"
  * `ILogger<MyCompany.ProductX.Thing>`: This would go to "InternalLogs"
  * `ILogger<MyCompany.Product1.Thing>`: This would go to "InternalProduct1Logs"
  * `ILogger<MyCompany.Product2.Thing>`: This would go to "InternalProduct2Logs"
  * `ILogger<MyCompany.Product2.Security.Alert>`: This would go to
    "InternalSecurityLogs"


joelcdpereira (approved):
Great additional capability to OneCollection

CodeBlanch (approved):
LGTM

PR #2026, opened 2024-08-27
SqlClient db metrics

matt-hensley:
## Changes

Opening draft PR for feedback.

Some TODOs, including flaky tests, likely due to state.


matt-hensley:
FYI @alanwest @gregkalapos

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlActivitySourceHelper.cs
```diff
@@ -22,6 +23,15 @@ internal sealed class SqlActivitySourceHelper
     public static readonly ActivitySource ActivitySource = new(ActivitySourceName, Assembly.GetPackageVersion());
     public static readonly string ActivityName = ActivitySourceName + ".Execute";
 
+    public static readonly string MeterName = AssemblyName.Name!;
+    public static readonly string MeterVersion = AssemblyName.Version!.ToString();
```
alanwest, line 27: The `ActivitySource` is using `Assembly.GetPackageVersion()`. Should probably use that to make these consistent.

This whole section would probably be better written something like

```csharp
    public static readonly Assembly Assembly = typeof(SqlActivitySourceHelper).Assembly;
    public static readonly AssemblyName AssemblyName = Assembly.GetName();
    public static readonly string InstrumentationScopeName = AssemblyName.Name!;
    public static readonly string InstrumentationScopeVersion = Assembly.GetPackageVersion();

    public static readonly ActivitySource ActivitySource = new(InstrumentationScopeName, InstrumentationScopeVersion);
    public static readonly Meter Meter = new(InstrumentationScopeName, InstrumentationScopeVersion);

```

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -148,24 +157,40 @@ public override void OnEventWritten(string name, object? payload)
...
+                    //{
+                    //    return;
+                    //}
+
+                    activity.Stop();
+
+                    if (SqlActivitySourceHelper.DbClientOperationDuration.Enabled)
                     {
-                        return;
+                        SqlActivitySourceHelper.DbClientOperationDuration.Record(activity.Duration.TotalSeconds);
```
alanwest, line 177: You're not currently adding any dimensions, but once you get to that point...

Assuming you plan on relying on the activity to retrieve the values, you may not have what you need here when both metrics and tracing is enabled and `activity.IsAllDataRequested = false`.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -60,7 +61,10 @@ public override void OnEventWritten(string name, object? payload)
                     if (activity == null)
                     {
                         // There is no listener or it decided not to sample the current request.
-                        return;
+                        activity = new Activity(SqlActivitySourceHelper.MeterName);
```
cijothomas, line 64: unsure why is `Activity` created this way here?

alanwest: Instead of creating an activity, could you incorporate the approach you used in your [previous PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1989/files#diff-339e87239afadd02c6b5e405f59714f3f538a1086db13d4ed983c4850baab098R58-R66)?

The idea being:
* When tracing is disabled use the stopwatch when recording the duration.
* When tracing is enabled use the activity when recording the duration so that the two are in sync.

cijothomas: +1. "Activity" is a bit heavy heap object, AsyncLocal-Stopwatch is much lighter.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -199,11 +206,103 @@ public override void OnEventWritten(string name, object? payload)
...
+        {
+            _ = this.commandFetcher.TryFetch(payload, out var command);
+
+            if (command != null)
+            {
+                if (hasError)
+                {
+                    if (this.exceptionFetcher.TryFetch(payload, out Exception? exception) && exception != null)
+                    {
+                        tags.Add(SemanticConventions.AttributeErrorType, exception.Message);
```
lmolkova, line 265: the message has high cardinality,  this should be the full name of the exception type.
We also added `db.response.status_code` recently - ptal at some details on how to capture it 
https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/database/sql.md

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -199,11 +206,103 @@ public override void OnEventWritten(string name, object? payload)
...
+        var tags = new TagList
+        {
+            { SemanticConventions.AttributeDbSystem, SqlActivitySourceHelper.MicrosoftSqlServerDatabaseSystemName },
+        };
+
+        if (activity != null && activity.IsAllDataRequested)
+        {
+            if (hasError)
+            {
+                var exceptionType = activity.Tags.SingleOrDefault(t => t.Key == SemanticConventions.AttributeExceptionType);
```
alanwest, line 233: The `exception.type` attribute is currently never set on the activity because it is a new attribute and the trace instrumentation has not yet been adapted to the new conventions.

Was your plan to update the trace instrumentation to follow the new conventions in this PR? Or were you looking to split the work up some how?

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlConnectionDetails.cs
```diff
@@ -0,0 +1,180 @@
...
+                sqlActivity.SetTag(SemanticConventions.AttributeServerAddress, connectionDetails.ServerHostName);
+            }
+            else
+            {
+                sqlActivity.SetTag(SemanticConventions.AttributeServerSocketAddress, connectionDetails.ServerIpAddress);
+            }
+
+            if (!string.IsNullOrEmpty(connectionDetails.Port))
+            {
+                // TODO: Should we continue to emit this if the default port (1433) is being used?
```
alanwest, line 175: No, it should not be set if it is the default. See the following footnote from the [conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md#metric-dbclientoperationduration)

> [11]: If using a port other than the default port for this DBMS and if server.address is set.

PR #2033, opened 2024-09-03
Awslambda instrumentation support application loadbalancer events

### LINKED ISSUE #1981: Add ApplicationLoadBalancerRequest parent extraction support to AWSLambda
### Component

OpenTelemetry.Instrumentation.AWSLambda

Currently the AWS Lambda instrumentation does not automatically extract the parent context when using `ApplicationLoadBalancerRequest` input type. This use-case is supported by https://www.nuget.org/packages/Amazon.Lambda.AspNetCoreServer

The parent context should be populated with the HTTP headers of the request, similar to `APIGatewayProxyRequest`

The workaround for this is to set the parent context manually.

### Additional context

_No response_

harveyrendell:
Fixes #1981
Design discussion issue #1981

## Changes

This change adds support for extracting parent context from Lambda functions configured as targets for Application Load Balancers. It follows the conventions from API gateway request types and extracts the same headers as those requests.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Text;
 using System.Web;
 using Amazon.Lambda.APIGatewayEvents;
+using Amazon.Lambda.ApplicationLoadBalancerEvents;
 using OpenTelemetry.Trace;
 
 namespace OpenTelemetry.Instrumentation.AWSLambda.Implementation;
@@ -42,6 +43,13 @@ internal class AWSLambdaHttpUtils
                 var hostHeaderV2 = AWSLambdaUtils.GetHeaderValues(requestV2, HeaderHost)?.LastOrDefault();
                 (hostName, hostPort) = GetHostAndPort(httpScheme, hostHeaderV2);
                 break;
+            case ApplicationLoadBalancerRequest albRequest:
+                httpScheme = AWSLambdaUtils.GetHeaderValues(albRequest, HeaderXForwardedProto)?.LastOrDefault();
+                httpTarget = string.Concat(albRequest.Path ?? string.Empty, GetQueryString(albRequest));
+                httpMethod = albRequest.HttpMethod;
+                var albHostHeader = AWSLambdaUtils.GetHeaderValues(albRequest, HeaderHost)?.LastOrDefault();
+                (hostName, hostPort) = GetHostAndPort(httpScheme, albHostHeader);
+                break;
             default:
                 return tags;
         }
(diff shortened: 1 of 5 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -102,6 +113,51 @@ internal static void SetHttpTagsFromResult(Activity? activity, object? result)
     internal static string? GetQueryString(APIGatewayHttpApiV2ProxyRequest request) =>
         string.IsNullOrEmpty(request.RawQueryString) ? string.Empty : "?" + request.RawQueryString;
 
+    internal static string? GetQueryString(ApplicationLoadBalancerRequest request)
+    {
+        // If the request has a query string value, one of the following properties will be set, depending on if
+        // "Multi value headers" is disabled on ELB Target Group.
+        if (request.QueryStringParameters != null)
```
Oberon00, line 120: I think it needs to be the other way round, i.e. you should check `MultiValueQueryStringParameters` first, at least that would match other event types. But please check what happens if you e.g. trigger an ALB event with a query like `?a=v1&a=v2&b=vb`. I would expect that everything is also contained in MultiValueQueryStringParameters while QueryStringParameters might only have partial information for key `a`.

Oberon00:
Thank you for contributing a new HTTP event type. Since these cloud instrumentations are always hard to unit test, could you please also confirm you tested this manually with a real AWS Lambda and real ALB trigger?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
harveyrendell: As far as I'm aware, these two fields are mutually exclusive so we shouldn't see a situation where both are set for the same request. I agree though that it makes sense to prioritise MultiValue here.

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#enable-multi-value-headers

harveyrendell:
> Thank you for contributing a new HTTP event type! Since these cloud instrumentations are always hard to unit test, could you please also confirm you tested this manually with a real AWS Lambda and real ALB trigger?

Hey, yes I've tested this using lambda as an ALB target and I'm getting the results that I expect.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
harveyrendell: I've tested both with multi value headers enabled and disabled on the target group, and it behaves as you've described.

Oberon00 (approved):
LGTM!

Oberon00:
(BTW, I cannot "approve & run" workflows, so somebody else would need to make sure the tests are run & successful before merging)

Kielek (commented):
Please add changelog entry.

harveyrendell:
Updated changelog and removed an additional blank line that was causing the format check to fail.

Kielek (approved):

Kielek:
@harveyrendell, thanks for your first contribution. Looking for more!

PR #2037, opened 2024-09-03
[Instrumentation.AWSLambda] Incoming FaaS Span attributes: detect and set attribute for cold start

rypdal:
Fixes #
Design discussion issue #

## Changes

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
@@ -162,14 +162,9 @@ public static class AWSLambdaWrapper
             }
         }
 
-        // No parallel invocation of the same lambda handler.
-        bool faasColdStart = isColdStart;
-        if (faasColdStart)
-        {
-            isColdStart = false;
-        }
-
-        var functionTags = AWSLambdaUtils.GetFunctionTags(input, context, faasColdStart);
+        // No parallel invocation of the same lambda handler expected.
+        var functionTags = AWSLambdaUtils.GetFunctionTags(input, context, isColdStart);
+        isColdStart = false;
         var httpTags = AWSLambdaHttpUtils.GetHttpTags(input);
 
         // We assume that functionTags and httpTags have no intersection.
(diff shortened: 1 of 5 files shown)


Kielek (approved):
LGTM.
Consider if any tests needs to be added/extended.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
```diff
@@ -160,7 +162,14 @@ public static Task<TResult> TraceAsync<TInput, TResult>(
             }
         }
 
-        var functionTags = AWSLambdaUtils.GetFunctionTags(input, context);
+        // No parallel invocation of the same lambda handler.
+        bool faasColdStart = isColdStart;
+        if (faasColdStart)
```
Oberon00, line 167: The `if` can be removed, it's fine to unconditionally set it to false here. Then we also might not need the local variable if you change it after calculating functionTags

Oberon00:
Yes, this should be unit-testable, please add or extend one to verify the coldstart attribute going from true on the first call to false on subsequent ones.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs
rypdal: Yes, we can do it. I removed the local variable and moved setting of the internal one after the calculation of functionTags.

rypdal:
> LGTM. Consider if any tests needs to be added/extended.

I added test for the cold start tag check.

Kielek:
One more thing, please extend changelog.

rypdal:
> One more thing, please extend changelog.

changelog has been updated.

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added detection of Lambda cold start and setting of `faas.coldstart` Activity tag.
```
```suggestion
* Add detection of Lambda cold start and set `faas.coldstart` Activity tag.
```

Oberon00 (approved):

Kielek (approved):

PR #2039, opened 2024-09-05
[OpenTelemetry.Instrumentation.AWS] Fix Memory Leak by Reusing ActivitySources, Meters, and Instruments

### LINKED ISSUE #2025: Memory leak with AWS instrumentation
### Component

OpenTelemetry.Instrumentation.AWS

### Package Version

| OpenTelemetry.Exporter.OpenTelemetryProtocol                     | 1.9.0  |
| OpenTelemetry.Instrumentation.AspNetCore                     | 1.9.0  |
| OpenTelemetry.Extensions.Hosting | 1.9.0   |
| OpenTelemetry.Instrumentation.AWS | 1.1.0-beta.5 |

### Runtime Version

net8.0

### Description

``OpenTelemetry.Instrumentation.AWS `` v ``1.1.0-beta.5`` appears to have a memory leak issue which leads to it continuously use up more memory on long running applications.

I am relatively confident in this by running an app with the only difference being in ``1.1.0-beta.5`` vs  ``1.1.0-beta.4``

Please note, it is is only very obvious under high load scenarios. For example I am able to notice it within the hour with an application 2.7k traces/hr (sampling 2% of ~135k requests in a web application for example).

It happens on very modest applications as well but isn't particularly noticeable for a lot longer.

### Steps to Reproduce

This will be tricky since its really minor at low load. 

I am happy to try come up with an experiment on how to replicate this, but it will certainly be tricky, the POC will presumably need to make AWS API calls to get the lib's code to be exercised.

### Expected Result

Steady state for memory use

### Actual Result

Memory use increases steadily without end.

For example this is within an hour or so of deployment
image

Not the greatest increase sure, but over a longer period of time (about 12 hours)

image

You can assume the blue valley in the graph is due to deployment

### Additional Context

I will investigate it a bit further as well, to be completely sure it is the lib, and if so where the leak is.

Will need to do some investigation on how to get memory snapshots out of the services I can replicate this issue in to try do that.

muhammad-othman:
Fixes #2025

## Changes
This pull request addresses issue #2025 a memory leak issue in `OpenTelemetry.Instrumentation.AWS`, particularly related to the new metrics features.
The memory leak was identified after running a significant number of requests (around 10,000). The root cause was found to be the non-reuse of `ActivitySources`, `Meters`, and Metrics Instruments. These objects were being instantiated repeatedly, leading to a gradual increase in memory usage.
This PR includes refactoring the code to reuse `ActivitySources`, `Meters`, and Metrics Instruments instead of creating new instances for each request.

Here are memory graph screenshots to illustrate the impact of the fix:
* Before the Fix (with memory leak):
    * The first snapshot was taken after running 200 requests to warm up the application.
    * The second snapshot was taken after running 10,000 requests, showing a significant memory increase due to the leak.
showing-memory-leak-2

* After the Fix (no memory leak):
    * The first snapshot was again taken after 200 requests.
    * The second snapshot was taken after 10,000 requests, showing stable memory usage with no signs of a leak related to `OpenTelemetry.Instrumentation.AWS`.
image


src/OpenTelemetry.Instrumentation.AWS/Implementation/Metrics/AWSHistogram.cs
```diff
@@ -9,11 +10,25 @@ namespace OpenTelemetry.Instrumentation.AWS.Implementation.Metrics;
 internal sealed class AWSHistogram<T> : Histogram<T>
     where T : struct
 {
+    private static readonly ConcurrentDictionary<string, System.Diagnostics.Metrics.Histogram<T>> HistogramsDictionary
```
ppittle, line 13: naive question - any concern with effectively caching Histograms, Meters, etc indefinitely?   Looking to confirm it is necessary to clean up any old objects.

muhammad-othman: I don't think there are major side effects other than the memory overhead considering that other instrumentations packages do reuse the instruments too (ex:https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInMetricsListener.cs#L36) 
though we do store more than just one histogram and meter.

For cleaning up old objects, we do create one meter per service the first time it is used and around a dozen metrics instruments each, the names of these instruments are constants that doesn't include any user defined value, so there will always be a chance that the old objects are needed later.

ppittle (approved):

Kielek (approved):
Approving per component owner review.

Kielek:
Merging as agreed with @ppittle.

cijothomas:
These are very common mistakes, and has now been explicitly called out in docs. See below for Metrics. Similar exists for Traces too.
https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics#metrics-api

PR #2040, opened 2024-09-05
[SemanticConvention] Migrate generation to Weaver and bump version to 1.26

### LINKED ISSUE #1944: Migrate semantic conventions generation to Weaver
### Component

OpenTelemetry.SemanticConventions

The Semantic conventions repo already made the switch to use [Weaver](https://github.com/open-telemetry/weaver) instead of the "old" build-tools. As part of this transition, the semconv group also would like to move all SIGs to also use Weaver for codegen, that is the semantic conventions constants. 

We have discussed in the group and want to volunteer work to each language SIG to help them move to Weaver. Since I worked on this package originally, I thought of coming back to it and help with the switch. 

Part of: https://github.com/open-telemetry/weaver/issues/227

Semconv codegen is generated via Weaver

N/A

### Additional context

_No response_

joaopgrassi:
Fixes #1944

## Changes

- Move code generation to Weaver instead of build-tools
- Bump semconv version to `v.1.26.0` as that's required for the Weaver move

My goal with this initial PR is to do the minimum changes to get Weaver to work. To bring this package to stabilization, we will need more PRs to separate stable to unstable attributes.


src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
```diff
@@ -2,8 +2,8 @@ $SCRIPT_DIR = $PSScriptRoot
 $ROOT_DIR = "${SCRIPT_DIR}/../"
 
 # freeze the spec version to make SemanticAttributes generation reproducible
-$SPEC_VERSION = "1.25.0"
-$GENERATOR_VERSION = "latest"
+$SEMCONV_VERSION = "1.26.0"
```
lmolkova, line 5: I believe weaver won't work properly with 1.25.0, so this is necessary.

Nit: could you please call semconv version update out in the PR title/description?

src/OpenTelemetry.SemanticConventions/scripts/templates/registry/SemanticConventionsAttributes.cs.j2
```diff
@@ -0,0 +1,52 @@
...
+using System;
+
+namespace OpenTelemetry.SemanticConventions;
+
+/// <summary>
+/// Constants for semantic attribute names outlined by the OpenTelemetry specifications.
+/// </summary>
+public static class {{my_class_name}}
+{
+{% for attribute in ctx.attributes | rejectattr("name", "in", ctx.excluded_attributes) | sort(attribute="name") %}
```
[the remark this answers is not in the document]
joaopgrassi: Neat!

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
joaopgrassi: Done!

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
```diff
@@ -2,28 +2,29 @@ $SCRIPT_DIR = $PSScriptRoot
...
```
cijothomas, line 5: could you add this version update to changelog?

src/OpenTelemetry.SemanticConventions/scripts/generate.sh
```diff
@@ -5,8 +5,8 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 ROOT_DIR="${SCRIPT_DIR}/../"
 
 # freeze the spec version to make SemanticAttributes generation reproducible
-SPEC_VERSION=1.25.0
-GENERATOR_VERSION=latest
+SPEC_VERSION=1.26.0
```
cijothomas, line 8:
```suggestion
SEMCONV_VERSION=1.26.0
```

cijothomas:
> To maintainers:
> 
> Can we not run the `Run python3 ./build/scripts/sanitycheck.py` for the Semconv package? For ex, now it is complaining with trailing space which is not so easy to fix within code gen. Also, given this is auto gen, seems a bit superfluous. image.
> 
> Also the non-ascii check is problematic. Will try to remove that from the semconv itself, but I still believe we can just skip this check for this package.

I think it is reasonable to modify the sanitycheck script to have a "ignore" list of directories to ignore and ask it to ignore sem conventions. (This can be done as a separate PR and merge before this PR to keep the overall scope focused)

src/OpenTelemetry.SemanticConventions/scripts/templates/registry/weaver.yaml
```diff
@@ -0,0 +1,48 @@
+params:
+
+  # excluded attributes will not be generated
+  # this behavior is fully controlled by jinja templates
+  excluded_attributes: ["messaging.client_id", "otel.library.name", "otel.library.version"]
```
joegoldman2, line 5: Any reason why you didn't include otel.library.name and otel.library.version? They are deprecated but has to be propagated according to https://github.com/open-telemetry/opentelemetry-specification/blob/v1.36.0/specification/common/mapping-to-non-otlp.md#instrumentationscope.

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
joegoldman2, line 5: Any reason why you're not using the latest version 1.27.0?

src/OpenTelemetry.SemanticConventions/scripts/templates/registry/weaver.yaml
joaopgrassi: Because there's no `brief` for them so there's no `summary` and the build fail due to not having docs for public members. I fixed it in semconv so once that is released, bumping the version and removing them from here will bring them back.

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
joaopgrassi: No specific reason, I wanted to keep the changes at minimum at this first PR.

joaopgrassi:
> I think it is reasonable to modify the sanitycheck script to have a "ignore" list of directories to ignore and ask it to ignore sem conventions. (This can be done as a separate PR and merge before this PR to keep the overall scope focused)

Alright, thanks @cijothomas! I will look into it and send a separated PR.

src/OpenTelemetry.SemanticConventions/.publicApi/PublicAPI.Unshipped.txt
Kielek, line 1: Please add to src/OpenTelemetry.SemanticConventions/.publicApi/PublicAPI.Shipped.txt and regenerate this file

```txt
#nullable enable

```

src/OpenTelemetry.SemanticConventions/Attributes/AndroidAttributes.cs
```diff
@@ -15,35 +15,35 @@ namespace OpenTelemetry.SemanticConventions;
 public static class AndroidAttributes
 {
     /// <summary>
-    /// Uniquely identifies the framework API revision offered by a version (<c>os.version</c>) of the android operating system. More information can be found <a href="https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels">here</a>.
+    /// Uniquely identifies the framework API revision offered by a version (<c>os.version</c>) of the android operating system. More information can be found <a href="https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels">here</a>
```
Kielek, line 18: Can we keep `.` at the end of documentation comment? Typically it leads to https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md

joaopgrassi: Yes, there's already a issue in Weaver to add support to adding dots at the end. https://github.com/open-telemetry/weaver/issues/353

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
```diff
@@ -2,8 +2,8 @@ $SCRIPT_DIR = $PSScriptRoot
...
+$GENERATOR_VERSION = "v0.9.1"
```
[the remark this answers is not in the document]
joaopgrassi: I mean we can always create a script to read both, find the `$GENERATOR_VERSION = "v0.9.2"` and validate against each other. 

What I'm thinking is if we actually even need both. Maybe we just need the PowerShell one? Given one can install Powershell in Linux..

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
joaopgrassi: Not sure I follow. What committed version you mean? Once things are committed, nothing is suppose to change or?

src/OpenTelemetry.SemanticConventions/.publicApi/PublicAPI.Unshipped.txt
joaopgrassi: Should I add to the `PublicAPI.Shipped.txt`? Or to the template where the `.cs` files are generated?

Kielek: `PublicAPI.Shipped.txt`. Nullability are enabled across solution.

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
Kielek: It will be great to have validation on CI that the code committed by the PR author is correctly generated, without any manual modifications.

I am asking to run script in CI-pipeline and check if there is no differences. Similar what we have for `dotnet format`.

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
joaopgrassi: Fixed. It still worked, but now the error is just ignored.

src/OpenTelemetry.SemanticConventions/.publicApi/PublicAPI.Unshipped.txt
Kielek: I think that there is some issue with PuglicAPI generator if whole public api is taken from generated files.
I played a bit with this, and the solution is to add `#nullable enable` to the template. Then all files are correctly annotated.

Attaching git patch. I have tried to push it, but I do not have permissions to do it (I can only merge changes from GH UI).

[0001-Fix-nullable-annotation-in-PublicAPI-files.patch](https://github.com/user-attachments/files/16988982/0001-Fix-nullable-annotation-in-PublicAPI-files.patch)

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
Kielek: If you are fine, we can do it in the separate PR. It will be easier to focus just on this script/pipeline.
IMO it is not hard-blocker as long as you want to provide it. :)

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
Kielek: It can be considered as follow up PR. Not a blocker here.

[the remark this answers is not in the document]
joaopgrassi: I created a follow up issue for it. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2068

src/OpenTelemetry.SemanticConventions/scripts/generate.ps1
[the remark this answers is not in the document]
joaopgrassi: Alright I will remove it from here and keep it as a follow up. I created an issue, feel free to assign it to me. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2068

src/OpenTelemetry.SemanticConventions/.publicApi/PublicAPI.Unshipped.txt
joaopgrassi: Yeah something weird with our fork org, I allowed maintainers edit but 🤷 . I changed it now to have it on the template and generated everything. Should be fine now!

PR #2054, opened 2024-09-10
[repo] Always execute build-test step

Kielek:
## Changes

Always execute build-tests step. It is used to determine if all steps passed.
Before changes there were possibility to merge changes even if previous steps were failing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -590,6 +590,7 @@ jobs:
       lint-misspell-sanitycheck,
       detect-changes,
       lint-md,
+      lint-yml,
       lint-dotnet-format,
       build-test-exporter-geneva,
       build-test-exporter-influxdb,
@@ -633,7 +634,8 @@ jobs:
       build-test-contrib-shared-tests,
       verify-aot-compat
     ]
-    if: always() && !cancelled() && !contains(needs.*.result, 'failure')
-    runs-on: windows-latest
+    if: always() && !cancelled()
+    runs-on: ubuntu-latest
     steps:
-    - run: echo 'build complete ✓'
+    - run: |
+          if ( ${{ contains(needs.*.result, 'failure') }} == true ); then echo 'build failed ✗'; exit 1; else echo 'build complete ✓'; fi


.github/workflows/ci.yml
```diff
@@ -633,7 +634,8 @@ jobs:
...
       verify-aot-compat
     ]
-    if: always() && !cancelled() && !contains(needs.*.result, 'failure')
-    runs-on: windows-latest
+    if: always() && !cancelled()
+    runs-on: ubuntu-latest
     steps:
-    - run: echo 'build complete ✓'
+    - run: |
+          if ( ${{ contains(needs.*.result, 'failure') }} == true ); then echo 'build failed ✗'; exit 1; else echo 'build complete ✓'; fi
```
CodeBlanch, line 641: @Kielek I'm kind of confused by this 😄 `build-test` is a required status. If it doesn't run (because something failed), that should block the merge by design. Having it run always seems like it would pass the status check even when there was a failure?

CodeBlanch: Current config for `main*`:

image

Kielek: Current status:
 When `${{ contains(needs.*.result, 'failure')` is true (one of the required step is failing) `build-status` step is not executed. It has `skipped` status. For some reasons in this case, GitHub unblocks `Merge` button. Eg. #2052 there were possibility to merge traditional way, without forcing it.

After changes:
`build-test` is always executed, but when `${{ contains(needs.*.result, 'failure')` is true it calls `exit 1`. It means that step is failing. The merge should be prevented.

CodeBlanch (approved):
LGTM

.github/workflows/ci.yml
CodeBlanch: Sounds good thanks for the explanation! We should probably do same over here eh? https://github.com/open-telemetry/opentelemetry-dotnet/blob/37535a5607ee7e4056c0e274ec01d1e0111a64be/.github/workflows/ci.yml#L189

Kielek: That's the plan. Same for auto-instrumentation repo.

Kielek:
Some random fail in this PR. Works as expected:
image

PR #1279, opened 2023-07-26
Downgrade `system.text.json` OpenTelemetry.ResourceDetectors.Azure

vishweshbankwar:
Fixes #.

## Changes

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [ ] Design discussion issue #
* [ ] Changes in public API reviewed

PR #2056, opened 2024-09-10
[Resources.Azure] NugetAudit - fix dependencies with known vulnerabilities

Kielek:
Follow up to #2034

## Changes

``` 
    <!-- System.Text.Encodings.Web is indirect reference. It is needed to upgrade it directly to avoid https://github.com/advisories/GHSA-ghhp-997w-qr28 -->
    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
```
Changelog based on https://github.com/open-telemetry/opentelemetry-dotnet/blob/37535a5607ee7e4056c0e274ec01d1e0111a64be/src/OpenTelemetry.Exporter.Console/CHANGELOG.md#L112-L114

### DIFF
--- a/src/OpenTelemetry.Resources.Azure/OpenTelemetry.Resources.Azure.csproj
+++ b/src/OpenTelemetry.Resources.Azure/OpenTelemetry.Resources.Azure.csproj
@@ -15,6 +15,8 @@
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
     <PackageReference Include="System.Text.Json" Version="4.7.2" />
+    <!-- System.Text.Encodings.Web is indirect reference. It is needed to upgrade it directly to avoid https://github.com/advisories/GHSA-ghhp-997w-qr28 -->
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/OpenTelemetry.Resources.Azure/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Azure/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added direct reference to `System.Text.Encodings.Web` with minimum version of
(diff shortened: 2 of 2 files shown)


src/OpenTelemetry.Resources.Azure/OpenTelemetry.Resources.Azure.csproj
```diff
@@ -15,6 +15,8 @@
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
     <PackageReference Include="System.Text.Json" Version="4.7.2" />
+    <!-- System.Text.Encodings.Web is indirect reference. It is needed to upgrade it directly to avoid https://github.com/advisories/GHSA-ghhp-997w-qr28 -->
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
```
rajkumar-rangaraj, line 19: The .NET Standard 2.0 version of `System.Text.Json` has an indirect dependency on `System.Text.Encodings.Web`. It is better to upgrade `System.Text.Json`, as was done in [OpenTelemetry.Exporter.OneCollector](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/4791960d8c49adb9e5b45b8900010e3337c06591/src/OpenTelemetry.Exporter.OneCollector/OpenTelemetry.Exporter.OneCollector.csproj#L27)

rajkumar-rangaraj: We may need to re-evaluate whether this is still a challenge with auto-instrumentation - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1279#issuecomment-1652182987

Kielek: If possible, I would be more conservative with versioning. This solution is applied in Exporter.Console. I do not see any reason we cannot follow this pattern.

rajkumar-rangaraj: We will explore it outside of this PR to understand System.Text.Json version could be bumped.

rajkumar-rangaraj (approved):

PR #2122, opened 2024-10-01
[onecollector] remove Author tag

Kielek:
Fixes https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2118#discussion_r1782129093

## Changes

[onecollector] remove Author tag
it is inherited from props files

I do not found any other places with this.

### DIFF
--- a/src/OpenTelemetry.Exporter.OneCollector/OpenTelemetry.Exporter.OneCollector.csproj
+++ b/src/OpenTelemetry.Exporter.OneCollector/OpenTelemetry.Exporter.OneCollector.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Description>An OpenTelemetry .NET exporter that sends telemetry to Microsoft OneCollector</Description>
-    <Authors>OpenTelemetry Authors</Authors>
     <TargetFrameworks>net8.0;net6.0;netstandard2.1;netstandard2.0</TargetFrameworks>
     <TargetFrameworks Condition="$(OS) == 'Windows_NT'">$(TargetFrameworks);net462</TargetFrameworks>
     <MinVerTagPrefix>Exporter.OneCollector-</MinVerTagPrefix>


src/OpenTelemetry.Exporter.OneCollector/OpenTelemetry.Exporter.OneCollector.csproj
```diff
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Description>An OpenTelemetry .NET exporter that sends telemetry to Microsoft OneCollector</Description>
-    <Authors>OpenTelemetry Authors</Authors>
```
reyang, line 6: On a related topic, do we allow component owners to put different authors info here or we enforce that all projects in this repo should use "OpenTelemetry Authors"?

reyang (approved):

src/OpenTelemetry.Exporter.OneCollector/OpenTelemetry.Exporter.OneCollector.csproj
Kielek: Nobody want tot change it. I would say that it has to be enforced. All code files higlight _OpenTelemetry Authors_ as authors. As I remember it is done by OTel guidance.

reyang: I think copyright notice in the source code is a different thing than the authors tag in the package metadata. Also, OpenTelemetry's guidance allow retaining previous copyright FYI https://github.com/open-telemetry/community/blob/c77b41e023291ac48ce325c6defc9783a72d4b5c/guides/contributor/processes.md?plain=1#L202-L206

Kielek: There were no such requests. We could handle it when needed : ).

IMO, if it was donated/vendore, it will be modified by someone from donators. In this case Third-party-file is good place to put all important credits.

PR #2130, opened 2024-10-02
[Instrumentation.EntityFrameworkCore] Update semantic conventions for stable release

joegoldman2:
## Changes

Add support for `OTEL_SEMCONV_STABILITY_OPT_IN` following https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md.


src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Some attributes have been updated to align with
+  the latest version of Semantic Conventions:
+  - Rename `db.name` to `db.namespace`
+  - Rename `db.statement` to `db.query.text`
```
Kielek, line 8: I think that this part should be postponed according to https://github.com/open-telemetry/semantic-conventions/blob/v1.27.0/docs/database/database-spans.md Switch should be done as a batch.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+  - Rename `peer.service` to `server.address`
```
Kielek, line 9: I am fine with this change as network sem. conv. is with us for a while.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
joegoldman2: I'm fine with that. We can wait for the db stabilization.

Kielek:
@joegoldman2, if you want to proceed with this PR, please consider https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/db-migration.md

There were semantic convention helper for HTTP in main repository. It can be easily adjusted: https://github.com/open-telemetry/opentelemetry-dotnet/pull/5270/files#diff-c94fc23f1212ef9513764dd888ec8c98d2d59b2d51a0aabb311fa0652ec8b059

Keep in mind that network related attributes can be fixed now, the problem is only with SQL releated part.

joegoldman2:
I tested the change manually on a test app, it is working.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -139,10 +142,19 @@ public override void OnEventWritten(string name, object? payload)
                         }
 
                         var dataSource = (string)this.dataSourceFetcher.Fetch(connection);
-                        activity.AddTag(AttributeDbName, database);
```
alanwest, line 142: We should still emit this tag when emitting the old attributes.

alanwest: Oh nevermind.. it moved below..

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -139,10 +142,19 @@ public override void OnEventWritten(string name, object? payload)
...
+                        }
+
+                        if (this.options.EmitOldAttributes)
+                        {
+                            activity.AddTag(AttributeDbName, database);
+                        }
+
+                        if (this.options.EmitNewAttributes)
+                        {
+                            activity.AddTag(AttributeDbNamespace, database);
```
alanwest, line 157: @trask @lmolkova what are your thoughts here?

Per [footnote 4 in the conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#common-attributes):

> Semantic conventions for individual database systems SHOULD document what db.namespace means in the context of that system.

Since this instrumentation is for an ORM and supports many database systems, this should respect the conventions declared for the target database system.

That said, not all systems this instrumentation supports have documentation for what `db.namespace` means.

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/DatabaseSemanticConventionHelperTests.cs
```diff
@@ -0,0 +1,79 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using Microsoft.Extensions.Configuration;
+using Xunit;
+using static OpenTelemetry.Internal.DatabaseSemanticConventionHelper;
+
+namespace OpenTelemetry.Internal.Tests;
+
+public class DatabaseSemanticConventionHelperTests
```
alanwest, line 10: These tests should be moved to the OpenTelemetry.Tests project. Similar to how we had it with the HTTP conventions https://github.com/open-telemetry/opentelemetry-dotnet/pull/5270/files#diff-ece803150e7b62366d8f76174609124fad529370a8c9a62021644dfe43837c45L10

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
trask: we have a similar issue in Java (https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jdbc/library/src/test/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcConnectionUrlParserTest.java)

technically, I think stable JDBC instrumentation should only emit stable instrumentation by default, which means only emitting instrumentation for stable `db.system` values

in practice, this would mean that we can't mark the JDBC instrumentation as stable even after we update it to emit the stable database semconv (unless we want to force lots of users to opt-in to the unstable `db.system` values, which I don't think is a great experience)

I'm hoping this will motivate us to push semconv definitions for `db.namespace` for all of these other databases quickly, we just haven't had the bandwidth to do it yet and didn't want to block initial database semconv stability on it

alanwest (approved):
This PR looks good to me. There's still things left to do, but I think this is a step forward, so approving/merging.

I will open issues, but here's a quick summary off hand for things left to do:

1. Collecting additional attributes
    * `error.type` / `db.response.status_code`
    * Adding option to parse `db.collection.name` / `db.operation.name` (this is something I'm working on in the context of the SQL instrumentation and should also be applicable here)
2. Adding option to sanitize `db.query.text` (also working on this for SQL instrumentation)
3. Defining our requirements for declaring this instrumentation stable. It's an ORM, so it may have some unique things to think through see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2130#discussion_r1803604186

PR #2140, opened 2024-10-03
[Instrumentation.AWSLambda] Replace .NET 6 target with .NET 8 and add .NET Standard 2.0

joegoldman2:
## Changes

Replace .NET 6 target, which will be very soon out of support, with .NET 8 and add support for .NET Standard 2.0.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -6,6 +6,8 @@
   ([#2037](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2037))
 * Add HTTP server span attributes for Application Loadbalancer triggers
   ([#2033](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2033))
+* Drop support for .NET 6 as this target is no longer supported and add .NET 8 target.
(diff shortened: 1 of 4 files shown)


Oberon00:
I would prefer waiting at least for the "Block function create" date (Feb 28, 2025)
image

joegoldman2:
> I would prefer waiting at least for the "Block function create" date (Feb 28, 2025) https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

.NET 6 itself will be out of support before. People who still use .NET 6 will be able to keep the current package version, right?

joegoldman2:
> I would prefer waiting at least for the "Block function create" date (Feb 28, 2025) https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

The addition of .NET Standard 2.0 target should be fine for people who are targeting .NET 6 and less.

Oberon00:
That looks fine to me. Do we even need .NET 8 separately if we have netstandard2.0 as target?

joegoldman2:
> That looks fine to me. Do we even need .NET 8 separately if we have netstandard2.0 as target?

For all packages, I'm replacing .NET 6 with .NET 8 and add .NET Standard 2.0 if not already present. I think some features are not available if only .NET Standard is targeted (but not sure if it's the case for this package).

CodeBlanch (approved):
LGTM

PR #2188, opened 2024-10-07
[Sampler.AWS] Wrapped root sampler with ParentBasedSampler

AsakerMohd:
Fixes #
Design discussion issue #

## Changes

The current implementation of the X-Ray sampler has 2 levels: The X-Ray Sampler itself which is made up two different samplers for X-Ray purposes. Those two different samplers are wrapped by parentBasedSampler so that the children inherit the parent sampling decision. However, on the CloudWatch console, we are adding the children sampling decision into the stats section. This means: if we have 1 parent and 9 children, the CW console will show that 10 spans are sampled which isn't correct. It should show only 1 span being sampled. This PR moves the ParentBasedSampler wrapper to be on the root level of the X-Ray sampler. This means that the builder doesn't return AWSXrayRemoteSampler anymore but returns ParentBasedSampler(new AWSXrayRemoteSampler). This keeps the same parentBasedSampling logic but is instead on the root level thus skipping the stat aggregation for the children.  


vastin (approved):
LGTM

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSamplerBuilder.cs
```diff
@@ -66,8 +67,17 @@ public AWSXRayRemoteSamplerBuilder SetEndpoint(string endpoint)
     /// <summary>
     /// Returns a <see cref="AWSXRayRemoteSampler"/> with configuration of this builder.
     /// </summary>
-    /// <returns>an instance of <see cref="AWSXRayRemoteSampler"/>.</returns>
-    public AWSXRayRemoteSampler Build()
+    /// <returns>an instance of <see cref="ParentBasedSampler"/>.</returns>
+    public ParentBasedSampler Build()
```
Kielek, line 71:
```suggestion
    public Sampler Build()
```

Shouldn't it be just `Sampler` to avoid any future breaking changes?

AsakerMohd: Good call. Updated the return type to be Trace.Sampler instead of ParentBasedSampler.

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSamplerBuilder.cs
```diff
@@ -66,8 +67,17 @@ public AWSXRayRemoteSamplerBuilder SetEndpoint(string endpoint)
...
+    public Trace.Sampler Build()
+    {
+        using var rootSampler = this.BuildXraySampler();
+        return new ParentBasedSampler(rootSampler);
+    }
+
+    // This is intended for testing to check that the XRayRemoteSampler is built with the correct attributes
+    // Should not be exposed to public as the public build method should return the sampler wrapped inside
+    // ParentBasedSampler.
+    internal AWSXRayRemoteSampler BuildXraySampler()
```
ppittle, line 80: Is there a way to avoid needing a special constructor just for testing?

Could the tests call `Build`, get the `ParentBasedSampler` and then access the `AWSXrayRemoteSampler` from that class some way?

AsakerMohd: Using reflection allows us to get the private sampler in the ParentBasedSampler class. However doing so is kindoff a code smell since at this point we are relying on the internal variable naming of ParentBasedSampler and if that changes, the test breaks. I would prefer using the special constructor in this case since it's being used in the actual public facing Build function and can be used to test the AWSXrayRemoteSampler being built properly.

Kielek:
@ppittle, do you want to merge this?

@AsakerMohd, I think that having CHANGELOG entry is good idea here.

AsakerMohd:
> @ppittle, do you want to merge this?
> 
> @AsakerMohd, I think that having CHANGELOG entry is good idea here.

Yeah let me do that.

src/OpenTelemetry.Sampler.AWS/AWSXRayRemoteSamplerBuilder.cs
```diff
@@ -66,8 +67,17 @@ public AWSXRayRemoteSamplerBuilder SetEndpoint(string endpoint)
     /// <summary>
     /// Returns a <see cref="AWSXRayRemoteSampler"/> with configuration of this builder.
     /// </summary>
-    /// <returns>an instance of <see cref="AWSXRayRemoteSampler"/>.</returns>
-    public AWSXRayRemoteSampler Build()
+    /// <returns>an instance of <see cref="Trace.Sampler"/>.</returns>
+    public Trace.Sampler Build()
+    {
+        using var rootSampler = this.BuildXraySampler();
```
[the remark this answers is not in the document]
AsakerMohd: removed.

ppittle (approved):

ppittle:
@AsakerMohd - LGTM - thanks for the changes!

src/OpenTelemetry.Sampler.AWS/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
 * Updated OpenTelemetry core component version(s) to `1.10.0`.
   ([#2317](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2317))
 
+* `AWSXRayRemoteSamplerBuilder.Build()` now returns `ParentBasedSampler` which
+  which is of type `Trace.Sampler` instead of `AWSXRayRemoteSampler`
+
```
Kielek, line 22:
```suggestion
* `AWSXRayRemoteSamplerBuilder.Build()` now returns `ParentBasedSampler` which
  which is of type `Sampler` instead of `AWSXRayRemoteSampler`.
  ([#2188](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2188))

```

Kielek (approved):
Approving ane mergin based on @ppittle comment.

PR #2262, opened 2024-10-28
[SqlClient] Add error.type and db.response.status_code attributes

### LINKED ISSUE #2239: [sql] Support `error.type` and `db.response.status_code` attributes
Refer to the [conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/mssql.md) for attribute descriptions.

alanwest:
Fixes #2239

### DIFF
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
@@ -4,10 +4,11 @@
 #if !NETFRAMEWORK
 using System.Data;
 using System.Diagnostics;
-using OpenTelemetry.Trace;
 #if NET
 using System.Diagnostics.CodeAnalysis;
 #endif
+using System.Globalization;
+using OpenTelemetry.Trace;
 
 namespace OpenTelemetry.Instrumentation.SqlClient.Implementation;
 
@@ -32,6 +33,7 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
     private readonly PropertyFetcher<CommandType> commandTypeFetcher = new("CommandType");
     private readonly PropertyFetcher<object> commandTextFetcher = new("CommandText");
     private readonly PropertyFetcher<Exception> exceptionFetcher = new("Exception");
+    private readonly PropertyFetcher<int> exceptionNumberFetcher = new("Number");
     private readonly SqlClientTraceInstrumentationOptions options;
 
     public SqlClientDiagnosticListener(string sourceName, SqlClientTraceInstrumentationOptions? options)
@@ -210,6 +212,13 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
                         {
                             if (this.exceptionFetcher.TryFetch(payload, out Exception? exception) && exception != null)
                             {
+                                activity.AddTag(SemanticConventions.AttributeErrorType, exception.GetType().FullName);
+
+                                if (this.exceptionNumberFetcher.TryFetch(exception, out var exceptionNumber))
+                                {
+                                    activity.AddTag(SemanticConventions.AttributeDbResponseStatusCode, exceptionNumber.ToString(CultureInfo.InvariantCulture));
+                                }
+
                                 activity.SetStatus(ActivityStatusCode.Error, exception.Message);
 
                                 if (this.options.RecordException)
--- a/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
@@ -25,7 +25,8 @@
   semantic conventions for
   [spans](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/database/database-spans.md).
   ([#2229](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2229),
-   [#2277](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2277))
+   [#2277](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2277),
+   [#2262](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2262))
 * **Breaking change**: The `peer.service` and `server.socket.address` attributes
   are no longer emitted. Users should rely on the `server.address` attribute
   for the same information. Note that `server.address` is only included when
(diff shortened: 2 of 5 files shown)


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -210,6 +212,13 @@ public override void OnEventWritten(string name, object? payload)
                         {
                             if (this.exceptionFetcher.TryFetch(payload, out Exception? exception) && exception != null)
                             {
+                                activity.AddTag(SemanticConventions.AttributeErrorType, exception.GetType().FullName);
```
alanwest, line 215: This is not very useful because `error.type` will always equal `Microsoft.Data.SqlClient.SqlException`.

Per [the conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/mssql.md), this is not necessarily wrong, but it's not ideal:

> The error.type SHOULD match the db.response.status_code returned by the database or the client library, or the canonical name of exception that occurred. When using canonical exception type name, instrumentation SHOULD do the best effort to report the most relevant type. For example, if the original exception is wrapped into a generic one, the original exception SHOULD be preferred. Instrumentations SHOULD document how error.type is populated.

Doing something like the following would be more useful and would align with the response status code.

```csharp
activity.AddTag(SemanticConventions.AttributeErrorType, exception.Message);
```

However exception message can have variable text and could cause error.type to be high cardinality. For example:

```sql         
SELECT ColumnThatDoesNotExist FROM MyTable
```

This query results in [error code 207](https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-207-database-engine-error?view=sql-server-ver16) and message:

> Invalid column name 'ColumnThatDoesNotExist'

It's tempting to do something like the following to replace anything in single-quotes with something.

```csharp
Regex.Replace(exception.Message, "'[^']*'", "'%'"));
```

But, glancing over the list of [error codes](https://learn.microsoft.com/sql/relational-databases/errors-events/database-engine-events-and-errors?view=sql-server-ver16), I'm not sure such a simple replacement will suffice.

Alternatively, we could maintain a dictionary of codes to messages... there are a lot.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -210,6 +212,13 @@ public override void OnEventWritten(string name, object? payload)
...
+
+                                if (this.exceptionNumberFetcher.TryFetch(exception, out var exceptionNumber))
+                                {
+                                    activity.AddTag(SemanticConventions.AttributeDbResponseStatusCode, exceptionNumber.ToString(CultureInfo.InvariantCulture));
```
alanwest, line 219: The error has a number, severity, and state.

The number is most descriptive of the problem that occurred, though I can see severity and state being useful.

From [the docs](https://learn.microsoft.com/sql/relational-databases/errors-events/understanding-database-engine-errors?view=sql-server-ver16)

Severity:
> The severity indicates how serious the error is. Errors that have a low severity, such as 1 or 2, are information messages or low-level warnings. Errors that have a high severity indicate problems that should be addressed as soon as possible. For more information about severities, see [Database Engine Error Severities](https://learn.microsoft.com/sql/relational-databases/errors-events/database-engine-error-severities?view=sql-server-ver16).

State:
> Some error messages can be raised at multiple points in the code for the Database Engine. For example, an 1105 error can be raised for several different conditions. Each specific condition that raises an error assigns a unique state code.
>
> When you are viewing databases that contain information about known issues, such as the Microsoft Knowledge Base, you can use the state number to determine whether the recorded issue is the same as the error you have encountered. For example, if a Knowledge Base Article describes an 1105 error that has a state of 2 and the 1105 error message you received had a state of 3, the error probably has a different cause than the one reported in the article.
>
> A Microsoft support engineer can also use the state code from an error to find the location in the source code where that error code is being raised. This information might provide additional ideas on how to diagnose the problem.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -210,6 +212,13 @@ public override void OnEventWritten(string name, object? payload)
...
+                                }
+
                                 activity.SetStatus(ActivityStatusCode.Error, exception.Message);
 
                                 if (this.options.RecordException)
```
alanwest, line 224: The instrumentation is currently setting the span's status to ERROR when an exception occurs. A span event is also recorded with the exception if the user opts into this.

I do not think the conventions currently specify setting span status or recording the exception event. Should they?

alanwest (commented):
@lmolkova @trask Can I get your thoughts on a couple things regarding `error.type` and `db.response.status_code`?

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
lmolkova: I believe there is a common practice to set message on the span description.

It's ok if for .NET it's uncommon to subclass exceptions and that `error.type` won't be super-interesting. We still have status code.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
lmolkova: could you please check out https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql.md?plain=1#L102-L136 ?

I believe MS SQL server doesn't support SQL STATE, but if you ever see both - state and code, you can concat them together - see the link above.

If severity is interesting, we can always add it later, but it seems there is a mapping from the code to severity and it might be redundunt.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
lmolkova: > The instrumentation is currently setting the span's status to ERROR when an exception occurs. A span event is also recorded with the exception if the user opts into this.

this seems great for exceptions that are thrown to the user. 

> I do not think the conventions currently specify setting span status or recording the exception event. Should they?

Great point. Let me create an issue.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
alanwest: The notion of state here is different from SQL STATE. From what I can tell you're correct that MSSQL does not support SQL STATE.

My read is that for a given MSSQL error code, that code may be raised from different locations in the DBMS and the state specifies exactly which location the code was raised. I think I'm going to just ignore state.

> If severity is interesting, we can always add it later, but it seems there is a 1:1 mapping from the error code to severity and it might be redundant.

I agree it's probably redundant.

CodeBlanch (approved):
LGTM

PR #2284, opened 2024-10-31
[sqlclient] Remove SetDbStatementForStoredProcedure option

alanwest:
Removes the `SetDbStatementForStoredProcedure` configuration option.

The [database semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/database/database-spans.md) are nearing stability, there are number of attributes which are required when they are readily available (i.e. no complex parsing or sanitization of query text is necessary). This includes `db.query.text`, `db.operation.name`, and `db.collection.name`.

In the case of stored procedures, no parsing or sanitization is necessary, so the instrumentation needs to capture these attributes.

Note that the readme has not yet been updated with documentation of the new conventions, so it currently only refers to the old `db.statement` attribute.


CodeBlanch (approved):
Looks like some lint/publicapi issues but otherwise LGTM

PR #2305, opened 2024-11-07
[sqlclient] Make SqlClientInstrumentation a singleton

alanwest:
In an upcoming PR we will be adding metric instrumentation. That instrumentation will need to leverage the same `DiagnosticSourceListener` that is used for traces. This PR makes `SqlClientInstrumentation` a singleton because multiple instances of the instrumentation would create duplicate activities.


Kielek:
@alanwest, I do not think that metrics should relay on activities. Please consider sampling scenario. It might lead to dropping measurements.

As I remember we have tried this approach in AspNetCore, then changed to separate listeners.

alanwest:
> I do not think that metrics should relay on activities. Please consider sampling scenario. It might lead to dropping measurements.
> 
> As I remember we have tried this approach in AspNetCore, then changed to separate listeners.

Metrics will not rely on activities. A single listener will be needed though because when both tracing and metrics are enabled, the duration of the span and metrics must be the same. A stopwatch will be used when only metrics are enabled or the span is sampled.

Aspnet core was a bit different as the span is always created by the native instrumentation.

CodeBlanch (approved):
LGTM

PR #2359, opened 2024-12-05
Added changes to AWSXRayPropagator

AsakerMohd:
## Changes

While doing some auto instrumentation testing, we found that adding the AWSXRayPropagator and AWS SDK instrumentation causes a conflict with the propagated `X-Amzn-Trace-Id `. This is a header that is used to propagate X-Ray trace context for trace linking. Lets assume we have an application that makes a call to AWS S3. There will be 2 activities in this case, the S3 activity created by the instrumentation `Amazon.AWS.AWSClientInstrumentation` and then a `System.Net.Http.HttpRequestOut` activity which is an http call to `s3.amazonaws.com`. The issue here is that when the httpRequest is created, the AWS SDK injects the traceHeader where the traceId and parent id is set to the `Amazon.AWS.AWSClientInstrumentation` S3 activity. However,  since `System.Net.Http.HttpRequestOut` is the last outgoing call, the propagator injects the traceHeader from `HttpRequestOut`, basically overwriting the previous context. 

To overcome this, a change is made is made so that If X-Amzn-Trace-Id already exists in the headers and the carrier is of HttpRequestMessage, This means that the request is coming from the AWS SDK Instrumentation library and in this case, we don't want to overwrite the propagation context from the AWS SDK Span with the context from the outgoing HttpRequestOut.


ppittle (approved):

Kielek:
Please double check if it does not require changelog entry. Based on the answer, it can be merged as is.

AsakerMohd:
> Please double check if it does not require changelog entry. Based on the answer, it can be merged as is.

Yes it doesn't require any changelog entry so can be merged as is.

Kielek (approved):
Approving/merging based on @ppittle review.

PR #2367, opened 2024-12-06
Upgrade Semantic Convention handling in AWS libraries.

ppittle:
Fixes #
Design discussion issue #

_NOTE:_ This PR is built on top of https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2363, https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2363 should be merged first.

## Updates 2

This PR will contain -no- change in Semantic Convention by default.  Users can opt-in to `SemanticConventionVersions.1.28`, but default is `SemanticConventionVersions.Legacy`.

Once merged, this PR will be released as a beta - allowing users a non-breaking way to test opting in-to 1.28.

There will be a fast follow up PR that removes `SemanticConventionVersions.Legacy` and changes the default to `SemanticConventionVersions.V1_28_0`.  That PR will be merged and released as a Release Candidate.

This follows the release strategy for other contrib packages that have made similar breaking changes around Semantic Convention stabilization.

## Updates 1

- The original design of having `SemanticConventionVersions` track the version of the OpenTelemetry.*.AWS Libraries has been scrapped in favor of tracking Semantic Convnetion repo release versions.
-  To simplify communications and development, all OpenTelemetry.*.AWS Libraries will be upgraded to use 
   Semantic Convention 1.28.
-  To demonstrate how new Semantic Convention versions will be added in the future, this PR also includes Semantic Convention 1.29.
-  To promote stability for users, OpenTelemetry.*.AWS Libraries will default to Semantic Convnetion 1.28 until all conventions used by the libraries are stabalized.  This means future versions of the OpenTelemetry.*.AWS Library will _not_ change `AWSSemanticConvention.DefaultSemanticConventionVersion`
	-  Users can opt-in to any version of Semantic Convention supported the current (or future) versions of OpenTelemetry.*.AWS Libraries.

## Changes

Implementation of the design discussed in the Nov 26th OpenTelemetry .NET SIG meeting. This design allows releasing the AWS OTel Contrib libraries as GA while still informing users that underlying Semantic Conventions are experimental.

This is exposed via the `SemanticConventionVersion` enum:

```csharp
/// <summary>
/// <para>
/// Collection of the Open Telemetry Semantic Conventions supported by the OpenTelemetry.*.AWS libraries.
/// Can be used to pin the version of Semantic Convention emitted.
/// </para>
/// <para>
/// While these libraries are intended for production use, they rely on several Semantic Conventions that are
/// still considered Experimental, meaning they may undergo additional changes before becoming Stable.  This can
/// impact the aggregation and analysis of telemetry signals in environments with multiple applications or microservices.
/// For example, a microservice using an older version of the Semantic Conventions for Http Attributes may emit
/// <c>"http.method"</c> with a value of GET,while a different microservice, using a new version of Semantic Convention may instead emit the GET as
/// <c>"http.request.method"</c>.
/// </para>
/// <para>
/// Future versions the OpenTelemetry.*.AWS libraries will include updates to the Semantic Convention, which may break compatibility with a previous version.
/// To opt-out of automatic upgrades, you can pin to a specific version:
/// <code>
/// <![CDATA[
///  using (var tracerProvider = Sdk.CreateTracerProviderBuilder()
///     .AddAWSLambdaConfigurations(opt =>
///     {
///         opt.SemanticConventionVersion = SemanticConventionVersion.v1_10_EXPERIMENTAL;
///     })
///     .Build()!)
/// ]]>
/// </code>
/// </para>
/// <para>
/// For additional details, see:
/// https://opentelemetry.io/docs/specs/otel/versioning-and-stability/.
/// </para>
/// </summary>
/// <remarks>
/// Once a Semantic Convention becomes Stable, OpenTelemetry.*.AWS libraries
/// will remain on that version until the next major version bump.
/// </remarks>
public enum SemanticConventionVersion
{
    /// <summary>
    /// Use Experimental Conventions until they become stable and then
    /// pin to stable.
    /// </summary>
    Latest = 0,

    /// <summary>
    /// Pin to the specific state of all Semantic Conventions as of the 1.28.0
    /// release. See:
    /// https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.28.0.
    /// </summary>
    V1_28_0_Experimental = 1,

    /// <summary>
    /// Pin to the specific state of all Semantic Conventions as of the 1.29.0
    /// release. See:
    /// https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.29.0.
    /// </summary>
    V1_29_0_Experimental = 2,
}
```

Users can choose how to handle this for their applications by either pinning to a specific Semantic Convention version or automatically getting updates the next time they upgrade the AWS Otel nuget packages.

Selection is made by updating the Options object in the builder method.

```csharp
using (var tracerProvider = Sdk.CreateTracerProviderBuilder()
           .AddAWSLambdaConfigurations(opt =>
           {
               // Added in this PR
               opt.SemanticConventionVersion = SemanticConventionVersion.V1_29_0_Experimental;
           })
           .AddInMemoryExporter(exportedItems)
           .Build()!)
```


src/Shared/AWS/SemanticConventionVersion.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// Pin to the specific state of all Semantic Conventions as of the 1.10 release of this library.
+    /// https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.27.0.
+    /// </summary>
+    v1_10_Experimental = 1,
+
+    /// <summary>
+    /// Pin to the specific state of all Semantic Conventions as of the 1.10.1 release of this library.
+    /// https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.29.0.
+    /// </summary>
+    v1_10_1_Experimental = 2,
```
Kielek, line 62: Consider using semantic convention versions as a names. IMO it will be less confusing.

ppittle: @Kielek - I debated whether to have the naming track the version of OpenTelemetry.*.AWS or the Semantic Convention releases.  

I landed on OpenTelemetry.*.AWS because I thought this would be more approachable/intuitive for users.  a) They're already seeing the version of the OTel library in nuget and in their csproj and b) the SemanticConvention and OpenTelemetry nuget packages update on an independent cadence, so there will inevitably be a miss match.  Ie, OpenTelemetry.*.AWS release include either 0 changes from SemanticConvention, or might use the newest version which is several versions ahead:

_Hypothetical:_
OpenTelemetry.*.AWS 1.10.0 uses SemanticConvnetion 1.27.0
OpenTelemetry.*.AWS 1.10.1 uses SemanticConvnetion 1.29.0
OpenTelemetry.*.AWS 1.10.2 uses SemanticConvnetion 1.29.0
OpenTelemetry.*.AWS 1.11.0 uses SemanticConvnetion 1.35.0


So **Option 1 (current strategy) - Track OpenTelemetry.*.AWS**
```csharp
public enum SemanticConventionVersion
{
   v1_10_0_Experimental, // tracks 1.27.0
   v1_10_1_Experimental, // tracks 1.29.0
   v1_11_0_Experimental, // tracks 1.35.0
}

```

Or **Option 2  - Track SemanticConvention**
```csharp
public enum SemanticConventionVersion
{
   v1_27_0_Experimental, // added in OpenTelemetry.*.AWS 1.10.0
   v1_29_0_Experimental, // added in OpenTelemetry.*.AWS 1.10.1
   v1_35_0_Experimental, // added in OpenTelemetry.*.AWS 1.11.0
}
```

**Very open to suggestion / comments here.  What would be the best user experience?**

Kielek: I would go with Optuon 2.

It is the common sense con Semantic Covention Version.

Option 1 is using same name for different purposes.

ppittle: Sounds good - I'll code up the change.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
```diff
@@ -27,9 +27,14 @@
...
   <ItemGroup>
+    <Compile Include="$(RepoRoot)\src\Shared\AWS\*.cs" Link="Includes\AWS\%(Filename).cs" />
     <Compile Include="$(RepoRoot)\src\Shared\AssemblyVersionExtensions.cs" Link="Includes\AssemblyVersionExtensions.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
-    <Compile Include="$(RepoRoot)\src\Shared\SemanticConventions.cs" Link="Includes\SemanticConventions.cs"/>
+    <Compile Include="$(RepoRoot)\src\Shared\SemanticConventions.cs" Link="Includes\SemanticConventions.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\OpenTelemetry.SemanticConventions\OpenTelemetry.SemanticConventions.csproj" />
```
ppittle, line 37: @Kielek - I am referencing the SemanticConventions project to bring in documentation in the `AWSSemanticConventions` class via `/// <inherit cref=""/>`.

Can you advise if this is the correct way to reference that project?

ppittle:
@Kielek - is the CHANGELOG auto-generated or should I manually add an entry for each impacted project?

src/Shared/AWS/AWSSemanticConventions.cs
```diff
@@ -0,0 +1,329 @@
...
+        return activity;
+    }
+
+    private static IDictionary<string, string> AddDic(IDictionary<string, string> dict, Func<AWSSemanticConventionsBase, string> attributeNameFunc, string value)
+    {
+        var semanticConventionVersionImpl = GetSemanticConventionVersion();
+
+        var attributeName = attributeNameFunc(semanticConventionVersionImpl);
+
+        if (!string.IsNullOrEmpty(attributeName))
```
Oberon00, line 303: Empty strings may be valid and meaningful depending on the concrete attribute.

src/Shared/AWS/AWSSemanticConventions.cs
```diff
@@ -0,0 +1,329 @@
...
+    /// <![CDATA[
+    /// public class AWSLambdaInstrumentationOptions
+    /// {
+    ///    /// <inheritdoc cref="OpenTelemetry.AWS.SemanticConventionVersion"/>
+    ///    public SemanticConventionVersion SemanticConventionVersion { get; set; } = AWSSemanticConventions.DefaultSemanticConventionVersion;
+    /// }
+    /// ]]>
+    /// </code>
+    /// </summary>
+    internal const SemanticConventionVersion DefaultSemanticConventionVersion = SemanticConventionVersion.Latest;
```
Oberon00, line 62: At least for one release, wouldn't it make sense to choose the default such that behavior remains unchanged and users have the option to upgrade at their own convenience?

src/Shared/AWS/AWSSemanticConventions.v1.27.0.cs
```diff
@@ -0,0 +1,105 @@
...
+        public override string AttributeAWSDynamoTableName => "aws.table_name";
+        public override string AttributeAWSSQSQueueUrl => "aws.queue_url";
+        public override string AttributeAWSBedrockAgentId => "aws.bedrock.agent.id";
+        public override string AttributeAWSBedrockDataSourceId => "aws.bedrock.data_source.id";
+        public override string AttributeAWSBedrockGuardrailId => "aws.bedrock.guardrail.id";
+        public override string AttributeAWSBedrockKnowledgeBaseId => "aws.bedrock.knowledge_base.id";
+        public override string AttributeAWSBedrock => "aws_bedrock";
+
+        // FAAS Attributes
+        public override string AttributeFaasID => "cloud.resource_id";
```
Oberon00, line 69: This was previously `"faas.id"` (matching the identifier's name). It seems this now always becomes cloud.resource_id. I don't know in which version faas.id was removed, was that already before 1.27? If so, I think since we are already going out of our way to offer different versions, we should also support the one that was previously used here.

Kielek:
> @Kielek - is the CHANGELOG auto-generated or should I manually add an entry for each impacted project?

Changelogs needs to be manually modified. The only exception is release process (adding version and release date).

src/Shared/AWS/AWSSemanticConventions.cs
muhammad-othman: I think we should always set `DefaultSemanticConventionVersion` to a specific version rather than `SemanticConventionVersion.Latest` and only update it when we introduce a new major version. This way users can safely upgrade minor versions without facing breaking changes.

That said, after glancing over other packages, it seems that none handle backward compatibility for semantic convention versioning. So I'm find with keeping it this way if we are ok with making maintaining backward compatibility a user responsibly using the configuration that was added.

ppittle: This request makes sense to me - we can make this more conservative and set the default to the 1.27, which mostly aligns to what the current library uses.  

This way users wont see any changes in Semantic Convention unless they opt-in.

I've made this update.

src/Shared/AWS/AWSSemanticConventions.v1.27.0.cs
ppittle: @Oberon00 - good catch. v1.27 is meant to track the current state of the OpenTelemeetry.*.AWS libraries - so I've reverted this back to `faas.id`.  v1.29 will upgrade to `cloud.resource_id`.

I rechecked the remaining faas attributes, but let me know if you see any other inconsistencies.

src/OpenTelemetry.Instrumentation.AWSLambda/OpenTelemetry.Instrumentation.AWSLambda.csproj
ppittle: nevermind, I copied over the xml documentation rather than take the reference on OpenTelemetry.SemanticConventions

ppittle:
@Oberon00 - I've addressed your feedback, could I trouble you to take another look?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
```diff
@@ -30,40 +33,70 @@ internal static IEnumerable<KeyValuePair<string, object>> GetHttpTags<TInput>(TI
...
+        tags.AddAttributeUrlPath(urlPath, addIfEmpty: true);
+
+        if (urlQuery != null)
+        {
+            tags.AddAttributeUrlQuery(urlQuery, addIfEmpty: true);
+        }
+
+        if (httpMethod != null)
+        {
+            tags.AddAttributeHttpMethod(httpMethod, addIfEmpty: true);
```
Oberon00, line 83: method vs request method is a matter of semantic convention version, so I would expect to have a single method (probably named after the newer semantic convention version) that does the right thing(s) depending on version.

There may be cases where the values differ depending on semconv version and then it makes sense to require different calls per version (and make some calls do nothing depending on semconv version) but does this apply to http method vs http request method?

ppittle: > There may be cases where the values differ depending on semconv version and then it makes sense to require different calls per version

You ask a good question.  I've been working to balance the maintainability of the `AWSSemanticConventions` class, so that's it easy to track how Semantic Convention changes over time and that the AddAttributeXXX methods continue to align with the actual attribute name; vs keeping the business logic class, like `AWSLambdaHttpUtils` free from needing to deal with supporting multiple Semantic Convention versions.

I originally implemented Http Method as you suggested, where `AWSSemanticConventions_v1_29_0` changed the name of `AttributeHttpMethod` to `"http.request.method"`.

However, **I decided to change course for any Attributes that are clearly marked Obsolete.**

As [`HttpAttributes.AttributeHttpMethod`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.SemanticConventions/Attributes/HttpAttributes.cs#L43-L44) is marked Obsolete in the spec, trying to adapt the `AddAttributeHttpMethod()` felt like it would cause more maintainability problems down the road vs marking that method obsolete and updating `AWSLambdaHttpUtils` to make an additional call to `AddAttributeHttpRequestMethod`.

A technical note: `AWSSemanticConventions.Add` does check if the `attributeName` is empty and does not add it.  So v1.27 will only output `"http.method"` and v1.29+ will only output `"http.request.method"`.  They're never both output.

Thoughts?

src/OpenTelemetry.Instrumentation.AWS/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -1,6 +1,12 @@
 #nullable enable
+OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion
+OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion.Latest = 0 -> OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion
+OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion.v1_29_0_Experimental = 2 -> OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion
+OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion.v1_27_0_Experimental = 1 -> OpenTelemetry.Instrumentation.AWS.SemanticConventionVersion
```
alanwest, line 5: Is there a reason you want to offer support for 1.27?

Given this instrumentation has not yet received a stable release, I think it would be permissible to update everything to to adhere to 1.29 conventions. Only once you declare the instrumentation stable do you need to pin it to a specific version by default.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
Oberon00: I don't think it's ideal, but probably good enough. Instrumentation developers need to know all the different versions they support and make sure to set all the relevant aliases/variants, otherwise the function will silently do nothing. For something only used locally within the AWS instrumentations I think it's acceptable, if we want to use that more widely, it would IMHO need reconsideration.

ppittle: @Oberon00 - I agree with both of your points.  For this PR, I think we should move forward.  Fortunately, everything is `internal` so we can tweak this implementation moving forward to best support maintainability.

src/Shared/AWS/AWSSemanticConventions.cs
ppittle: :)

src/OpenTelemetry.Instrumentation.AWS/.publicApi/PublicAPI.Unshipped.txt
ppittle: @alanwest - 

> Is there a reason you want to offer support for 1.27

You're right that we could bump to the latest as part of this PR as the library is not yet stable.  But I wanted to include at least 2 versions of Semantic Convention in this PR so there's a clear pattern established on how to add additional versions moving forward.

Additionally, there was some discussion ([link](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367#discussion_r1875510194)) where we decided to default to the Semantic Convention currently in use (~1.27) and then users can opt-in to selecting a higher version.  This way, there's less disruptions to end users.

Thoughts?

src/Shared/AWS/SemanticConventionVersion.cs
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// because this class is shared and public, each project that exposes it must
+// ensure that it has a unique namespace, otherwise users could face type collisions
+// if they import multiple projects.
+#if Instrumentation_AWSLambda
+namespace OpenTelemetry.Instrumentation.AWSLambda;
+#elif Instrumentation_AWS
+namespace OpenTelemetry.Instrumentation.AWS;
+#elif Resources_AWS
```
Kielek, line 11: Typically in C# alle markers for conditional compilation are `UPPERCASE`. Consider this for consistency.

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
+  provides a mechanism for developers to opt-in to newer versions of the
+  of the Open Telemetry Semantic Conventions. ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
+
```
Kielek, line 8:
```suggestion
* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
  provides a mechanism for developers to opt-in to newer versions of the
  of the Open Telemetry Semantic Conventions.
  ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
+  provides a mechanism for developers to opt-in to newer versions of the
+  of the Open Telemetry Semantic Conventions. ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```
Kielek, line 7:
```suggestion
* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
  provides a mechanism for developers to opt-in to newer versions of the
  of the Open Telemetry Semantic Conventions.
  ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```

src/OpenTelemetry.Resources.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Introduce `AWSResourceBuilderOptions.SemanticConventionVersion` which
+  provides a mechanism for developers to opt-in to newer versions of the
+  of the Open Telemetry Semantic Conventions. ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```
Kielek, line 7:
```suggestion
* Introduce `AWSResourceBuilderOptions.SemanticConventionVersion` which
  provides a mechanism for developers to opt-in to newer versions of the
  of the Open Telemetry Semantic Conventions.
  ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```

src/Shared/AWS/SemanticConventionVersion.cs
```diff
@@ -0,0 +1,90 @@
...
+namespace OpenTelemetry.Resources.AWS;
+#endif
+
+#pragma warning disable SA1300
```
Kielek, line 15: Consider removing this pragma and `v1_27_0_Experimental` to `V1_27_0_Experimental`. It is still readable. but does not violate C# namign convention.

I do not see good possibility for the underscores.

Kielek (commented):
Only technical comment from my side. For the details implementation I would relay on component owners judgement.

You can consider also documenting differences in sem. conv in README files.

src/Shared/AWS/SemanticConventionVersion.cs
ppittle: agreed - changed case.

src/Shared/AWS/SemanticConventionVersion.cs
ppittle: agreed - renamed enums and class names to match this convention.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
ppittle: @alanwest - 

Yes, the very first entry in `SemanticConventionsVersions` is a little odd.  I intended it to represent the current Semantic Conventions being used, which is really a mixture of many different versions.

Originally, `SemanticConventionsVersions` tracked the version of the OpenTelemetry.*.AWS libraries themselves, rather then a specific Semantic Convention, so this aspect made a bit more sense.  But after some [discussion](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367#discussion_r1873338621) we decided to track the Semantic Convention version instead.

**I'm very open to suggestions on how to tweak naming / documentation to make this clearer. **

I could also upgrade the library to use the full version of 1.27 - but I don' think that would help any users:  anyone who wants the latest convention should use `1.29` and anyone that wants backwards compatibility with the existing betas wouldn't have an option.

Perhaps we could rename `V1_27_0_Experimental` -> `Legacy_Experimental`.  But this feels like it might be more confusing then it's worth.

**Thoughts?**

alanwest: > Perhaps we could rename V1_27_0_Experimental -> Legacy_Experimental. But this feels like it might be more confusing then it's worth.

Yes, I was thinking the same kind of thing.. one idea is that you could release things as follows

**Next non-stable release**

```csharp
enum SemanticConventionVersions
{
    Legacy = -1, // This would reflect the telemetry as it is today and be the default while not stable
    Latest,
    V1_29_0_Experimental,
}
```

** Future release candidate / stable releases **

Drop the `Legacy` option. This is a breaking change, but that's ok. Folks who still depend on the legacy behavior can continue to use prior non-stable versions until they are ready to upgrade.

```csharp
enum SemanticConventionVersions
{
    Latest,
    V1_29_0_Experimental, // This would become the new default
}
```

We had a release plan very similar to this when we stabilized ASP.NET Core and HttpClient instrumentation. The last non-stable release defaulted to the legacy conventions, though you could opt in to the new HTTP conventions using the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable. Then, when we shipped the instrumentation stable, we removed the support for `OTEL_SEMCONV_STABILITY_OPT_IN` and defaulted to the new, stable HTTP conventions.

ppittle: @alanwest - I've made a number of changes based on your feedback.  It deviates slightly but I think captures the spirit of your suggestions:

### `Legacy`
I've dropped support for `Legacy` Semantic Conventions.  This made more sense when the design was for `SemanticConventionVersions` to track the version of the OpenTelemetry.*.AWS Libraries.  But now that I've switched to tracking the OpenTelemetry.SemanticConventions instead, `Legacy` is more complicated then it's worth.  

This addresses your concern of `1.27` not _really_ being `1.27`.  

And, as you pointed out, "Folks who still depend on the legacy behavior can continue to use prior non-stable versions until they are ready to upgrade."

### Experimental
During the SIG meeting discussion, I walked away that there was a strong desire amongst the quorum to communicate to users that this GA library was using Experimental conventions.  My opinion is it will be difficult to communicate to users that some Semantic Convention is stable, while others are experimental; it's safer to keep `Experimental` in the name up until all Conventions used by the OpenTelemetry.*.AWS Libraries are stable.

### Release Plan

Dropping `Legacy` simplifies things:

**_1.10.0-beta.next_**

```
enum SemanticConventionVersions
{
    Latest,
    V1_28_0_Expiremental,  // Default
    V1_29_0_Expiremental
}
```

_**1.10.0** // first prod release_

```
enum SemanticConventionVersions
{
    Latest,
    V1_28_0_Expiremental,  // Default
    V1_29_0_Expiremental
}
```

_**1.11.0** // vNext beta/prod_
```
enum SemanticConventionVersions
{
    Latest,
    V1_28_0_Expiremental,  // Default
    V1_29_0_Expiremental,
    v1_32_0_Expiormental // assuming 1.32 is the most recent Sem Conv come release time for 1.11
}
```

The library continues to default to `V1_28_0_Expiremental` to offer users stability, but continues to add support for (some) newer Semantic Conventions for users that wish to opt-in.

**Is this something you can support?**

srprash (approved):
Overall this makes sense to me. I would highly recommend checking with @AsakerMohd on how this may affect consumers like ADOT.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaHttpUtils.cs
alanwest: > During the SIG meeting discussion, I walked away that there was a strong desire amongst the quorum to communicate to users that this GA library was using Experimental conventions.

Communication is important, though I do not think adding `Experimental` to the names of the enum values help. If anything, it makes things more confusing. I think you can do everything you need to communication wise through documentation. I like the pattern this PR is proposing. If we were to apply it to other instrumentation in the contrib repo that I am a code owner for, I will strongly advocate not suffixing the name of versions with `Experimental`. I think the versions speak clearly enough for themselves.

In the SIG meeting, I emphasized the thing of primary importance to end users. It essentially boils down to: If you declare instrumentation stable then the telemetry that instrumentation produces must not change by default in a breaking way until a major version bump. From the standpoint of my support and review of this PR, this is the only lens I'm reviewing it through. I am not familiar with the AWS instrumentation to weigh in on any other aspect, so I'm relying on the approval and support of the code owners of these components. Though, personally, I support your approach. I think it meets the requirements of the specification with regards to telemetry stability.

As for your release plan, it's up to you which versions of the conventions you want to initially support. However, I'm still not sure I understand why you have chosen to support 1.28 and 1.29 to begin with. It seems to me the simplest thing would be to initially support only whatever the latest version is at the time of your stable release.

---

As I've thought about this PR a bit more, I want to call out one more additional consideration that came to mind. Here's a potential limitation of the approach you're taking:

Let's say you ship stable and default to the v1.28 of the conventions. Now imagine two things happen when v1.32 of the conventions are released:

1. A new attribute is introduced for HTTP spans. Adding span attributes is considered a non-breaking change, so it is allowed for instrumentation to implement this change without a major version bump.
2. The faas conventions evolve in breaking ways and are still not yet considered stable.

Let's say an end user wants the new HTTP attribute but is not yet ready for the breaking changes to the faas telemetry. They're essentially stuck because selecting `V1_32_0` 

I personally do not think this limitation is a showstopper, but I just wanted to call it out to make sure you agree this is a consequence of the approach you're taking and to bring more clarity to others reviewing this PR.

ppittle: **"Experimental"**
I think your argument about the Enum not including "Experimental" make sense.  Upon reflection, I think it best to defer to your experienced opinion in this matter.  I've updated the Enum to drop the "Experimental" suffix.

**Semantic Convention Granularity**
Per the call out of the SemanticConventionVersions not being fine grained enough to allow selecting the version of each Semantic Convention document, I did think through this when designing.  Because the AWS libraries use ~10 different documents, I felt offering fine grained control, while technically feasible, would be too overwhelming for users and it would be a better approach to stick with a single value in the Options class.  

If there was sufficient customer demand for better control, we could offer it in a future version.  

But in the end, hopefully this is all just a short term measure until Semantic Convention stabilizes.

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
+  provides a mechanism for developers to opt-in to newer versions of the
+  of the Open Telemetry Semantic Conventions.
+  ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```
alanwest, line 8:
```suggestion
* Introduce `AWSClientInstrumentationOptions.SemanticConventionVersion` which
  provides a mechanism for developers to opt-in to newer versions of the
  of the OpenTelemetry Semantic Conventions. Currently, you need to opt-in
  to these new conventions. In the upcoming stable release of this library,
  the new conventions will be enabled by default, and the conventions this library
  currently emit will no longer be supported.
  ([#2367](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2367))
```

alanwest (approved):
Discussed this PR at length with @ppittle today.

The plan is to release one more beta version of these instrumentations that supports the existing conventions they emit. By default the telemetry they emits does not change with this PR, but a user can now opt in to specific versions of the semantic conventions.

A release candidate will ship shortly after the last beta version. The release candidate will default to the new semantic conventions and remove support for the "legacy" conventions.

There are some active reviewers of this PR that have not had a opportunity to approve it yet (@Oberon00 and @Kielek). Since this PR should not be changing any behavior by default, I've decided to approve and merge it so that we can release the next beta version. @ppittle will open up the follow up PR in preparation of the release candidate soon. That should give folks the chance to weigh in with any additional feedback on this plan.

PR #2138, opened 2024-10-03
[Instrumentation.AspNetCore] Remove .NET 6 target

joegoldman2:
## Changes

Remove .NET 6 target which will be very soon out of support.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #2403, opened 2024-12-09
[aspnetcore] Restore metrics instrumentation in netstandard builds

alanwest:
#2360 needs further discussion. It removes metric functionality from a supported target, so it is a breaking change that warrants a major version bump. That said, I don't see any reason to have removed this functionality at all.

This change needs to be reverted. Also, it appears this was released as 1.10.0. A version 1.10.1 will need to be released to fix this breaking change.

Lastly, I noted that #2360 was merged without the approval of an approver or maintainer.

@Kielek @CodeBlanch

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationMeterProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationMeterProviderBuilderExtensions.cs
@@ -1,6 +1,10 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+#if !NET
+using OpenTelemetry.Instrumentation.AspNetCore;
+using OpenTelemetry.Instrumentation.AspNetCore.Implementation;
+#endif
 using OpenTelemetry.Internal;
 
 namespace OpenTelemetry.Metrics;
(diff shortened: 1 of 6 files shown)


Kielek:
There was a decision to fully drop .NET6/.NET7 support (without making. the major upgrade here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2138/files#diff-9fc8151d477e5c83d4231a1da766f08e0f087e880e42e19695175598e4069b4a).
I do not think that dropping code which is officially not supported is something bad. IMO, it is even better to do this now, instead of just after release. The message is clear for end-users. What is more, it is sync with https://github.com/open-telemetry/opentelemetry-dotnet/blob/145e7ade6f3d518280d139624951822117b60ebd/README.md#supported-net-versions

If we speaking about lack of technical approval, I assumed that @CodeBlanch is fine with changes, just after fixing the feedback. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2360#discussion_r1874367855. Maybe I was wrong, if so, I am sorry for this.

alanwest:
> I do not think that dropping code which is officially not supported is something bad. IMO, it is even better to do this now, instead of just after release.

[Our stance on .NET Standard](https://github.com/open-telemetry/opentelemetry-dotnet/discussions/3767) was discussed at length a couple years ago. While I think your position is a valid one, it is not the one we settled on. Until we decide to change that, and discuss how, we need to stick with our prior decisions.

Kielek (approved):

CodeBlanch (approved):
LGTM

PR #2422, opened 2024-12-15
[GenevaExporter] Add ability to export Log Exception using StackTrace

### LINKED ISSUE #2421: [Exporter.Geneva] Allow to export only stack trace in logs
### Component

OpenTelemetry.Exporter.Geneva

### Package Version

| OpenTelemetry.Api                     | 1.10.0   |
| OpenTelemetry                         | 1.10.0   |

### Runtime Version

net9.0

### Description

I am looking at the Geneva Exporter and there seems to be bug:

image

Exception.ToString() is not the stack trace.... it has a ton more than that...
Can we please fix and limit to the stack trace?
 
Please consider using Exception.StackTrace here which is also cheaper to generate.

### Steps to Reproduce

Log and exception and you see it is not the stack trace that gets exported.

### Expected Result

Just the stack trace to be exported for this property.

### Actual Result

Full Exception Message.

### Additional Context

_No response_

cijothomas:
## Changes
Adds a new option to use `StackTrace` property of Exception be used while exporting Logs.
Fixes #2421


src/OpenTelemetry.Exporter.Geneva/ExceptionStackExportMode.cs
```diff
@@ -14,9 +14,17 @@ public enum ExceptionStackExportMode
...
+    /// Though formatted in culture-agnostic way, this is more targetted towards human readability.
+    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.stacktrace"/>.
     /// </summary>
     ExportAsString,
 
+    /// <summary>
+    /// Exception stack traces are exported as string, using the StackTrace property of the Exception.
+    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.tostring"/>.
+    /// </summary>
+    ExportAsStackTraceString,
```
cijothomas, line 27: Alternate option is to do this as a privatepreviewfeature, in the connectionstring. But this seems a safe thing to do, and is opt-in too, so I think this is reasonable approach to take.

src/OpenTelemetry.Exporter.Geneva/ExceptionStackExportMode.cs
```diff
@@ -14,9 +14,17 @@ public enum ExceptionStackExportMode
...
     /// <summary>
-    /// Exception stack traces are exported as strings.
+    /// Exception stack traces are exported as string, using a culture agnostic representation of the results of ToString() implementation of the Exception.
+    /// Though formatted in culture-agnostic way, this is more targetted towards human readability.
+    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.stacktrace"/>.
     /// </summary>
     ExportAsString,
 
+    /// <summary>
+    /// Exception stack traces are exported as string, using the StackTrace property of the Exception.
```
cijothomas, line 24: need to mention this does not do anything with InnerException or AggregateExcetion etc..where ToString() typically covers them.

cijothomas: Also, need to confirm if this eliminates stacks for async/await machinery, which ToString() eliminates by default now.

src/OpenTelemetry.Exporter.Geneva/ExceptionStackExportMode.cs
```diff
@@ -14,9 +14,17 @@ public enum ExceptionStackExportMode
...
+    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.tostring"/>.
```
rajkumar-rangaraj, line 25:
```suggestion
    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.stacktrace"/>.
```

src/OpenTelemetry.Exporter.Geneva/ExceptionStackExportMode.cs
```diff
@@ -14,9 +14,17 @@ public enum ExceptionStackExportMode
     Drop,
 
     /// <summary>
-    /// Exception stack traces are exported as strings.
+    /// Exception stack traces are exported as string, using a culture agnostic representation of the results of ToString() implementation of the Exception.
+    /// Though formatted in culture-agnostic way, this is more targetted towards human readability.
+    /// See <see href="https://learn.microsoft.com/dotnet/api/system.exception.tostring"/>.
```
cijothomas, line 19: TODO: Mention that ToString() can be overridden by users, so this could return completely irrelevant information too!

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -417,22 +417,32 @@ internal ArraySegment<byte> SerializeLogRecord(LogRecord logRecord)
...
-                // keep top and bottom.
                 var exceptionStack = logRecord.Exception.ToInvariantString();
                 cursor = MessagePackSerializer.SerializeAsciiString(buffer, cursor, "env_ex_stack");
                 cursor = MessagePackSerializer.SerializeUnicodeString(buffer, cursor, exceptionStack);
                 cntFields += 1;
             }
+            else if (this.exportExceptionStack == ExceptionStackExportMode.ExportAsStackTraceString)
+            {
+                var exceptionStack = logRecord.Exception.StackTrace;
+                if (exceptionStack != null)
```
rajkumar-rangaraj, line 439: Is there a reason why we check here only for null, while on TLD we use `string.IsNullOrEmpty`?

src/OpenTelemetry.Exporter.Geneva/ExceptionStackExportMode.cs
cijothomas: same with stackTrace too.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
cijothomas: good catch! Don't think we need either checks and serialize them as-is, as already done.

rajkumar-rangaraj (approved):

CodeBlanch (approved):
LGTM

reyang:
I don't understand the direction here, if this is what we want to give to the user, what's the plan for other exporters (e.g. OTLP)?

cijothomas:
> I don't understand the direction here, if this is what we want to give to the user, what's the plan for other exporters (e.g. OTLP)?

Otel SDK has to offer mechanism to let user pick how they want stack trace represented.
`Activity` in runtime already offer mechanism to let user decide this. ILogger does not, but the change would likely be in Otel SDK/Exporter, as ILogger hands the `Exception` object itself to the providers.

reyang:
> > I don't understand the direction here, if this is what we want to give to the user, what's the plan for other exporters (e.g. OTLP)?
> 
> Otel SDK has to offer mechanism to let user pick how they want stack trace represented. `Activity` in runtime already offer mechanism to let user decide this. ILogger does not, but the change would likely be in Otel SDK/Exporter, as ILogger hands the `Exception` object itself to the providers.

Is there a tracking issue and design discussions? My concern is around lack of overall design/thinking. I don't see why we want to rush here.

cijothomas:
> > > I don't understand the direction here, if this is what we want to give to the user, what's the plan for other exporters (e.g. OTLP)?
> > 
> > 
> > Otel SDK has to offer mechanism to let user pick how they want stack trace represented. `Activity` in runtime already offer mechanism to let user decide this. ILogger does not, but the change would likely be in Otel SDK/Exporter, as ILogger hands the `Exception` object itself to the providers.
> 
> Is there a tracking issue and design discussions? My concern is around the lack of overall design/thinking. I don't see why we want to rush here.

Nothing open in Otel .NET repo, but spec has seen asks before on having the ability to control the strackTrace representation:
https://github.com/open-telemetry/opentelemetry-specification/issues/2839

PR #2429, opened 2024-12-18
[Instrumentation.AspNetCore] Listen to SignalR server activities

### LINKED ISSUE #1998: Consider 1st class support for SignalR tracing
### Component

OpenTelemetry.Instrumentation.AspNetCore

In .NET 9, SignalR has added telemetry support which uses the name `"Microsoft.AspNetCore.SignalR.Server"` on the server-side and `"Microsoft.AspNetCore.SignalR.Client"` on the client-side. It would be nice if `AddAspNetCoreInstrumentation` could enable the server-side tracing as SignalR is part of ASP.NET Core.

Calling `tracing.AddAspNetCoreInstrumentation();` sets up SignalR (at least server-side) tracing.

Right now you would need to call `tracing.AddSource("Microsoft.AspNetCore.SignalR.Server");`. It also looks like there is more logic based on specific activity names
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/88645a913eab3906c7e618961e1b16fbd3aedfce/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L21
Not sure what exactly is done there but there might be more work to be done for listening to SignalR activites too.

Alternatively, there could be a new `AddSignalRInstrumentation()` similar to other libraries like `AddGrpcCoreInstrumentation()`.

For the client-side, there might also want to be a separate method for it similar to `AddHttpClientInstrumentation`.

### Additional context

https://github.com/dotnet/aspnetcore/issues/51557 tracks the ASP.NET Core side for SignalR instrumentation and includes cross-linking to relevant PRs and other information.

BrennanConroy:
Fixes #1998
Design discussion issue #1998

## Changes

Adds activity source "Microsoft.AspNetCore.SignalR.Server" when calling `AddAspNetCoreInstrumentation`.
This activity source is new in .NET 9.0.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
@@ -121,5 +121,12 @@ public static class AspNetCoreInstrumentationTracerProviderBuilderExtensions
             builder.AddSource(HttpInListener.ActivitySourceName);
             builder.AddLegacySource(HttpInListener.ActivityOperationName); // for the activities created by AspNetCore
         }
+
+        // SignalR activities first added in .NET 9.0
+        if (Environment.Version.Major >= 9)
+        {
+            // https://github.com/dotnet/aspnetcore/blob/6ae3ea387b20f6497b82897d613e9b8a6e31d69c/src/SignalR/server/Core/src/Internal/SignalRServerActivitySource.cs#L13C35-L13C70
+            builder.AddSource("Microsoft.AspNetCore.SignalR.Server");
+        }
     }
 }
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
@@ -130,7 +130,7 @@ public static class AspNetCoreInstrumentationTracerProviderBuilderExtensions
         if (Environment.Version.Major >= 9)
         {
             var options = serviceProvider?.GetRequiredService<IOptionsMonitor<AspNetCoreTraceInstrumentationOptions>>().Get(optionsName);
-            if (options is null || !options.DisableAspNetCoreSignalRSupport)
+            if (options is null || options.EnableAspNetCoreSignalRSupport)
             {
                 // https://github.com/dotnet/aspnetcore/blob/6ae3ea387b20f6497b82897d613e9b8a6e31d69c/src/SignalR/server/Core/src/Internal/SignalRServerActivitySource.cs#L13C35-L13C70
                 builder.AddSource("Microsoft.AspNetCore.SignalR.Server");
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
@@ -68,7 +68,7 @@ public static class AspNetCoreInstrumentationTracerProviderBuilderExtensions
         {
             deferredTracerProviderBuilder.Configure((sp, builder) =>
             {
-                AddAspNetCoreInstrumentationSources(builder, sp);
+                AddAspNetCoreInstrumentationSources(builder, name, sp);
             });
         }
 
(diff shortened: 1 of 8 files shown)


src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreInstrumentationTracerProviderBuilderExtensions.cs
```diff
@@ -121,5 +121,12 @@ private static void AddAspNetCoreInstrumentationSources(
             builder.AddSource(HttpInListener.ActivitySourceName);
             builder.AddLegacySource(HttpInListener.ActivityOperationName); // for the activities created by AspNetCore
         }
+
+        // SignalR activities first added in .NET 9.0
+        if (Environment.Version.Major >= 9)
+        {
+            // https://github.com/dotnet/aspnetcore/blob/6ae3ea387b20f6497b82897d613e9b8a6e31d69c/src/SignalR/server/Core/src/Internal/SignalRServerActivitySource.cs#L13C35-L13C70
+            builder.AddSource("Microsoft.AspNetCore.SignalR.Server");
```
cijothomas, line 129: are these long (as in, really long) running server spans? I recall ApplicationInsights users complained about recording such long running spans, as that'd skew the latency calculations, and ended up ignoring those spans by default... 

https://github.com/dotnet/aspnetcore/pull/32084

cijothomas: If this is the case, then we should not do this by default, as there is no API to "remove" a source once added to the provider....

BrennanConroy: The only long running span should be the http one that has the long running tag. All of the added SignalR spans are only for a specific operation.
{4852D1D1-3D32-4C73-8B65-823B1B13C9B6}

Most of the spans added by SignalR add the long running span as a link, but individually they are short lived operations.
{84468919-B48C-4E8C-9E4B-4793E39D5756}
The `OnConnectedAsync` and `OnDisconnectedAsync` spans are the only 2 that are parented under the long running span, but they are both very short lived spans themselves.

rajkumar-rangaraj: A few years back, I collaborated with your team to introduce the `http.long_running` attribute in SignalR to help filter long-running HTTP requests in Application Insights .NET SDK. This effort was driven by significant customer feedback about long-running SignalR requests distorting request duration metrics.

With `http.long_running` part of the default ASP.NET Core instrumentation, customers might face similar issues if long-running spans are automatically included without a suppression mechanism. I recommend considering an option to filter these spans by default or providing a toggle to enable them explicitly before rolling this out.

BrennanConroy: Added an opt-out, not sure how this repo likes to do things like this, so might be completely wrong.

Also, the option as-is, will show up for all TFMs, even though it's only used for .NET 9.0+

BrennanConroy: @rajkumar-rangaraj @cijothomas thoughts?

BrennanConroy: Ok, how does one do that.

rajkumar-rangaraj: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AspNetCore#filter

src/OpenTelemetry.Instrumentation.AspNetCore/AspNetCoreTraceInstrumentationOptions.cs
```diff
@@ -93,6 +93,11 @@ internal AspNetCoreTraceInstrumentationOptions(IConfiguration configuration)
     /// </remarks>
     public bool RecordException { get; set; }
 
+    /// <summary>
+    /// Gets or sets a value indicating whether SignalR activities are recorded.
+    /// </summary>
+    public bool DisableAspNetCoreSignalRSupport { get; set; }
```
noahfalk, line 99: Naming suggestion:
```suggestion
    public bool EnableAspNetCoreSignalRSupport { get; set; }
```
A couple thoughts in there:
1. Naming the property Enable rather than Disable avoids reasoning with double-negatives when the value is false.
2. It matches the pattern for EnableGrpcAspNetCoreSupport below.

PR #2437, opened 2024-12-20
[aws] default to use Semantic Convention 1.28 and remove option Legacy

ppittle:
Fixes #
Design discussion issue #  Continuation of #2367

## Changes

Update the OpenTelemetry.*.AWS libraries to to use Semantic Convention 1.28 and remove the option `SemanticConventionVerisons.Legacy`

## Background

We determined that #2367, should not change Semantic Conventions by default, allowing users a non-breaking way to test opting in-to 1.28.

This is a fast follow up PR that removes `SemanticConventionVersions.Legacy` and changes the default to SemanticConventionVersions.V1_28_0. This PR is to be released as a Release Candidate.

This follows the release strategy for other contrib packages that have made similar breaking changes around Semantic Convention stabilization.

### DIFF
--- a/src/OpenTelemetry.Resources.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.AWS/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* BREAKING: Change default Semantic Convention to 1.28
+
 ## 1.10.0-beta.1
 
 Released 2024-Dec-20
--- a/src/OpenTelemetry.Resources.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.AWS/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * BREAKING: Change default Semantic Convention to 1.28
+* BREAKING: Remove option to use Legacy semantic conventions (the old default)
 
 ## 1.10.0-beta.1
 
--- a/src/Shared/AWS/AWSSemanticConventions.cs
+++ b/src/Shared/AWS/AWSSemanticConventions.cs
@@ -19,9 +19,6 @@ using OpenTelemetry.Resources.AWS;
 #pragma warning disable SA1201
 #pragma warning disable SA1623
 
-// disable obsolete warning - remove in RC
-#pragma warning disable CS0618
-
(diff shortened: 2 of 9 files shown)


alanwest (approved):

rajkumar-rangaraj (approved):

ppittle:
@Oberon00 - this PR come out of a discussion you and @alanwest and I were having in #2367.

Would you mind reviewing?

muhammad-othman (approved):

src/OpenTelemetry.Resources.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* BREAKING: Change default Semantic Convention to 1.28
```
Oberon00, line 5:
```suggestion
* BREAKING: Change default Semantic Convention to 1.28
* BREAKING: Remove option to use Legacy semantic conventions (the old default)
```

The second point is also important to mention.

src/Shared/AWS/AWSSemanticConventions.cs
```diff
@@ -494,9 +488,6 @@ private AWSSemanticConventionsBase GetSemanticConventionVersion()
             case SemanticConventionVersion.V1_28_0:
                 return new AWSSemanticConventions_V1_28_0();
 
-            case SemanticConventionVersion.Legacy:
-                return new AWSSemanticConventionsLegacy();
```
Oberon00, line 498: Shouldn't the implementation class AWSSemanticConventionsLegacy also be removed? Is it still needed somewhere?

ppittle: The AWSSemenaticConventions classes use inheritance so they only need to capture the diff between versions.  So 1.28 inherits from Legacy and 1.29 inherits from 1.28.

```csharp
private class AWSSemanticConventions_V1_28_0 : AWSSemanticConventionsLegacy
```
and 
```csharp
private class AWSSemanticConventions_V1_29_0 : AWSSemanticConventions_V1_28_0
```

But I did mark `AWSSemanticConventionsLegacy` as `asbtract` to further highlight that it shouldn't be used directly.

PR #2439, opened 2024-12-20
[Instrumentation.EntityFrameworkCore] server.address attribute to follow sem. conv.

### LINKED ISSUE #2438: [bug] EntityFrameworkCore spans are not populating server.address following the otel conventions
### Component

OpenTelemetry.Instrumentation.EntityFrameworkCore

### Package Version

| OpenTelemetry.Instrumentation.EntityFrameworkCore                     |   1.10.0-beta.1  |

### Runtime Version

net8.0,net9.0, any

### Description

As reported in the OTEL docs, `server.address` should be populated with the server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. i.e.  `example.com`, `10.1.2.80`, `/tmp/my.sock`.
Meanwhile, the latest version of **OpenTelemetry.Instrumentation.EntityFrameworkCore** populate the field prefixing it with the connection protocol `tcp`.

https://opentelemetry.io/docs/specs/semconv/attributes-registry/server/#server-attributes

### Steps to Reproduce

Configure the instrumentation and record some traces

### Expected Result

`server.address` to be populated as `my.domain.example` 

### Actual Result

`server.address` to populated as `tcp:my.domain.example` 

### Additional Context

_No response_

overbit:
Fixes #2438

Update the `server.address` field in EntityFrameworkCore spans to populate with the server domain name without the `tcp` prefix.

## Changes

* Modify `src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs` to fetch the `host` and `port` properties from the `connection` object and set them to the `server.address` field, with a fallback to use the `dataSource` property if the `host` property is not available.
* Update `src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md` to reflect the changes made to the `server.address` field.
* Add tests in `src/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs` to verify the `server.address` field is populated correctly without the `tcp` prefix.


src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -140,10 +139,24 @@ public override void OnEventWritten(string name, object? payload)
...
+                            {
+                                activity.AddTag(AttributeServerAddress, dataSource);
+                            }
+                        }
+
+                        var port = (string)this.portFetcher.Fetch(connection);
+                        if (!string.IsNullOrEmpty(port))
                         {
-                            activity.AddTag(AttributeServerAddress, dataSource);
+                            activity.AddTag(AttributeServerAddress, port);
```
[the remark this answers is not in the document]
overbit: good spot! Fixed

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -1,6 +1,3 @@
-// Copyright The OpenTelemetry Authors
-// SPDX-License-Identifier: Apache-2.0
-
```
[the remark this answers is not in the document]
overbit: fixed

overbit:
@Kielek let me know if there is anything else

overbit:
> > @Kielek let me know if there is anything else
> 
> There is still couple of errors in CI pipeline.

can I get some support with this https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/12784788454/job/35638594583?pr=2439#step:7:79 ? 
I struggle to see the correlation with the actual change.

Kielek:
@overbit, sorry for delay.

When you are executing tests, after your changes the exception occurrs
```
This exception was originally thrown at this call stack:
    OpenTelemetry.Instrumentation.PropertyFetcher<T>.Fetch(object) in PropertyFetcher.cs
    OpenTelemetry.Instrumentation.EntityFrameworkCore.Implementation.EntityFrameworkDiagnosticListener.OnEventWritten(string, object) in EntityFrameworkDiagnosticListener.cs
    OpenTelemetry.Instrumentation.DiagnosticSourceListener.OnNext(System.Collections.Generic.KeyValuePair<string, object>) in DiagnosticSourceListener.cs
```

It means that the code related to setting db name is never executed.

overbit:
@Kielek fixed. Thanks for the help

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics;
+using Xunit;
+
+namespace OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests;
+
+public class EntityFrameworkInstrumentationTests
```
Kielek, line 9: This test class is testing only itself. There is no reference to the _production_ code.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -40,6 +41,8 @@ internal sealed class EntityFrameworkDiagnosticListener : ListenerHandler
     private readonly PropertyFetcher<CommandType> commandTypeFetcher = new("CommandType");
     private readonly PropertyFetcher<string> commandTextFetcher = new("CommandText");
     private readonly PropertyFetcher<Exception> exceptionFetcher = new("Exception");
+    private readonly PropertyFetcher<string> hostFetcher = new("Host");
+    private readonly PropertyFetcher<string> portFetcher = new("Port");
```
Kielek, line 45: https://github.com/open-telemetry/semantic-conventions/blob/v1.30.0/docs/database/database-spans.md

port should be int not string.

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    public void ServerAddressWithoutProtocolPrefix()
+    {
+        var activity = new Activity("TestActivity");
+        activity.Start();
+
+        var connection = new
+        {
+            Host = "my.domain.example",
+            DataSource = "tcp:my.domain.example",
+            Port = "5432",
```
Kielek, line 21: As I understand, you are DuckTyping some class here. What is your exact type?

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fixed the `server.address` field in EntityFrameworkCore spans to populate
+  with the server domain name without the `tcp` prefix.
+  ([#2439](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2439))
```
Kielek, line 7: Worth to mention also that `server.port` is also introduced.

Kielek (commented):
Sorry for lack of response for a long time. I have a lot of another tasks.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
overbit: Good point.

overbit: done

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs
overbit: removed

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs
Kielek: This question is still valid

Kielek:
@overbit, do you have any updates here? I would like to help with merging, but I need some more information.

test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkInstrumentationTests.cs
overbit: This file was removed

overbit:
> You were trying to fetch Port as a string. Now you have changed it to the int. What is the real class which is source of the information. What is the contract there? Int/string/long/something other?

The port if define as number, integer due to TCP / UDP protocol providing a 16bits for the port number.
The change was tested with npgsql library and the type can be checked here:
https://github.com/npgsql/npgsql/blob/main/src/Npgsql/NpgsqlConnection.cs#L424 
@Kielek

overbit:
@Kielek any updates?

Kielek:
@overbit, sorry for long waiting for the response.
Topic is not so simple.

In general, we should be able to handle most/all EntifyFrameworkCore providers. It means that we need to support data sources most common for all packages. As you mentioned, your solution is very specific for Npgsql package.

Do you see any possibility to make a exercise and test code/provide comprehensive solution for [all packages](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L89-L135)?

Ideally, it should realys on data in [DbConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbconnection?view=net-9.0)/[IDbConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.idbconnection?view=net-9.0).

It will be great if you can find generic source of the needed properties.
If data source for attributes is not common, we need a list of supported packages.

Kind of table with summary can be helpful.

Please let me know if my message is clear enough.

overbit:
> @overbit, sorry for long waiting for the response. Topic is not so simple.
> 
> In general, we should be able to handle most/all EntifyFrameworkCore providers. It means that we need to support data sources most common for all packages. As you mentioned, your solution is very specific for Npgsql package.
> 
> Do you see any possibility to make a exercise and test code/provide comprehensive solution for [all packages](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L89-L135)?
> 
> Ideally, it should realys on data in [DbConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbconnection?view=net-9.0)/[IDbConnection](https://learn.microsoft.com/en-us/dotnet/api/system.data.idbconnection?view=net-9.0).
> 
> It will be great if you can find generic source of the needed properties. If data source for attributes is not common, we need a list of supported packages.
> 
> Kind of table with summary can be helpful.
> 
> Please let me know if my message is clear enough.

@Kielek it is a tricky topic indeed. I agree that we should always rely on IDbConnection but unfortunately, the OTEL specification around this topic doesn't overlap with that as we can only read ConnectionString and Database that are both strings. I would also challenge that we should now rely on DbConnection properties as libraries that extend Ef Core could just implement the interface and not extend the DbConnection class. 

In this scenario, unfortunately, it's not possible to define a final table of support as some of the library supported are close source.

We face now 2 options:
- close the current PR and live with the mismatch. This will cause confusion in users that rely on OTEL for many services in different languages, i.e. NodeJs app and .NET app, as they will see 2 different resources in their service map (one prefixed with the protocol and one that follows the specification)
- add an additional documentation piece / xml comment that clarify that only libraries that extend DbConnection with Server and Port properties will be supported

WDYT?

Kielek:
@overbit, one more idea.
The set of well known EFC Providers is big, but still limited. See the way how `db.system` is calculated.
Maybe we can do similar logic for all/parf of the other attributes. Including server.address and server.port.

There is a chance that some attributes will be not available in part of the providers.

There is a couple of problems here:
1. Testing effort, initial and creating CI tests. Potentially requires integration tests.
2. Solution needs to be performant, needs to be considered when creating _resolvers solution._

Issue #2356, opened 2024-12-03
[feature request] Make Kafka instrumentation more amenable to unit testing


IGx89:
### Component

OpenTelemetry.Instrumentation.ConfluentKafka

### Is your feature request related to a problem?

I want to be able to unit test my application logic using mock IConsumer's instead of only being able to do E2E testing against a live Kafka instance.

I have code using the `IConsumer.ConsumeAndProcessMessageAsync` extension method and want to unit test it. I'm injecting an IConsumer mock into the DI container to facilitate that. That extension method throws an exception if consumer is not InstrumentedConsumer<TKey, TValue>, an internal class, preventing me from unit testing that code.

### What is the expected behavior?

There are multiple possible solutions, including making InstrumentedConsumer public or removing the unnecessary `consumer is not InstrumentedConsumer<TKey, TValue>` check from the extension method.

### Which alternative solutions or features have you considered?

Writing E2E tests (much slower and more complicated) or not testing my app logic, neither of which is appealing.

IGx89:
I found another workaround that involves using reflection to set the "consumer" field on InstrumentedConsumer to my mock consumer. That unblocks me for now, but still would be nice to have a non-hacky solution.

Issue #1935, opened 2024-07-01
Add instructions to configure ConfluentKafka instrumentation


g7ed6e:
It would be good to link the example from here. or have some basic examples on how to configure the instrumentation. (Not a blocker - can be a follow up PR)

_Originally posted by @vishweshbankwar in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1493#discussion_r1659150221_

t03apt:
@g7ed6e 
I'd love to use this package. I am trying to use the InstrumentedProducerBuilder in my project but I see no way to extend it with my additional ProducerBuilder methods.
```csharp
using Confluent.Kafka;

public static class ProducerBuilderExtensions
{
    public static ProducerBuilder<TKey, TValue> UseDefaultErrorHandler<TKey, TValue>(this ProducerBuilder<TKey, TValue> builder, ILogger logger)
    {
        ArgumentNullException.ThrowIfNull(builder);
        ArgumentNullException.ThrowIfNull(logger);

        builder.SetErrorHandler((self, e) =>
        {
            // TODO: figure out how periodic Local_AllBrokersDown could be avoided via configuration
            var logLevel = e.Code == ErrorCode.Local_AllBrokersDown ? LogLevel.Debug : LogLevel.Error;
            logger.Log(logLevel, "Kafka producer error: ErrorCode:{ErrorCode}, Reason:{Reason}, Name:{Name}", e.Code, e.Reason, self.Name);
        });
        return builder;
    }

    public static ProducerBuilder<TKey, TValue> UseDefaultLogHandler<TKey, TValue>(this ProducerBuilder<TKey, TValue> builder, ILogger logger)
    {
        ArgumentNullException.ThrowIfNull(builder);
        ArgumentNullException.ThrowIfNull(logger);

        builder.SetLogHandler((_, log) => log.LogWithLogger("Kafka producer", logger));
        return builder;
    }
}
```

I could have my own `ProducerBuilder` implementation and create the InstrumentedProducer, but it is internal. Could you make both the InstrumentedProducer and InstrumentedConsumer public?

t03apt:
I just realized that the InstrumentedProducerBuilder derives from the Confluent ProducerBuilder and has all the methods I need. Please ignore my previous comment.

t03apt:
[...]

PR #2568, opened 2025-02-15
Kafka updates

### LINKED ISSUE #2567: [feature request] Kafka instrumentation should follow the Semantic Conventions
### Component

None

Since the Kafka instrumentation was created prior to the current conventions (marked as Development right now), it's not following the conventions. We should attempt to get this implementation up to the right level.

***note:** I have a PR for this for review*

All conventions are followed as closely as possible

None

### Additional context

I have a PR that is quite large for addressing this

martinjt:
Fixes #2567, #2356, and #1935

This is a rather larger update than I first thought, and it's not done yet so this is more for feedback on the approach.

A large part of the changes are around changing the interface such that the presence of a listener is what we need for enabling metrics and tracing, and the fact that we've used an instrumented producer/consumer means that propagation should happen.

Further to that, I've moved to using Activity's that are created and tracked, over ones that are created at the end with a start/finish. This presented the issue around the idea that when we start an activity, we don't know the topic, and if the receive action fails, we should still create the activity.

I believe this is more inline with what the Semantic Conventions that are in development are looking for.

/cc @g7ed6e @lmolkova


src/OpenTelemetry.Instrumentation.ConfluentKafka/OpenTelemetryConsumeResultExtensions.cs
```diff
@@ -15,65 +16,36 @@ namespace Confluent.Kafka;
...
+    /// <param name="startActivityAsLink">true if you want the <see cref="System.Diagnostics.Activity" /> to be started as a Link rather than a child of the message's parent</param>
     /// <typeparam name="TKey">The type of key of the <see cref="ConsumeResult{TKey,TValue}"/>.</typeparam>
     /// <typeparam name="TValue">The type of value of the <see cref="ConsumeResult{TKey,TValue}"/>.</typeparam>
     /// <returns>A <see cref="ValueTask"/>.</returns>
     public static ValueTask<ConsumeResult<TKey, TValue>?> ConsumeAndProcessMessageAsync<TKey, TValue>(
         this IConsumer<TKey, TValue> consumer,
-        OpenTelemetryConsumeAndProcessMessageHandler<TKey, TValue> handler) =>
-        ConsumeAndProcessMessageAsync(consumer, handler, default);
+        OpenTelemetryConsumeAndProcessMessageHandler<TKey, TValue> handler,
+        bool startActivityAsLink = false) =>
```
matt-hensley, line 31: The resque instrumentation has [a convenient `propagation_style` parameter](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-resque/v0.6.0/instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb#L28) with values of link, child, or none. Switching `startActivityAsLink` from bool to an enum would allow for a similar set of options.

martinjt: I'm not sure about none here. If you're using the instrumented version (which is a choice), you're saying you want traces to be propagated.

I don't hate the enum though. If this PR becomes the approach I'll look at changing it.

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
lmolkova, line 1: which version of semconv are you targeting? 1.24.0 ?  could we document it somewhere - maybe just a comment somewhere?

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -308,70 +293,59 @@ private static void GetTags(string topic, out TagList tags, int? partition = nul
...
+        if (activity == null || consumeResult == null)
+        {
+            return;
+        }
+
+        if (!consumeResult.IsPartitionEOF)
+        {
+            activity.SetTag(SemanticConventions.AttributeMessagingKafkaDestinationPartition, consumeResult.TopicPartition.Partition.Value);
+            activity.SetTag(SemanticConventions.AttributeMessagingDestinationName, consumeResult.Topic);
+            activity.SetTag(SemanticConventions.AttributeMessagingKafkaMessageOffset, consumeResult.Offset.Value);
```
lmolkova, line 307: should we add links to received messages here?

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -308,70 +293,59 @@ private static void GetTags(string topic, out TagList tags, int? partition = nul
...
     {
+        if (!ConfluentKafkaCommon.ReceiveMessagesCounter.Enabled &&
+            !ConfluentKafkaCommon.ReceiveDurationHistogram.Enabled)
+        {
+            return;
+        }
+
         GetTags(topicPartition.Topic, out var tags, partition: topicPartition.Partition, errorType);
 
         ConfluentKafkaCommon.ReceiveMessagesCounter.Add(1, in tags);
```
lmolkova, line 321: is it always 1? If so, there is there a point in reporting it - it's derived from duration

> This metric is [required](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/metrics/metric-requirement-level.md#required) for batch receive operations. It's [opt-in](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.21.0/specification/metrics/metric-requirement-level.md#opt-in) when the messaging system does not support batch receive since the message count can be derived from the messaging.receive.duration histogram.

https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/messaging/messaging-metrics.md#metric-messagingreceivemessages

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -108,23 +89,27 @@ public void SetSaslCredentials(string username, string password)
...
             result = this.consumer.Consume(timeout);
             consumeResult = ExtractConsumeResult(result);
+            UpdateActivityWithConsumeResults(activity, result);
             return result;
         }
         catch (ConsumeException e)
         {
             (consumeResult, errorType) = ExtractConsumeResult(e);
+            activity?.AddException(e);
+            activity?.SetStatus(ActivityStatusCode.Error);
```
lmolkova, line 104: pls add `error.type` to activity and exception message to status description

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -308,70 +293,59 @@ private static void GetTags(string topic, out TagList tags, int? partition = nul
...
+
+        if (!consumeResult.IsPartitionEOF)
+        {
+            activity.SetTag(SemanticConventions.AttributeMessagingKafkaDestinationPartition, consumeResult.TopicPartition.Partition.Value);
+            activity.SetTag(SemanticConventions.AttributeMessagingDestinationName, consumeResult.Topic);
+            activity.SetTag(SemanticConventions.AttributeMessagingKafkaMessageOffset, consumeResult.Offset.Value);
+        }
+    }
+
     private static void RecordReceive(TopicPartition topicPartition, TimeSpan duration, string? errorType = null)
```
lmolkova, line 311: nit, maybe `RecordReceiveMetrics` for better readability ?

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -308,70 +293,59 @@ private static void GetTags(string topic, out TagList tags, int? partition = nul
...
+        }
+
         GetTags(topicPartition.Topic, out var tags, partition: topicPartition.Partition, errorType);
 
         ConfluentKafkaCommon.ReceiveMessagesCounter.Add(1, in tags);
         ConfluentKafkaCommon.ReceiveDurationHistogram.Record(duration.TotalSeconds, in tags);
     }
 
-    private void InstrumentConsumption(DateTimeOffset startTime, DateTimeOffset endTime, ConsumeResult consumeResult, string? errorType)
+    private static void InstrumentConsumption(DateTimeOffset startTime, DateTimeOffset endTime, ConsumeResult consumeResult, string? errorType)
```
lmolkova, line 325: nit (and very subjective) - is this method necessary at all now?

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -48,56 +48,37 @@ public void SetSaslCredentials(string username, string password)
...
             result = this.consumer.Consume(cancellationToken);
             consumeResult = ExtractConsumeResult(result);
+            UpdateActivityWithConsumeResults(activity, result);
             return result;
         }
         catch (ConsumeException e)
         {
             (consumeResult, errorType) = ExtractConsumeResult(e);
+            activity?.AddException(e);
+            activity?.SetStatus(ActivityStatusCode.Error);
```
lmolkova, line 72: a common complaint on Azure messaging that might be applicable here is that we record timeout as error, while all it's saying that someone polled and there were no messages on the topic to consume (which is absolutely fine).

do you know how it manifests on Kafka .NET? Would we record this as error?

PR #2653, opened 2025-03-10
Set initial AWS Semantic Convention list capacity to guard against list resizing

ppittle:
Fixes # #2652 
Design discussion issue #

## Changes

Set initial AWS Semantic Convention list capacity to guard against list resizing.  This should have a very small performance improvement and help in concurrent unit tests.

Note: this code is not originally intended to be thread safe:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/f4dff45c5fdde913e192f07ce37b189ae7d56a45/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaWrapper.cs#L168-L169

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [x] Changes in public API reviewed (if applicable)

PR #2658, opened 2025-03-13
Publish NuGet packages to Artifacts

martincostello:
See discussion from here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2653#issuecomment-2713844502

## Changes

Publish NuGet packages from CI builds as workflow run artifacts.

### DIFF
--- a/.github/workflows/Component.BuildTest.yml
+++ b/.github/workflows/Component.BuildTest.yml
@@ -140,3 +140,9 @@ jobs:
         flags: ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }}
         name: Code Coverage for ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }} on [${{ matrix.os }}.${{ matrix.version }}]
         codecov_yml_path: .github/codecov.yml
+
+    - name: Publish NuGet packages to Artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ inputs.project-name }}-packages
+        path: 'src\**\*.*nupkg'


.github/workflows/Component.BuildTest.yml
```diff
@@ -140,3 +140,9 @@ jobs:
         flags: ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }}
         name: Code Coverage for ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }} on [${{ matrix.os }}.${{ matrix.version }}]
         codecov_yml_path: .github/codecov.yml
+
+    - name: Publish NuGet packages to Artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{ inputs.project-name }}-packages
+        path: 'src\**\*.*nupkg'
```
martincostello, line 148:
```suggestion
        path: '.\src\**\*.*nupkg'
```

.github/workflows/Component.BuildTest.yml
```diff
@@ -140,3 +140,11 @@ jobs:
...
         name: Code Coverage for ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }} on [${{ matrix.os }}.${{ matrix.version }}]
         codecov_yml_path: .github/codecov.yml
+
+    - name: Publish NuGet packages to Artifacts
+      uses: actions/upload-artifact@v4
+      if: ${{ matrix.os == 'windows-latest' && inputs.pack }}
+      with:
+        name: ${{ inputs.project-name }}${{ matrix.version }}--packages
+        path: '.\src\**\*.*nupkg'
+        if-no-files-found: error
```
martincostello, line 150:
```suggestion
        name: ${{ inputs.project-name }}-${{ matrix.version }}-packages
        path: '.\src\**\*.*nupkg'
```

TimothyMothra:
I'm fine with publishing artifacts, but I'm concerned about the storage size. I'm not sure about the max size or if open-telemetry gets billed for storage.

Given that workflows are run multiple times daily, I think we need to set a retention period for artifacts. I think 1 week would be sufficient for anyone actively working on changes. (PRs are automatically closed after 1 week of inactivity).
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#configuring-a-custom-artifact-retention-period

martincostello:
AFAIK for public repos there's no cost for artifact storage (or if there's a free limit, I'm not aware of anything hitting it).

trask:
I haven't heard of workflow artifact cost being an issue, and [the docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) say

> Storing artifacts uses storage space on GitHub. GitHub Actions usage is free for standard GitHub-hosted runners in public repositories.

so I'd suggest go for it and don't worry about setting retention period unless you want to do that for other reasons besides cost

martincostello:
Pending answers to questions posed in description.

Kielek:
> * Every Windows TFM produces packages - should that be narrowed to a specific job, or is the duplication (and need to have specifically named artifacts) ok?

It is duplication. The reason for packaging on Windows is the support for .NET Framework. The good default can be publish from net462 target with possibility to override setting for packages without this target (AspNetCore, EF Core).

> * Two projects don't produce packages (contrib-shared-tests and semanticconventions), so an earlier commit that raised an error if there was nothing to upload broke the build for these two. Should these be changed to specify `pack: false` so that can be restored?

There is nothing to publish in contrib-shared-tests. It covers tests for shared, internal classes. OpenTelemetry.SemanticConvention should be packagable to the nuget. If it is not, it should be fixed earlier.

martincostello:
I get that packaging needs to happen on Windows at least once, but do you need the same packages three times?

martincostello:
So I should change the existing workflow to not do that? It already does that, it's just now more apparent if each job also makes the packages available as artifacts.

Kielek:
> So I should change the existing workflow to not do that? It already does that, it's just now more apparent if each job also makes the packages available as artifacts.

I think that it will be great if we reduce amount of produced useless artifacts.

martincostello:
I'll work on this next week.

martincostello:
Opened #2680 to deal with `dotnet pack` being run for projects that don't produce anything.

.github/workflows/Component.BuildTest.yml
```diff
@@ -140,3 +150,12 @@ jobs:
...
         name: Code Coverage for ${{ inputs.code-cov-prefix }}-${{ inputs.code-cov-name }} on [${{ matrix.os }}.${{ matrix.version }}]
         codecov_yml_path: .github/codecov.yml
+
+    - name: Publish ${{ steps.resolve-project.outputs.name }} NuGet packages to Artifacts
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      # Only publish packages from the first job, which should be net462 for Windows in most cases, which is preferred for .NET Framework support
+      if: ${{ matrix.os == 'windows-latest' && inputs.pack && strategy.job-index == 0 }}
+      with:
+        name: ${{ steps.resolve-project.outputs.name }}-packages
+        path: '.\src\**\*.*nupkg'
```
Kielek, line 160: It it smart way to publish all necessary files, but it leads to some deep structure of the archive, especially if you have more nuget packages files in the archive, it can be annoying to flat the results.

Do you see any easy option to convert
`OpenTelemetry.Instrumentation.AWS-packages.zip\OpenTelemetry.Extensions.AWS\bin\Release`

and remove at least `bin\Release` from the paths?

Kielek (approved):

Kielek:
@CodeBlanch, could you please check?

.github/workflows/Component.BuildTest.yml
martincostello: If it were my project, I would use [Artifacts output layout](https://learn.microsoft.com/dotnet/core/sdk/artifacts-output) which was added in .NET 8, then every NuGet package would be placed in the same single `./artifacts/package/release` directory, so the entire directory could be uploaded as-is and it would be a flat structure.

This is what we do in [Polly](https://github.com/App-vNext/Polly/blob/main/.github/workflows/build.yml#L119) and [Swashbuckle.AspNetCore](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/22c8c36374bffbcdc5867fa92375622bf31c0a7d/.github/workflows/build.yml#L118) (as an example, you can see the artifacts [here](https://github.com/domaindrivendev/Swashbuckle.AspNetCore/actions/runs/14294202225) which contain 19 packages each).

That's a much bigger change though, as this would affect your _entire_ CI/CD pipeline to adopt as it's essentially a global setting that affects build, pack, test and publish (which is why I haven't previously suggested it).

I think that's a far better way that writing custom code to glob all the packages, move them, then upload from there. But, it's probably best done by someone who understands your entire CI/CD process as an enhancement to your engineering setup (and in the non-contrib repo).

TimothyMothra (approved):
LGTM.

I'm also fine making iterative changes as needed.

rajkumar-rangaraj (approved):

.github/workflows/Component.BuildTest.yml
martincostello: #2816

PR #2674, opened 2025-03-30
[Bugfix] Fix IPv6 literal parsing from connection string (#2673)

### LINKED ISSUE #2673: [bug] SqlClient Instrumentation does not correctly report IPv6 literal address
### Component

### Package Version

| OpenTelemetry.Instrumentation.SqlClient | 1.11.0-beta.2    |

### Runtime Version

net8.0

### Description

The server details are not correctly reported when using an IPv6 literal address. For example the localhost address "[::1]" is reported as "1]".

### Steps to Reproduce

1. Ensure SQL server is running on IPv6, e.g. on localhost.
2. Configure a connection string to use the IPv6 literal, e.g. "server=[::1],14330;..."
3. Generate some trace records
4. Observe the logger server details is "1]"

### Expected Result

The logged server name should be the IPv6 literal "[::1]".

### Actual Result

The incorrect server details are logged as "1]", with the first part of the address missing (it is assumed to be a protocol).

Note that for a longer IPv6 address this is even worse, e.g. "[2001:db8::1]" would also be reported as "1]"

### Additional Context

The code uses a regex to parse the protocol out, using "(.*\s*:\\{0,2})?". A regex is greedy and so this consumes everything up until the last ":" (i.e. assuming it is all part of the protocol).

There are several ways to approach, e.g. explicitly listing protocols "(tcp|np|lpc|...etc)", stopping at the first ":" (non-greedy), or not allowing IPv6 literal "[" starting characters.

sgryphon:
Fixes #2673

## Changes

- Add some unit test data for IPv6 literal addresses, with expected results.
- Fix parsing of connection string (minimal fix, to just not allow "[")

* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes - trivial
* [ ] Changes in public API reviewed (if applicable) - no change


Kielek:
Could you please add CHANGELOG.md entry under unreleased section?

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlConnectionDetails.cs
```diff
@@ -18,17 +18,19 @@ internal sealed class SqlConnectionDetails
      *
      * [ ] can be any number of white-space, SQL allows it for some reason.
      *
-     * Optional "protocol" can be "tcp", "lpc" (shared memory), or "np" (named pipes). See:
+     * Optional "protocol" can be "tcp", "lpc" (shared memory), or "np" (named pipes),
+     * or more rarely "spx", "adsp", "rpc", "vines". See:
```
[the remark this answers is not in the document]
sgryphon: Removed the change in comments, so no longer referenced.

I really only care about the IPv6 fix. I don't know anything about these protocols (e.g. what they look like); just saw them when looking into the best approach for the fix, i.e. whether a list of (tcp|lpc|np) would work, and when I found references to other protocols I changed the approach to just a simple [^[].

sgryphon:
> Could you please add CHANGELOG.md entry under unreleased section?

I thought the change was trivial (which the instructions say don't worry about changelog), but easy enough to add. Done.

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fix data source regex to support IPv6 literal addresses
```
Kielek, line 5:
```suggestion
* Fix data source regex to support IPv6 literal addresses.
```

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fix data source regex to support IPv6 literal addresses.
+  ([#2674](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2674))
```
alanwest, line 6:
```suggestion
* Fix issue where IPv6 addresses were improperly parsed from the the
  connection's `DataSource` when used to populate the `server.address`
  attribute.
  ([#2674](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2674))
```

alanwest (approved):
LGTM thanks @sgryphon!

PR #2686, opened 2025-04-07
[repo] RegexGenerator for .NET

### LINKED ISSUE #2234: [sql] Consider using regex source generator
Discussed here in the context of parsing sql connection details: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2233#discussion_r1811106842

> There is a RegEx source generator available now. Doesn't have to be part of this PR, but we should probably move to that on targets where it is available. Reason being is I assume it gives best perf across the board but especially when AoT is in play and compilation is not available (assuming RegEx falls back to interpreter when dynamic code isn't available).
>
> Some reasons listed here why it should be used:
https://learn.microsoft.com/dotnet/standard/base-types/regular-expression-source-generators#when-to-use-it

Kielek:
Fixes #2234 across all production code

## Changes

Use RegexGenerator in production code for .NET targets


ppittle (approved):

alanwest (approved):
LGTM. Need to resolve conflict from merging #2674 (the regex changed).

Also would be good to get an approval from the OneCollector owners since it touches that. @rajkumar-rangaraj @CodeBlanch

Kielek:
Conflicts solved.

CodeBlanch (approved):
LGTM

Issue #2238, opened 2024-10-22
[sql] Generate `db.query.summary`


alanwest:
Instrumentation that supports parsing `db.query.text` should [generate a summary of the query text](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#generating-a-summary-of-the-query-text).

This is used in the `db.query.summary` attribute. It is also used, when available, in the span name (#2237).

stevejgordon:
@alanwest As discussed, I'm going to try to pick this up and close it off. I'm first going to update/add tests using the test cases from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/database-test-cases/db-sql-test-cases.json and then tweak the summary logic as required to handle them. I'll also take a look at benchmarking the current implementation and seek any obvious optimisations.

martincostello:
While working on #3011 I noticed that EFCore doesn't change its span name like SqlClient does here:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/52fc1c91606bdfeb19eb00867ab0247919a4b57a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L181

I guess as part of this issue we should also update this code to use `SqlProcessor.GetSanitizedSql()` and set `db.query.summary` too?

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/52fc1c91606bdfeb19eb00867ab0247919a4b57a/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L149-L171

stevejgordon:
@alanwest / @martincostello Do you think we can close this now?

martincostello:
Seems reasonable to me.

PR #2689, opened 2025-04-09
[sql] SQL parsing for generating `db.query.summary`

alanwest:
Towards #2238

This PR introduces the parsing necessary to generate the `db.query.summary` attribute. I will wire up the instrumentation to add the attribute in a follow up PR.

It also adds additional parsing for DDL operations like `CREATE TABLE`.


src/Shared/SqlProcessor.cs
```diff
@@ -233,32 +236,136 @@ private static bool SanitizeNumericLiteral(string sql, ref int index)
...
+                state.DbQuerySummary.Append(sql[i]);
+            }
+
+            return true;
+        }
+
+        return false;
+    }
+
+    private static bool LookAhead(string compare, string sql, ref int index, SqlProcessorState state, bool isOperation = true, bool captureNextTokenAsTarget = false)
```
TimothyMothra, line 319: I would feel more comfortable if we had some unit tests for the LookAhead method.
I understand that this is covered by the TestCases, but the testcases have been difficult to debug.

TimothyMothra: Sure. Both the Http and AspNetCore tests use a similar TestCase.json.
The last time I needed to work in one of those projects, I found it difficult to debug a specific test case.

Here in the Sql project, I would like to see a couple unit tests for the `LookAhead` method. 
I'm certain that your implementation works based on the scenarios I looked over in the test-case.json.
But I'm thinking that in the future when someone needs to inspect how `LookAhead` works, that future maintenance would be easier with a dedicated unit test.

alanwest: I'd like to avoid writing tests against `LookAhead` for a couple reasons, at least for now.

1. I expect I will need to continue to evolve this implementation especially as I evolve it to support other SQL dialects. I'd like to do so with less friction reducing the test surface area I'll have to maintain.
2. While sure this is an internal class, I generally try to avoid exposing private methods only for tests. I know there are times we do this, but I try to avoid it anyways. Additionally, this class may ultimately be pulled into other instrumentations, and I would like its interface to be clear.

That said, I do hear you with respect to debuggability. For this test suite I have been leaning on conditional breakpoints filtering to the name of the test case.

When I implemented the ASP.NET Core test suite, I originally had a feature where you could mark a test with a `debug` flag, here's from an old [commit](https://github.com/open-telemetry/opentelemetry-dotnet/blob/7d552b34080a7baa8679e5dd30990fc39143064b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RouteTestData.cs#L44-L53). I guess I didn't land this feature since I ultimately gave all the tests names, but if you like this I wouldn't be opposed to adding it back.

TimothyMothra: > When I implemented the ASP.NET Core test suite, I originally had a feature where you could mark a test with a debug flag, here's from an old [commit](https://github.com/open-telemetry/opentelemetry-dotnet/blob/7d552b34080a7baa8679e5dd30990fc39143064b/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RouteTestData.cs#L44-L53). I guess I didn't land this feature since I ultimately gave all the tests names, but if you like this I wouldn't be opposed to adding it back.

ooooooo, thanks for sharing. This would have made some previous work easier. I'd be supportive of a feature like this. But a lower priority since conditional breakpoints are an option.

TimothyMothra (approved):

PR #2693, opened 2025-04-11
[sql] Implement updates to new semantic conventions for stored procedures

alanwest:
Implementing the new [`db.stored_procedure.name` attribute](https://github.com/open-telemetry/semantic-conventions/pull/2037).

The `db.operation.name`, `db.collection.name`, and `db.query.text` attributes have been removed when `CommandType` = `StoredProcedure` because they are no longer applicable.

Side note, we have an odd scenario for .NET Framework users. `CommandType` is not known, so `db.query.text` will still be used even for stored procedures. This requires documentation, but is beyond the scope of this PR.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
@@ -115,9 +115,7 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
 
                                     if (options.EmitNewAttributes)
                                     {
-                                        activity.SetTag(SemanticConventions.AttributeDbOperationName, "EXECUTE");
-                                        activity.SetTag(SemanticConventions.AttributeDbCollectionName, commandText);
-                                        activity.SetTag(SemanticConventions.AttributeDbQueryText, commandText);
+                                        activity.SetTag(SemanticConventions.AttributeDbStoredProcedureName, commandText);
                                     }
 
                                     break;
@@ -278,8 +276,7 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
                 {
                     if (this.commandTextFetcher.TryFetch(command, out var commandText))
                     {
-                        tags.Add(SemanticConventions.AttributeDbOperationName, "EXECUTE");
-                        tags.Add(SemanticConventions.AttributeDbCollectionName, commandText);
+                        tags.Add(SemanticConventions.AttributeDbStoredProcedureName, commandText);
                     }
                 }
             }
--- a/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
@@ -11,7 +11,7 @@
   When `CommandType` is `StoredProcedure`, the `db.stored_procedure.name` has
   been added and the `db.query.text`, `db.operation.name`, and
   `db.collection.name` attributes have been removed.
-  ([#TBD](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TBD))
+  ([#2693](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693))
 
 ## 1.11.0-beta.2
 
--- a/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
@@ -6,6 +6,12 @@
   connection's `DataSource` when used to populate the `server.address`
   attribute.
   ([#2674](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2674))
+* Updates have been made to adhere to recent changes in the new semantic
+  conventions. These affect you if you have opted in to the new conventions.
+  When `CommandType` is `StoredProcedure`, the `db.stored_procedure.name` has
+  been added and the `db.query.text`, `db.operation.name`, and
+  `db.collection.name` attributes have been removed.
+  ([#TBD](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TBD))
 
 ## 1.11.0-beta.2
 
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlActivitySourceHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlActivitySourceHelper.cs
@@ -34,10 +34,9 @@ internal sealed class SqlActivitySourceHelper
     internal static readonly string[] SharedTagNames =
     [
         SemanticConventions.AttributeDbSystem,
-        SemanticConventions.AttributeDbCollectionName,
         SemanticConventions.AttributeDbNamespace,
+        SemanticConventions.AttributeDbStoredProcedureName,
         SemanticConventions.AttributeDbResponseStatusCode,
-        SemanticConventions.AttributeDbOperationName,
         SemanticConventions.AttributeErrorType,
         SemanticConventions.AttributeServerPort,
         SemanticConventions.AttributeServerAddress,
--- a/src/Shared/SemanticConventions.cs
+++ b/src/Shared/SemanticConventions.cs
@@ -141,13 +141,12 @@ internal static class SemanticConventions
     // New database conventions as of commit:
     // https://github.com/open-telemetry/semantic-conventions/blob/25f74191d749645fdd5ec42ae661438cf2c1cf51/docs/database/database-spans.md#common-attributes
     public const string AttributeDbSystem = "db.system";
-    public const string AttributeDbCollectionName = "db.collection.name";
     public const string AttributeDbNamespace = "db.namespace";
-    public const string AttributeDbOperationName = "db.operation.name";
     public const string AttributeDbResponseStatusCode = "db.response.status_code";
     public const string AttributeDbOperationBatchSize = "db.operation.batch.size";
     public const string AttributeDbQuerySummary = "db.query.summary";
     public const string AttributeDbQueryText = "db.query.text";
+    public const string AttributeDbStoredProcedureName = "db.stored_procedure.name";
 
 #pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
 }
--- a/test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientTests.cs
+++ b/test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientTests.cs
@@ -482,9 +482,7 @@ public class SqlClientTests : IDisposable
 
                 if (emitNewAttributes)
                 {
-                    Assert.Equal("EXECUTE", activity.GetTagValue(SemanticConventions.AttributeDbOperationName));
-                    Assert.Equal(commandText, activity.GetTagValue(SemanticConventions.AttributeDbCollectionName));
-                    Assert.Equal(commandText, activity.GetTagValue(SemanticConventions.AttributeDbQueryText));
+                    Assert.Equal(commandText, activity.GetTagValue(SemanticConventions.AttributeDbStoredProcedureName));
                 }
 
                 break;


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -115,9 +115,7 @@ public override void OnEventWritten(string name, object? payload)
 
                                     if (options.EmitNewAttributes)
                                     {
-                                        activity.SetTag(SemanticConventions.AttributeDbOperationName, "EXECUTE");
-                                        activity.SetTag(SemanticConventions.AttributeDbCollectionName, commandText);
-                                        activity.SetTag(SemanticConventions.AttributeDbQueryText, commandText);
+                                        activity.SetTag(SemanticConventions.AttributeDbStoredProcedureName, commandText);
```
alanwest, line 118: @trask @lmolkova @maryliag

Wanted to check with you all to confirm this meets the spirt of the recent changes regarding stored procedures.

1. Removing `db.operation.name`. I was arbitrarily setting it to `EXECUTE` before since there was no "operation" present in the query text.
2. Removing `db.collection.name` and `db.query.text` since their values would simply be the same as `db.stored_procedure.name` making them redundant.

alanwest: Also, once I add `db.query.summary` should `db.query.summary` = `db.stored_procedure.name`?

trask: `db.collection.name` shouldn't be captured since it's not a collection

I would expect `db.query.text` to be captured if the stored proc is run via a query API (where you pass SQL), but from a non-query API where you explicitly pass a stored proc name

`db.operation.name` is a good question. I think if it's not a query API (where you pass SQL), then we should capture `db.operation.name`

trask: I don't think Java has a non query API for calling stored procedures

https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#prepareCall-java.lang.String-

alanwest: The difference between using the "non-query" vs "query" API when invoking a stored procedure in C# is subtle.

### "Non query API"

This will be the more common way to invoke a stored procedure.

```csharp
var command = new SqlCommand();
command.CommandType = CommandType.StoredProcedure;
command.CommandText = "some_stored_procedure"
command.Parameters.AddWithValue("@some_parameter", "some value");
```

This is what I think you're proposing:

`db.stored_procedure.name` = `some_stored_procedure`
`db.query.text` = `some_stored_procedure`
`db.operation` = `EXECUTE` or `CALL` or `EXEC` or ... maybe this is where the spec could be more clear
`db.query.summary` = `EXECUTE some_stored_procedure` (or `CALL` or `EXEC`)

### "Query API"

This will likely be less common, but still possible.

```charp
var command = new SqlCommand();
command.CommandType = CommandType.Text;
command.CommandText = "some_stored_procedure @some_parameter"
command.Parameters.AddWithValue("@some_parameter", "some value");
```

I guess in this scenario we might only be able to set `db.query.text`.

`db.query.text` = "some_stored_procedure"
`db.operation` = ???
`db.query.summary` = ??? (it _might_ be possible to parse the query text and safely infer that it is likely a stored procedure)

trask: ### "Non-query API"

I think not capturing `db.query.summary` since not capturing `db.query.text`

### "Query API"

I think an ideal parser could capture both `db.stored_procedure.name` and `db.query.summary`? But it's also ok if a parser doesn't do this.

alanwest: > I think not capturing `db.query.summary` since not capturing `db.query.text`

Ah interesting, ok I agree that makes sense.

> I think an ideal parser could capture both db.stored_procedure.name and db.query.summary? But it's also ok if a parser doesn't do this.

I suppose if a parser could capture `db.stored_procedure.name` then I'd be inclined to not set `db.query.text` and `db.query.summary` just as in the non-query API case. Though, I'm not yet sure this will be possible... I'll consider it.

trask: > I suppose if a parser could capture `db.stored_procedure.name` then I'd be inclined to not set `db.query.text` and `db.query.summary` just as in the non-query API case.

I think it's ok to capture `db.query.text` and `db.query.summary` in this case, since the instrumentation is at the query text level in this case.

It's kind of a fine line, e.g. in Java, Redis client has a structured API, but then emits a query syntax. The Java instrumentation is written at the lower level query layer and so captures `db.query.text` (and `db.query.summary` once that's implemented). However if the instrumentation was written at the structured API layer, then I think it wouldn't capture `db.query.text` (or `db.query.summary`).

I'm not sure this is right, but it's how it makes (some) sense in my mind. I'll open a tracking issue, I think this is something we should try to decide on and clarify in the spec.

trask: https://github.com/open-telemetry/semantic-conventions/issues/2098

TimothyMothra (approved):
LGTM

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
matt-hensley: Calling stored procedures via EF Core the "Query API" path looks to be commonly used.

[Example from the docs](https://learn.microsoft.com/en-us/ef/core/querying/sql-queries?tabs=sqlserver#passing-parameters):
```
var user = "johndoe";
var blogs = await context.Blogs
    .FromSql($"EXECUTE dbo.GetMostPopularBlogsForUser {user}")
    .ToListAsync();
```

lmolkova: @alanwest @matt-hensley @trask wdyt about this in the context of https://github.com/open-telemetry/semantic-conventions/pull/2200?

I think it applies to this case in the following way:

```csharp
var command = new SqlCommand();
command.CommandType = CommandType.StoredProcedure;
command.CommandText = "some_stored_procedure"
command.Parameters.AddWithValue("@some_parameter", "some value");
```

is a form of convenience API.

The proposal in this PR is to apply instrumentation on the convenience layer where operation-level things are available.

We already have a language that requires to populate `db.operation.name` when it's readily available. I think in scope of this instrumentation (Ms.Data.SqlClient) it's always `EXEC` for stored procedures. So we should set it.

----

Now should we also set `db.query.text`, summary and parameters?

**If we do** it wouldn't be possible to tell if I used convenience or executed a raw query. According to [chatgpt](https://chatgpt.com/share/68119770-94e8-800b-a788-d3b3f0d454d9), these two are not equivalent and may have different perf and behavior esp when using ADO.NET abstractions over different databases or similar thing in JDBC. So  there probably should be a way to tell the difference and we might not need the same telemetry for non-query vs query calls

**If we don't**:  
- do we still want to record parameters on non-query API? I think we should, even though `db.query.text` is not there.
- should we adjust the span name section and tell to use stored procedure name as one of the possible fallbacks (when summary is not available)?
- we'd end up with inconsistency (which I think is fine)
  - `db.operation.name` and `db.stored_procedure.name` are set when 1) non-query API is used, 2) query api is used and instr does parsing
  - only `db.query.*` are set when query API is used and instr does not parse it

I'd rather **not** set `db.query.text|summary` for non-query API to avoid duplication and reduce perf overhead of sanitization and parsing. It also can be a hint for the end-users to understand which API they called.
I can add the span name adjustments to the PR.

alanwest: Summary of comments from @lmolkova and @trask:

| | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
| --- | --- | --- |
| `db.query.summary` | ✖️ | ✖️ |
| `db.query.text` |  ✔️ if parsing disabled | ✖️ |
| `db.operation.name` | ✔️ if parsing enabled and only if @trask is ok changing language | ✔️ |
| `db.stored_procedure.name` | ✔️ if parsing enabled | ✔️ |

One concern when using `CommandType.Text` and parsing is disabled is that metrics become very limited since `db.query.text` will usually not be present. I guess this would just encourage people to either enable parsing or use `CommandType.StoredProcedure` which is more performant anyways. I haven't yet thought deeply about how I'd reliably parse `db.stored_procedure.name` from the query text. Though, besides these concerns, I'm ok with this proposal and I will update this PR to reflect it.

> So there probably should be a way to tell the difference and we might not need the same telemetry for non-query vs query calls

I think the presence or lack of certain attributes is an awfully subtle way to suggest which API was used. So, if you think distinguishing between which API was used is important, I'm not sure this is the greatest solution. Also, when parsing is enabled, the attributes present would be identical, so there would be no indication of which API was used.

> we probably need to add db.operation.name to the sql conventions for this case since we've left it out so far
👍

> we could record those parameters using db.operation.parameter.<key>?
👍

alanwest: I'd be ok setting `db.query.summary` for `CommandType.Text`, though correct me if I misremember @trask, I think you raised a concern that we'd artificially insert an operation in the summary that was not actually present in the text.

For example,
`db.operation.name` = `EXECUTE`
`db.stored_procedure.name` = `some_stored_procedure`
`db.query.text` = `some_stored_procedure`
`db.query.summary` = `EXECUTE some_stored_procedure`

---

Separate question: sanitization vs. parsing enabled... I was not planning on having separate configs for these. To your knowledge, are other languages implementing separate configs? I figured if you're scanning the entire query text to sanitize it then you may as well also construct the summary.

trask: > I think the presence or lack of certain attributes is an awfully subtle way to suggest which API was used. So, if you think distinguishing between which API was used is important, I'm not sure this is the greatest solution.

reminds me of https://github.com/open-telemetry/semantic-conventions/issues/754#issuecomment-2836940272

knowing whether a prepared statement was used or not also seems a "nice to have", but I think ok to address both post-stability when someone actively wants it

trask: @lmolkova do you think we would record parameters under `db.operation.parameter.<key>` in the non-query case? if so, we could have separate rows in the table for `db.query.parameter.<key>` and `db.operation.parameter.<key>`

lmolkova: > @lmolkova do you think we would record parameters under `db.operation.parameter.<key>` in the non-query case? if so, we could have separate rows in the table for `db.query.parameter.<key>` and `db.operation.parameter.<key>`

I'd prefer to use the same attribute for both - `db.query.parameter.<key>`.
In my mental model operation parameters are meta/config (e.g. operation timeout, max rows to read on the client side)

lmolkova: > Separate question: sanitization vs. parsing enabled...

I think it's more about instrumentation capabilities?

I.e. if instrumentation supports sanitization, it should populate `db.query.text` - it does not need to do full parsing and understand QL syntax. If may also implement parsing and then, by default should also parse. It seems important to let users disable parsing (if they are perf-sensitive). Disabling sanitization leaves them very bling, so at this point you probably just don't get much value from the instrumentation to start with?

lmolkova: > I'd be ok setting db.query.summary for CommandType.Text,

If we have a query, we'd set `db.query.text`, right? If we set query text, we also want to set `db.query.summary` (when we parse the query) - in this case stored procedure query is the same as any other query.

trask: > I'd be ok setting `db.query.summary` for `CommandType.Text`, though correct me if I misremember @trask, I think you raised a concern that we'd artificially insert an operation in the summary that was not actually present in the text.
> 
> For example,
> `db.operation.name` = `EXECUTE`
> `db.stored_procedure.name` = `some_stored_procedure`
> `db.query.text` = `some_stored_procedure`
> `db.query.summary` = `EXECUTE some_stored_procedure`

do you mean for `CommandType.StoredProcedure`?

lmolkova: looking into the @trask comment above, it'd be either-or (operation/stored procedure or query text/summary)


| | Query API (`CommandType.Text`) | Non-query API (`CommandType.StoredProcedure`) |
| --- | --- | --- |
| `db.query.summary` | ✔️ (if parsing enabled) | ✖️ |
| `db.query.text` |  ✔️ (if sanitization enabled) | ✖️ |
| `db.query.parameter.<key>` | ✔️ (opt-in) | ✔️ (opt-in) |
| `db.operation.name` | ✖️ | ✔️ |
| `db.stored_procedure.name` | ✖️ | ✔️ |

trask: > if instrumentation supports sanitization, it should populate db.query.text - it does not need to do full parsing and understand QL syntax

Java sanitization is done via the same parsing that we use to extract operation and table names

lmolkova: we chatted with @trask about this and came up with a few PRs:

- https://github.com/open-telemetry/semantic-conventions/pull/2206
- https://github.com/open-telemetry/semantic-conventions/pull/2205
- not filed yet: add operation and collection name to SQL

@alanwest @matt-hensley ptal when you have a moment, thanks!

alanwest: 👍 I'm good with [this proposal](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2693#discussion_r2069225825), and also the PRs updating the conventions.

Circling back to one thing...

> do you mean for CommandType.StoredProcedure?

No, I meant `CommandType.Text`. Even with the "query API" The literal query text can simply be the stored procedure name without the keyword `exec`. In this case, should `db.query.summary` be `some_stored_procedure` or should it be `exec some_stored_procedure`?

This is a super minor point in my opinion, but just raising it to clarify what I meant.

trask: is `exec` optional in general when executing a stored procedure in SQLServer?

alanwest: > is exec optional in general when executing a stored procedure in SQLServer?

Yes. The following queries are identical:

```sql
some_stored_procedure @some_param = 123
exec some_stored_procedure @some_param = 123
execute some_stored_procedure @some_param = 123
```

trask: thanks

> In this case, should `db.query.summary` be `some_stored_procedure` or should it be `exec some_stored_procedure`?

I'm thinking it should be `some_stored_procedure` based on:

> Instrumentations that support query parsing SHOULD parse the query and extract a
list of operations and targets from the query. It SHOULD set `db.query.summary`
attribute to the value formatted in the following way:
>
> ```
> {operation1} {target1} {operation2} {target2} {target3} ...
> ```

trask: > * not filed yet: add operation and collection name to SQL

now filed: https://github.com/open-telemetry/semantic-conventions/pull/2207

lmolkova: > I'm thinking it should be some_stored_procedure based on:

hmmm. but the query sent to DB would be `EXEC some_stored_procedure` and operation.name would be EXEC, so summary should  be `EXEC some_stored_procedure`.

alanwest: > the query sent to the DB would just be `some_stored_procedure`

Yup, that's correct.

trask:
thanks all 🎉

Issue #2682, opened 2025-04-03
[feature request] Solution to ungraceful shutdown of a process - PartialActivityProcessor


mladjan-gadzic:
### Component

OpenTelemetry.Extensions

### Is your feature request related to a problem?

An ungraceful shutdown of a process can cause some spans to never be pushed to a collector, resulting in missing crucial data. This is really frustrating for long-running processes.

### What is the expected behavior?

It already works as expected. This change would provide more flexibility to those in need.

### Which alternative solutions or features have you considered?

At G-Research, we’ve considered adding a custom PartialActivityProcessor that keeps track of active spans. It sends a heartbeat log during startup and while active and a stop log when the span ends. These logs are used by a custom collector that acts as both a receiver and an exporter.

We already have this implemented as a standalone extension of dotnet sdk, but we'd like to contribute this back to a community.

We're open to suggestions and further discussions to finding a home for PartialActivityProcessor.

martinjt:
I'm not sure that this is something that we should be presenting as a generalised approach and providing tooling around.

It's a known mode of operation that spans are only generated at the end (due to the requirement for start/end times).

This would be a good usecase for the new Events sub-signal, but I wouldn't advise building artificial spans from the events, it would be upto the observability backend to represent that data in a useful way.

xBis7:
We raised this issue on the dotnet community call yesterday April 8. Here is a summary of what was discussed.

* Issue title and description
    * The initial title and description didn't fully explain the scope or the intent of this issue.
    * Both of them have been updated accordingly. The main problem that we are trying to address is gaining visibility on active spans of very long-running processes (many hours or even days).
* Action items
    * Contribute the processor under the `Extensions` package on the `dotnet-contrib`
    * Add documentation with instructions on how people can implement something similar and also how they can use it
    * Bring it up again with the otel spec to determine whether this can be added to the main repo or not, depending on how useful it turns out to be to users

PR #2704, opened 2025-04-17

mladjan-gadzic:
Fixes 

Design discussion issue 

## Changes

PartialActivityProcessor is added. It helps by sending heartbeats (logs) while span is active and stop log when span ends. Log body is used to json serialize span per OpenTelemetry specification.


src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+    private void ExporterProc()
+    {
+        var triggers = new WaitHandle[] { this.shutdownTrigger };
+
+        while (true)
+        {
+            try
+            {
+                WaitHandle.WaitAny(triggers, this.scheduledDelayMilliseconds);
+                this.Heartbeat();
```
pellared, line 242: What is the purpose of sending heartbeat events?
If there was a "start" event then if there was not "end" event yet then it means that the span is in progress.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+        List<KeyValuePair<string, object?>> logRecordAttributesToBeAdded)
+    {
+        var buffer = new byte[750000];
+
+        var result = writeTraceDataMethod.Invoke(
+            null,
+            [buffer, 0, sdkLimitOptions, null!, new Batch<Activity>(data)]);
+        var writePosition = result as int? ?? 0; // Use a default value if null
+
+        var logRecord = (LogRecord)logRecordConstructor.Invoke(null);
```
pellared, line 155: I think we should also set the `Severity` and `EventId`.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: Sending heartbeat events serves the purpose of indicating that a span is still in progress. This improves observability by showing that an operation is active, even if it’s long-running. The heartbeat interval can typically be configured through the constructor.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+
+    private static List<KeyValuePair<string, object?>> GetLogRecordAttributes() =>
+    [
+        new("telemetry.logs.cluster", "partial"),
+        new("telemetry.logs.project", "span"),
+    ];
+
+    private static List<KeyValuePair<string, object?>> GetStopLogRecordAttributes() =>
+    [
+        new("partial.event", "stop"),
```
pellared, line 230: maybe

```suggestion
        new("span.state", "ended"),
```

We usually say that span has "started" and "ended".
I know that it is not great, but I think that ship has sailed: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#interface-definition.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: What about `INFO` for `Severity`?

About `EventId`, it get a little complicated. In docs it says that it should be only set when used through `ILogger` - and this is not the case. From where/what would it value be derived from?

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: Sure. This was more of internal idea of naming convention.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: We had some internal discussions about this and our conclusion was more/less the same. This is the first iteration just to get the community feedback on this.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
pellared: > What about INFO for Severity?

Works for me.

> From where/what would it value be derived from?

This should be something which defines the name of the event. Maybe simply `span.state`.

Example: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-events.md

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        this.scheduledDelayMilliseconds = scheduledDelayMilliseconds;
+
+        this.activeActivities = new ConcurrentDictionary<ActivitySpanId, Activity>();
+        this.endedActivities = new ConcurrentQueue<KeyValuePair<ActivitySpanId, Activity>>();
+
+        this.shutdownTrigger = new ManualResetEvent(false);
+
+        // Access OpenTelemetry internals as soon as possible to fail fast rather than waiting for the first heartbeat
+        AccessOpenTelemetryInternals(
```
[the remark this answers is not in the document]
pellared: Sorry, I am not a maintainer/approver here :wink: 
CC @open-telemetry/dotnet-approvers

pellared:
PTAL @lmolkova

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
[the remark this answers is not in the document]
cijothomas: Don't think it is a good idea to rely on any internals. If your intent to send a signal at Activity start/end, accept a ILoggerFactory/iLogger, and then use that to emit the log, just like how end users would do.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
swythan: The heartbeat helps detect cases where the process that started the span has crashed. If the span is _expected_ to last hours you can use the heartbeat to detect orphaned spans much more quickly.

I agree that it could easily be optional and/or be very infrequent.

swythan: OK. I can see your point.

My main concern is how to cross-reference the heartbeat events with the partial span events. The obvious thing to use would be `service.instance.id`, I suppose. 

When I started typing this response I was worried that it would need to be truly per-process unique to catch problems like a k8s pod getting OOMKilled and restarting. I have seen this not be true with the instance id injected by the OTEL k8s operator, for example. In that case it is unique to the _pod_ but not a single run of the container (because it injects env vars to the pod). 

Reading the [conventions](https://opentelemetry.io/docs/specs/semconv/attributes-registry/service/#service-instance-id), though, I'm now thinking maybe that's just a problem with relying on the k8s operator. I guess if you've got enough control to add a PartialActivityProcessor then you can just call `AddService()` and let it generate a random instance id each time.

lmolkova: sorry if I miss some other part of this discussion, but do we need a generic heartbeat? Most spans are super-short and won't benefit from it.

But if I know that I have a long-running job, I can instrument that long-running job and add domain-specific events (progress updates).

E.g. 

`timestamp = X, event_name = starting_job, attributes: {"job.name": "foo", "job.id": 42, "total_items": 100500, ...}`
`timestamp = X + 1, event_name = progress_update, attributes: {"job.id": 42, "done_items": 10, ...}`
....

Wouldn't it be much more useful than generic span heartbeat?

swythan: This isn't really for reporting progress, this is to enable downstream processing that can recover from long-running (multi minute/hour/day) spans never completing (due to problems from _outside_ the application) by rehydrating the unfinished span from this log-event data and sending it with an error state. This allows the proper visualisation/processing of spans hierarchies that contain failures of this type.

Our main use case is for long-running, complex, multi-level batch processes (think ML training, but we've got a very heterogeneous set of workloads) where individual steps suffer from random failures (viewed at scale). I've seen another use case for the same functionality for mobile apps, though, where the app can be suspended by the OS at any time.

The span heartbeats allow the processing to mark a span as failed (and send it) if it's not seen a heartbeat when it expects one. If we had a process/service heartbeat we could _probably_ use that to mark all uncompleted spans we'd seen from that process as errors if the heartbeats stopped arriving. I've not been involved in the implementation of our PoC, though. I'll defer to @mladjan-gadzic

mladjan-gadzic:
> > I think that processor can be used as a prototype for:
> > 
> > * [Define SpanProcessor emitting span lifecycle events opentelemetry-specification#4483](https://github.com/open-telemetry/opentelemetry-specification/issues/4483)
> > * [Define events semantic conventions for span lifecycle semantic-conventions#2133](https://github.com/open-telemetry/semantic-conventions/issues/2133)
> > 
> > I am not sure what is the stability policy in this project and if it accepts "Development/Experimental" features. Defering to @open-telemetry/dotnet-maintainers
> > Still I think it would be good to cleanup this PR so that it is in a "mergable" state and that it proposes some semantic conventions.
> 
> thanks for a feedback @pellared! i'll be off until Monday, but i'd clean it up on Monday if not before.

@pellared cleaned it up as much as i could

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
swythan: Note that the other advantage of per-span heartbeats is that we can pick up changes to the span attributes during its lifetime. It's a bit of an edge-case, but it's a fundamental difference in what's possible.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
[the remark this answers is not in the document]
swythan: What we really need is just a way to serialise all the Span content into a log message so that if the Span never completes normally (e.g. due to a process crash) we can rehydrate it in the component that has been recording these partial span events (including all the original attributes, etc.), mark it as finished in an error state and send it back in to our OTEL pipeline so it's not lost.

Using the serialiser from `OpenTelemetry.Exporter.OpenTelemetryProtocol` seems the obvious way to do that without missing something, but that code's not on the public API. 

FWIW I agree that accessing this via reflection is far from ideal (versioning issues worry me). Does anyone know why this isn't public in the upstream exporter?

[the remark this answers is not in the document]
pellared: Serializing to OTLP is coupling the processor to certain exporter implementation. We should simply encode the span as [a complex log attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/api.md#emit-a-logrecord) named `span`.

MikeGoldsmith:
Not sure why this was assigned to me. I don't hold any role in this SIG and searching doesn't show me listed in any workflow. Probably best to re-assign to one of the @open-telemetry/dotnet-contrib-approvers

pellared:
> Not sure why this was assigned to me. I don't hold any role in this SIG and searching doesn't show me listed in any workflow. Probably best to re-assign to one of the @open-telemetry/dotnet-contrib-approvers

Because you are the codeowner of `OpenTelemetry.Extensions` since https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1659.

pellared:
> Not sure why this was assigned to me. I don't hold any role in this SIG and searching doesn't show me listed in any workflow. Probably best to re-assign to one of the @open-telemetry/dotnet-contrib-approvers

@MikeGoldsmith, because you are the code owner of `OpenTelemetry.Extensions` since https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1659.

MikeGoldsmith:
Ah, that makes sense. I thought the workflow was picking me as one of the contrib approvers / maintainers. 

I'm happy to continue as a codeowner for that package 👍🏻

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
```diff
@@ -21,6 +21,12 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="System.Text.Json" Version="9.0.4" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="nupkgs\OpenTelemetry.Extensions.1.11.0-beta.1.67.snupkg" />
```
MikeGoldsmith, line 29: Can you explain a little more about why we need this?

src/OpenTelemetry.Extensions/Trace/ActivitySpec.cs
```diff
@@ -0,0 +1,134 @@
...
+        this.Name = activity.DisplayName;
+        this.ActivitySpecContext = new ActivitySpecContext(activity.Context);
+        this.ParentId = activity.ParentSpanId == default
+            ? string.Empty
+            : activity.ParentSpanId.ToString();
+        this.StartTime = FormatTimestamp(activity.StartTimeUtc);
+        this.EndTime = signal == Signal.Heartbeat
+            ? string.Empty
+            : FormatTimestamp(activity.StartTimeUtc.Add(activity.Duration));
+        this.StatusCode = activity.Status.ToString();
```
MikeGoldsmith, line 50: Do we need a default value here if Status was [Unset](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitystatuscode?view=net-9.0)?

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,130 @@
...
+#if NET
+        ArgumentNullException.ThrowIfNull(logExporter);
+#else
+        if (logExporter == null)
+        {
+            throw new ArgumentOutOfRangeException(nameof(logExporter));
+        }
+#endif
+        this.logExporter = logExporter;
+        this.logProcessor = new SimpleLogRecordExportProcessor(logExporter);
```
MikeGoldsmith, line 39: I believe it's generally preferred to use a [BatchLogRecordExportProcessor](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8c1e63894f0286b1c48b0448a866fdb3bd603fe8/src/OpenTelemetry/Logs/Processor/BatchLogRecordExportProcessor.cs) over a simple one processor for efficiency.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,130 @@
...
+            case Timeout.Infinite:
+                return this.logExporter.Shutdown() && this.logProcessor.Shutdown();
+            case 0:
+                return this.logExporter.Shutdown(0) && this.logProcessor.Shutdown(0);
+        }
+
+        var sw = Stopwatch.StartNew();
+        var timeout = timeoutMilliseconds - sw.ElapsedMilliseconds;
+        return this.logExporter.Shutdown((int)Math.Max(timeout, 0)) &&
+               this.logProcessor.Shutdown((int)Math.Max(timeout, 0));
```
MikeGoldsmith, line 109:
```suggestion
        var timeout = (int)Math.Max(timeoutMilliseconds - sw.ElapsedMilliseconds, 0);
        return this.logExporter.Shutdown(timeout) &&
               this.logProcessor.Shutdown(timeout);
```

src/OpenTelemetry.Extensions/Trace/ActivitySpec.cs
```diff
@@ -0,0 +1,134 @@
...
+        this.StatusMessage = activity.StatusDescription ?? string.Empty;
+        this.Attributes = activity.TagObjects
+            .ToDictionary(tag => tag.Key, tag => tag.Value)!;
```
MikeGoldsmith, line 53: Is TagObjects guaranteed to be set? Should we have a default here too?

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        using (this.logger.BeginScope(GetStartLogRecordAttributes()))
+        {
+            this.logger.LogInformation(
+                ActivitySpec.Json(new ActivitySpec(data, ActivitySpec.Signal.Heartbeat)));
+        }
+    }
+
+    /// <inheritdoc />
+    public override void OnEnd(Activity data)
```
MikeGoldsmith, line 72: This means we'd send a heartbeat on _every_ activity. This will result in a lot of heartbeat logs for regular actions that don't take a long time. Also, it means a long running action that doesn't continually create activities won't generate heartbeat messages.

MikeGoldsmith (commented):
Overall the code looks good and I've left some minor suggestion.

My concern is that the current design will generate a lot of heartbeat messages, one per activity, but won't generate track long running actions that don't continually create new activities.

Maybe a possible design could be to create some synchronised dictionary of active trace IDs and on some interval heartbeat messages are generated?

mladjan-gadzic:
> Overall the code looks good and I've left some minor suggestion.
> 
thanks for the review! @MikeGoldsmith 

> My concern is that the current design will generate a lot of heartbeat messages, one per activity, but won't generate track long running actions that don't continually create new activities.
> 
There is going to be heartbeat mechanism in the future. This is just the first step, to make this pr review friendly https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2704#discussion_r2049106325.

> Maybe a possible design could be to create some synchronised dictionary of active trace IDs and on some interval heartbeat messages are generated?
>
The idea is to keep two dicts, started and ended traces. Then on heartbeat interval all entries from ended that are still in started are going to be removed and ended dict is going to be emptied. For those remaining in started dict, heartbeat log is going to be pushed.

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
mladjan-gadzic: probably a leftover from testing. it will be removed.

src/OpenTelemetry.Extensions/Trace/ActivitySpec.cs
mladjan-gadzic: i am in the middle of transitioning of serialization from https://opentelemetry.io/docs/concepts/signals/traces/#spans to https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto. instead of just sending a span i'd send everything like it is in proto spec but it will be in json. this will be fixed.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: i used simple log record export processor because i'd like to have logs pushed asap. do you think i can achieve the same with batch log record export processor?

src/OpenTelemetry.Extensions/Trace/ActivitySpec.cs
mladjan-gadzic: i guarded this in another implmentation which will be available soon.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
mladjan-gadzic: i'm going to introduce heartbeat dealy which is intended for those regular actions that don't take long time. nice observation!

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor.cs
MikeGoldsmith: Yes, the batch processor still sends fairly quickly (the default [batch window](https://github.com/open-telemetry/opentelemetry-dotnet/blob/8c1e63894f0286b1c48b0448a866fdb3bd603fe8/src/OpenTelemetry/BatchExportProcessor.cs#L18) is 5s but could be configured lower if needed). It means the SDK doesn't need to establish and send each log record one at a time, instead it waits a short period and sends a group together.

To avoid having to expose those settings in this extension, you could take the processor as an argument so someone can choose how to configure it. Eg use a simple processor, change batch times, etc. I'd recommend the default be the Batch processor if one isn't provided.

pellared:
> The idea is to keep two dicts, started and ended traces. Then on heartbeat interval all entries from ended that are still in started are going to be removed and ended dict is going to be emptied. For those remaining in started dict, heartbeat log is going to be pushed.

And this can be done by a separate processor, right?

mladjan-gadzic:
> > The idea is to keep two dicts, started and ended traces. Then on heartbeat interval all entries from ended that are still in started are going to be removed and ended dict is going to be emptied. For those remaining in started dict, heartbeat log is going to be pushed.
> 
> And this can be done by a separate processor, right?
> 
> One can emit span start/end logs. Second the heartbeats logs in the way you just described.

I don't see any obvious blockers for that, so i'd yes.

mladjan-gadzic:
would it make more sense to accept ILogger or ILoggerFactory instead of BaseExporter? that would ultimately give more flexibility to end user.

mladjan-gadzic:
@MikeGoldsmith i've refactored pr to be inline with your feedback. the idea about heartbeats is to have another pr which will introduce delay and scheduled heartbeats. that way it should be more review friendly. what do you think about it?

pellared:
> @MikeGoldsmith i've refactored pr to be inline with your feedback. the idea about heartbeats is to have another pr which will introduce delay and scheduled heartbeats. that way it should be more review friendly. what do you think about it?

Was pushing https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2704/commits/d29c158aec6e24e0fa693cf3786c9b1667f8b92c a mistake?

mladjan-gadzic:
> > @MikeGoldsmith i've refactored pr to be inline with your feedback. the idea about heartbeats is to have another pr which will introduce delay and scheduled heartbeats. that way it should be more review friendly. what do you think about it?
> 
> Was pushing [d29c158](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/commit/d29c158aec6e24e0fa693cf3786c9b1667f8b92c) a mistake?

thanks for checking out this pr :)

after trying to introduce initial delay for start event, so that short lived spans are not spamming with unnecessary start/stop info, i've noticed that it is almost 80% into heartbeat implementation and just went with it - it makes more sense in this case.

pellared:
> after trying to introduce initial delay for start event, so that short lived spans are not spamming with unnecessary start/stop info, 

Why do you find this unnecessary? I think there still can be cases when these are nice to have. Even though the spans are "short lived" they can still not be finished e.g. because there was an application crash.

mladjan-gadzic:
> > after trying to introduce initial delay for start event, so that short lived spans are not spamming with unnecessary start/stop info,
> 
> Why do you find this unnecessary? I think there still can be cases when these are nice to have. Even though the spans are "short lived" they can still not be finished e.g. because there was an application crash.

"unnecessary" might be strong word, but the thing is that without delay, system (collector) gets quickly overwhelmed. that's why we find initial delay "must have" in our case. initial delay can be configured so that there is no initial delay and events will be sent as they happen. i agree with you on this logic, that's why i went ahead and added heartbeat because it goes so well with initial delay / or no initial delay for that matter.

pellared:
> "unnecessary" might be strong word, but the thing is that without delay, system (collector) gets quickly overwhelmed. 

How is it being overwhelmed by logs and it is not by spans?

Maybe you should consider adding some log processing/filtering on the collector.

mladjan-gadzic:
> > "unnecessary" might be strong word, but the thing is that without delay, system (collector) gets quickly overwhelmed.
> 
> How is it being overwhelmed by logs and it is not by spans?
> Maybe you should consider adding some log processing/filtering on the collector.
>

we've implemented custom collector that did some gymnastics with log body data and that was a choke point. team that was stress testing this decided to move on with idea to add initial heartbeat delay because of this and it seems to be just right. 

adding structure to enable initial start log delay was in line what was needed for heartbeat anyways.

if you mind having heartbeat as part of this processor or have concerns for that matter, flag can be added that enables/disables heartbeat. by default it can be disabled. what do you think about that?

swythan:
> > "unnecessary" might be strong word, but the thing is that without delay, system (collector) gets quickly overwhelmed.
> 
> How is it being overwhelmed by logs and it is not by spans?

It's not about the logs themselves, it the system that handles them. The system that handles these span state events (start/stop/heartbeat) is necessarily stateful. It has to keep track of all the open partial spans, record heartbeats against them, drop them from state when they complete normally and periodically "rehydrate" failed incomplete spans that have expired and send them as (in error state) to the normal collectors. This essentially means it needs a backing store, which then adds a lot more overhead to the handling of each span. 

Note that we're working with jobs running on a large compute farm. There are many, many thousands of jobs running at any one time, generating spans at various rates, depending on the job. A very significant proportion of spans are for HTTP requests, DB calls, etc. Those are typically very quick, so the statistical chance of the system failing whilst they are incomplete is low (and often the impact of losing them is also low). 

The spans that inspired us to do this work are the ones that last several minutes up to several days. If one of those is lost then all the child spans lose their relationship to each other (there can be tens of thousands of them in this case). 

Given all the above, the idea of the initial delay is to allow the trade-off to be made between missing short-lived spans and the resource requirements of the back-end. TBH we've not tuned that trade-off yet. All we know so far is that as soon as we rolled out our version the code in this PR (in a fairly limited way) it more-or-less instantly overwhelmed our backend.  (This is a common occurrence when we point our farm at a service for the first time! 😀)

PR #2708, opened 2025-04-22
[Exporter.Geneva] Fix AFDCorrelationId retrieval exception handling in Geneva Exporter

rajkumar-rangaraj:
Fixes #
Design discussion issue #

The `AFDCorrelationIdLogProcessor` was throwing unhandled exceptions when attempting to retrieve the `AFDCorrelationId `from `RuntimeContext` if the value had not been previously set by customers. This caused processing failures in the telemetry pipeline.
Solution.

## Changes

* Added a state tracking mechanism using `GenevaAfdCorrelationIdStateTracker` to remember when access failures occur
* Created a RuntimeContextSlot to track correlation ID access state
* Added try-catch block in GetRuntimeContextValue() to handle exceptions gracefully
* Implemented a check to skip attempting to access the correlation ID if previous attempts failed
* Added proper resource cleanup in the Dispose method


dmitry-kharchenko:
Overall LGTM

src/OpenTelemetry.Exporter.Geneva/Internal/AFDCorrelationIdLogProcessor.cs
```diff
@@ -42,6 +50,37 @@ public override void OnEnd(LogRecord data)
...
+        }
+
+        try
+        {
+            return RuntimeContext.GetValue<string>(AFDCorrelationId);
+        }
+        catch
+        {
+            GenevaAfdCorrelationIdStateTracker.Set(true);
+            return null;
```
TimothyMothra, line 83: nit: need an EventSource log here

src/OpenTelemetry.Exporter.Geneva/Internal/AFDCorrelationIdLogProcessor.cs
```diff
@@ -42,6 +50,37 @@ public override void OnEnd(LogRecord data)
...
+        }
+
+        try
+        {
+            return RuntimeContext.GetValue<string>(AFDCorrelationId);
+        }
+        catch
+        {
+            GenevaAfdCorrelationIdStateTracker.Set(true);
```
TimothyMothra, line 82: nit: it's not clear what `true` and `false` mean for "StateTracker".
To improve readability consider naming this something like "IsGenevaAfdCorrelationIdAvailable"

rajkumar-rangaraj: If there is an issue, the customer could set this from their app. Want to avoid "Is" in the name.

src/OpenTelemetry.Exporter.Geneva/Internal/AFDCorrelationIdLogProcessor.cs
TimothyMothra: just a follow up question, should we include any unique identifiers when logging here? 
I'm thinking along the lines of, if we need to correlate this specific exception with the web request it occurred within.

ChatGPT offered a suggestion, that logging the Activity.Current.TraceId might be useful.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterAFDCorrelationTests.cs
```diff
@@ -0,0 +1,392 @@
...
+            }
+            else
+            {
+                path = GenerateTempFilePath();
+                exporterOptions.ConnectionString = "Endpoint=unix:" + path + ";PrivatePreviewEnableAFDCorrelationIdEnrichment=true";
+                var endpoint = new UnixDomainSocketEndPoint(path);
+                senderSocket = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.IP);
+                senderSocket.Bind(endpoint);
+                senderSocket.Listen(1);
+            }
```
TimothyMothra, line 43: nit: it looks like this block is repeated in multiple tests. consider moving this to a static helper method.

rajkumar-rangaraj: We could follow up on this, as it's part of the other exporter test too. It needs refactoring.

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterAFDCorrelationTests.cs
```diff
@@ -0,0 +1,392 @@
...
+
+            // Verify the correct number of logs were created
+            Assert.Equal(threadCount, exportedItems.Count);
+            Assert.Equal(threadCount / 2, countWithCorrelationId);
+            Assert.Equal(threadCount / 2, countWithoutCorrelationId);
+
+            // Check that no exceptions were thrown
+            // If our implementation is correct, logs from threads without correlation ID
+            // should have been processed without exceptions
```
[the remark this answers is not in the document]
rajkumar-rangaraj: It's a statement indicating that no exceptions were thrown. Since the existing implementation threw an exception, I had to add it, instead of making test complex.

xiang17 (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterAFDCorrelationTests.cs
```diff
@@ -0,0 +1,392 @@
...
+                File.Delete(path);
+            }
+            catch
+            {
+            }
+        }
+    }
+
+    [Fact]
+    public void AFDCorrelationIdLogProcessor_WithoutCorrelationId_HandlesGracefully()
```
TimothyMothra, line 182: nit: the WithoutCorrelationId and WithCorrelationId tests are very similar. 
These could probably be combined into a `[Theory]` with a bool parameter to simplify this test file.

not blocking, could be done in a follow up PR.

TimothyMothra (approved):

PR #2709, opened 2025-04-22
[Instrumentation.SqlClient] Introducing context propagation to SQL Server

sincejune:
…on to propagate trace info to SQL Server

## Changes
This PR introduces a new flag, `ContextPropagationLevel`, in SqlClientInstrumentation, which enables the library to propagate `traceparent` information to the SQL Server database.

The ContextPropagationLevel can be set to either `trace` or `disabled` (the default value). When set to `trace`, the instrumentation library uses the [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16) command to pass a `traceparent` ID of the current Activity to the SQL Server.

### Next steps:
I plan to introduce a "service" level that will pass service information via SQL commands via [sqlcommenter](https://google.github.io/sqlcommenter/).

* [x] Unit tests updated: Enable this flag by default in integration tests


sincejune:
> @sincejune, could you please rise issue under [semantic convention](https://github.com/open-telemetry/semantic-conventions/) repository? It is not something we should accept without common standard for all supported languages (including other db clients).

@Kielek Sure, filed https://github.com/open-telemetry/semantic-conventions/issues/2162

Kielek:
Reopening, per request in private channel. https://github.com/open-telemetry/semantic-conventions/pull/2363 is now merged

src/OpenTelemetry.Instrumentation.SqlClient/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -9,6 +9,8 @@ OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.Rec
 OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.SetDbStatementForText.get -> bool
 OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.SetDbStatementForText.set -> void
 OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.SqlClientTraceInstrumentationOptions() -> void
+OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.ContextPropagationLevel.get -> string!
+OpenTelemetry.Instrumentation.SqlClient.SqlClientTraceInstrumentationOptions.ContextPropagationLevel.set -> void
```
Kielek: @alanwest, as I remember you are in the middle of stabilization process.
This feature is not stable on semantic level. I think that it should not be part of the public API. The alternative approach could be env var `OTEL_EXPERTIMENTAL_DOTNET_SQLCLIENTINSTRUMENTATION_CONTEXTPROPAGATIONLEVEL`. What is your opinion?

sincejune: Updated, naming it as `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATION_LEVEL` to follow the existing convention.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -82,6 +93,25 @@ public override void OnEventWritten(string name, object? payload)
                         return;
                     }
 
+                    if (options.ContextPropagationLevel == SqlClientTraceInstrumentationOptions.ContextPropagationLevelTrace &&
+                        command is IDbCommand { CommandType: CommandType.Text, Connection.State: ConnectionState.Open } iDbCommand)
+                    {
+                        var setContextCommand = iDbCommand.Connection.CreateCommand();
+                        setContextCommand.CommandText = SetContextSql;
```
zacharycmontoya, line 108: If there's a transaction we need to ensure the new set context_info query runs in the same transaction, otherwise the ExecuteNonQuery will throw. For reference, you can see a similar Datadog C# implementation here: https://github.com/DataDog/dd-trace-dotnet/blob/49c43b5bc1ca03a6fc5da0ce5077c308ac4cca06/tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs#L332-L333

```suggestion
                        setContextCommand.Transaction = iDbCommand.Transaction;
                        setContextCommand.CommandText = SetContextSql;
```

zacharycmontoya (commented):
Before approving, I'd like to see my comment resolved regarding copying over the Transaction to the new set context_info query. Without this change, I expect to see exceptions thrown when queries are executed within a transaction.

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -10,6 +10,16 @@
...
 
+* Added `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATION_LEVEL`
+  environment variable to propagate trace context to SQL Server databases.
+  ([#2709](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709))
+
+  Valid options for this environment variable are:
+  * `trace`: Propagate `traceparent` information to SQL Server databases
+    (see [SET CONTEXT_INFO](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-context-info-transact-sql?view=sql-server-ver16)).
+    Note that this option incurs an additional round-trip to the database.
+  * `disabled`: No context propagation performed, default value.
```
alanwest: Why `trace` and `disabled`? If it's just a matter of this feature being enabled or not, then wouldn't a simple `true`/`false` value be sufficient?

alanwest: The name `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATION_LEVEL` seems to suggest there might be other "levels" in the future? If not then I'd suggest the name `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_CONTEXT_PROPAGATION`.

Kielek: I would say that, there is high chance that there will be support also for `service.name`. Please check https://github.com/open-telemetry/semantic-conventions/pull/2495

alanwest: Ok I see. The name `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATION_LEVEL` still doesn't seem quite right to me, though.

It seems possible that users may want to enable propagating both `service.name` and `traceparent`.

Two ideas come to my mind.

1. Rename `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATION_LEVEL` to `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_CONTEXT_PROPAGATORS`. This config would accept a comma-separated list of propagators (e.g., `traceparent,servicename`), Empty list would be the default.
2. Two separate configuration options.
    * `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION=true`
    * `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_SERVICE_NAME_PROPAGATION=true`

Without better understanding the direction the spec is taking on context propagation for database instrumentation, I kind of prefer the second option.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -82,6 +93,26 @@ public override void OnEventWritten(string name, object? payload)
                         return;
                     }
 
+                    if (options.ContextPropagationLevel == SqlClientTraceInstrumentationOptions.ContextPropagationLevelTrace &&
+                        command is IDbCommand { CommandType: CommandType.Text, Connection.State: ConnectionState.Open } iDbCommand)
+                    {
+                        var setContextCommand = iDbCommand.Connection.CreateCommand();
+                        setContextCommand.Transaction = iDbCommand.Transaction;
```
martincostello, line 107: An integration should be added if possible that verifies transactions work.

sincejune: Updated.

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
sincejune: Updated.

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
martincostello: If this line is commented-out does the new test fail with an exception as suggested by https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709#pullrequestreview-3002594486?

Kielek: Yes: image

Kielek (changes requested):
There is some issues on .NET Framework 4.6.2 (locally executed) in integration tests
2 cases are failing on

```csharp
        if (commandText == GetContextInfoQuery)
        {
            Assert.NotEqual(commandResult, DBNull.Value); // <---- failing here
            Assert.True(commandResult is byte[]);
            var contextInfo = Encoding.ASCII.GetString((byte[])commandResult).TrimEnd('\0');
            Assert.Equal(contextInfo, activity.Id);
        }
```
.NET 8 and 9 are passing.

martincostello:
> There is some issues on .NET Framework 4.6.2 (locally executed) in integration tests

Do the .NET Framework tests not run in CI (I thought they did)? That might explain why https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2826#discussion_r2152736963 happened.

Kielek:
> > There is some issues on .NET Framework 4.6.2 (locally executed) in integration tests
> 
> Do the .NET Framework tests not run in CI (I thought they did)? That might explain why [#2826 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2826#discussion_r2152736963) happened.

It requires docker for linux. It is not available on the GH runners on Windows systems. So itnegration tests are not executed for such cases.

cuichenli:
> There is some issues on .NET Framework 4.6.2 (locally executed) in integration tests

The current implementation does not yet support .NET Framework. We may need to explore how to integrate it later.
 https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07105a510a5e8c93afd84c48f9b9dc4b6127b36e/src/OpenTelemetry.Instrumentation.SqlClient/SqlClientInstrumentation.cs#L51-L55

The logic is currently implemented in the `SqlClientDiagnosticListener.cs` file, but it's only used when running on .NET.
Given that, the most straightforward solution for now is to disable the context info check in .NET Framework tests.

Would that be acceptable? If not, we can continue working on support in this PR. Otherwise, we can defer it to a follow-up PR. Thanks!

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
```diff
@@ -66,24 +77,37 @@ public void SuccessfulCommandTest(
...
         catch
         {
         }
 
+        transaction?.Commit();
+
         Assert.Single(activities);
         var activity = activities[0];
 
+#if !NETFRAMEWORK
```
Kielek: Why? Does it mean that this feature will be working only on .NET?
If so, you need to adjust CHANGELOG and the documentation. I think that you should consider also conditional compilation for _production_ code.

cuichenli: >Does it mean that this feature will be working only on .NET?

I attempted to implement support for .NET Framework, but it appears there’s no mechanism on the instrumentation side to retrieve the current connection. So for now, we’ll limit support to .NET. 

CHANGELOG and README updated.

> should consider also conditional compilation for production code.

The entire `SqlClientDiagnosticListener.cs` file is wrapped in conditional compilation and won't compile under .NET Framework.

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -10,6 +10,15 @@
   the new conventions.
   ([#2811](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2811))
 
+* Added `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION`
+  environment variable to propagate trace context to SQL Server databases.
+  It is now only available on .NET 8 and newer.
+  ([#2709](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709))
```
alanwest:
```suggestion
* Added `OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION`
  environment variable to propagate trace context to SQL Server databases.
  This will remain experimental while the [specification](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/sql-server.md#context-propagation)
  remains in development.
  It is now only available on .NET 8 and newer.
  ([#2709](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2709))
```

alanwest (approved):

YayBurritos:
@sincejune , @alanwest , @Kielek : Any concerns over using SET CONTEXT_INFO with connection pooling? Is there a risk of the TraceId that's propagated "contaminating" unrelated SQL queries? It's great that the instrumentation library provides support for TraceId propagation, but I'd like to have some confidence that if our applications use this, we can rely on the TraceIds associated with our SQL queries to be accurate. Any info you can provide would be appreciated. Thanks!

PR #2716, opened 2025-04-25
[Instrumentation.Wcf] Restore aspnet headers read-only state

tchowice:
Fixes

- AspNetParentSpanCorrector always sets the request headers to read-only, even if it was originally writable. This breaks any existing application code that makes further changes to the request headers.

## Changes

AspNetParentSpanCorrector.OnRequestStarted now records the original state of `HttpContext.Request.Headers.IsReadOnly`, and always restores the original value to the property when done writing the headers.


Kielek:
@tchowice. could you please add CHANGELOG entry?
I think it can be important for other users also.

Kielek (approved):
LGTM, please ignore my previous comment. I have added changelog entry.

PR #2724, opened 2025-04-29
[Instrumentation.AspNetCore] Handle cases where the `HttpContext` is a property of the payload (ASP.NET Core 2.x support)

mathew-tolj:
Fixes #4126, #5019

## Changes

Fixes an issue where using `OpenTelemetry.Instrumentation.AspNetCore` with ASP.NET Core 2.x does not export any metrics or traces for incoming requests. This is achieved by checking if the `HttpContext` is a property of the payload in cases where the payload cannot be converted to directly to `HttpContext`. This allows ASP.NET Core 2.x (the only version still officially support on .NET Framework) to proceed with processing the event instead of short circuiting.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
@@ -47,6 +47,7 @@ internal class HttpInListener : ListenerHandler
     };
 
     private static readonly PropertyFetcher<Exception> ExceptionPropertyFetcher = new("Exception");
+    private static readonly PropertyFetcher<HttpContext> HttpContextPropertyFetcher = new("HttpContext");
 
     private readonly AspNetCoreTraceInstrumentationOptions options;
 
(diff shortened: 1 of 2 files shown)


Kielek:
@mathew-tolj, thank you for the contribution.
I think that I need to clarify couple things with @alanwest and @rajkumar-rangaraj.

Based on this https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/d1e3207695d7f88da048ca54df383133e65315f8/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/OpenTelemetry.Instrumentation.AspNetCore.Tests.csproj#L4

This package was never intended to work with .NET Framework. In fact, it will not if we will do not have .NET Standard 2.0 target for all packages in contrib.

What is your perspective on adding such support for this legacy stuff?

@mathew-tolj , If other maintainers will be fine with extending support we need to do couple other things
1. Consider `net462` as an additional target framework in the csproj.
2. Extending test coverage for .NET Framework. See already linked file.
3. Updating documentation, probably in both cases, whether we will be adding this support or not.

mathew-tolj:
I did notice that there weren't any existing tests for the .NET Standard target that this PR is for so that is why I had not pushed any changes to the test projects. It is pretty easy to add a test to the .NET 8 and 9 that tests extraction of the `HttpContext` from an anonymous object but I wasn't sure that was the way to go as that doesn't represent the real world scenario accurately.

I have also attempted to add a .NET Framework target to the test projects and that is mostly working although it is not without it's own challenges and I am not sure if the code is of a good enough standard to push. As of now there are 3 remaining obstacles:
- There appears to be an issue with compilation of the Razor pages under .NET Framework that I have not figured out yet.
- One of the tests focused on exception handling fails because the exception is bubbling up and short circuiting the test.
- The biggest issue is that there is no way to the route template in the OpenTelemetry code when using ASP.NET 2.x as endpoint routing support was not added until ASP.NET Core 3.0. This mean that the activity name is not updated and the `http.route` tag is not added. The missing tag is especially problematic as a lot of the tests use it as part of their assertions.

PR #2727, opened 2025-05-01
[SqlClient] Support new DB convention: `db.system.name`

alanwest:
Introduces the new `db.system.name` attribute when the new conventions are enabled.

We're still lacking great test coverage, and this PR does not seek to improve upon that. Coming soon, though!


rajkumar-rangaraj (approved):
LGTM

PR #2798, opened 2025-05-17
[SqlClient] Refactor SqlClient test

alanwest:
I think this is the last PR refactoring SqlClient tests. The tests still need to be expanded, but it should be much clearer now what is tested. Adding additional coverage will be far easier.


test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
```diff
@@ -104,6 +105,126 @@ public void SuccessfulCommandTest(
         }
     }
 
+    private static void VerifyActivityData(
```
alanwest, line 108: I moved some methods to this file because the integration test was the only remaining thing using them.

I still need to decide what to do with this integration test. The new test suite could fairly easily be turned into proper integration tests replacing this one.

rajkumar-rangaraj (approved):

PR #2822, opened 2025-06-09
Support .NET 10

### LINKED ISSUE #2459: [bug] Remove Explicit Dependency on System.Text.Json in OpenTelemetry.Exporter.OneCollector on net8.0
### Component

OpenTelemetry.Exporter.OneCollector

### Package Version

| OpenTelemetry.Exporter.OneCollector   | 1.10.0   |

### Runtime Version

net8.0

### Description

On net8.0 and above, an explicit reference to System.Text.Json is not needed. It is included in the Framework. Removing this reference reduces explicit transitive dependencies.

### Steps to Reproduce

Include OpenTelemetry.Exporter.OneCollector in a project using net8.0

### Expected Result

OpenTelemetry.Exporter.OneCollector uses STJ from the framework and doesn't use an explicit dependency.

### Actual Result

OpenTelemetry.Exporter.OneCollector uses STJ from an explicit dependency.

### Additional Context

I can submit a fix if this is accepted.

martincostello:
## Changes

Add support for building and testing with .NET 10 using .NET 10 RC2.

Also resolves #2459 as the .NET 10 SDK has new features that allows System.Text.Json to be detected as a shipping part of the parent application so we can remove the reference and allow the application to bring its own version with it.


martincostello:
Need to investigate macOS test failure with .NET 10. I don't have a Macbook so might be tricky:

```text
[xUnit.net 00:00:00.10]   Finished:    OpenTelemetry.Resources.OperatingSystem.Tests
  Failed OpenTelemetry.Resources.OperatingSystem.Test.OperatingSystemDetectorTests.TestOperatingSystemAttributes [7 ms]
  Error Message:
Test Run Failed.
   Assert.Contains() Failure: Sub-string not found
String:    "macOS 14.7.6"
Not found: "Darwin"
  Stack Trace:
     at OpenTelemetry.Resources.OperatingSystem.Test.OperatingSystemDetectorTests.TestOperatingSystemAttributes() in /Users/runner/work/opentelemetry-dotnet-contrib/opentelemetry-dotnet-contrib/test/OpenTelemetry.Resources.OperatingSystem.Tests/OperatingSystemDetectorTests.cs:line 49
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

martincostello:
> Need to investigate macOS test failure with .NET 10

Due to changes in https://github.com/dotnet/runtime/pull/113041.

martincostello:
Need to investigate the new .NET 10 failures for the HTTP instrumentation.

martincostello:
The .NET 10 tests are failing due to https://github.com/dotnet/runtime/pull/115959 which fixed https://github.com/dotnet/runtime/issues/108050.

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.Basic.cs
```diff
@@ -621,7 +621,13 @@ public async Task ReportsExceptionEventForNetworkFailuresWithGetAsync()
 
         // Exception is thrown and collected as event
         Assert.True(exceptionThrown);
+
+#if NET10_0_OR_GREATER
+        // .NET 10 reports its own activity, but does not set the status description
+        Assert.Contains(exportedItems[0].Events, evt => evt.Name.Equals("exception"));
+#else
```
martincostello, line 628: Alternatively, we could just change the assertion to `Contains()` instead of `Single()` for all TFMs.

ghost:
@martincostello with dotnet/runtime#116269 and dotnet/runtime#115959 available in .NET 10, the HttpClient native instrumentation now records exceptions. I think you can condition the following lines to apply them only for versions before 10:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/fabb87f0f6ad7eeb047b8c3eb64b8b6acc040e24/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L181-L202

In the same way that some attributes are set only for versions before 9 (because runtime does it from this version):
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/fabb87f0f6ad7eeb047b8c3eb64b8b6acc040e24/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L140

Directory.Packages.props
```diff
@@ -142,5 +146,10 @@
     <PackageVersion Update="Microsoft.AspNetCore.Mvc.Testing" Version="9.0.9" />
     <PackageVersion Update="Microsoft.AspNetCore.SignalR.Client" Version="9.0.9" />
   </ItemGroup>
+  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0'))" Label="Non-production packages, such as for examples and tests, for .NET 9">
```
martincostello:
```suggestion
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net10.0'))" Label="Non-production packages, such as for examples and tests, for .NET 10">
```

Directory.Build.props
```diff
@@ -3,4 +3,8 @@
     <ArtifactsPath>$([System.IO.Path]::Combine('$(MSBuildThisFileDirectory)', 'artifacts'))</ArtifactsPath>
     <UseArtifactsOutput>true</UseArtifactsOutput>
   </PropertyGroup>
+  <!-- HACK Workaround for https://github.com/dotnet/sdk/issues/51265 -->
+  <PropertyGroup>
+    <RestoreEnablePackagePruning>false</RestoreEnablePackagePruning>
+  </PropertyGroup>
```
martincostello, line 9:
```suggestion
  </PropertyGroup>
  <ItemGroup>
    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-5rrx-jjjq-q2r5" />
  </ItemGroup>
```

martincostello, line 9: Opened #3340 to track removing this for GA.

rajkumar-rangaraj (changes requested):
PR looks good, we identified a regression with Geneva and need a release in a day or two.

rajkumar-rangaraj:
> PR looks good, we identified a regression with Geneva and need a release in a day or two.

Completed the release.

Kielek (approved):
LGTM, Thanks @martincostello!

martincostello:
I think I just need to update the routing test cases due to #3350, then this should be ready to go if we want to do an RC before .NET 10 GA.

martincostello:
Will also need updating after #3338 is merged.

rajkumar-rangaraj (approved):

.github/workflows/integration.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [net8.0]
+        version: [net8.0, net10.0]
```
xiang17, line 33: Why does this not have net9.0 and line 42 only have net8.0?

martincostello: .NET 9 has never been in the test matrix. That's an unrelated change to adding .NET 10 support.

The Cassandra leg was missed as when the PR was started it didn't exist, and then I didn't update it in a later rebase/merge.

xiang17 (approved):

PR #2838, opened 2025-06-11
[Infra] Update Dockerfiles with dependabot

martincostello:
## Changes

Update dependabot configuration to keep the SDK versions in the Dockerfiles up-to-date on a weekly basis to match the .NET SDK in `global.json`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,11 @@ updates:
       day: "wednesday"
     labels:
       - "infra"
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-major"
+          - "version-update:semver-minor"
   - package-ecosystem: "docker"
     directory: "test/OpenTelemetry.Instrumentation.ConfluentKafka.Tests"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,6 +6,42 @@ updates:
       interval: "daily"
     labels:
       - "infra"
+  - package-ecosystem: "docker"
+    directory: "test/OpenTelemetry.Instrumentation.Cassandra.Tests"
```
TimothyMothra: just curious, would it make sense to scan the entire "test/*" directory?

martincostello: I genuinely don't know if it recurses or not.

TimothyMothra: Looks like it can :)
I think we should give this a try.


https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#:~:text=To%20specify%20manifests%20in%20the%20current%20directory%20and%20recursive%20subdirectories


```yaml
# Specify all directories from the current layer and below recursively,
# using globstar, for locations of manifest files

version: 2
updates:
  - package-ecosystem: "composer"
    directories:
      - "**/*"
    schedule:
      interval: "weekly"
```

Kielek (approved):

alanwest (approved):
Approving/merging. As discussed in today's SIG, this solution halfway fixes the issue.

[Here's an example](https://github.com/open-telemetry/opentelemetry-dotnet/pull/6339#discussion_r2153278584) from the main repo where this solution bumps the .NET 8 SDK but not the .NET 9 SDK, so manual intervention is still required.

But halfway is better than noway!

PR #2826, opened 2025-06-09
[SqlClient] Fix test for .NET Framework

martincostello:
## Changes

Update test cases for different behaviour between .NET Framework and .NET.

Discovered after I ran `dotnet test` on my Windows 11 machine.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #2850, opened 2025-06-17
[SqlClient] Unify .NET/.NET Framework tests

alanwest:
Another PR bringing some sanity to SqlClient tests. The main purpose is to leverage the same test suite introduced in #2798 for .NET Framework.


test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
```diff
@@ -27,22 +27,19 @@ public SqlClientIntegrationTests(SqlClientIntegrationTestsFixture fixture)
     [InlineData(CommandType.Text, "select 1/1")]
     [InlineData(CommandType.Text, "select 1/1", true, "select ?/?")]
     [InlineData(CommandType.Text, "select 1/0", false, null, true)]
-    [InlineData(CommandType.Text, "select 1/0", false, null, true, false, false)]
-    [InlineData(CommandType.Text, "select 1/0", false, null, true, true, false)]
+    [InlineData(CommandType.Text, "select 1/0", false, null, true, true)]
```
martincostello, line 30: Do you get the same failure locally as I get as fixed in #2826?

martincostello (approved):

test/OpenTelemetry.Instrumentation.SqlClient.Tests/SqlClientIntegrationTests.cs
alanwest: No, these tests were running fine for me on a Windows 11 machine. A test case was removed here because I removed the `shouldEnrich` parameter. I do not feel that it is important that this integration test tests enrich. We have that coverage elsewhere.

rajkumar-rangaraj (approved):

PR #2869, opened 2025-06-30
WIP: Use source generators for semantic conventions

### LINKED ISSUE #2856: [feature request] source generator using semantic conventions
### Component

None

I want to eliminate the need to define a collection of strings representing my attribute keys.

I define a struct/class in my project like below

```
[attributes("disk")]
internal static partial struct DiskAttributeNames
{
}

```

Once generator runs you end with
```
internal static partial struct DiskAttributeNames
{
  internal static partial struct disk
  {
    internal static partial struct io
    {
      internal const string direction = "disk.io.direction";
    }
  }
}
```

And you can use them via `DiskAttributeNames.disk.io.direction`

Use nuget package of all semconv attributes but adds a transitive dependency to customer apps which can cause breaking changes

### Out of Scope

- Generation of metric names
- Generation of meters
- Generation of event names
- Generation of event data structs
- Generation of extension methods for activities including type checking
```
span.SetDiskIoDirectionTag("read")
// maps to
span?.SetTag("DiskIoDirection", "read");

```

- Generation of extension methods for spans including type checking
```
span.SetDiskIoDirectionAttribute("read")
// maps to
span.SetAttribute("DiskIoDirection", "read");

```

thompson-tomo:
Fixes #2856 
Design discussion issue #

## Changes

This PR introduces a new client package `OpenTelemetry.SemanticConventions.Abstractions` which is designed to ease the implementation of otel by reducing the need for implementors to be defining their own keys.

The approach here overcomes the limitation of the current approach by Scoping the generated code to the users namespace and the accessibility modifier being internal. At the same time it is only generating the necessary namespace.


thompson-tomo:
Keep open

thompson-tomo:
Still waiting for review feedback,

thompson-tomo:
Awaiting feedback

thompson-tomo:
Awaiting feedback

PR #2873, opened 2025-07-02
Support for Blazor diagnostics in Net10

pavelsavara:
## Changes

- adds `Microsoft.AspNetCore.Components` and `Microsoft.AspNetCore.Components.Server.Circuits` tracing sources into `AspNetCoreInstrumentationTracerProviderBuilderExtensions.AddAspNetCoreInstrumentation`
    - can be disabled by new `AspNetCoreTraceInstrumentationOptions.EnableAspNetCoreBlazorSupport` = false
- adds `Microsoft.AspNetCore.Components` and `Microsoft.AspNetCore.Components.Server.Circuits` into `AspNetCoreInstrumentationMeterProviderBuilderExtensions.AddAspNetCoreInstrumentation`

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/.publicApi/PublicAPI.Shipped.txt
@@ -3,6 +3,8 @@ OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions
 OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.AspNetCoreTraceInstrumentationOptions() -> void
 OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnableAspNetCoreSignalRSupport.get -> bool
 OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnableAspNetCoreSignalRSupport.set -> void
+OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnableAspNetCoreBlazorSupport.get -> bool
+OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnableAspNetCoreBlazorSupport.set -> void
 OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnrichWithException.get -> System.Action<System.Diagnostics.Activity!, System.Exception!>?
 OpenTelemetry.Instrumentation.AspNetCore.AspNetCoreTraceInstrumentationOptions.EnrichWithException.set -> void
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Instrumentation.AspNetCore/.publicApi/PublicAPI.Shipped.txt
martincostello, line 1: These should be in the `Unshipped` file.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/BasicTests.cs
```diff
@@ -1213,6 +1214,56 @@ void ConfigureTestServices(IServiceCollection services)
     }
 #endif
 
+#if NET10_0_OR_GREATER
```
martincostello: Do these tests pass if incorporated on top of #2822?

pavelsavara: The whole project is Net9 right now, so this ifdef makes code ignored until SDK upgrade.

I tested the test locally on Net9 and it passed because the this test is only testing the negative case.
I also use fake `ActivitySource` because I think that doing full Blazor app and integration test with browser client is overkill on this repo.

martincostello: ~~Sure, but as main doesn't compile for .NET 10 just checking that they actually pass if it does, which that PR adds the TFM for. I wouldn't have asked otherwise as the CI would fail if they didn't.~~

Just realised I'd pasted the wrong PR number.

pavelsavara: The new test in this PR would pass anyway as it's negative test.

The Blazor feature is getting more and more stable. There are some changes in Net10 preview 6 which is not released yet. After this repo is at Net10 preview 6, it would be possible to also implement positive test, hitting real activity source in the Blazor code.

When do you plan to merge Net10 PR ?

martincostello: I thinking more on the basis of _"does the test compile, pass and not break the CI"_ - there's no validation of that right now.

Current intention is not to merge that PR .NET 10 ships stable in November.

We can put a positive test into that PR, but some manual validation the negative test doesn't fail if the .NET SDK is updated to 10 would be good if the intention is to merge this before .NET 10 is released.

pavelsavara: > I thinking more on the basis of _"does the test compile, pass and not break the CI"

I added the test to run also on Net9, so that the code doesn't get stale until you merge the Net10 PR. There is no harm in that.

> if the intention is to merge this before .NET 10 is released.

I think people will use Net9 version of this library to play with Net10 SDK previews. 
I used it in context of Aspire this way. 

So I think merging it now would be good.

pavelsavara:
Paused this PR until https://github.com/dotnet/aspnetcore/issues/62555 and https://github.com/dotnet/aspnetcore/issues/62556 pass formal review

martincostello (approved):

martincostello:
Closing in favour of #3012.

Issue #2878, opened 2025-07-05
[feature request] Track exceptions from WCF services


AngleOSaxon:
### Component

OpenTelemetry.Instrumentation.Wcf

### What is the expected behavior?

When a WCF service throws an unhandled exception during the processing of a request, that exception is logged as an error and associated with the current span.

### Which alternative solutions or features have you considered?

We can sometimes capture the exception information by utilizing the existing clientside instrumentation via the `RecordException` option, but that requires instrumenting the client and having the `IncludeExceptionDetailInFaults` setting active.

The service can be updated to perform its own logging, but this would require additional work on each service we plan to instrument.

PR #2880, opened 2025-07-05
Allow WCF services to record exceptions if `RecordException` is set

AngleOSaxon:
Implements #2878 

## Changes

Adds an `IErrorHandler` implementation (`TracingErrorHandler`) that gets added to the `ChannelDispatcher` instances for any instrumented services.  If the `RecordExceptions` option is set, it associates the current `Activity` with `OperationContext.Current` during the initial message inspection.  When an exception is triggered, the `TracingErrorHandler` loads the `Activity` and adds the exception to it.

Although the documentation [specifically notes](https://learn.microsoft.com/en-us/dotnet/api/system.servicemodel.dispatcher.ierrorhandler?view=netframework-4.8.1#remarks) that the `HandleError` method should be used to perform activities such as logging, I'm performing the logging in the `ProvideFault` method.  This is being done for reasons of timing--`ProvideFault` is called before the message is returned to the client and therefore before the `Activity` is stopped in `TelemetryDispatchMessageInspector.BeforeSendReply`.  This allows the exception to be properly placed inside the recorded span, prevents it from being attached to an already completed `Activity`, and ensures the `OperationContext` with the relevant information still exists.

It also simplifies the testing considerably, because I have not been able to find an event that only fires after all `HandleError` implementations have been called, so I'm forced to simply rely on predefined waits to see if the error is recorded.

As far as I am aware, the biggest potential issue with using `ProvideFault` is it will block completing the request and sending the response to the client.  I don't expect that to be an issue here, but it will rely on the good behavior of the `ActivityListener.ExceptionRecorder` implementations.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Wcf/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Wcf/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added server instrumentation support for `RecordException` option.
+
 ## 1.12.0-beta.1
 
 Released 2025-May-06
--- a/src/OpenTelemetry.Instrumentation.Wcf/Implementation/TelemetryDispatchMessageInspector.cs
+++ b/src/OpenTelemetry.Instrumentation.Wcf/Implementation/TelemetryDispatchMessageInspector.cs
@@ -104,6 +104,11 @@ internal class TelemetryDispatchMessageInspector : IDispatchMessageInspector
                 }
             }
 
+            if (WcfInstrumentationActivitySource.Options.RecordException)
+            {
+                OperationContext.Current?.Extensions.Add(new WcfOperationContext(activity));
+            }
+
             if (textMapPropagator is not TraceContextPropagator)
             {
                 Baggage.Current = ctx.Baggage;
--- a/src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
+++ b/src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
@@ -1,0 +1,48 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NETFRAMEWORK
+using System.Diagnostics;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Dispatcher;
+
+namespace OpenTelemetry.Instrumentation.Wcf.Implementation;
+
+internal class TracingErrorHandler : IErrorHandler
+{
+    public bool HandleError(Exception error)
+    {
+        return false;
+    }
+
+    public void ProvideFault(Exception error, MessageVersion version, ref Message fault)
+    {
+        // By rights this should be in `HandleError` instead, which would keep it from
+        // interfering with the response to the client.
+        // However, by the time `HandleError` fires, the Activity has already been stopped
+        // so the error appears after the Activity has completed.  Additionally, sometimes
+        // the context has already been disposed or otherwise lost, preventing association
+        // at all.
+        // Also it becomes very difficult to unit-test, because there is no easy `ErrorsHandled`
+        // event to listen for before checking to see whether the error was logged.
+
+        if (!WcfInstrumentationActivitySource.Options?.RecordException ?? false)
+        {
+            return;
+        }
+
+        // OperationContext.Current *should* be reliable even in async calls at .NET 4.6.2+.
+        // In older versions it may not be.
+        var context = OperationContext.Current?.Extensions.FirstOrDefault(item => item is WcfOperationContext) as WcfOperationContext;
+        var activity = context?.Activity ?? WcfInstrumentationActivitySource.ActivitySource.StartActivity(WcfInstrumentationActivitySource.UnassociatedExceptionActivityName, ActivityKind.Internal);
+
+        activity?.AddException(error);
+
+        if (activity != context?.Activity)
+        {
+            activity?.Stop();
+        }
+    }
+}
+#endif
--- a/src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfOperationContext.cs
+++ b/src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfOperationContext.cs
@@ -1,0 +1,25 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics;
+using System.ServiceModel;
+
+namespace OpenTelemetry.Instrumentation.Wcf.Implementation;
+
+internal class WcfOperationContext : IExtension<OperationContext>
+{
+    public WcfOperationContext(Activity activity)
+    {
+        this.Activity = activity;
+    }
+
+    public Activity Activity { get; }
+
+    public void Attach(OperationContext owner)
+    {
+    }
+
+    public void Detach(OperationContext owner)
+    {
+    }
+}
--- a/test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
+++ b/test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
@@ -205,6 +205,116 @@ public class TelemetryDispatchMessageInspectorTests : IDisposable
             Assert.Empty(stoppedActivities);
         }
     }
+
+    [Theory]
+    [InlineData(true, true, true)]
+    [InlineData(true, true, false)]
+    [InlineData(true, false, true)]
+    [InlineData(false, true, true)]
+    [InlineData(true, false, false)]
+    [InlineData(false, false, true)]
+    [InlineData(false, true, false)]
+    [InlineData(false, false, false)]
+    public async Task RecordExceptionTest(
+        bool recordException,
+        bool triggerException,
+        bool runAsync)
+    {
+        List<Activity> stoppedActivities = [];
+        List<Activity> startedActivities = [];
+
+        List<Exception> recordedExceptions = [];
+        using var activityListener = new ActivityListener
+        {
+            ShouldListenTo = activitySource => true,
+            ActivityStarted = startedActivities.Add,
+            ActivityStopped = stoppedActivities.Add,
+        };
+
+        activityListener.ExceptionRecorder += (Activity activity, Exception ex, ref TagList tags) =>
+        {
+            recordedExceptions.Add(ex);
+        };
+
+        ActivitySource.AddActivityListener(activityListener);
+
+        var tracerProvider = Sdk.CreateTracerProviderBuilder()
+            .AddWcfInstrumentation(options =>
+            {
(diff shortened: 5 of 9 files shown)


src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+        // event to listen for before checking to see whether the error was logged.
+
+        if (!WcfInstrumentationActivitySource.Options?.RecordException ?? false)
+        {
+            return;
+        }
+
+        // OperationContext.Current *should* be reliable even in async calls at .NET 4.6.2+.
+        // In older versions it may not be.
+        var context = OperationContext.Current?.Extensions.FirstOrDefault(item => item is WcfOperationContext) as WcfOperationContext;
```
KalleOlaviNiemitalo: [IExtensionCollection\<T\>.Find\<E\>](https://learn.microsoft.com/dotnet/api/system.servicemodel.iextensioncollection-1.find?view=netframework-4.8) would be more concise.

AngleOSaxon: Fixed

src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfOperationContext.cs
```diff
@@ -0,0 +1,25 @@
...
+    public Activity Activity { get; }
+
+    public void Attach(OperationContext owner)
+    {
+    }
+
+    public void Detach(OperationContext owner)
+    {
+    }
+}
```
Kielek: I hope I do not overlooked anything, could you please decorate it with conditional compilation?

```suggestion

#if NETFRAMEWORK
using System.Diagnostics;
using System.ServiceModel;

namespace OpenTelemetry.Instrumentation.Wcf.Implementation;

internal class WcfOperationContext : IExtension<OperationContext>
{
    public WcfOperationContext(Activity activity)
    {
        this.Activity = activity;
    }

    public Activity Activity { get; }

    public void Attach(OperationContext owner)
    {
    }

    public void Detach(OperationContext owner)
    {
    }
}
#endif
```

src/OpenTelemetry.Instrumentation.Wcf/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added server instrumentation support for `RecordException` option.
```
Kielek, line 5:
```suggestion
* Added server instrumentation support for `RecordException` option.
  ([#2880](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2880))
```

Kielek (commented):
@mconnew, is there any chance that you find some time to review general concept in this PR?

test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
```diff
@@ -205,6 +205,116 @@ await client.ExecuteAsync(
...
+        {
+            ShouldListenTo = activitySource => true,
+            ActivityStarted = startedActivities.Add,
+            ActivityStopped = stoppedActivities.Add,
+        };
+
+        activityListener.ExceptionRecorder += (Activity activity, Exception ex, ref TagList tags) =>
+        {
+            recordedExceptions.Add(ex);
+        };
```
KalleOlaviNiemitalo, line 237: I considered whether this ExceptionRecorder should check that `activity.Source.Name` is as expected.  Just to verify that the exception is being recorded by the OpenTelemetry WCF instrumentation rather than any WCF built-in feature.  However, if WCF somehow got such a feature, then the `recordException: false` test cases would fail.  So it's not necessary to check the ActivitySource here.

test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
```diff
@@ -205,6 +205,116 @@ await client.ExecuteAsync(
...
+        }
+
+        Assert.NotEmpty(stoppedActivities);
+        Assert.Single(stoppedActivities);
+
+        var activity = stoppedActivities[0];
+
+        if (recordException && triggerException)
+        {
+            Assert.Collection(recordedExceptions, e => Assert.IsType<Exception>(e));
```
KalleOlaviNiemitalo, line 309: OK, this verifies the type of the exception is exactly Exception as thrown from `test/OpenTelemetry.Instrumentation.Wcf.Tests/WCF/Service.netfx.cs`, not translated to any FaultException.

src/OpenTelemetry.Instrumentation.Wcf/Implementation/TelemetryDispatchMessageInspector.cs
```diff
@@ -104,6 +104,11 @@ internal TelemetryDispatchMessageInspector(IDictionary<string, ActionMetadata> a
                 }
             }
 
+            if (WcfInstrumentationActivitySource.Options.RecordException)
+            {
+                OperationContext.Current?.Extensions.Add(new WcfOperationContext(activity));
+            }
```
KalleOlaviNiemitalo, line 110: I have been doing something similar, and yes, OperationContext.Current will be more reliable for this use than Activity.Current, because when the WCF dispatcher pauses an operation and resumes another operation, it restores only OperationContext and not the whole ExecutionContext.

My implementation additionally has an ICallContextInitializer that finds the (already started) Activity from OperationContext.Current and sets it as Activity.Current.  I added that because I suspected that  _throttling_ in the WCF dispatcher might mess up the AsyncLocal\<T\> states.  However, I haven't yet tested whether that actually happens.

This OpenTelemetry WCF instrumentation doesn't seem to have such an ICallContextInitializer. Perhaps one should be added; but this PR does not make the problem worse.

test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
```diff
@@ -205,6 +205,116 @@ await client.ExecuteAsync(
...
+            tracerProvider?.Shutdown();
+            tracerProvider?.Dispose();
+
+            WcfInstrumentationActivitySource.Options = null;
+        }
+
+        Assert.NotEmpty(stoppedActivities);
+        Assert.Single(stoppedActivities);
+
+        var activity = stoppedActivities[0];
```
matt-hensley: nit
```suggestion
        var activity = Assert.Single(stoppedActivities);
```

KalleOlaviNiemitalo:
Should mention the feature around here, rather than say it "is available only on the client side": <https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/44486139bfa11ea0c17affd156c579222b327f38/src/OpenTelemetry.Instrumentation.Wcf/README.md?plain=1#L232-L237>

AngleOSaxon:
> Should mention the feature around here, rather than say it "is available only on the client side":
> 
> https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/44486139bfa11ea0c17affd156c579222b327f38/src/OpenTelemetry.Instrumentation.Wcf/README.md?plain=1#L232-L237

Done

mconnew:
@kiel, sorry I missed the mention earlier. I'll take a look this week.

src/OpenTelemetry.Instrumentation.Wcf/Implementation/TelemetryDispatchMessageInspector.cs
mconnew: I believe you are correct that the throttles would mess up the AsyncLocal\<T\> state as it essentially maintains a list of continuations to run in order as throttles are released and the ExecutionContext is captured with context flow disabled. Definitely worth adding this as a follow up item.

src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    public void ProvideFault(Exception error, MessageVersion version, ref Message fault)
+    {
+        // By rights this should be in `HandleError` instead, which would keep it from
+        // interfering with the response to the client.
+        // However, by the time `HandleError` fires, the Activity has already been stopped
+        // so the error appears after the Activity has completed.  Additionally, sometimes
+        // the context has already been disposed or otherwise lost, preventing association
+        // at all.
+        // Also it becomes very difficult to unit-test, because there is no easy `ErrorsHandled`
```
mconnew, line 27: This is perfectly fine to do here. HandleError allows you to prevent the session/context/connection from being aborted, and ProvideFault enable you to provde a custom fault message. Neither one of them are designed specifically for what you're doing, so there's no "right" place to do this. HanndleError is no more the right place than ProvideFault, so if this method works better due to the activity being stopped, then it's the right method to use.

test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
mconnew: WCF in .NET Framework isn't adding anything like this. We were going to initially, but the goal posts kept moving on what telemetry was going to look like, and they keep changing in small subtle ways (evolving), so adding telemetry externally via this package is going to be the best place for it to exist for a long time. And the more time passes, the less interest there is in adding anything to .NET Framework as all the focus is on .NET.

src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            return;
+        }
+
+        // OperationContext.Current *should* be reliable even in async calls at .NET 4.6.2+.
+        // In older versions it may not be.
+        var context = OperationContext.Current?.Extensions.Find<WcfOperationContext>();
+        var activity = context?.Activity ?? WcfInstrumentationActivitySource.ActivitySource.StartActivity(WcfInstrumentationActivitySource.UnassociatedExceptionActivityName, ActivityKind.Internal);
+
+        activity?.AddException(error);
```
mconnew, line 40: Do you need the null coalescing here? I presume StartActivity is guaranteed to return a non-null activity, which means `activity` can't be null here. Same with the call to Stop below.

mconnew (approved):

mconnew:
I've approved the changes, but I don't have write access to this repo, so it's still blocked.

src/OpenTelemetry.Instrumentation.Wcf/Implementation/TracingErrorHandler.cs
KalleOlaviNiemitalo: > I presume StartActivity is guaranteed to return a non-null activity

It's not.  StartActivity returns null if the ActivitySource has no listeners or if no ActivityListener chooses to sample the activity.

PR #2882, opened 2025-07-07

mladjan-gadzic:
Fixes 

Historical PR:
- #2704

Design discussion issue 

## Changes

PartialActivityProcessor is added. It helps address visibility for long running spans or processes that shut down ungracefully by sending heartbeats (logs) while span is active and stop log when span ends. Initial heartbeat delay is set to 5s to reduce noise of short lasting spans. Log body is used to serialize span in json per OpenTelemetry specification in case collectors need to process it further.


jcmfernandes:
Hi @mladjan-gadzic, thanks for this PR! I'm working on a golang implementation, but I was unable to find the part of the specification covering the ingestion of partial spans. Can you please point me to it?

mladjan-gadzic:
> Hi @mladjan-gadzic, thanks for this PR! I'm working on a golang implementation, but I was unable to find the part of the specification covering the ingestion of partial spans. Can you please point me to it?

hi @jcmfernandes, thanks for the effort to make partial spans more visible!

afaik there is no specification that covers ingestion of partial spans. for our own usecase we've implemented custom collector that uses log record body, which has serialized span in json per spec, to recreate span if necessary.

this pr aims to provide span visibility during long-running jobs or ungraceful shutdown. how someone decides to use data from logs, specifically log record body, is up to them.

jcmfernandes:
Thanks for the swift reply @mladjan-gadzic!

So maybe I'm misunderstanding the goal here. I'm working with long-running processes, and I would like to visualize spans in grafana/datadog/x-ray/etc. before having to close them, as that may only happen in days/weeks. Isn't this the goal? If so, doesn't the specification need to cover how these partial spans are encoded and transferred? E.g., don't we need to define the meaning of [`"span.state": "heartbeat"`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2882/files#diff-a87bbb360bcf07e75f4359cea671aef6d6f5c1afdfea81b4818fbfaa84c1a84bR200)?

PR #2883, opened 2025-07-07
[infra] Fix Codecov baseline reporting

### LINKED ISSUE #2881: Codecov not uploading new baseline report leading to inaccurate reporting
It has emerged that Codecov is comparing pull request's to an out of date baseline resulting in a report which is not providing an accurate picture of the changes.

An example of this is https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2840 which shows a reduction in code coverage but in fact there should be no change as only the sdk was bumped.

### expectation
When a pull request is merged the results are uploaded & the pr Codecov summary should only be considering the changes in that PR.

pratstick:
# [updated] Fix Codecov baseline reporting by updating existing CI for main branch

Fixes: #2881

This PR fixes the issue where Codecov was comparing pull requests against outdated baselines, leading to inaccurate coverage reporting. The root cause was that no CI workflow was running on `main` branch pushes to upload fresh baseline coverage data.

- [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
- [N/A] Unit tests added/updated
- [N/A] Appropriate `CHANGELOG.md` files updated for non-trivial changes
- [N/A] Changes in public API reviewed (if applicable)


.github/workflows/ci-main.yml
```diff
@@ -0,0 +1,467 @@
+name: Main Branch CI
```
martincostello: What does this workflow do differently to the PR workflow? It would be much less to maintain if that just added an extra trigger, rather than duplicating everything.

pratstick: I thought I dont have to touch the PR workflow, so I decided to make a separate one.

pratstick: I will change it

pratstick: Fixed - removed the duplicate ci-main.yml workflow and consolidated everything into ci.yml.
Thanks for the feedback and sorry for the inconvenience.

Kielek (approved):
Checked with @CodeBlanch.

It should be fine to execute in this way. Merging any changes for GitHub actions definitions (including this one) should trigger codecov for whole repository. Merges related to scoped PRs should be responsible for refreshing data for its projects.

PR #2893, opened 2025-07-10
Add in alot more process attributes #2892

### LINKED ISSUE #2892: [feature request] update processruntime resource detectors
### Component

OpenTelemetry.Resources.ProcessRuntime

I want the information about my resources to be as rich as possible.

The process resource detector provides more of the attributes which can easily be obtained from the sdk

### Additional context
N/a

thompson-tomo:
Fixes #2892 

## Changes

The resource detector will now set the following attributes:

- process.executable.path
- process.working.directory
- process.args.count
- process.creation.time
- process.executable.name
- process.interactive
- process.title

With the below only set if enabled (defaults to false)

- process.command_line
- process.command_args

### DIFF
--- a/src/OpenTelemetry.Resources.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Process/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## Unreleased
 
-* Extended `ProcessDetector` to include more of the supported attributes.
+* Extended `ProcessDetector` to include more of the recommend/required attributes.
+ ([#2893](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2893))
+
+* Extended `ProcessDetector` to support sending the 2 opt-in command line attributes
+when the IncludeCommand property is true.
  ([#2893](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2893))
 
 ## 1.12.0-beta.1
--- a/src/OpenTelemetry.Resources.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Process/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Extended `ProcessDetector` to include more of the supported attributes.
+
 ## 1.12.0-beta.1
 
 Released 2025-May-06
--- a/src/OpenTelemetry.Resources.Process/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Process/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Extended `ProcessDetector` to include more of the supported attributes.
+ ([#2893](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2893))
 
 ## 1.12.0-beta.1
 
--- a/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
+++ b/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
@@ -8,18 +8,23 @@ namespace OpenTelemetry.Resources.Process;
 /// </summary>
 internal sealed class ProcessDetector : IResourceDetector
 {
+    private readonly ProcessDetectorOptions options;
+
+    internal ProcessDetector(ProcessDetectorOptions? options)
+    {
+        this.options = options ?? new ProcessDetectorOptions();
+    }
+
     /// <summary>
     ///     Detects the resource attributes for process.
     /// </summary>
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
-        var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
-        return new Resource(new List<KeyValuePair<string, object>>(2)
+        using var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+        var attributes = new List<KeyValuePair<string, object>>(11)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
-            new(ProcessSemanticConventions.AttributeProcessCommandLine, Environment.CommandLine),
-            new(ProcessSemanticConventions.AttributeProcessCommandArgs, Environment.GetCommandLineArgs()),
             new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
             new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString("O") ?? string.Empty),
             new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
@@ -30,15 +35,17 @@ internal sealed class ProcessDetector : IResourceDetector
 #if NET
             new(ProcessSemanticConventions.AttributeProcessPid, Environment.ProcessId),
             new(ProcessSemanticConventions.AttributeProcessExecPath, Environment.ProcessPath ?? string.Empty),
-        });
+        };
 #else
-            new(ProcessSemanticConventions.AttributeProcessPid, GetProcessPid()),
-        });
-        static int GetProcessPid()
-        {
-            using var process = System.Diagnostics.Process.GetCurrentProcess();
-            return process.Id;
-        }
+            new(ProcessSemanticConventions.AttributeProcessPid, currentProcess.Id),
+        };
 #endif
+        if (this.options.IncludeCommand)
+        {
+            attributes.Add(new(ProcessSemanticConventions.AttributeProcessCommandLine, Environment.CommandLine));
+            attributes.Add(new(ProcessSemanticConventions.AttributeProcessCommandArgs, Environment.GetCommandLineArgs()));
+        }
+
+        return new Resource(attributes);
     }
 }
--- a/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
+++ b/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
@@ -15,14 +15,13 @@ internal sealed class ProcessDetector : IResourceDetector
     public Resource Detect()
     {
         var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
-#pragma warning disable CA1305 // Specify IFormatProvider
         return new Resource(new List<KeyValuePair<string, object>>(2)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
             new(ProcessSemanticConventions.AttributeProcessCommandLine, Environment.CommandLine),
             new(ProcessSemanticConventions.AttributeProcessCommandArgs, Environment.GetCommandLineArgs()),
             new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
-            new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString() ?? string.Empty),
+            new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString("O") ?? string.Empty),
             new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
             new(ProcessSemanticConventions.AttributeProcessWorkingDir, Environment.CurrentDirectory),
 
@@ -32,7 +31,6 @@ internal sealed class ProcessDetector : IResourceDetector
             new(ProcessSemanticConventions.AttributeProcessPid, Environment.ProcessId),
             new(ProcessSemanticConventions.AttributeProcessExecPath, Environment.ProcessPath ?? string.Empty),
         });
-#pragma warning restore CA1305 // Specify IFormatProvider
 #else
             new(ProcessSemanticConventions.AttributeProcessPid, GetProcessPid()),
         });
--- a/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
+++ b/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
@@ -14,12 +14,25 @@ internal sealed class ProcessDetector : IResourceDetector
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
+        var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+#pragma warning disable CA1305 // Specify IFormatProvider
         return new Resource(new List<KeyValuePair<string, object>>(2)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
+            new(ProcessSemanticConventions.AttributeProcessCommandLine, Environment.CommandLine),
+            new(ProcessSemanticConventions.AttributeProcessCommandArgs, Environment.GetCommandLineArgs()),
+            new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
+            new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString() ?? string.Empty),
+            new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
+            new(ProcessSemanticConventions.AttributeProcessWorkingDir, Environment.CurrentDirectory),
+
+            new(ProcessSemanticConventions.AttributeProcessExecName, currentProcess.ProcessName),
+            new(ProcessSemanticConventions.AttributeProcessInteractive, Environment.UserInteractive),
 #if NET
             new(ProcessSemanticConventions.AttributeProcessPid, Environment.ProcessId),
+            new(ProcessSemanticConventions.AttributeProcessExecPath, Environment.ProcessPath ?? string.Empty),
         });
+#pragma warning restore CA1305 // Specify IFormatProvider
 #else
             new(ProcessSemanticConventions.AttributeProcessPid, GetProcessPid()),
         });
--- a/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
+++ b/src/OpenTelemetry.Resources.Process/ProcessDetector.cs
@@ -10,9 +10,9 @@ internal sealed class ProcessDetector : IResourceDetector
 {
     private readonly ProcessDetectorOptions options;
 
-    internal ProcessDetector(ProcessDetectorOptions? options)
+    internal ProcessDetector(ProcessDetectorOptions options)
     {
-        this.options = options ?? new ProcessDetectorOptions();
+        this.options = options;
     }
 
     /// <summary>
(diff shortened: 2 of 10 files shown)


src/OpenTelemetry.Resources.ProcessRuntime/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Removed `process.runtime.description` which has never been a support semconv attribute
```
Kielek: Please check this section: https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/resource/process.md#process-runtimes

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,12 +14,25 @@ internal sealed class ProcessDetector : IResourceDetector
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
+        var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+#pragma warning disable CA1305 // Specify IFormatProvider
```
martincostello: These CA13xx warnings should be fixed by using the invariant culture, not suppressed.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
```diff
@@ -19,39 +20,32 @@ internal sealed class ProcessRuntimeDetector : IResourceDetector
...
-        var frameworkDescription = RuntimeInformation.FrameworkDescription;
-        string netRuntimeName;
+        string netRuntimeVersion = string.Empty;
 #if NETFRAMEWORK
         var netFrameworkVersion = GetNetFrameworkVersionFromRegistry();
-#endif
-
-        var lastSpace = frameworkDescription.LastIndexOf(' ');
-        if (lastSpace != -1)
+#pragma warning disable CS8602 // Dereference of a possibly null reference.
```
martincostello: It's better to use `!` as the less verbose built-in language feature to do this if the code can't be refactored to not produce false-positives.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,11 +14,22 @@ internal sealed class ProcessDetector : IResourceDetector
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
+        var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
```
martincostello:
```suggestion
        using var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
```

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,11 +14,22 @@ internal sealed class ProcessDetector : IResourceDetector
...
         return new Resource(new List<KeyValuePair<string, object>>(2)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
+            new(ProcessSemanticConventions.AttributeProcessCommandLine, Environment.CommandLine),
+            new(ProcessSemanticConventions.AttributeProcessCommandArgs, Environment.GetCommandLineArgs()),
+            new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
```
martincostello: Should these be opt-in/opt-out in case the arguments passed to the process include any secrets?

martincostello: From [the documentation](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/):

> SHOULD NOT be collected by default unless there is sanitization that excludes sensitive data.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
Kielek, line 1: It will be great to have unreleased changes in dedicated PRs.
2 reasons for this:
1. Smaller PRs --> easier reviews --> faster merges
2. Changelog references contains references to located changes, There is no need to review unrelated stuff.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
Kielek: I am fully supporting these comments. Security is crucial here.

thompson-tomo: Changes made.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -8,26 +8,49 @@ namespace OpenTelemetry.Resources.Process;
...
+
     /// <summary>
     ///     Detects the resource attributes for process.
     /// </summary>
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
-        return new Resource(new List<KeyValuePair<string, object>>(2)
+        using var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+        var attributes = new List<KeyValuePair<string, object>>(2)
```
martincostello: The dictionary is having its capacity pre-allocated here, but the number hasn't been updated to account for the new properties.

src/OpenTelemetry.Resources.Process/ProcessDetectorOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+/// <remarks>Use this class to specify settings that control the behavior of the process resource detector.</remarks>
+public class ProcessDetectorOptions
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ProcessDetectorOptions"/> class.
+    /// </summary>
+    public ProcessDetectorOptions()
+    {
+        this.IncludeCommand = false;
+    }
```
martincostello: This is redundant as it's just setting the property to the default value.

src/OpenTelemetry.Resources.Process/ProcessDetectorOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether should the command used to start the process should be included as a resource attribute.
+    /// </summary>
+    /// <remarks> Should only be set to <c>true</c> if non-sensitive data will not be passed. Defaults to <c>false</c>.</remarks>
```
martincostello:
```suggestion
    /// <remarks>Should only be set to <c>true</c> if non-sensitive data will not be passed. Defaults to <c>false</c>.</remarks>
```

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -8,26 +8,49 @@ namespace OpenTelemetry.Resources.Process;
...
+
     /// <summary>
     ///     Detects the resource attributes for process.
     /// </summary>
     /// <returns>Resource with key-value pairs of resource attributes.</returns>
     public Resource Detect()
     {
-        return new Resource(new List<KeyValuePair<string, object>>(2)
+        using var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+        var attributes = new List<KeyValuePair<string, object>>(11)
```
martincostello, line 25:
```suggestion
        var attributes = new List<KeyValuePair<string, object>>(12)
```

thompson-tomo: Why 12 as per unit tests max is 11

martincostello: I just blindly counted how many `new(`s there were without accounting for the `#if NET`.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -8,26 +8,49 @@ namespace OpenTelemetry.Resources.Process;
...
+
+            new(ProcessSemanticConventions.AttributeProcessExecName, currentProcess.ProcessName),
+            new(ProcessSemanticConventions.AttributeProcessInteractive, Environment.UserInteractive),
 #if NET
             new(ProcessSemanticConventions.AttributeProcessPid, Environment.ProcessId),
-        });
+            new(ProcessSemanticConventions.AttributeProcessExecPath, Environment.ProcessPath ?? string.Empty),
+        };
 #else
             new(ProcessSemanticConventions.AttributeProcessPid, GetProcessPid()),
```
martincostello:
```suggestion
            new(ProcessSemanticConventions.AttributeProcessPid, currentProcess.Id),
```

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -8,26 +8,49 @@ namespace OpenTelemetry.Resources.Process;
...
-        });
+        };
         static int GetProcessPid()
         {
             using var process = System.Diagnostics.Process.GetCurrentProcess();
             return process.Id;
         }
```
martincostello:
```suggestion
```

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
thompson-tomo: No worries.

src/OpenTelemetry.Resources.Process/ProcessResourceBuilderExtensions.cs
```diff
@@ -15,10 +15,11 @@ public static class ProcessResourceBuilderExtensions
     /// Enables process resource detector.
     /// </summary>
     /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <param name="options">The <see cref="ProcessDetectorOptions"/> used to control the behaviour configured.</param>
     /// <returns>The instance of <see cref="ResourceBuilder"/> being configured.</returns>
-    public static ResourceBuilder AddProcessDetector(this ResourceBuilder builder)
+    public static ResourceBuilder AddProcessDetector(this ResourceBuilder builder, ProcessDetectorOptions? options = default)
```
Kielek: Public API should not be using optional parameters. It it not maintinable.
Cosnider 2 separte methodw
```csharp
    public static ResourceBuilder AddProcessDetector(this ResourceBuilder builder)
    public static ResourceBuilder AddProcessDetector(this ResourceBuilder builder, Action<ProcessDetectorOptions>? configure)
```

src/OpenTelemetry.Resources.Process/ProcessDetectorOptions.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.Resources.Process;
+
+/// <summary>
+/// Provides configuration options for the process resource detector.
+/// </summary>
+/// <remarks>Use this class to specify settings that control the behavior of the process resource detector.</remarks>
+public class ProcessDetectorOptions
```
Kielek, line 10: I think that it will deserve separate discussion/confirmation on the SIG.

There is a couple types of attributes
- required - always enabled, no option to opt-out
- conditionally required - probably the most complex cases, both opt-in/opt-out possible with different default values
- recommended - enabled by default, opt-out possible
- opt-in - disabled by default, opt-in possible

It is common pattern across all resource detectors, we need to create recommendation how to handle such cases for all resource detector packages.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
thompson-tomo: Are you asking me to split the small refactoring within the process namespace into a seperate PR. I recon it will be longer to split then if it is reviewed as is.

Also it is confusing as to why it is in a separate package. To me the packages should be based on the primary namespace, you could have seperate methods for each entity/resource.

Kielek: Yes, I am asking for to split refactor. See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07105a510a5e8c93afd84c48f9b9dc4b6127b36e/CONTRIBUTING.md#tips-and-best-practices-for-pull-requests

src/OpenTelemetry.Resources.Process/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Extended `ProcessDetector` to include more of the supported attributes.
```
Kielek: PR link. It is applicable for all PRs with significant changes.

src/OpenTelemetry.Resources.Process/ProcessResourceBuilderExtensions.cs
thompson-tomo: Please add an analyzer rule to enforce that rulenas alot of apis have default parameters.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
thompson-tomo: Yes I have read it & I would say that I was compliant based on "single concern/issue" which was that resource detector for the process namespace should be accurately setting as many attributes as possible.

It is confusing that the runtime is in a seperate package given that pattern is not followed elsewhere of a package per resource. It should be merged

src/OpenTelemetry.Resources.Process/ProcessResourceBuilderExtensions.cs
```diff
@@ -17,8 +17,32 @@ public static class ProcessResourceBuilderExtensions
...
         Guard.ThrowIfNull(builder);
-        return builder.AddDetector(new ProcessDetector());
+        return builder.AddDetector(new ProcessDetector(options));
+    }
+
+    /// <summary>
+    /// Enables process resource detector.
+    /// </summary>
+    /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <param name="configure">Optional callback action for configuring <see cref="ProcessDetectorOptions"/> which control's the behaviour of the resource detector.</param>
```
martincostello:
```suggestion
    /// <param name="configure">Callback action for configuring <see cref="ProcessDetectorOptions"/> which controls the behavior of the resource detector.</param>
```

src/OpenTelemetry.Resources.Process/ProcessResourceBuilderExtensions.cs
```diff
@@ -17,8 +17,32 @@ public static class ProcessResourceBuilderExtensions
...
     public static ResourceBuilder AddProcessDetector(this ResourceBuilder builder)
+    {
+        return builder.AddProcessDetector(new ProcessDetectorOptions());
+    }
+
+    /// <summary>
+    /// Enables process resource detector.
+    /// </summary>
+    /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <param name="options">The <see cref="ProcessDetectorOptions"/> which control's the behaviour of the resource detector.</param>
```
martincostello:
```suggestion
    /// <param name="options">The <see cref="ProcessDetectorOptions"/> which controls the behavior of the resource detector.</param>
```

src/OpenTelemetry.Resources.Process/ProcessResourceBuilderExtensions.cs
```diff
@@ -17,8 +17,32 @@ public static class ProcessResourceBuilderExtensions
...
         Guard.ThrowIfNull(builder);
-        return builder.AddDetector(new ProcessDetector());
+        return builder.AddDetector(new ProcessDetector(options));
+    }
+
+    /// <summary>
+    /// Enables process resource detector.
+    /// </summary>
+    /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <param name="configure">Optional callback action for configuring <see cref="ProcessDetectorOptions"/> which controls the behavior of the resource detector.</param>
```
martincostello, line 40: I wouldn't call this optional because it's not an optional parameter - it's just that it allows null.

Personally, I don't think it makes sense to allow null here. There's an overload that doesn't take an options, so you'd do `AddProcessDetector()` rather than `AddProcessDetector(null)` (and that's ambiguous unless you know the type of assign null to a variable of type `Action<ProcessDetectorOptions>?`).

thompson-tomo: Looking at the code this approach here it is consistent with the most common approach. There is 8 instances which are not nullable but an additional 34 nullable. Mentioning varies based on the library.

ysolomchenko:
Comments from this [PR](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2895) also apply here regarding the README.md, PR title, and PR description.

I also suggest adding a note to CHANGELOG.md specifying which attributes can be optionally enabled.

src/OpenTelemetry.Resources.Process/ProcessDetectorOptions.cs
ysolomchenko: I believe it would make sense to simplify this into just two categories (even the "required" attributes might not be strictly necessary for end users in all cases):

- Enabled by default and can be disabled: includes required, conditionally required, and recommended attributes 
- Disabled by default and can be enabled: includes opt-in attributes 

I recall seeing an issue or discussion around making resource detector attributes configurable, but I can't seem to find it right now. This approach could align well with that direction and provide a consistent pattern across all resource detectors.

The difference between these types is more about how strongly they are expected to be implemented in the package:
- `required` – MUST be implemented
- `conditionally required` – MUST be implemented if the condition is met and it’s feasible to do
- `recommended` – SHOULD be implemented
- `opt-in` – CAN be implemented

here is documentation for this https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
Kielek: It has even own section on the semantic convention. I am tend to keep it as is.

thompson-tomo: So the process namespace is the only one that follows that idea of 1 attribute group = 1 package but multiple have multiple detectors producing different attribute groups within the 1 package & 1 even has 1 detector producing different attribute groups based on the machine.

thompson-tomo:
Still awaiting feedback

ysolomchenko:
@thompson-tomo , @Kielek, I created an issue for standardizing attribute enablement: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2962

@thompson-tomo, if you want to move this topic forward faster, please participate in the SIG and propose your solution.

I also recommend creating sub-issues for implementing attribute enable/disable functionality for each resource detector. As @Kielek mentioned: Smaller PRs --> easier reviews --> faster merges.

src/OpenTelemetry.Resources.Process/ProcessDetectorOptions.cs
thompson-tomo: This question has been addressed by updated docs by semconv https://opentelemetry.io/docs/specs/semconv/general/attribute-requirement-level/#requirement-levels

PR #2895, opened 2025-07-10
Add network info for hosts #2894

### LINKED ISSUE #2894: [feature request] add in network info for host
### Component

OpenTelemetry.Resources.Host

_No response_

Both ip & mac addresses are reported as attributes.

N/a

### Additional context

_No response_

thompson-tomo:
Fixes #2894 
Design discussion issue #

## Changes

Allows both the ip addresses as well as the mac addresses of the host to be set if the applicable setting is set to true (false by default). It also enables the host name to be explicitly set rather than looked up.

### DIFF
--- a/src/OpenTelemetry.Resources.Host/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Host/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 ## Unreleased
 
-* Added support for `host.ip` and `host.mac` attributes within the resource detector
+* Added ability for `host.ip` and `host.mac` attributes to be enabled
+which results in them being determined by the resource detector
+  ([#2895](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2895))
+
+* Added the ability to explicitly set the `host.name` rather than having it
+looked up from the system.
   ([#2895](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2895))
 
 ## 1.12.0-beta.1
--- a/src/OpenTelemetry.Resources.Host/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Host/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Added support for `host.ip` and `host.mac` attributes within the resource detector
+  ([#2895](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2895))
 
 ## 1.12.0-beta.1
 
--- a/src/OpenTelemetry.Resources.Host/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Host/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added support for `host.ip` and `host.mac` attributes within the resource detector
+
 ## 1.12.0-beta.1
 
 Released 2025-May-06
--- a/src/OpenTelemetry.Resources.Host/HostDetector.cs
+++ b/src/OpenTelemetry.Resources.Host/HostDetector.cs
@@ -257,6 +257,11 @@ internal sealed class HostDetector : IResourceDetector
 
     private string? GetMachineName()
     {
+        if (!string.IsNullOrEmpty(this.options.Name))
+        {
+            return this.options.Name;
+        }
+
         try
         {
             return Environment.MachineName;
--- a/src/OpenTelemetry.Resources.Host/HostDetector.cs
+++ b/src/OpenTelemetry.Resources.Host/HostDetector.cs
@@ -2,6 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics;
+using System.Globalization;
+using System.Net;
+using System.Net.NetworkInformation;
 #if !NETFRAMEWORK
 using System.Runtime.InteropServices;
 #endif
@@ -93,6 +96,37 @@ internal sealed class HostDetector : IResourceDetector
             if (machineId != null && !string.IsNullOrEmpty(machineId))
             {
                 attributes.Add(new(HostSemanticConventions.AttributeHostId, machineId));
+            }
+
+            try
+            {
+                var hostEntry = Dns.GetHostEntry(Environment.MachineName);
+                var ips = hostEntry.AddressList.Where(x => !IPAddress.IsLoopback(x))
+                    .Select(x => x.ToString())
+                    .Distinct()
+                    .Where(x => !string.IsNullOrEmpty(x))
+                    .ToArray();
+                attributes.Add(new(HostSemanticConventions.AttributeHostIp, ips));
+            }
+            catch (Exception ex)
+            {
+                HostResourceEventSource.Log.ResourceAttributesExtractException(nameof(HostDetector), ex);
+            }
+
+            try
+            {
+                var nics = NetworkInterface.GetAllNetworkInterfaces();
+                var macs = nics.Select(x => string.Join(":", x.GetPhysicalAddress()
+                        .GetAddressBytes()
+                        .Select(y => y.ToString("X2", CultureInfo.InvariantCulture))))
+                    .Distinct()
+                    .Where(x => !string.IsNullOrEmpty(x))
+                    .ToArray();
+                attributes.Add(new(HostSemanticConventions.AttributeHostMac, macs));
+            }
+            catch (Exception ex)
+            {
+                HostResourceEventSource.Log.ResourceAttributesExtractException(nameof(HostDetector), ex);
             }
 
             return new Resource(attributes);
--- a/src/OpenTelemetry.Resources.Host/HostDetector.cs
+++ b/src/OpenTelemetry.Resources.Host/HostDetector.cs
@@ -27,18 +27,21 @@ internal sealed class HostDetector : IResourceDetector
     private readonly Func<IEnumerable<string>> getFilePaths;
     private readonly Func<string?> getMacOsMachineId;
     private readonly Func<string?> getWindowsMachineId;
+    private readonly HostDetectorOptions options;
 
     /// <summary>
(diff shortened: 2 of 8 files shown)


src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -95,6 +98,37 @@ public Resource Detect()
...
 
+            try
+            {
+                var hostEntry = Dns.GetHostEntry(Environment.MachineName);
+                var ips = hostEntry.AddressList.Where(x => !IPAddress.IsLoopback(x))
+                    .Select(x => x.ToString())
+                    .Distinct()
+                    .Where(x => !string.IsNullOrEmpty(x))
+                    .ToArray();
+                attributes.Add(new(HostSemanticConventions.AttributeHostIp, ips));
```
martincostello, line 120: According [to the documentation](https://opentelemetry.io/docs/specs/semconv/resource/host/) these are both opt-in, so shouldn't be collected by default.

src/OpenTelemetry.Resources.Host/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added support for `host.ip` and `host.mac` attributes within the resource detector
```
Kielek: Missing link to PR. Please check other entries.

src/OpenTelemetry.Resources.Host/HostDetectorOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the mac ID of the host is to be included as a resource attributes.
+    /// </summary>
+    public bool IncludeMac { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the ip of the host is to be included as a resource attributes.
+    /// </summary>
+    public bool IncludeIp { get; set; }
```
martincostello: I believe `IncludeIP` would be the correct casing by .NET conventions (e.g. [`IPAddress`](https://learn.microsoft.com/dotnet/api/system.net.ipaddress), [`IPEndPoint`](https://learn.microsoft.com/dotnet/api/system.net.ipendpoint)).

src/OpenTelemetry.Resources.Host/HostDetectorOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace OpenTelemetry.Resources.Host;
+
+/// <summary>
+/// Provides options for configuring the host resource detector.
+/// </summary>
+public class HostDetectorOptions
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the mac ID of the host is to be included as a resource attributes.
```
martincostello:
```suggestion
    /// Gets or sets a value indicating whether the MAC addresses of the host are to be included as a resource attributes.
```

src/OpenTelemetry.Resources.Host/HostDetectorOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public bool IncludeMac { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the ip of the host is to be included as a resource attributes.
```
martincostello:
```suggestion
    /// Gets or sets a value indicating whether the IP addresses of the host are to be included as a resource attributes.
```

src/OpenTelemetry.Resources.Host/HostResourceBuilderExtensions.cs
```diff
@@ -15,10 +15,11 @@ public static class HostResourceBuilderExtensions
     /// Enables host resource detector.
     /// </summary>
     /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <param name="options">The <see cref="HostDetectorOptions"/> used to control the behaviour of the resource detector.</param>
```
martincostello:
```suggestion
    /// <param name="options">The <see cref="HostDetectorOptions"/> used to control the behavior of the resource detector.</param>
```

src/OpenTelemetry.Resources.Host/HostDetectorOptions.cs
```diff
@@ -0,0 +1,20 @@
...
+/// </summary>
+public class HostDetectorOptions
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the MAC addresses of the host are to be included as a resource attributes.
+    /// </summary>
+    public bool IncludeMac { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the IP addresses of the host ares to be included as a resource attributes.
```
martincostello:
```suggestion
    /// Gets or sets a value indicating whether the IP addresses of the host are to be included as a resource attributes.
```

ysolomchenko:
I suggest changing the PR title to: [Resources.Host] Add opt-in network attributes

ysolomchenko:
Update README.md: add an entry about `host.ip` and `host.mac` attributes, with an example of how to add them.

ysolomchenko:
Also, please update in README.md, CHANGELOG.md, and in the PR description that [host.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/) can now be set manually (allowed by the semantic conventions)

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -82,28 +93,50 @@ internal HostDetector(
...
         {
-            // Handling InvalidOperationException due to https://learn.microsoft.com/en-us/dotnet/api/system.environment.machinename#exceptions
-            HostResourceEventSource.Log.ResourceAttributesExtractException(nameof(HostDetector), ex);
+            var nics = NetworkInterface.GetAllNetworkInterfaces();
+            var macs = nics?.Select(x => string.Join(":", x.GetPhysicalAddress()
+                    .GetAddressBytes()
+                    .Select(y => y.ToString("X2", CultureInfo.InvariantCulture))))
+                .Distinct()
+                .Where(x => !string.IsNullOrEmpty(x))
+                .ToArray();
```
ysolomchenko, line 132: Fix `tabs` (indentation)

thompson-tomo: What do u mean? The getAddressBytes & the following select is running within the first select, hence the double indenting. If it was 1 it would given the wrong impression that the following distinct was operating at the same level.

ysolomchenko: Mark this comment as resolved

ysolomchenko (commented):
README.md: Update entry about attributes . List also host.ip and host.mac

example:
your application is running:
- **HostDetector**: `host.id` (when running on non-containerized systems), `host.name`, `host.ip`, `host.mac`.

thompson-tomo:
@ysolomchenko they are already listed

> Added ability for host.ip and host.mac attributes to be enabled which results in them being determined by the resource detector

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -13,8 +13,25 @@
...
+added & enabled in your project.
+
+### HostDetector
+
+|Attribute| Comment |
+|--- | --- |
+|[`host.id`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/#host-id)| Only set when running on non-containerized systems. |
+|[`host.ip`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/#host-ip)| Needs to be enabled via the `IncludeIP` setting. |
+|[`host.mac`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/#host-mac)| Needs to be enabled via the `IncludeMac` setting. |
+|[`host.name`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/#host-name)| Can be explicitly set via the `Name` setting. |
```
ysolomchenko: I think you should create an issue for this change, as it needs to be applied across all resource projects. It would also be great to have more discussion about what that table should contain. In my opinion, this change makes sense, but it should be moved to a separate PR

thompson-tomo: The problem is your commented solution is not obvious to a viewer of the readme and also lacks important details. I wouldn't be concerned about formatting consistency of readme given we don't have technical consistency/packaging with resource detectors. As such this should not block this pr.

thompson-tomo:
Still waiting for feedback.

Kielek (commented):
You are mixing 2 separate functionalities within one PR.

1. Possibility to override host.name - if you want to do, you can configure it manually on the Resource builder. It is not needed here.
2. Optional parameters - same comment as https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2893#discussion_r2197353238 - Please create issue with good description, examples, etc. You can bring the discussion to the SIG. When you will have it, we can proceed with review of PublicAPI.

thompson-tomo:
Disagree on 1, as a user i should have **1** location to manage the detection of my host resource information especially if i add in a library to manage that resource for me. By having it on the options it can disable the automated lookup as per the specification and hence making it more efficent due to the elimination of the un-necessary lookup of name. It avoids scenarios of where the defined name would be used and others it is not as the clear guidance is if you are using the host resource detector your options reside there. 
Further to that this approach is consistent with entities where attributes are needing to be grouped based on id's and then sorted into 1 of 2 buckets based on role. Hence is the most streamlined solution.

In relation 2, that is news to me that another issue is needing to be raised especially as the comment said discussion no mention of seperate issue to discuss a feature of a pr, usually the pr in question is just used as a disucssion point or even the issue but that was prematurely closed as duplicate when it wasn't  but rather a more focussed subset of functionality.  But any way sure additional issues can be raised.

Kielek (changes requested):
Blocking merge until there will be decision for public contact for all resource detectors for https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2893/files#r2197353238

PR #2902, opened 2025-07-11
[Resources.ProcessRuntime] cleanup logic

thompson-tomo:
Progresses #2892
Design discussion issue #

## Changes

This cleans up the logic for determining runtime to improve accuracy. This is done by reducing the amount of string manipulation which is occurring.

### DIFF
--- a/src/OpenTelemetry.Resources.ProcessRuntime/OpenTelemetry.Resources.ProcessRuntime.csproj
+++ b/src/OpenTelemetry.Resources.ProcessRuntime/OpenTelemetry.Resources.ProcessRuntime.csproj
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" Condition="'$(TargetFramework)' == 'netStandard2.0'" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
     <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" Condition="'$(TargetFramework)' == '$(NetFrameworkMinimumSupportedVersion)'" />
   </ItemGroup>


test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
```diff
@@ -12,18 +12,18 @@ public void TestProcessRuntimeAttributes()
     {
         var resource = ResourceBuilder.CreateEmpty().AddProcessRuntimeDetector().Build();
 
-        var resourceAttributes = resource.Attributes.ToDictionary(x => x.Key, x => (string)x.Value);
+        var resourceAttributes = resource.Attributes.ToDictionary(x => x.Key, x => x.Value);
```
Kielek, line 15: Could you please explain, why do you think that new version is better than old onr?
casting in 3 places instead of one?

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
```diff
@@ -20,38 +20,26 @@ internal sealed class ProcessRuntimeDetector : IResourceDetector
     public Resource Detect()
     {
         var frameworkDescription = RuntimeInformation.FrameworkDescription;
-        string netRuntimeName;
+        string netRuntimeVersion = Environment.Version.ToString();
```
ysolomchenko: NIT: Use `var` for initialization instead of `string` to maintain file consistency. Applies in several places.

test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
thompson-tomo: This approach enables the type of the value to be checked to ensure it is compliant as "6" is not the same as 6 which would have in old solution been converted just for the test.

thompson-tomo:
Previous solution focused on doing string manipulation of the description rather than relying on constants to determine the framework when possible with the exception being when compiled for NetStandard as that could be used by both.

thompson-tomo:
Awaiting feedback.

test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
```diff
@@ -12,18 +12,24 @@ public void TestProcessRuntimeAttributes()
...
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeName]);
         Assert.Equal(".NET Framework", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeName]);
-        Assert.Matches(@"^4.[1-9](.[1-2])?$", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
+        Assert.Matches(@"^4.(0.0|[5-9](.[1-2]))?$", (string)resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
 #else
-        Assert.Matches(@"^\.NET \d+\.\d+\.\d+(\-(preview|rc)\.\d+\.\d+\.\d+)?$", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
+        Assert.Matches(@"^\.NET \d+\.\d+\.\d+(\-(preview|rc)\.\d+\.\d+\.\d+)?$", (string)resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
```
Kielek: It applies in all cases, if you decided to check types.

```suggestion
        var processRuntimeDescription = Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
        Assert.Matches(@"^\.NET \d+\.\d+\.\d+(\-(preview|rc)\.\d+\.\d+\.\d+)?$", processRuntimeDescription);
```

src/OpenTelemetry.Resources.ProcessRuntime/README.md
```diff
@@ -13,8 +13,24 @@
 > Resources detected by this packages are defined by [experimental semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/resource/process.md#process-runtimes).
 > These resources can be changed without prior notification.
 
+## Attribute Utilization
```
Kielek: Please revert whole file.

For sure separate PR/PRs, for all Resources.* packages. I would like to have readmes consistent. I cam agree that some links to semantic convention might be useful, but it should not go to the registry. Implementation realays on very specific resource detector description. ---> links should go to https://opentelemetry.io/docs/specs/semconv/resource/process/#process-runtimes

test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
```diff
@@ -12,18 +12,24 @@ public void TestProcessRuntimeAttributes()
...
 
 #if NETFRAMEWORK
-        Assert.Matches(@"^\.NET Framework \d+\.\d+\.\d+\.\d+$", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
+        Assert.Matches(@"^\.NET Framework \d+\.\d+\.\d+\.\d+$", (string)resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeDescription]);
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeName]);
         Assert.Equal(".NET Framework", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeName]);
-        Assert.Matches(@"^4.[1-9](.[1-2])?$", resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
+        Assert.IsType<string>(resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
+        Assert.Matches(@"^4.(0.0|[5-9](.[1-2]))?$", (string)resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
```
Kielek: What is the use case for `4.0.0`? Especially in the context of supported versions https://dotnet.microsoft.com/en-us/learn/dotnet/what-is-dotnet-framework (without .NET Framework 3.5).

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
```diff
@@ -20,38 +20,31 @@ internal sealed class ProcessRuntimeDetector : IResourceDetector
...
-            netRuntimeName = "unknown";
-#if NETFRAMEWORK
-            netFrameworkVersion ??= "unknown";
-#endif
+            netRuntimeVersion = netFrameworkVersion;
         }
+#elif NET
+        var netRuntimeName = ".NET";
+#else
+        var pos = frameworkDescription.LastIndexOf(netRuntimeVersion, StringComparison.InvariantCultureIgnoreCase);
```
Kielek: I thini that you should be able to detect if .NET/.NET Fx in runtime. possibly, you can call GetNetFrameworkVersionFromRegistry here if .NET Fx.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
```diff
@@ -20,38 +20,31 @@ internal sealed class ProcessRuntimeDetector : IResourceDetector
     public Resource Detect()
     {
         var frameworkDescription = RuntimeInformation.FrameworkDescription;
-        string netRuntimeName;
+        var netRuntimeVersion = Environment.Version.ToString();
```
Kielek: .NET Framework case - for most/all cases this call is redundant.

src/OpenTelemetry.Resources.ProcessRuntime/README.md
thompson-tomo: As mentioned previously those documents are slated for removal so we shouldn't be linking to them. In fact the attributes table on the mentioned page also links to the same pages which I have done here.

test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
thompson-tomo: It was a matter of switching to a valid regex for dotnet, hence removing the options of 4.1.0 etc which are not valid frameworks. Happy to remove net 4.0.0 if you would like.

Kielek: I suppose that this should cover all supported cases:

```suggestion
        Assert.Matches(@"^4.([6-9](.[1-2]))?$", (string)resourceAttributes[ProcessRuntimeSemanticConventions.AttributeProcessRuntimeVersion]);
```

test/OpenTelemetry.Resources.ProcessRuntime.Tests/ProcessRuntimeDetectorTests.cs
thompson-tomo: Done.

src/OpenTelemetry.Resources.ProcessRuntime/ProcessRuntimeDetector.cs
thompson-tomo: Has been reworked.

src/OpenTelemetry.Resources.ProcessRuntime/OpenTelemetry.Resources.ProcessRuntime.csproj
```diff
@@ -15,6 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" Condition="'$(TargetFramework)' == 'netStandard2.0'" />
```
martincostello:
```suggestion
    <PackageReference Include="Microsoft.Win32.Registry" Version="5.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
```

martincostello:
You don't need to instantly re-tag me for a review - I get GitHub notifications when PRs/issues I interact with change.

PR #2918, opened 2025-07-16
[Infra] Remove AssemblyInfo files

martincostello:
## Changes

Remove the repetitive `AssemblyInfo.cs` files from the projects, particularly with the strong-naming key, and instead generate the relevant attributes using MSBuild.


martincostello:
Moving back to draft as I added a test to verify the signing (and key) and I'm not getting the results I expected for all the assemblies.

martincostello:
> Moving back to draft as I added a test to verify the signing (and key) and I'm not getting the results I expected for all the assemblies.

Hangfire's tests opt-out of strong-naming, which is why I was getting a failure.

test/Shared/StrongNameTests.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma warning disable IDE0005
```
TimothyMothra: please add the comment here that describes what is being disabled.

test/Shared/StrongNameTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        // Arrange
+        var assembly = typeof(StrongNameTests).Assembly;
+        var name = assembly.GetName();
+
+        if (name.Name == "OpenTelemetry.Instrumentation.Hangfire.Tests")
+        {
+            // Hangfire tests are not strong named (SignAssembly=false), so skip
+            return;
+        }
```
TimothyMothra: if Hangfire later adds strong naming, is there any risk of getting false positives from this test?

I wonder if it would be more resilient to have a property in individual projects to opt ouf of this test. Something like: `<SkipStrongNameValidation>true</SkipStrongNameValidation>`

test/Shared/StrongNameTests.cs
martincostello: Will do; TL;DR - in some projects `Xunit` seemed to be already available as an implicit using, which caused the warning.

build/Common.props
```diff
@@ -18,6 +17,17 @@
...
+    <StrongNamePublicKey>002400000480000094000000060200000024000052534131000400000100010051c1562a090fb0c9f391012a32198b5e5d9a60e9b80fa2d7b434c9e5ccb7259bd606e66f9660676afc6692b8cdc6793d190904551d2103b7b22fa636dcbb8208839785ba402ea08fc00c8f1500ccef28bbf599aa64ffb1e1d5dc1bf3420a3777badfe697856e9d52070a50c3ea5821c80bef17ca3acffa28f89dd413f096f898</StrongNamePublicKey>
+  </PropertyGroup>
+
+  <ItemGroup Condition=" '$(AssemblyIsCLSCompliant)' != '' ">
+    <AssemblyAttribute Include="System.CLSCompliant">
+      <_Parameter1>$(AssemblyIsCLSCompliant)</_Parameter1>
+      <_Parameter1_IsLiteral>true</_Parameter1_IsLiteral>
+    </AssemblyAttribute>
+  </ItemGroup>
+
```
TimothyMothra, line 30: 👍

test/Shared/StrongNameTests.cs
TimothyMothra: one more nit; can you add a comment with this test that references the property that can be used to disable the test?

I'm thinking if someone creates a new project, this test is failing, what do they do next.
If we have some instructions... do X to enable strong naming, or use Skip Property to disable this check.

I want to avoid a situation where someone tries to introduce a custom hack to disable the test.

martincostello:
> is there anywhere to document that by default assembly internals are visible to

Where would such documentation go, other than a comment immediately above the declaration of the convention that adds it?

test/Shared/StrongNameTests.cs
Kielek: I removed this pragma in 3bfd5d4201c8d51948a86150df3ced87f53f0415. Only one Gcp package was affected - containing one file.
We can add Xunit globally for impicit usings in tests, but in the separate PR.

Kielek (approved):

PR #2920, opened 2025-07-17
CONTRIBUTING - update RR with new example

Kielek:
Replaces #2919, I needed to make hard cleanup in my fork after some tests

## Changes

CONTRIBUTING - update PR with example of new project
the previous one is outdated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -218,8 +218,8 @@ shipped through CI.
 > [!NOTE]
 > It is generally helpful to reference a previous pull request when adding a new
   project to the repository. A good example to follow is the pull request which
-  added the `OpenTelemetry.Resources.OperatingSystem` project:
-  [#1943](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1943).
+  added the `OpenTelemetry.OpAmp.Client` project:
+  [#2917](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2917).
 
 * Based on what your project is, you may need to depend on the [OpenTelemetry
   SDK](https://www.nuget.org/packages/OpenTelemetry) or the [OpenTelemetry


martincostello (approved):

CONTRIBUTING.md
```diff
@@ -218,8 +218,8 @@ shipped through CI.
 > [!NOTE]
 > It is generally helpful to reference a previous pull request when adding a new
   project to the repository. A good example to follow is the pull request which
-  added the `OpenTelemetry.Resources.OperatingSystem` project:
-  [#1943](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1943).
+  added the `OpenTelemetry.OpAmp.Client` project:
```
thompson-tomo, line 221: I would call that a bad example as the build consumes over 6hrs of build time & runs over 200 actions hence not efficient. I wouldn't want more projects following that example, hence a better example should be selected.

thompson-tomo (changes requested):

CONTRIBUTING.md
Kielek: It is about document current state of the implementation. If you see possibility to improve, we are opened for this.
If changes are touch only one project, only small subset of builds is executed. See: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/.github/workflows/ci.yml#L25-L74

thompson-tomo: i am aware of that massive if etc. If you look at: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2926 you will see what i am seeing as that did a massive run and we are yet to see a case of a non massive run. The only change to a file not in the opamp folder is the removal of a project from the verify-aot step but it still ran a massive long list of actions hence far from efficent. It appears to be simply treating any yaml changes in the github folder as grounds for an entire build, hence lots of room for improvement. I have already put in a PR to adopt modules which detangles responsibility for build process and is all documented.

thompson-tomo (approved):

PR #2885, opened 2025-07-09
Update OS to follow latest semconv #1958

thompson-tomo:
Fixes #1958 
Design discussion issue #

## Changes

This serves as the prototype for the changes to semconv, leverages the description field correctly and adds support for more os.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #2923, opened 2025-07-18
[infra] Adjust stale pr handling to cater for proposals/drafts

thompson-tomo:
## Changes

This removes stale label from PR when activity occurs and from draft pr's


.github/workflows/stale.yml
```diff
@@ -25,3 +25,5 @@ jobs:
           days-before-issue-stale: -1
           days-before-pr-close: 7
           days-before-issue-close: -1
+          exempt-draft-pr: true
+          remove-pr-stale-when-updated: true
```
Kielek, line 30: This one should be fine.

.github/workflows/stale.yml
```diff
@@ -25,3 +25,5 @@ jobs:
           days-before-issue-stale: -1
           days-before-pr-close: 7
           days-before-issue-close: -1
+          exempt-draft-pr: true
```
Kielek, line 28: I do not want to manually close abadoned draft PRs. It is possible to reopen/recreate PR when it is needed.
``

```suggestion
```

Kielek (changes requested):

.github/workflows/stale.yml
thompson-tomo: Draft pr's should have different requirements as they can be being used for proposals, hence have implemented different rules for drafts.

TimothyMothra: What's the underlying reason for this change? 
Is there a recent problem that you're trying to solve?

I think exempting drafts would be problematic as it now becomes a manual task to watch and close. 
It currently requires 14 days of idle activity before any PR is closed.

I'm not sure that "proposals" is a good enough justification. Users can use Issues for proposals which don't expire as quickly. Code samples can be referenced in a draft pr even after it's been closed. And as @Kielek already mentioned, it's possible to reopen a PR when needed.

thompson-tomo: The problem is i should be able to make protypes to accompy semconv changes and 2 weeks of in-activity is no where near enough, time. By having them as draft, feedback can be provided, suggestions made to ensure that the implementation and spec is in alignment. Look at https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2885 where no feedback was provided other than changing 1 comparison.

> I think exempting drafts would be problematic as it now becomes a manual task to watch and close.
It currently requires 14 days of idle activity before any PR is closed.

The pr doesn't exempt drafts, it lengthens the period before they are closed. 

> And as @Kielek already mentioned, it's possible to reopen a PR when needed.

That is wrong, a pr can not be re-opened by the raiser see below from https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2885
image

thompson-tomo:
@rajkumar-rangaraj I am sorry but I fail to see any actual additional burden this change creates. It slightly increase the queue size but that is addressable via filters to ignore drafts.

One of the benefit of having it as a draft it is to shorten the cycle to get the change merged, just look at how long it is taking to get additional process/host attributes set on the resource detectors. Both of which have had issues around for over a year with no activity and it wasn't until the pr came around did a discussion start about configuring of opt-in attributes. 

At the same time having a draft pr enables easy identication of all the changes needed to achieve it & invites others to provide feedback etc.

rajkumar-rangaraj:
Closing in favor of #2953

PR #2872, opened 2025-07-02
[WIP] OpAMP Client

RassK:
## What

Is a early phase of OpAMP Client implementation for OTel agents. 
OpAMP spec is here: https://opentelemetry.io/docs/specs/opamp/

## ToDo for first alpha

* [ ] Add public api DTOs
* [ ] Add diagnostic logging (`System.Diagnostics.Tracing.EventSource`)
* [ ] Add implementation for sending large websockets frames
* [ ] Restructure public and internal DTOs.
* [x] Consider renaming the namespace (OpenTelemetry.OpAmp.Client) as there may be other extensions.
* [ ] Add readme

## ToDo for later

* [ ] Add configuration management (missing file based configuraions is a blocker)
* [ ] Add (static) plugins to extend custom messages and capabilities handling

## Hard requirements

* OTel .NET Auto Instrumentation requires minimal dependencies.
* Consider a second package for interfacing with` Micrososft.Extensions.*`

## Merge requirement checklist

* [ ] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #2926, opened 2025-07-18
[OpAMP.Client] Add support for plain http transport

RassK:
## Changes

Adds support for plain http transport (send & receive).


src/OpenTelemetry.OpAmp.Client/Protos/opamp.proto
```diff
@@ -0,0 +1,1061 @@
...
+// OpAMP: Open Agent Management Protocol (https://github.com/open-telemetry/opamp-spec)
+
+syntax = "proto3";
+
+package opamp.proto;
+
+import "anyvalue.proto";
+
+option go_package = "github.com/open-telemetry/opamp-go/protobufs";
+option csharp_namespace = "OpAmp.Protocol";
```
RassK: Still under review https://github.com/open-telemetry/opamp-spec/pull/243

src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
```diff
@@ -2,6 +2,5 @@
 
 ## Unreleased
 
-* Initial release of
-  `OpenTelemetry.OpAmp.Client`
-  project.
+* Initial release of `OpenTelemetry.OpAmp.Client` project.
+* Added support for OpAMP Plain Http transport.
```
martincostello: Also needs a link to this PR.

```suggestion
* Added support for OpAMP Plain HTTP transport.
```

src/OpenTelemetry.OpAmp.Client/FrameBuilder.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using Google.Protobuf;
+using OpAmp.Protocol;
+
+namespace OpenTelemetry.OpAmp.Client;
+
+internal class FrameBuilder : IFrameBuilder
```
martincostello: Unless there's inheritance happening, internal classes can be `sealed` as this can help the JIT create better code.

src/OpenTelemetry.OpAmp.Client/FrameBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        if (this.currentMessage != null)
+        {
+            throw new InvalidOperationException("Message base is already initialized.");
+        }
+
+        var message = new AgentToServer()
+        {
+            InstanceUid = this.instanceUid,
+            SequenceNum = ++this.sequenceNum,
```
martincostello, line 33: Just checking, the intention is to start at 1 and count up?

src/OpenTelemetry.OpAmp.Client/FrameBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+        };
+
+        this.currentMessage = message;
+        return this;
+    }
+
+    AgentToServer IFrameBuilder.Build()
```
martincostello, line 40: Why explicit rather than implicit?

src/OpenTelemetry.OpAmp.Client/FrameProcessor.cs
```diff
@@ -0,0 +1,123 @@
...
+    public void OnServerFrame(ReadOnlySequence<byte> sequence)
+    {
+        this.Deserialize(sequence);
+    }
+
+    private void Deserialize(ReadOnlySequence<byte> sequence)
+    {
+        var message = ServerToAgent.Parser.ParseFrom(sequence);
+
+        if (message.ErrorResponse != null)
```
martincostello, line 58: Would it be possible to collect all the messages to dispatch, and then send them together as a batch?

src/OpenTelemetry.OpAmp.Client/Listeners/IOpAmpMessage.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Listeners;
+
+internal interface IOpAmpMessage
+{
+}
```
martincostello, line 8: Is this just intended to be a marker interface?

src/OpenTelemetry.OpAmp.Client/Listeners/Messages/AgentIdentificationMessage.cs
```diff
@@ -0,0 +1,16 @@
...
+
+internal class AgentIdentificationMessage : IOpAmpMessage
+{
+    public AgentIdentificationMessage(AgentIdentification agentIdentification)
+    {
+        this.AgentIdentification = agentIdentification;
+    }
+
+    public AgentIdentification AgentIdentification { get; set; }
+}
```
martincostello, line 16: Could make all these be a bit less boiler-plate-y with primary constructors?

```suggestion
internal class AgentIdentificationMessage(AgentIdentification agentIdentification) : IOpAmpMessage
{
    public AgentIdentification AgentIdentification { get; set; } = agentIdentification;
}
```

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
```diff
@@ -0,0 +1,43 @@
...
+    public Uri ServerUrl { get; set; } = new("https://localhost:4320/v1/opamp");
+}
+
+/// <summary>
+/// Specifies the type of transport protocol to be used for communication.
+/// </summary>
+/// <remarks>This enumeration defines the available transport protocols for communication. Use <see
+/// cref="WebSocket"/> for WebSocket-based communication, or <see cref="Http"/> for
+/// HTTP-based communication.</remarks>
+#pragma warning disable SA1201 // Elements should appear in the correct order
```
martincostello: Fix rather than suppress?

src/OpenTelemetry.OpAmp.Client/OpenTelemetry.OpAmp.Client.csproj
```diff
@@ -13,4 +13,32 @@
...
+    <PackageReference Include="Grpc.Tools" Version="2.72.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <!-- .NET Framework supporting packages -->
+  <ItemGroup Condition="'$(TargetFramework)' == '$(NetFrameworkMinimumSupportedVersion)'">
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
+    <PackageReference Include="System.Collections.Immutable" Version="9.0.7" />
```
martincostello: Considering https://github.com/open-telemetry/opentelemetry-dotnet/pull/6327, I suggest starting with the lowest version possible (e.g. `8.0.0`).

src/OpenTelemetry.OpAmp.Client/Transport/Http/PlainHttpTransport.cs
```diff
@@ -0,0 +1,70 @@
...
+        byteContent.Headers.Add("Content-Type", "application/x-protobuf");
+
+        var response = await this.httpClient
+            .PostAsync(this.uri, byteContent, cancellationToken: token)
+            .ConfigureAwait(false);
+
+        if (!response.IsSuccessStatusCode)
+        {
+            throw new HttpRequestException($"Failed to send message: {response.ReasonPhrase}");
+        }
```
martincostello: Just use `EnsureSuccessStatusCode()`?

src/OpenTelemetry.OpAmp.Client/Transport/Http/PlainHttpTransport.cs
```diff
@@ -0,0 +1,70 @@
...
+            .ConfigureAwait(false);
+
+        this.processor.OnServerFrame(responseMessage.AsSequence());
+    }
+
+    public void Dispose()
+    {
+        this.httpClient?.Dispose();
+        this.handler?.Dispose();
+    }
```
martincostello, line 66: The `HttpClient` should be able to be constructed so that it can be told to dispose of the handler itself, then this instance doesn't need to do that or keep a reference to it.

src/OpenTelemetry.OpAmp.Client/FrameBuilder.cs
RassK: Should indicate it better as it contains full experience:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2872

src/OpenTelemetry.OpAmp.Client/FrameBuilder.cs
RassK: yes, as a sequence and not an index. Just in case rechecked protocol spec also, no suggested initial value there.

src/OpenTelemetry.OpAmp.Client/Listeners/IOpAmpMessage.cs
RassK: yes, due some partials in ServerToAgent aren't IMessages, just flags.

src/OpenTelemetry.OpAmp.Client/Listeners/Messages/AgentIdentificationMessage.cs
RassK: May want to leave it for now, since these are supposed to be public api dtos, but currently are unmapped from protos and needs rework anyway.

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
RassK: I was mistakenly looking that SA1201 doesn't like it either way, basically prohibits multiple classes in the same file, but didn't want to extract it (yet, or at least not into root).

src/OpenTelemetry.OpAmp.Client/FrameProcessor.cs
RassK: Basically `ServerToAgent` is a "batch" in this case. It's deconstructed (fan-out) into smaller bits, so all interested parties can subscribe only to a necessary part. One `ServerToAgent` does not contain multiple same types.

src/OpenTelemetry.OpAmp.Client/Transport/Http/PlainHttpTransport.cs
RassK: Then you need to suppress CA2000 manually. `disposeHandler` is there but static analysis does not pass.

RassK:
FYI: PTO 🌴, Will be back in 4th of August. Feel free to push changes and merge if critical stuff got resolved.  

(you can find context in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2872 if there is something looking very weird, there is kind of a big picture)

RassK:
Q: are you tracking aot incompatible libs somewhere? - seems protobuf is not compatible, therefore opamp also.

martincostello:
There's an issue in Google.Protobuf that broke some native AoT compatibility that I'm aware of: https://github.com/protocolbuffers/protobuf/issues/21824

RassK:
@martincostello do you see anything else that requires attention, since we wait for spec to commit changes.

src/OpenTelemetry.OpAmp.Client/FrameProcessor.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using OpAmp.Proto.V1;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Listeners;
+using OpenTelemetry.OpAmp.Client.Listeners.Messages;
+
+namespace OpenTelemetry.OpAmp.Client;
+
+internal class FrameProcessor
```
martincostello: Can this (and various other new types being added) be made `sealed`?

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace OpenTelemetry.OpAmp.Client;
+
+/// <summary>
+/// Specifies the type of transport protocol to be used for communication.
+/// </summary>
+/// <remarks>This enumeration defines the available transport protocols for communication. Use <see
+/// cref="WebSocket"/> for WebSocket-based communication, or <see cref="Http"/> for
+/// HTTP-based communication.</remarks>
+internal enum ConnectionType
```
martincostello, line 12: This should be in its own file.

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    WebSocket = 1,
+}
+
+internal class OpAmpClientSettings
+{
+    /// <summary>
+    /// Gets or sets the unique identifier for the current instance.
+    /// </summary>
+    public Guid InstanceUid { get; set; } = Guid.NewGuid(); // TODO: use Guid.CreateVersion7() with .NET 9+
```
martincostello: Just do this now?

.github/workflows/ci.yml
```diff
@@ -594,7 +594,6 @@ jobs:
       || contains(needs.detect-changes.outputs.changes, 'instrumentation-servicefabricremoting')
       || contains(needs.detect-changes.outputs.changes, 'instrumentation-sqlclient')
       || contains(needs.detect-changes.outputs.changes, 'instrumentation-stackexchangeredis')
-      || contains(needs.detect-changes.outputs.changes, 'opamp-client')
```
martincostello, line 597: How come being removed?

RassK: AoT does not work, errors pointing to Google.Protobuf. Did I miss something?

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
RassK: where would you put it? ... I did not like root.

.github/workflows/ci.yml
martincostello: Ah sorry, I didn't notice this was AoT specific.

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
martincostello: I would put it in `ConnectionType.cs` in the root, as otherwise I think the analyzers would complain about the style (having more than one type defined in the same file).

src/OpenTelemetry.OpAmp.Client/OpenTelemetry.OpAmp.Client.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(NetMinimumSupportedVersion);$(NetStandardMinimumSupportedVersion);$(NetFrameworkMinimumSupportedVersion)</TargetFrameworks>
+    <TargetFrameworks>$(NetMinimumSupportedVersion);net9.0;$(NetStandardMinimumSupportedVersion);$(NetFrameworkMinimumSupportedVersion)</TargetFrameworks>
```
martincostello: Newest should go first as some tooling (I think it's Rider) doesn't work as well otherwise.

```suggestion
    <TargetFrameworks>net9.0;$(NetMinimumSupportedVersion);$(NetStandardMinimumSupportedVersion);$(NetFrameworkMinimumSupportedVersion)</TargetFrameworks>
```

src/OpenTelemetry.OpAmp.Client/OpAmpClientSettings.cs
RassK: So far no errors like in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2926#discussion_r2216726464

In the [PoC branch](https://github.com/RassK/opentelemetry-dotnet-contrib/tree/opamp-client/src/OpenTelemetry.OpAmp.Client/Settings) I added them to `Settings` (since the subs would bloat root).
Those will become public and after that are not movable (not without creating a mess in namespaces).-
So need to get that internal / public boundary right.

src/OpenTelemetry.OpAmp.Client/OpenTelemetry.OpAmp.Client.csproj
```diff
@@ -13,4 +13,32 @@
     <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Google.Protobuf" Version="3.31.1" />
+    <PackageReference Include="Grpc.Tools" Version="2.72.0">
```
Kielek, line 18: Just a note: It will not work on ARM, if anybody decide to compile on this machine.

https://github.com/grpc/grpc/issues/38538

src/OpenTelemetry.OpAmp.Client/OpenTelemetry.OpAmp.Client.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>$(NetMinimumSupportedVersion);$(NetStandardMinimumSupportedVersion);$(NetFrameworkMinimumSupportedVersion)</TargetFrameworks>
+    <TargetFrameworks>net9.0;$(NetMinimumSupportedVersion);$(NetStandardMinimumSupportedVersion);$(NetFrameworkMinimumSupportedVersion)</TargetFrameworks>
```
Kielek, line 4: Note for myself, uid should be generated based on Guid.V7. It is available on .NET9+.

Kielek (approved):
LGTM. I will merge as is. It is good groundwork for further changes. I would like to avoid any more huge PRs.

Changes related to the proto namespace can be easily fixed later.

PR #2928, opened 2025-07-21
[AWSLambda] Add faas.instance and faas.max_memory

martincostello:
## Changes

Set the [`faas.instance`](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/faas/faas-spans.md) and [`faas.max_memory`](https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/model/faas/registry.yaml#L54) attributes for AWS Lambda functions, which are both recommended.


Kielek:
@philasmar, could you please check as you are seems the owner of https://github.com/aws/aws-lambda-dotnet? It seems that Philip is no longer affiliated to the AWS and it will be great to have someone with AWS expertise to cover ownership of this package.

Kielek (approved):
LGTM. It should be fine to add this for all SemConv versions based on https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0

Kielek:
@normj, FYI, as you declared that you can help with reviews here

normj (approved):
PR looks good. I was a worried at first of using the stream name for the `faas.instance` but I see that is what is being used in the Ruby and Python Lambda otel providers.

martincostello:
> using the stream name for the faas.instance

The main reason I used that was because it's the example in the Semantic Conventions, but good to know it's correct if that's what those other providers use.

Kielek:
@rypdal, @Oberon00, do you have any comments here?

rypdal:
> @rypdal, @Oberon00, do you have any comments here?

LGTM

rypdal (approved):

PR #2917, opened 2025-07-16
[OpAMP.Client] Project base

RassK:
## Changes

Provides a basis for a new OpAMP client project.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [x] Changes in public API reviewed (if applicable)

PR #2949, opened 2025-07-30
Introduce Assembly Metadata Resource Detector

austindrenski:
I'd like to offer a new resource detector that I've been using for a little while now that has a somewhat novel¹ approach for embedding/extracting build-time resource attributes via assembly metadata.

_¹ It feels a little silly calling it novel, since basically all of the heavy lifting is done by an [existing MSBuild target](https://github.com/dotnet/sdk/blob/v9.0.303/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets), but it differs from the current set of resource detectors which are all focused on a specific set of resource attributes, whereas this is instead intended to let users hook build-time details into telemetry._

## Changes

The Assembly Metadata Resource Detector uses an MSBuild target to include user-defined key-value pairs as assembly metadata that can be extracted at runtime by the `AssemblyMetadataDetector` type.

This approach provides an explicit and flexible approach for users to record a wide variety of interesting data at build-time without resorting to hard-coding values in code.

The inspiration for this resource detector came about while working on a project where there was a desire to record certain VCS details during a build process. The need was partly attestation, partly for telemetry.

On the telemetry side, this is pretty normal, and you'll often see users embed repo urls and commit SHAs directly in the dockerfiles (e.g. `ENV DD_GIT_COMMIT_SHA=${DD_GIT_COMMIT_SHA}`).

But from the attestation perspective, this leaves something to be desired. Setting a default env var in a container image works, but there's nothing to stop it from being replaced when the container is started.

_(Note this isn't a story of bad actors; on more than one occassion I've been on both sides of this where I can't tell at runtime if the current value of an env var is sourced from the image baking process or if its being superseded by an ECS task definition, etc.)_

So I went looking for a native dotnet solution for baking arbitrary data into an assembly, inspired by golang's `ldflags=-X=main.FIELD` magic tricks.

There are tons of ways we could implement this, including something fancy like a source generator, but eventually I stumbled back across some target files between dotnet/sdk and dotnet/sourcelink that were adding somewhat arbitrary metadata using the [`GetAssemblyAttributes` target](https://github.com/dotnet/sdk/blob/v9.0.303/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L75-L130) from [`Microsoft.NET.GenerateAssemblyInfo.targets`](https://github.com/dotnet/sdk/blob/v9.0.303/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets) which also handles the caching and writing of all that metadata to intermediate output files (i.e. `obj/{Project}.AssemblyInfo.cs`).

So that's basically the story. We could make this more exciting, add a source generator, maybe impose some restrictions on what users can add through this mechanism, or maybe even opinionatedly detect and embed some attributes by default (e.g. several `vcs.*` semconv attrs can be populated from SourceLink properties).

But I think there's something nice about how this works right now. It's simple and fairly unambiguous, provided you're vaguely familiar with MSBuild targets/syntax.

I've been using something pretty close to this in production for a few weeks now, and, while not terribly novel, it just works great.

I'm not sure if the benefits will be immediately obvious to most users, but I suspect (hope? fear??) there may be others who may be interested in offloading some attestation-esque metadata stamping to MSBuild.


src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+                continue;
+            }
+
+            if (!attr.Key.StartsWith(Prefix, StringComparison.Ordinal))
+            {
+                continue;
+            }
+
+            attrs ??= new Dictionary<string, object>(StringComparer.Ordinal);
+            attrs[attr.Key.Substring(Prefix.Length)] = attr.Value;
```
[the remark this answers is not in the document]
austindrenski: That would require ifdefs, and I'm not sure that the IL would actually be any more performant. Happy to change this is a human reviewer weighs in tho.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+/// &lt;ItemGroup Condition="'$(OutputType)' == 'exe'"&gt;
+///   &lt;!-- https://opentelemetry.io/docs/specs/semconv/registry/attributes/vcs/ --&gt;
+///   &lt;OtelResourceAttribute Include="vcs.ref.head.name" Value="$(VCS_REF_HEAD_NAME)" /&gt;
+///   &lt;OtelResourceAttribute Include="vcs.ref.head.revision" Value="$(VCS_REF_HEAD_REVISION)" /&gt;
+///   &lt;OtelResourceAttribute Include="vcs.repository.url.full" Value="$(VCS_REPOSITORY_URL_FULL)" /&gt;
+/// &lt;/ItemGroup&gt;
+/// </code>
+/// </example>
+/// <seealso href="https://opentelemetry.io/docs/specs/semconv/registry/attributes/"/>
+internal sealed class AssemblyMetadataDetector : IResourceDetector
```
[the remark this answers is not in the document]
austindrenski: Marked this internal to match surrounding patterns. Happy to change it if a human reviewer weighs in.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System.Reflection;
+
+namespace OpenTelemetry.Resources.AssemblyMetadata;
+
+/// <summary>
+/// Detects resource attributes applied to an assembly at build-time by the <c>GetOtelResourceAttributes</c> MSBuild target.
+/// </summary>
+/// <example>
+/// The following snippet shows how environment variables can be used to define OTel VCS resource attributes in a <c>.csproj</c> file:
```
martincostello: This arguably would work in any .NET project file (`.vbproj`, `.fsproj`, `Directory.Build.targets`), so shouldn't be so specific.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+/// &lt;/ItemGroup&gt;
+/// </code>
+/// </example>
+/// <seealso href="https://opentelemetry.io/docs/specs/semconv/registry/attributes/"/>
+internal sealed class AssemblyMetadataDetector : IResourceDetector
+{
+    private const string Prefix = "otel:";
+
+    /// <inheritdoc />
+    public Resource Detect() => Detect(Assembly.GetEntryAssembly());
```
martincostello: I would have the extension allow the assemblies to be specified, but default to this behaviour with an overload that assumes the entry assembly. The detector would then just receive a collection of assemblies to detect from.

This would enable use cases like plug-ins/shared assemblies etc.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// <inheritdoc cref="Detect()" />
+    internal static Resource Detect(Assembly? assembly)
+    {
+        if (assembly is null)
+        {
+            return Resource.Empty;
+        }
+
+        var attrs = default(Dictionary<string, object>?);
```
martincostello: This style feels an odd way of stating this compared to `Dictionary<string, object>? attrs = null`.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        foreach (var attr in assembly.GetCustomAttributes<AssemblyMetadataAttribute>())
+        {
+            if (attr is not { Key: not null, Value: not null })
```
martincostello: Is it possible for a key to be empty?

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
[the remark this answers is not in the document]
martincostello: Depends on whether the code analysers in the IDE pick up on this - if they do, then it's noise and would either need fixing or suppressing, and in that case fixing is better.

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataResourceBuilderExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+/// </summary>
+public static class AssemblyMetadataResourceBuilderExtensions
+{
+    /// <summary>
+    /// Enables the assembly metadata resource detector.
+    /// </summary>
+    /// <param name="builder">The <see cref="ResourceBuilder"/> being configured.</param>
+    /// <returns>The instance of <see cref="ResourceBuilder"/> being configured.</returns>
+    /// <inheritdoc cref="AssemblyMetadataDetector" />
+    public static ResourceBuilder AddAssemblyMetadataDetector(this ResourceBuilder builder)
```
martincostello, line 20: As noted in the other file, I suggest adding an extension to support the assemblies being specified, with the entry assembly as the default behaviour.

Whether that's added or not, that behaviour should be included in the XML documentation so someone doesn't need to look at the code to determine how it chooses the assemblies.

src/OpenTelemetry.Resources.AssemblyMetadata/OpenTelemetry.Resources.AssemblyMetadata.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Include="_.targets" Pack="true" PackagePath="buildTransitive\$(TargetFramework)\$(AssemblyName).targets" />
+  </ItemGroup>
```
martincostello: I feel it's adding unnecessary complexity to define a new item type to re-use SDK-built-in functionality (i.e. the `AssemblyMetdata` item).

src/OpenTelemetry.Resources.AssemblyMetadata/_.targets
martincostello, line 1: I would just remove this and instead just document that `AssemblyMetadata` items whose key is prefixed with `otel:` will be included by the detector.

test/OpenTelemetry.Resources.AssemblyMetadata.Tests/OpenTelemetry.Resources.AssemblyMetadata.Tests.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="$(RepoRoot)\src\OpenTelemetry.Resources.AssemblyMetadata\_.targets" />
+
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
```
martincostello, line 4: This comment is redundant if .NET Framework isn't included in the test matrix.

```suggestion
```

Is there a reason why this isn't including .NET Framework like below?

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/da1620d0b3a2d0de4c056bde47f0bb3138a128e8/test/OpenTelemetry.Resources.Host.Tests/OpenTelemetry.Resources.Host.Tests.csproj#L4-L6

src/OpenTelemetry.Resources.AssemblyMetadata/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+You need to install the `OpenTelemetry.Resources.AssemblyMetadata` package to be able to use the Assembly Metadata Resource Detector.
+
+```shell
+dotnet add package --prerelease OpenTelemetry.Resources.AssemblyMetadata
+```
+
+## Usage
+
+You can add the Assembly Metadata Resource Detector to the `ResourceBuilder` with the following example:
```
martincostello:
```suggestion
You can add the Assembly Metadata Resource Detector to the `ResourceBuilder` as shown by the following example:
```

src/OpenTelemetry.Resources.AssemblyMetadata/README.md
```diff
@@ -0,0 +1,137 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.Services.AddOpenTelemetry()
+    .ConfigureResource(static builder => builder.AddAssemblyMetadataDetector())
+    .WithLogging(static builder => /* ... */)
+    .WithMetrics(static builder => /* ... */)
+    .WithTracing(static builder => /* ... */);
+```
+
+The resource detector will record attributes defined on the [entry assembly](https://learn.microsoft.com/dotnet/api/system.reflection.assembly.getentryassembly) using [AssemblyMetadataAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assemblymetadataattribute) with keys prefixed by `otel:`.
```
martincostello: This would also need to be updated if the overload for specifying the assemblies were added.

```suggestion
The resource detector will add resource attributes defined on the [entry assembly](https://learn.microsoft.com/dotnet/api/system.reflection.assembly.getentryassembly) using [AssemblyMetadataAttribute](https://learn.microsoft.com/dotnet/api/system.reflection.assemblymetadataattribute) with keys prefixed by `otel:`.
```

src/OpenTelemetry.Resources.AssemblyMetadata/README.md
```diff
@@ -0,0 +1,137 @@
...
+```csproj
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>...</PropertyGroup>
+
+  <ItemGroup>
+    <OtelResourceAttribute Include="vcs.ref.head.name" Value="$(SourceBranchName)" />
+    <OtelResourceAttribute Include="vcs.ref.head.revision" Value="$(SourceRevisionId)" />
+    <OtelResourceAttribute Include="vcs.repository.url.full" Value="$(RepositoryUrl)" />
+  </ItemGroup>
```
martincostello: I think this specific use-case would be better served by a resource detector that is [SourceLink-aware](https://learn.microsoft.com/visualstudio/debugger/how-to-improve-diagnostics-debugging-with-sourcelink) and automatically adds these resource attributes for the user based on the presence of the known properties associated with it that are part of the [Semantic Conventions](https://github.com/open-telemetry/semantic-conventions).

For example the part after the `+` in `AssemblyInformationalVersion` would map to `vcs.ref.head.revision`, `AssemblyMetadata` where `Key=RepositoryUrl` would map to `vcs.repository.url.full`, etc.

Or do it automatically via a built-in target that's doing what the example is.

I think as-is, the PR is sort of half way between too-custom and not plug-and-play enough, if you get what I mean.

src/OpenTelemetry.Resources.AssemblyMetadata/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry.Resources.AssemblyMetadata" />
+  </ItemGroup>
+
+</Project>
+```
+
+You can check on the output from these items by inspecting the `$(AssemblyName).AssemblyInfo.cs` file in your intermediate output path (e.g. `SomeProject/obj/debug/SomeProject.AssemblyInfo.cs`):
```
martincostello: It feels too detailed to include this generated code example in the README, IMHO. For me, this is more of a compiler implementation detail than one for the resource detector.

martincostello (commented):
I like the idea of the resource detector (plus you've inspired me to glue more SourceLink metadata into my own compiled assemblies...), but my first thought is how this fits against the fact that code in this repository is typically implementing the [Semantic Conventions](https://github.com/open-telemetry/semantic-conventions).

I feel as-is it's currently a bit too detached from a specific semantic convention, but would be more of a fit to the above if it were automatically lighting up the VCS/CI/CD-related semantic conventions for the user automatically.

It might be there's prior art for resource detectors not tied to a specific semantic convention in which case disregard the above, but that was my first thought on the detector.

Kielek (commented):
I agree with Martin, that creating `target` file is an overkill. Good documentation with `otel:` prefix for metadata attributes should be more than enough.

As it non-implementing any direct semantic convention , it will be great if you can rise this topic on the next SIG meeting. The good input can be research for other languages, is similar functionality already available for any other language: Java, C++, etc? 

When the overall concept will be accepted please go through https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#new-projects and consider creating empty project just for the scaffolding, like in #2917

austindrenski:
Thanks for the feedback @martincostello, @Kielek !

I'm working my way through your review comments this morning and have some ideas about how we can strike a better balance on the too-generic "what is this for?" spectrum, while also providing some nicer defaults out of the box, as @martincostello suggested, similar to the default detections that are baked into the SourceLink targets.

Should have something we can look at together here shortly, stay tuned.

src/OpenTelemetry.Resources.AssemblyMetadata/_.targets
```diff
@@ -0,0 +1,172 @@
...
+    ==================================================================
+    OtelCicdResourceAttributes
+    ==================================================================
+    This target extends the OtelResourceAttributes target by including
+    assembly metadata for `cicd.*` resource attributes.
+
+    To disable this target, set `OtelCicdResourceAttributes=false`.
+    ==================================================================
+  -->
+  <Target Name="OtelCicdResourceAttributes"
```
austindrenski: This and the target for VCS attributes could be shipped as separate nuget packages, if there was concern about one resource detector package providing multiple attribute domains.

I've left everything in this one common targets file for now because it seems like a lot of ceremony to have separate packages with just targets files, but happy to do the legwork on that if there are concerns around domain boundaries.

In a similar vein, we could also not provide these default targets in the box, and instead document suggested or default targets in the documentation showing how to populate things from your GitHub Actions, variables, etc. 

And perhaps in the opposite vein, we could forgo providing _any_ targets files or defaults and just have this nuget package provide the detector that grabs `otel:`-prefixed assembly metadata, and leave it entirely up to the user to opt-in each resource attribute they want included on their assembly. (This is my least preferred option, but it is _an_ option, so mentioning it for others to consider.)

test/OpenTelemetry.Resources.AssemblyMetadata.Tests/OpenTelemetry.Resources.AssemblyMetadata.Tests.csproj
austindrenski: Ope, no reason at all. I copy/pasta'd one of the other test projects to use as a template, so either it didn't include that, or (more likely) I dropped it by mistake.

Bringing this back now...

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
austindrenski: Just a habit from usually working in `var`-all-the-things projects. NREs don't exist if we just don't say `null`, right?? 😄 

I swapped the dictionary for a list while ifdef'ing this file, and also replaced the `var attr = default(...?);` with a simple `List<...>? attrs = null;`

src/OpenTelemetry.Resources.AssemblyMetadata/AssemblyMetadataDetector.cs
austindrenski: Good question. It _shouldn't_ happen from attrs generated by MSBuild, but looking at the source for `AssemblyMetadataAttribute`, there doesn't seem to be anything stopping you from doing this manually:

```csharp
[assembly: AssemblyMetadata(null!, null)]
```

But both ifdef branches in the current version of this file now guard against null and empty keys as part of asserting the keys start with `otel:`, so should be safe.

src/OpenTelemetry.Resources.AssemblyMetadata/README.md
austindrenski: > I think as-is, the PR is sort of half way between too-custom and not plug-and-play enough, if you get what I mean.

This was great feedback, thank you!

> For example the part after the `+` in `AssemblyInformationalVersion` would map to `vcs.ref.head.revision`, `AssemblyMetadata` where `Key=RepositoryUrl` would map to `vcs.repository.url.full`, etc.
>
> Or do it automatically via a built-in target that's doing what the example is.

What you're describing is actually very close to the original implementation that this PR is derived from, but got squashed along the way as I tried to avoid creating too tailored a detector (e.g. "please merge this detector, it detects exactly what my project needs!" 😄).

I've brought that logic back, and with fresh eyes, I think it's actually in a much better shape. 

Specifically, I've added an initial layer of assignment attempts that read `OTEL_VCS_*` environment variables to give users (e.g., me) an easy, code-change-free option for preempting specific values (e.g. if you're not happy with what we'd otherwise detect from SourceLink).

I've also added back a layer of assignment attempts that read `GITHUB_*`environment variables that come after the SourceLink attempts. This section is incomplete, but I've added the low-hanging fruit, and figure we can improve/extend coverage over time.

There's also the matter of supporting other, non-GitHub CI environment variables, but I think that would also be best left for a follow-up PR or contribution from other community members with more familiarity with those systems.

src/OpenTelemetry.Resources.AssemblyMetadata/_.targets
```diff
@@ -0,0 +1,175 @@
...
+      <AssemblyMetadata Include="otel:cicd.pipeline.task.run.result" Value="$(OtelCicdPipelineTaskRunResult)" Condition="'$(OtelCicdPipelineTaskRunResult)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.pipeline.task.run.url.full" Value="$(OtelCicdPipelineTaskRunUrlFull)" Condition="'$(OtelCicdPipelineTaskRunUrlFull)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.pipeline.task.type" Value="$(OtelCicdPipelineTaskType)" Condition="'$(OtelCicdPipelineTaskType)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.system.component" Value="$(OtelCicdSystemComponent)" Condition="'$(OtelCicdSystemComponent)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.worker.id" Value="$(OtelCicdWorkerId)" Condition="'$(OtelCicdWorkerId)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.worker.name" Value="$(OtelCicdWorkerName)" Condition="'$(OtelCicdWorkerName)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.worker.state" Value="$(OtelCicdWorkerState)" Condition="'$(OtelCicdWorkerState)' != ''" />
+      <AssemblyMetadata Include="otel:cicd.worker.url.full" Value="$(OtelCicdWorkerUrlFull)" Condition="'$(OtelCicdWorkerUrlFull)' != ''" />
```
austindrenski: Actually, after some further thought, I don't think we should ship a target that embeds `cicd.*` attributes, at least not by default. 

As it stands, embedding these attributes in the assembly breaks deterministic builds, and I'm not sure it's responsible to ship a nuget package that actively works against the reproducible build initiative.

So I'm going to remove this target and maybe include a snippet of it in the docs, but then we're not on the hook for maintaining something that—while interesting—will almost certainly not be appropriate for most projects.

austindrenski: Also, once I remove the `cicd.*` target, that just leaves the `vcs.*` target and I think that brings us back to the question of whether this package should ship any targets file at all.

I'm open to hearing other opinions, but right now it seems to me that it does not make sense to include the `vcs.*` target in this package, but may make sense for a separate package, `OpenTelemetry.Resources.AssemblyMetadata.Vcs`, that takes a dependency on this package and includes the targets file to automatically populate what we can from sourcelink, etc

austindrenski: I'll try to find some time today to remove the one target, and split the other target into a separate project.

I'll open a separate PR for that separate project based on top of this PR.

That way we can separate the questions about the new detector from what it's detecting. I think that might also help address some of the concerns about this package not being tied to a specific semantic convention domain.

austindrenski: FYI opened a draft of this^ on my fork for reference: austindrenski/opentelemetry-dotnet-contrib#1

Issue #2960, opened 2025-07-30
Make CodeQL required for PRs


martincostello:
With #2955 CodeQL now runs on all PRs.

To avoid new issues being introduced, we should make the `CodeQL / codeql` status required.

martincostello:
Is this something you can enable @Kielek?

Kielek:
We could do it as a separate job in github config, but it requires commit to separate repository.
IMO the best option is to add this codeql step to mandatory under: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/40bd5493a2efd76e4d316ef68db1095ba6fed288/.github/workflows/ci.yml#L624

martincostello:
The downside to that approach is that we lose the ability to run it on a schedule (unless you wanted to re-run the entire CI on a schedule too).

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/40bd5493a2efd76e4d316ef68db1095ba6fed288/.github/workflows/codeql-analysis.yml#L9

Kielek:
@martincostello, maybe we can move definitions of strps to `codeql-analysis-steps.yml`.

And trigger it both from ci.yml and brand new codeql-analysis.yml triggered by cron and workflow_dispatch?

PR #2961, opened 2025-07-30
[Infra] Fix labelling for blank issues

martincostello:
## Changes

- If a blank issue is opened (which has no tags) (e.g. #2960), avoid the labelling workflow [failing](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/actions/runs/16629013955).
- Use environment variables instead of interpolation.

### DIFF
--- a/.github/workflows/add-labels.yml
+++ b/.github/workflows/add-labels.yml
@@ -24,18 +24,19 @@ jobs:
       - name: Add labels for component found in bug issue descriptions
         shell: pwsh
         run: |
-          Import-Module .\build\scripts\add-labels.psm1
+          if (-Not [string]::IsNullOrEmpty($env:ISSUE_LABELS)) {
+            Import-Module .\build\scripts\add-labels.psm1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/add-labels.yml
```diff
@@ -24,15 +24,19 @@ jobs:
       - name: Add labels for component found in bug issue descriptions
         shell: pwsh
         run: |
-          Import-Module .\build\scripts\add-labels.psm1
+          if (-Not [string]::IsNullOrEmpty($env:ISSUE_LABELS)) {
```
Kielek: I think that the following method should be changed:

```pwsh
function AddLabelsOnIssuesForComponentFoundInBody {
  param(
    [Parameter(Mandatory=$true)][int]$issueNumber,
    [Parameter(Mandatory=$true)][string]$issueLabels, #add [AllowEmptyString()], it requires probably additional changes in the code, as it also allows $nulll, not only ""
    [Parameter(Mandatory=$true)][string]$issueBody
  )
```

Kielek (approved):

Issue #2947, opened 2025-07-28
[Instrumentation.AWS] Add cloud.region attribute to AWS SDK Client instrumentation


rypdal:
In OpenTelemetry the `aws.region` attribute has been replaced by the [cloud.region](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud). AWS SDK Client instrumentation for DynamoDB lacks `cloud.region` span attribute despite it's marked as "Recommend".

martincostello:
See also #1599.

martincostello:
`cloud.region` was added by [v1.34.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.34.0/docs/cloud-providers/aws-sdk.md).

I'll look at adding this as part of some changes to implement #1609, that also needs to add v1.34.0 support.

Issue #1467, opened 2023-12-07
Get rid of Moq dependency from the repo


utpilla:
This is related to the same effort in main repo: https://github.com/open-telemetry/opentelemetry-dotnet/issues/4758

PR #2969, opened 2025-07-31
[Instrumentation.AWS] added cloud.region attribute to activity tags.

rypdal:
Issue #2947: [Instrumentation.AWS] Add cloud.region attribute to AWS SDK Client instrumentation for DynamoDB.

## Changes

Set `cloud.region` tag on Activity for all AWS SDK Clients.


test/OpenTelemetry.Instrumentation.AWS.Tests/OpenTelemetry.Instrumentation.AWS.Tests.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="AWSSDK.BedrockAgentRuntime" Version="4.0.0" />
     <PackageReference Include="AWSSDK.BedrockRuntime" Version="4.0.0" />
     <PackageReference Include="AWSSDK.S3" Version="4.0.0" />
+    <PackageReference Include="Moq" Version="4.20.72" />
```
martincostello: Moq was removed by #1467.

For the minimal usage just create a hand-rolled mock/stub.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
```diff
@@ -114,4 +114,31 @@ internal static string GetAWSOperationName(IRequestContext requestContext)
...
+    internal static string? GetDynamoDbCloudRegion(IRequestContext requestContext)
+    {
+        string? region = requestContext.ClientConfig?.RegionEndpoint?.SystemName;
+
+        // Use ServiceURL as fallback.
+        region ??= GetCloudRegionFromDynamoServiceUrl(requestContext.ClientConfig?.ServiceURL);
+        return region;
+    }
+
+    private static string? GetCloudRegionFromDynamoServiceUrl(string? serviceUrl)
```
Oberon00: I think this URL parsing should not be necessary. The region should be available directly in the parameters/request context. Maybe you can dig up the old code that used to collect the `aws.region` before it was removed?

Then we should also consider adding this generically to all calls instead of only for DynamoDB.

Oberon00 (changes requested):

test/OpenTelemetry.Instrumentation.AWS.Tests/OpenTelemetry.Instrumentation.AWS.Tests.csproj
rypdal: I got rid of Moq reference and introduced custom mock for IRequestContext which is actually has most of methods and properties as not implemented as they are not required for such tests. In the future it can be extended if someone needs  more properties.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
rypdal: I think the fallback to URL still could be required. See **New Region using the service client configuration class** chapter in https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/net-dg-region-selection.html (ignore that example is for AmazonEC2Config as it's also applicable for DynamoDB).

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        Assert.Equal("eu-central-1", cloudRegion);
+    }
+
+    [Fact]
+    public void GetDynamoDbCloudRegion_RegionIsMissingInServiceUrl_CloudRegionIsNull()
+    {
+        var clientConfig = new AmazonDynamoDBConfig
+        {
+            ServiceURL = "https://mycompany.com/test",
```
Kielek: I think that you are missing here something like `https://fooo.bar.mycompany.com/test` - region will be `bar`.
Same for `https://for.baz.bar.mycomany.com/aaaa/bbbb`

Kielek (commented):
Should this tag be applicable for all semantic convention versions you are supporting? Only part of them? Maybe new one should be introduced?

Please add also CHANGELOG.

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
Oberon00: > I think the fallback to URL still could be required.

Ok makes sense.
Still, it should be possible to restore the removed functionality with changed attribute name as it was before https://github.com/open-telemetry/opentelemetry-dotnet-contrib/commit/131ccf35efcfd21584d7daecf2bbd2c4de196fcc#diff-639d7a5b1e9f3830007b96156d189aaf80d457867122f759ba269adb8387066e (check `src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSTracingPipelineHandler.cs` on that commit). This way it would be available for all services, not only DynamoDB.

rypdal:
> Should this tag be applicable for all semantic convention versions you are supporting? Only part of them? Maybe new one should be introduced?
> 
> Please add also CHANGELOG.

Judging by [this method](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/465b68a412b2d82e1ff469877fc4866c7104993d/src/Shared/AWS/AWSSemanticConventions.cs#L486) 

> Should this tag be applicable for all semantic convention versions you are supporting? Only part of them? Maybe new one should be introduced?
> 
> Please add also CHANGELOG.

As far as I see there was the only change in the v1.34.0 related to interpretation of values in cloud.region. But it should be fine. I don't see the reason why it should be supported only in part of semconv.
https://github.com/open-telemetry/semantic-conventions/blob/main/CHANGELOG.md
@Oberon00 : do you see any issue with supporting it in all semantic convention versions ?

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
rypdal: I'm not sure exactly for which reason setting of region has been removed for all AWS services. But I would rather focus on DynamoDB in this PR until the attribute will be required for other AWS services.

Oberon00: To me it looks like an accident, or maybe because at some point it was not contained in the semantic conventions (cetainly the original name aws.region was removed). At least I could not find a release note entry about it, other than generic semantic convention version notes. But now cloud.region is documented as span attribute for such client calls, so it should be fine & semconv-conformant to use.

>But I would rather focus on DynamoDB in this PR until the attribute will be required for other AWS services.

Why build a custom solution for DynamoDB only when we can restore already working & tested code? Does the custom DynamoDB solution make anything easier vs the old generic solution?

Oberon00:
>  @Oberon00 : do you see any issue with supporting it in all semantic convention versions ?

Yes, I think we should be careful with the versions. In older versions cloud.region was only specified to mean the region the telemetry emitter is running on itself. The meaning of a target region is younger than the attribute name. (OTOH, aws.region was also removed completely, so from practical perspective maybe it's fine?)

src/OpenTelemetry.Instrumentation.AWS/Implementation/AWSServiceHelper.cs
rypdal: I added code setting of `cloud.region`  to the common place, i.e. the method `ProcessBeginRequest(...)` where there was `aws.region` removed  from. It will result in setting of `cloud.region` not only for DynamoDB Client, but also for all other AWS SDK Clients.

rypdal:
> Should this tag be applicable for all semantic convention versions you are supporting? Only part of them? Maybe new one should be introduced?
> 
> Please add also CHANGELOG.

The change has been added to CHANGELOG.

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
Oberon00: Since we are now rely on AWSSDKUtils for the URL parsing which is provided by the SDK itself, I think we don't need this. We could even consider reducing test coverage for this aspect more.

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added `cloud.region`to activity tags based on
```
Oberon00:
```suggestion
* Added `cloud.region` to activity tags based on
```

muhammad-othman (approved):

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
rypdal: I would prefer to keep this code as it "documents" default returned by `AWSSDKUtils`. The default is the concrete region us-east-1 that surprised me.

rypdal:
Hello @ppittle and @srprash. This PR needs your review/approval. Thank you.

rypdal:
This PR still needs review/approval from @ppittle and @srprash.

martincostello:
As far as we're aware @ppittle no longer works at AWS so I don't think a review from them will be forthcoming.

rypdal:
> As far as we're aware @ppittle no longer works at AWS so I don't think a review from them will be forthcoming.

Thanks for the info, @martincostello. Somehow automation stated that @ppittle or @srprash can be approves. I was a bit confused thinking they are the only approvers.
As far as I see @martincostello and @Kielek can also approve this PR. Do you have more notes for the PR ?

src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added `cloud.region` to activity tags based on
+  [semantic convention v1.36.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/registry/attributes/cloud.md#cloud-region).
+  ([#2947](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2947))
```
martincostello:
```suggestion
  ([#2969](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2969))
```

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using Xunit;
+
+namespace OpenTelemetry.Instrumentation.AWS.Tests.Implementation;
+
+public class AWSServiceHelperTests
+{
+    [Fact]
+    public void ExtractCloudRegion_RegionEndpointIsSet_CloudRegionIsCorrect()
+    {
+        var region = "eu-central-1";
```
martincostello: Make this data driven and add a few more regions?

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        var clientConfig = new AmazonDynamoDBConfig
+        {
+            ServiceURL = serviceUrl,
+        };
+        var requestContext = new MockRequestContext(clientConfig);
+
+        var cloudRegion = AWSServiceHelper.ExtractCloudRegion(requestContext);
+
+        Assert.Equal("eu-central-1", cloudRegion);
```
martincostello: I would have the expected value be an input, and then vary the region between the example URLs to use different regions for some variety.

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        var clientConfig = new AmazonDynamoDBConfig
+        {
+            ServiceURL = "https://mycompany.com/test",
+        };
+        var requestContext = new MockRequestContext(clientConfig);
+
+        var cloudRegion = AWSServiceHelper.ExtractCloudRegion(requestContext);
+
+        Assert.Equal("us-east-1", cloudRegion); // AWS SDK default
```
martincostello: Could move this to a constant at the class level and then remove the need for the comment, e.g.:

```csharp
private const string DefaultAwsRegion = "us-east-1";

Assert.Equal(DefaultAwsRegion, cloudRegion);
```

test/OpenTelemetry.Instrumentation.AWS.Tests/Tools/MockRequestContext.cs
martincostello, line 1: We already have a fake for this: [TestRequestContext](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c0167a091aa860500a8807fa70a4933b94bd8d41/test/OpenTelemetry.Instrumentation.AWS.Tests/TestRequestContext.cs)

martincostello (commented):
FYI I can review and approve, but I can't merge.

Typically the code owners reviews are sought/required for non-trivial changes to functionality for a specific library.

Kielek:
> > As far as we're aware @ppittle no longer works at AWS so I don't think a review from them will be forthcoming.
> 
> Thanks for the info, @martincostello. Somehow automation stated that @ppittle or @srprash can be approves. I was a bit confused thinking they are the only approvers. As far as I see @martincostello and @Kielek can also approve this PR. Do you have more notes for the PR ?

Technically, they are. If you have any good contacts in the AWS team, it will be great to make them aware, that packages are abandoned by them.

martincostello:
@normj Are you able to help us out here with regards to ownership of these AWS-specific libraries?

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
rypdal: I added more regions.

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
rypdal: I added more URL and regions variety. Also added expected region as test parameter.

test/OpenTelemetry.Instrumentation.AWS.Tests/Implementation/AWSServiceHelperTests.cs
rypdal: I added class constant and reused everywhere incl. test parameters.

test/OpenTelemetry.Instrumentation.AWS.Tests/Tools/MockRequestContext.cs
rypdal: Thanks. I overlooked it. I reused this class and removed the new one.

normj:
I'm reviewing the PR. 

@Kielek it is true @ppittle is no longer at AWS. I'm happy to step in take AWS component ownership of the package in the repo. I'm unsure what the process for being added to the component ownership for the AWS packages.

normj (approved):

rypdal:
> If I do not missed anything, `cloud.region` attribute will be added to all semantic conventions, despite the version.
> 
> Based on https://github.com/open-telemetry/semantic-conventions/blob/v1.34.0/docs/cloud-providers/aws-sdk.md and https://github.com/open-telemetry/semantic-conventions/blob/v1.33.0/docs/cloud-providers/aws-sdk.md 1.34.0 is the very first version recommending it. I think that it potentially is a breaking change for older semantic conventions. As this package has stable release, you should consider adding new semantic convention only, and do not touch the older (if it was not mentioned there).
> 
> I hope I do not missed anything in the adding tags process.
> 
> Probably this tag can be added only for DynamoDB for older semantic conventions, but it needs to be double checked.

If I understand the current versioning implementation correctly it's necessary to introduce new class AWSSemanticConventions_V1_34_0 (`cloud.region` appeared from 1.34). Then extend corresponding enum and return the new class from this method:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/b84a2859fea60ecfd304a5ec9310c222fb2f8589/src/Shared/AWS/AWSSemanticConventions.cs#L484
Then it assumes that users initialisation code should be adapted to configure the new  `SemanticConventionVersion.V1_34_0` because `SemanticConventionVersion.V1_28_0` is the default. Also it means that the Latest != V1_29_0 anymore because the latest becomes V1_34_0. This may sound more risky. 
Just adding the new attribute `cloud.region` to all version starting from V1_28_0 doesn't look like a breaking change unless user code set tag with the same name and our code overwrites it. But it sounds like implausible scenario.

Kielek:
You might be right that it is not a breaking change. It just violates semantic convention.
As long package was in beta stage, and there were no concept of switching semantic conventions, this kind of changes was fully fine. With the stable packages, and the possibility to choice the sem.conv. version, it should not affect older versions, where the attribute was not known (not at the level of code, but at the level of sem conv.).

BTW if you decide to introduce new sem. conv, it should cover all changes in sem conv. I would suggest to create 1.36/1.37 (if released) and make full compliance check.

I am not in favor of blocking this PR or any other to AWS. I am just trying to show, where you can expect issues from the developers/end-users.

rypdal:
> You might be right that it is not a breaking change. It just violates semantic convention. As long package was in beta stage, and there were no concept of switching semantic conventions, this kind of changes was fully fine. With the stable packages, and the possibility to choice the sem.conv. version, it should not affect older versions, where the attribute was not known (not at the level of code, but at the level of sem conv.).
> 
> BTW if you decide to introduce new sem. conv, it should cover all changes in sem conv. I would suggest to create 1.36/1.37 (if released) and make full compliance check.
> 
> I am not in favor of blocking this PR or any other to AWS. I am just trying to show, where you can expect issues from the developers/end-users.

Do you mean to introduce the latest sem.conv. 1.36 in code.  Then align all attributes (means not only adding the cloud.region) with this version of sem.conv. Then make a full compliance check ? It might be a big overhead in comparison to the initial scope of the PR.

Kielek:
> Do you mean to introduce the latest sem.conv. 1.36 in code. Then align all attributes (means not only adding the cloud.region) with this version of sem.conv. Then make a full compliance check ? It might be a big overhead in comparison to the initial scope of the PR.

I think that it is the cost of having stable package relaying on non-stable sem. conv..

rypdal:
> > Do you mean to introduce the latest sem.conv. 1.36 in code. Then align all attributes (means not only adding the cloud.region) with this version of sem.conv. Then make a full compliance check ? It might be a big overhead in comparison to the initial scope of the PR.
> 
> I think that it is the cost of having stable package relaying on non-stable sem. conv..

I understand. If we claim 1.36 then we guarantee it's fully complaint. It means moving from 29 to 36, i.e. 7 versions ahead. This may induce a lot changes with potential bugs. Could we still have some kind of experimental/unstable version before major release of a stable  package ?

Kielek:
> > > Do you mean to introduce the latest sem.conv. 1.36 in code. Then align all attributes (means not only adding the cloud.region) with this version of sem.conv. Then make a full compliance check ? It might be a big overhead in comparison to the initial scope of the PR.
> > 
> > 
> > I think that it is the cost of having stable package relaying on non-stable sem. conv..
> 
> I understand. If we claim 1.36 then we guarantee it's fully complaint. It means moving from 29 to 36, i.e. 7 versions ahead. This may induce a lot changes with potential bugs. Could we still have some kind of experimental/unstable version before major release of a stable package ?

Sure! In scope of alpha/beta, we can have even breaking changes (but for sure, we should avoid this).

rypdal:
> > > > Do you mean to introduce the latest sem.conv. 1.36 in code. Then align all attributes (means not only adding the cloud.region) with this version of sem.conv. Then make a full compliance check ? It might be a big overhead in comparison to the initial scope of the PR.
> > > 
> > > 
> > > I think that it is the cost of having stable package relaying on non-stable sem. conv..
> > 
> > 
> > I understand. If we claim 1.36 then we guarantee it's fully complaint. It means moving from 29 to 36, i.e. 7 versions ahead. This may induce a lot changes with potential bugs. Could we still have some kind of experimental/unstable version before major release of a stable package ?
> 
> Sure! In scope of alpha/beta, we can have even breaking changes (but for sure, we should avoid this).

I had a brief look. Unfortunately such a big version difference results in many changes. If you add new attributes implementation effort it may take a long time to finish. It's not only names change, it also getting of new attributes from different not always clear sources.

alanwest:
> I had a brief look. Unfortunately such a big version difference results in many changes. If you add new attributes implementation effort it may take a long time to finish. It's not only names change, it also getting of new attributes from different not always clear sources.

This is exactly how AWS decided to design this package. I agree it leads to more work, but AWS perceived it more beneficial to stabilize the package with the trade off of taking on this additional complexity.

The enumeration of versions becomes meaningless if `SemanticConventionVersion.V1_28_0` comes to mean "version 1.28 but also with a little bit of 1.34 sprinkled in".

It _might_ be possible to consider adopting only a portion of 1.34 conventions to begin with, but this would have to be carefully considered. Generally speaking, adding new attributes to spans is not considered a breaking change. Adding new attributes to metrics _is_ considered a breaking change, but we're not dealing with metrics here.

So for example, it might be possible to add `SemanticConventionVersion.V1_34_0` and only add `cloud.region` for now. In follow up work additional things new to 1.34 may be able to be added without being considered a breaking change. Also, I would consider implementing `SemanticConventionVersion.V1_36_0` since that is the latest rather than `SemanticConventionVersion.V1_34_0`. I don't personally see a reason for the enumeration to support _all_ the versions.

> Then it assumes that users initialisation code should be adapted to configure the new SemanticConventionVersion.V1_34_0 because SemanticConventionVersion.V1_28_0 is the default.

Yes, this is the intent. It would be a breaking change to change the default. Consumers would need to explicitly opt into 1.34.

> Also it means that the Latest != V1_29_0 anymore because the latest becomes V1_34_0. This may sound more risky.

This is also the intent. The idea was that consumers choosing `Latest` are inherently taking on the risk that the telemetry produced by the library may change as they upgrade to a new version.

> Just adding the new attribute cloud.region to all version starting from V1_28_0 doesn't look like a breaking change unless user code set tag with the same name and our code overwrites it. But it sounds like implausible scenario.

It may not technically be a breaking change today, but it could lead to a breaking change later. The AWS SDK conventions are still under development. If the meaning or name of `cloud.region` were to later change then making that change could be breaking. The idea of introducing the enumeration of versions gives consumers selecting a specific version a guarantee that there will be no breaking changes.

rypdal:
@alanwest thanks for your explanation.

rypdal:
I would put this PR on hold (not sure if it's possible), or even close it and re-open later. Perhaps it make sense to collect a portion of 1.34  (or 1.36) convention attributes and then merge them together.

Kielek:
@rypdal, I am closing PR for now. Feel free to reopen/create new and link with the closed when you are ready to proceed.

PR #2971, opened 2025-08-01
#2892 implement additional process attributes

thompson-tomo:
Progresses #2892
Design discussion issue #

## Changes

Implemented the other recommended process attributes into the resource detector based on the definition of a resource as well as additional attributes from the process namespace.


thompson-tomo:
Not sure where you looked but they are both in fact listed in the sem conv registry: https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/

Docs will be updated.

Kielek:
Full search text for `process.working.directory` in semantic convention repository. Shows nothing. On the page you have linked the same.

The manual check. `.` is not the same as `_`.

thompson-tomo:
my apologies, no idea what resulted in the symbol being switched. Have corrected it to what it should've been.

thompson-tomo:
Still awaiting feedback after typo fix.

src/OpenTelemetry.Resources.Process/ProcessSemanticConventions.cs
```diff
@@ -7,4 +7,11 @@ internal static class ProcessSemanticConventions
 {
     public const string AttributeProcessOwner = "process.owner";
     public const string AttributeProcessPid = "process.pid";
+    public const string AttributeProcessExecPath = "process.executable.path";
+    public const string AttributeProcessWorkingDir = "process.working_directory";
+    public const string AttributeProcessArgsCount = "process.args_count";
+    public const string AttributeProcessStartTime = "process.creation.time";
+    public const string AttributeProcessExecName = "process.executable.name";
+    public const string AttributeProcessInteractive = "process.interactive";
+    public const string AttributeProcessTitle = "process.title";
```
Kielek, line 16: Some of this attributes are not listed under https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/resource/process.md

Resource definition is not the same as otel semantic convention registry. Resource detectors implementation should relay https://github.com/open-telemetry/semantic-conventions/tree/v1.36.0/docs/resource rather than https://github.com/open-telemetry/semantic-conventions/tree/v1.36.0/docs/registry.

What potentially can be considered, when we will be close to stable state of resource detectors (and sem conv) is to add ability to opt in into non-standard attributes. I will not accept this kind of changes right now.

Double checked recommendation of implementation with @lmolkova

PR #2804, opened 2025-05-22
Add `Filter` option to `StackExchangeRedisInstrumentationOptions`

aradalvand:
Fixes #2803

## Changes
- Added a `Filter` option (in line with conventions that most libraries in this repo follow) to `StackExchangeRedisInstrumentationOptions`.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [x] Changes in public API reviewed (if applicable)

PR #2977, opened 2025-08-02
Include `ParentActivity` in `Filter` and `Enrich` of `StackExchangeRe…

aradalvand:
I'm the contributor creating this earlier PR: #2804
I did not know that activities are created for profiled commands in an asynchronous fashion, which means that doing the following wouldn't work as expected:
```cs
.AddRedisInstrumentation(b =>
{
	b.Filter = e =>
	{
		return Activity.Current is not null; // PROBLEM: This is always `false`
	};
})
```
This is arguably the most common reason for adding a `Filter`, and it was certainly my motivation for #2804 — this is so that "orphan" Redis commands, that is, Redis commands that are not performed in the context of some larger unit of work, are discarded and don't make it to the OTEL back-end, hammering it with not-so-useful spans.

What I did is that instead of passing only the `IProfiledCommand` which is currently being passed to `Filter` in v1.12.0-beta.2, a so-called `RedisInstrumentationContext` is being passed which additionally includes the `ParentActivity` — this construct is very similar to [NATS's `NatsInstrumentationContext`](https://github.com/nats-io/nats.net/blob/c7756475acc4f1c148817662971dc4e3ce8651cd/src/NATS.Client.Core/NatsInstrumentationContext.cs).

### DIFF
--- a/src/OpenTelemetry.Instrumentation.StackExchangeRedis/.publicApi/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.Instrumentation.StackExchangeRedis/.publicApi/PublicAPI.Unshipped.txt
@@ -1,13 +1,15 @@
+OpenTelemetry.Instrumentation.StackExchangeRedis.RedisInstrumentationContext
+OpenTelemetry.Instrumentation.StackExchangeRedis.RedisInstrumentationContext.ProfiledCommand.get -> StackExchange.Redis.Profiling.IProfiledCommand!
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentation
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentation.AddConnection(StackExchange.Redis.IConnectionMultiplexer! connection) -> System.IDisposable!
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentation.AddConnection(string! name, StackExchange.Redis.IConnectionMultiplexer! connection) -> System.IDisposable!
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentation.Dispose() -> void
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions
-OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions.Enrich.get -> System.Action<System.Diagnostics.Activity!, StackExchange.Redis.Profiling.IProfiledCommand!>?
+OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions.Enrich.get -> System.Action<System.Diagnostics.Activity!, OpenTelemetry.Instrumentation.StackExchangeRedis.RedisInstrumentationContext!>?
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions.Enrich.set -> void
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions.EnrichActivityWithTimingEvents.get -> bool
 OpenTelemetry.Instrumentation.StackExchangeRedis.StackExchangeRedisInstrumentationOptions.EnrichActivityWithTimingEvents.set -> void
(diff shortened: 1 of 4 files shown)


global.json
```diff
@@ -1,6 +1,5 @@
 {
   "sdk": {
-    "rollForward": "latestFeature",
-    "version": "9.0.303"
+    "rollForward": "latestFeature"
```
martincostello: Please revert this change.

aradalvand: Sorry, my mistake.
Done.

matt-hensley:
This is a good fix and I something I also missed with the async behavior. I agree that a breaking change is OK here.

The context object has the benefit of not needing breaking changes in the future if additional context data would be useful in the callbacks.

Alternatively, we could potentially hide that the conversion is async by setting `Activity.Current` temporarily. The [HttpModule for ASP.NET](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs#L169) and [gRPC instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.GrpcCore/ClientTracingInterceptor.cs#L371) already do this.

I like the idea of the callbacks working "as expected", but I'm not sure if manipulating `Activity.Current` would fall outside best practices.

aradalvand:
> The context object has the benefit of not needing breaking changes in the future if additional context data would be useful in the callbacks.

Yeah exactly.

> Alternatively, we could potentially hide that the conversion is async by setting Activity.Current temporarily. The [HttpModule for ASP.NET](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/ActivityHelper.cs#L169) and [gRPC instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.GrpcCore/ClientTracingInterceptor.cs#L371) already do this.

@matt-hensley Hmm, that's interesting too, could be a better option as a matter of uniformity and principle of least surprise.
Let me know which approach you think we should go for. I could go ahead and make that change.

matt-hensley:
@aradalvand considered this some more, and I personally like setting `Activity.Current` to make usage uniform.

I having trouble coming up with downsides as that pattern has worked fine in some other instrumentation for quite some time.

aradalvand:
@matt-hensley Actually, I just tried to implement this and found out it's not possible:
I did this in `src/OpenTelemetry.Instrumentation.StackExchangeRedis/Implementation/RedisProfilerEntryToActivityConverter.cs`:
```cs
public static Activity? ProfilerCommandToActivity(Activity? parentActivity, IProfiledCommand command, StackExchangeRedisInstrumentationOptions options)
{
    Console.WriteLine($"parentActivity: {parentActivity}");
    Activity.Current = parentActivity;
    Console.WriteLine($"Activity.Current: {Activity.Current}");
    try
    {
        if (options.Filter != null && !options.Filter(new(command)))
        {
            return null;
        }
    }
    catch
    {
        return null;
    }
    ...
```
I get this output in the console:
```
parentActivity: System.Diagnostics.Activity
Activity.Current:
```
Notice that `Activity.Current` remains `null`. I digged into it and found out it's because the `Current` property's setter looks like this:
```cs
/// <summary>
/// Gets or sets the current operation (Activity) for the current thread. This flows
/// across async calls.
/// </summary>
public static Activity? Current
{
    get { return s_current.Value; }
    set
    {
        if (ValidateSetCurrent(value))
        {
            SetCurrent(value);
        }
    }
}
```
And `ValidateSetCurrent` looks like this:
```cs
private static bool ValidateSetCurrent(Activity? activity)
{
    bool canSet = activity == null || (activity.Id != null && !activity.IsStopped);
    if (!canSet)
    {
        NotifyError(new InvalidOperationException(SR.ActivityNotRunning));
    }

    return canSet;
}
```
Notice the `activity.IsStopped` check. This means that if the activity you're setting `Activity.Current` to had been stopped earlier (which is true in this case), the setter refuses it.

src/OpenTelemetry.Instrumentation.StackExchangeRedis/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Introduce `RedisInstrumentationContext` and use it as context for `Filter`
```
Kielek:
```suggestion
* **Breaking change** Introduce `RedisInstrumentationContext` and use it
  as context for `Filter`
```

matt-hensley (approved):

Kielek:
@aradalvand, thanks! Sorry for delay with merging your PR.

PR #2979, opened 2025-08-03
#2978 Remove azure vm sku attribute

### LINKED ISSUE #2978: [feature request] removal of azure.vm.sku attribute
### Component

OpenTelemetry.Resources.Azure

Not all the attributes are defined in semantic conventions.

The `azure.vm.sku` attribute is removed as it is redundant when considering the already defined attributes in the OS namespace and additions in review.

Defining attribute in registry.

### Additional context

Other attribute missing are added via https://github.com/open-telemetry/semantic-conventions/pull/2589

thompson-tomo:
Fixes #2978
Design discussion issue #

## Changes

This removes the `azure.vm.sku` attribute from the resource detector which is not a member of the semantic convention registry and is redundant when considering the existing OS namespace.


src/OpenTelemetry.Resources.Azure/AzureVmMetadataResponse.cs
```diff
@@ -77,9 +77,6 @@ internal string GetValueForField(string fieldName)
             case ResourceAttributeConstants.AzureVmScaleSetName:
                 amsValue = this.VmScaleSetName;
                 break;
-            case ResourceAttributeConstants.AzureVmSku:
-                amsValue = this.Sku;
```
martincostello, line 81: If this is being removed, then the `Sku` property can also be removed as nothing else appears to use it. The `ResourceGroupName` and `SubscriptionId` also appear to be used by nothing.

thompson-tomo: Ok have cleaned that up as well.

martincostello (approved):

rajkumar-rangaraj:
> This removes the azure.vm.sku attribute from the resource detector which is not a member of the semantic convention registry and is redundant when considering the existing OS namespace.

Could you provide a bit more detail and explain why this is considered redundant?

thompson-tomo:
When you consider the OS namespace we have `os.name` which would be "windows server 2019" and `os.description` would include if it is datacenter edition etc. With the extensions to sem conv in review the OS namespace would have an attribute `os.variant` for indicating the edition.

The OS resource detector would be able to provide this info.

thompson-tomo:
@rajkumar-rangaraj is anything else needed to progress this?

rajkumar-rangaraj (approved):

Issue #2973, opened 2025-08-01
[feature request] Have db systems match sem conv


thompson-tomo:
### Component

OpenTelemetry.Instrumentation.EntityFrameworkCore

### What is the expected behavior?

The db system is set as following:

- cosmosdb -> azure.cosmosdb
- firebird -> firebirdsql
- mssql -> microsoft.sql_server
- oracle -> oracle.db
- spanner -> gcp.spanner

### Which alternative solutions or features have you considered?

N/a

### Additional context

https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/

martincostello:
Looks like these changed as part of the [v1.30](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.30.0) of the Semantic Conventions from https://github.com/open-telemetry/semantic-conventions/pull/1734.

The code that does the mapping is here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/465b68a412b2d82e1ff469877fc4866c7104993d/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L97-L163

martincostello:
The provider also still uses [`db.system`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/465b68a412b2d82e1ff469877fc4866c7104993d/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L22) instead of `db.system.name`, so changing the provider names should be done as part of a wider effort to update the semantic convention version in the instrumentation in general, rather than just _specifically_ updating the values.

martincostello:
@Kielek Is there typically a coordinated effort to update the version of the Semantic Conventions used in the projects in this repo, or are they done on a case-by-case basis?

Kielek:
For db sem conv for SQL Client @alanwest created a milestone: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/milestone/6

There is an alternative approach available - create main issue with sub-issues like for ASP.NET stabilization https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2909 - both are fine IMO

For sure, there is a need to switch whole convention at once, I see 2 possibilities
[...]

PR #2985, opened 2025-08-07
Use latest semconv for redis #2976

### LINKED ISSUE #2302: [feature request] ElasticSearch - update network semantic convention
### Component

OpenTelemetry.Instrumentation.ElasticsearchClient

Network semantic convention uses deprecated attributes.
Its need to be updated to the latest one.

Instrumentation is using new semantic convention for network attributes. It is related to `net.peer.ip`, `net.peer.port`, and `net.peer.name`. All of them should be replaced by `network.peer.port` and `network.peer.address`.

For the reference see https://github.com/open-telemetry/semantic-conventions/blob/v1.28.0/docs/database/database-spans.md

Code: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/4301c9497887509c6fd996e4560d2ac7172de249/src/OpenTelemetry.Instrumentation.ElasticsearchClient/Implementation/ElasticsearchRequestPipelineDiagnosticListener.cs#L193-L205

N/A

### Additional context

_No response_

### LINKED ISSUE #2976: [feature request] replace usages of net.peer.* with server.*
### Component

OpenTelemetry.Instrumentation.ElasticsearchClient

_No response_

The server.* namespace ie server.address is used instead of net.peer.ip etc

Also use network.peer namespace.

### Additional context

_No response_

thompson-tomo:
Fixes #2976
Closes #2302
Design discussion issue #

## Changes


martincostello:
Same here with reference to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2973#issuecomment-3158394268 and the opt-in.

thompson-tomo:
Interesting as the definition does not talk about the need for opt-in https://opentelemetry.io/docs/specs/semconv/database/elasticsearch/ and in fact it explicitly changes the span definition to be experimental rather than stable.

Also i I am not sure how that could even work in this case as Elasticsearch is not stable itself in this case, if you were to set OTEL_SEMCONV_STABILITY_OPT_IN with a value of database, you would get nothing as span is not stable hence makes no sense to have a flag which would result in nothing.

I can understand the need for it on sql server etc which have been stabilized but not the case for elasticsearch.

martincostello:
That's what was requested in the SIG, and EFCore isn't stable either.

thompson-tomo:
efcore is not a db system but rather mysql etc is. Some of the db systems which ef uses are stable hence the stability opt-in is valid for those system. here that is not the case as we have a single db system which is not stable. This is no different to what was done for redis when updating redis to 1.23.0.

martincostello:
It's not, no, but it's using the database conventions, and the PR is changing the names being used for the attributes.

thompson-tomo:
I am trying to work out what this would look like in terms of configured options given it is not a stable signal.

- `database`: nothing emitted as span/span-name is not stable & also to top it off required attributes can not be emitted as the value is not stable.
- `database/dup`:
  - db.system.name
  - db.system
  - db.query.summary
  - db.query.text
  - db.operation.name
  - db.statement
- 'empty'
  - db.system.name -> uses a development value hence development
  - db.system
  - db.statement

Based on the above do we have database/dup as the standard given that the instrumentation is not stable and remove the deprecated when it goes rc?

The scenarios change when you are dealing with a stable db.system.name value as you can now generate a stable span.

martincostello:
Maybe one of the maintainers with more experience can clarify, but my understanding from the SIG meeting this week was that the opt-in is only affecting the names when things are emitted, not whether or not they are emitted.

Assuming that's the correct interpretation, then the span would still be emitted whatever the setting's value it.

thompson-tomo:
So the [semconv page](https://opentelemetry.io/docs/specs/semconv/database/database-spans/) mentions:
> Conventions include (but are not limited to) attributes, metric and span names, and unit of measure.

Hence more broad than just names. 

I feel that the documentation needs improvement as it is not clear and in some cases not possible without conflicting over conventions. Will also raise it in semconv.

stevejgordon:
I think it makes sense to have these renames behind the `OTEL_SEMCONV_STABILITY_OPT_IN` flag. See https://opentelemetry.io/docs/specs/semconv/database/database-spans/

thompson-tomo:
@stevejgordon i am aware of that document which is why I am attempting to find a way to be compliant with it when the signal is not stable as per the spec which makes a difference. Note See above summary of scenarios and how it can result in issues if followed as written for stable signals.

I have a semconv PR to be more descriptive about how to handle attribute deprecation but not getting feedback. 😞

thompson-tomo:
Further to that if you were to do a progressive upgrade:
- Upgrade to 1.25 first including supporting the env variable for that point in time
- Then upgrade to the latest which is allowed

You would end up with only the networking attributes being managed via stability variable and the rest can be directly updated without a phased migration.

Also as this is not a stable instrumentation library the following applies https://opentelemetry.io/docs/specs/otel/telemetry-stability/.

martincostello:
> Also as this is not a stable instrumentation library the following applies https://opentelemetry.io/docs/specs/otel/telemetry-stability/.

Neither are the SqlClient or EntityFramework instrumentation libraries, but they still use the opt-in to avoid breaking their users, which is why you've been asked to use the same approach here.

thompson-tomo:
@martincostello yes I am aware that they have recently added that environment variable. it doesn't mean that they are correctly following the spec especially in light of the above questions & findings. With the most interesting now being how few attributes the variable should actually control.

martincostello:
According to the CHANGELOG for SqlClient, it was added in December, so not that recent.

The consensus among the maintainers seems pretty clear - the environment variable should be added so that we give users a choice until the libraries have their versions stabilised.

Even though these packages have prerelease versions as far as the .NET ecosystem goes, many users use them as if they were stable as otherwise if they only used stable NuGet package versions a good proportion of the OpenTelemetry packages would become unusable to them.

Considering I've just applied this approach myself in a recent change for EFCore, I don't think this is a large or unreasonable amount of effort that's being asked of you here. I think more energy has been spent debating the request that addressing the feedback would have been.

Unless the maintainers change their opinion (I specifically asked them about this in a recent SIG meeting, solely because of issues/PRs you were creating) or grant this specific library an exemption, I don't think this PR is going to make any further progress forward in its current state.

thompson-tomo:
The problem is not implementing the environment variable, it is about doing an implementation which is following the documented specification which is what i have been asked on multiple occasions to ensure that i am doing. That is why i have raised questions multiple times about the expected behaviour as it doesn't match the documented specs but I have not gotten any answer to my questions. As such if i was to add the environment variable it wouldn't be possible to actually complete until those questions are resolved hence why i am trying to get to the bottom of how this is actually supposed to function.

SQL client is for a stable singal definition hence not a valid comparison to this one which is clearly desiginated as experimental and then for efcore a number of the db providers are stable hence different again.

PR #2994, opened 2025-08-12

mladjan-gadzic:
Fixes 

Historical PR:
- #2704

Design discussion issue 

## Changes

PartialActivityProcessor is added. It helps address visibility for long running spans or processes that shut down ungracefully by sending heartbeats (logs) while span is active and stop log when span ends. Initial heartbeat delay is set to 5s to reduce noise of short lasting spans. Log body is used to serialize span in json per OpenTelemetry specification in case collectors need to process it further.


.editorconfig
```diff
@@ -159,6 +159,16 @@ dotnet_diagnostic.IDE0005.severity = warning
 # RS0041: Public members should not use oblivious types
 dotnet_diagnostic.RS0041.severity = suggestion
 
+# RS0016:
+# dotnet_diagnostic.RS0016.severity = suggestion
+
+# CS1591
+# dotnet_diagnostic.CS1591.severity = suggestion
```
martincostello: These are commented out so should be removed?

.editorconfig
```diff
@@ -159,6 +159,16 @@ dotnet_diagnostic.IDE0005.severity = warning
...
+
+# CA1848-CA2254: Needs to be suppressed because interpolated messages cannot be used.
+dotnet_diagnostic.CA1848.severity = suggestion
+dotnet_diagnostic.CA2254.severity = suggestion
```
martincostello: I think it's better to suppress in-place where relevant, rather than turn it off everywhere.

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
```diff
@@ -21,6 +21,8 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="System.Text.Json" Version="9.0.4" />
```
martincostello: There are specific properties to use to determine which version to use:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/5298d31b00b40fe32ee9fe9f12e813cebe52ccdb/build/Common.props#L65-L69

Example usage:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/5298d31b00b40fe32ee9fe9f12e813cebe52ccdb/src/OpenTelemetry.Exporter.InfluxDB/OpenTelemetry.Exporter.InfluxDB.csproj#L21-L22

This package shouldn't force all TFMs to use the latest version as it can cause issues (e.g. https://github.com/open-telemetry/opentelemetry-dotnet/issues/5973).

test/OpenTelemetry.Extensions.Tests/OpenTelemetry.Extensions.Tests.csproj
```diff
@@ -18,6 +18,7 @@
   <ItemGroup>
     <PackageReference Include="OpenTelemetry.Exporter.InMemory" Version="$(OpenTelemetryCoreLatestVersion)" />
     <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.10.0" />
+    <PackageReference Include="Moq" Version="4.20.72" />
```
martincostello: Please remove usage of Moq. See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1467.

martincostello (commented):
Just a quick initial pass - I haven't reviewed most of the files yet.

mladjan-gadzic:
hey @martincostello, thanks for a quick glance and feedback. i've addressed all of your comments. please resolve conversations once you're ok with it.

.editorconfig
```diff
@@ -159,6 +159,8 @@ dotnet_diagnostic.IDE0005.severity = warning
 # RS0041: Public members should not use oblivious types
 dotnet_diagnostic.RS0041.severity = suggestion
 
+
+
```
martincostello:
```suggestion
```

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
martincostello, line 1: This is quite a lot of new APIs 😅

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
```diff
@@ -21,6 +21,8 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="$(OpenTelemetryCoreLatestVersion)" />
```
martincostello: Not sure that the extensions should be bringing in the dependency of all the OTLP exporters.

Rather than living in `OpenTelemetry.Extensions`, this should be its own new self-contained library?

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/AnyValue.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.Extensions.Trace.PartialActivityProcessor;
+
+/// <summary>
+/// AnyValue per spec.
```
martincostello: For things declaring they implement part of the spec, they should provide a link to the relevant part.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/EventPerSpec.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// <summary>
+    /// Gets or sets the name of the event.
+    /// </summary>
+    public string? Name { get; set; }
+
+    /// <summary>
+    /// Gets the attributes of the event.
+    /// </summary>
+    public Collection<KeyValue> Attributes { get; } = [];
```
martincostello: Any reason to use a specific concrete collection, rather than an interface such as `ICollection<T>`?

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/EventPerSpec.cs
```diff
@@ -0,0 +1,48 @@
...
+                Value = new AnyValue(activityEventTag.Value?.ToString()),
+            };
+            this.Attributes.Add(keyValue);
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets the time in Unix nanoseconds when the event occurred.
+    /// </summary>
+    public ulong? TimeUnixNano { get; set; }
```
martincostello: Unless the name is exactly matching something in the spec, `UnixTimeNanoseconds` would be more fitting with terminology used elsewhere in .NET, e.g. [`DateTimeOffset.FromUnixTimeMilliseconds()`](https://learn.microsoft.com/dotnet/api/system.datetimeoffset.fromunixtimemilliseconds)

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/Processor.cs
```diff
@@ -0,0 +1,349 @@
...
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace OpenTelemetry.Extensions.Trace.PartialActivityProcessor;
+
+/// <summary>
+/// The PartialActivityProcessor is an OpenTelemetry span processor that emits logs at regular intervals (referred to as "heartbeats") during the lifetime of a span and when the span ends.
+/// This processor is useful for monitoring long-running spans by providing periodic updates.
+/// </summary>
+public class Processor : BaseProcessor<Activity>
```
martincostello: `PartialActivityProcessor` is probably a better name, rather than using a generic name and relying on the namespace for differentiation.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/Processor.cs
```diff
@@ -0,0 +1,349 @@
...
+
+    private readonly Queue<(ActivitySpanId SpanId, DateTime InitialHeartbeatTime)>
+        delayedHeartbeatActivities;
+
+    private readonly HashSet<ActivitySpanId> delayedHeartbeatActivitiesLookup;
+
+    private readonly Queue<(ActivitySpanId SpanId, DateTime NextHeartbeatTime)>
+        readyHeartbeatActivities;
+
+    private readonly Thread processorThread;
```
martincostello: https://github.com/open-telemetry/opentelemetry-dotnet/pull/6379 is making changes to processors to support use of `Task` as well as `Thread`, so if that gets merged this should probably aim to provide the same.

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
mladjan-gadzic: it's mostly because of tests since i have to expose those in order to test it properly..

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
mladjan-gadzic: good catch. that one was removed since it was a leftover.

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
martincostello: Most (all?) of the projects use `[InternalsVisibleTo]` so it shouldn't be necessary if that's the only reason.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/AnyValue.cs
mladjan-gadzic: added.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/EventPerSpec.cs
mladjan-gadzic: there is not. fixed.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/EventPerSpec.cs
mladjan-gadzic: it is matching something in spec https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L232.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/Processor.cs
mladjan-gadzic: i couldn't name directory the same as file inside it. i'd rename both.

src/OpenTelemetry.Extensions/Trace/PartialActivityProcessor/Processor.cs
mladjan-gadzic: makes sense to me.

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
mladjan-gadzic: seems that this one wasn't using, so i added it.

src/OpenTelemetry.Extensions/AssemblyInfo.cs
```diff
@@ -1,4 +1,12 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Runtime.CompilerServices;
+
+#if SIGNED
+[assembly: InternalsVisibleTo("OpenTelemetry.Extensions.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010051c1562a090fb0c9f391012a32198b5e5d9a60e9b80fa2d7b434c9e5ccb7259bd606e66f9660676afc6692b8cdc6793d190904551d2103b7b22fa636dcbb8208839785ba402ea08fc00c8f1500ccef28bbf599aa64ffb1e1d5dc1bf3420a3777badfe697856e9d52070a50c3ea5821c80bef17ca3acffa28f89dd413f096f898")]
+#else
+[assembly: InternalsVisibleTo("OpenTelemetry.Extensions.Tests")]
+#endif
+
```
martincostello: This should be dealt with automatically by this:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/7befd2db5ca77819f0a196fa8c477a0941ffec52/build/Common.prod.props#L30-L32

If there's a specific reason that doesn't add it that isn't a bug, then it should be added to the project file like this:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/7befd2db5ca77819f0a196fa8c477a0941ffec52/src/OpenTelemetry.Exporter.Geneva/OpenTelemetry.Exporter.Geneva.csproj#L35-L38

.editorconfig
```diff
@@ -159,6 +159,7 @@ dotnet_diagnostic.IDE0005.severity = warning
 # RS0041: Public members should not use oblivious types
 dotnet_diagnostic.RS0041.severity = suggestion
 
+
```
martincostello: This file shouldn't be in the diff at all anymore as nothing needs to be changed.

```suggestion
```

src/OpenTelemetry.Extensions/AssemblyInfo.cs
mladjan-gadzic: just rebased on top of main and added it as specified in csproj and it was fixed.

src/OpenTelemetry.Extensions/OpenTelemetry.Extensions.csproj
```diff
@@ -21,6 +21,11 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryCoreLatestVersion)" />
+    <PackageReference Include="System.Text.Json" Version="$(SystemTextJsonLatestNet8OutOfBandPkgVer)" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="OpenTelemetry.Extensions.Tests" PublicKey="$(StrongNamePublicKey)" />
   </ItemGroup>
```
martincostello: If rebased onto main then this _shouldn't_ be needed at all as the convention is to add `$(Name).Tests` to all of the shipping assemblies.

mladjan-gadzic: hmm, let me try that instead.

mladjan-gadzic: thanks for pointing this out. removed.

PR #3023, opened 2025-08-21
Sql summary tests

stevejgordon:
Contributes to #2238. 

We discussed benchmarking and optimising this code with a view to making it opt-out (summary enabled by default). I will do that in a follow up PR. As this relies on `StringBuilder` and heap allocated state, this can likely be optimised with a move to `Span`-based parsing. 

## Changes

- Update the tests to use the test cases from the semantic conventions repository
- Normalize concurrent whitespace in the summary
- Support cross join using comma-separated targets

cc @alanwest @martincostello


test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
```diff
@@ -15,36 +18,68 @@ public static class SqlProcessorTestCases
         Converters = { new JsonStringEnumConverter() },
     };
 
-    public static IEnumerable<object[]> GetTestCases()
+    private static readonly Uri Uri = new("https://raw.githubusercontent.com/open-telemetry/semantic-conventions/refs/heads/main/docs/non-normative/database-test-cases/db-sql-test-cases.json");
```
martincostello: I think it would be better to keep the existing in-repo copy read from resources, then in a follow-up I can look to add a GitHub Actions workflow to periodically open a PR to update the content if it drifts. Then we're isolated from CI breaking if some drastic refactoring happens in open-telemetry/semantic-conventions.

We can then also avoid the use of HttpClient and the `#if`s to deal with calling it sync in the `[MemberData]` method.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
```diff
@@ -15,36 +18,68 @@ public static class SqlProcessorTestCases
...
+    private static readonly HashSet<string> DbSystemTestCasesToExecute = ["other_sql"];
+
+    public static IEnumerable<object[]> GetSemanticConventionsTestCases()
```
martincostello: If you use `TheoryData<>` we can make things strongly typed.

stevejgordon: As the existing implementation was using `MemberData` I stuck with that. Happy to update as part of this PR.

martincostello: I mean change the method pointed to by `[MemberData]` to return a `TheoryData<T1, T2, T...>` instead of `IEnumerable<object[]>`.

martincostello: Like this: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/eddd79f5f9657b4c45ab165bc6e208ac7012a3c8/test/OpenTelemetry.Instrumentation.EntityFrameworkCore.Tests/EntityFrameworkDiagnosticListenerTests.cs#L32

stevejgordon: It's a bit more complex as you can't use the `yield` keyword in that case. I'm not sure it offers a huge benefit to refactor this.

martincostello: OK, I can look to refactor it separately.

I don't think the tests are performance sensitive enough to need to worry about using `yield` (rather than just return them all) and I think it would be better to enjoy the type safety if it's a choice between the two.

martincostello (approved):

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
stevejgordon: I can update

stevejgordon: I've updated to `TheoryData`. For now I preferred using the `TestCase` type directly. This results in nicer display in VS for example:

image

The downside is needing to implement `IXunitSerializable`. The alternative is passing back `string` instances instead, which works but is less pleasant in the UI.

martincostello: I think that's good enough as it lets you easily find/see the specific test cases.

alanwest (approved):
Thanks @stevejgordon!

> We discussed benchmarking and optimising this code with a view to making it opt-out (summary enabled by default).

I'm proposing we remove the configuration option altogether. Just opened #3026

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
martincostello: #3030

PR #3031, opened 2025-08-22
[SqlClient] Add benchmarks for SqlProcessor and implement initial quick wins

stevejgordon:
Contributes to #2238
Follow up from #3023

## Changes
Implement some initial quick wins to avoid `SqlProcessor` overhead.

- Avoid boxing using `Dictionary` rather than `Hashtable`
- Set initial `StringBuilder` capacity
- Avoid `SubString` on NET targets

These changes approximately half the time required when a statement is cached (moving to Dictionary). For non-cached statements they save approx 30% if the allocations and reduce execution time by approx 36%.

### Before (with cache disabled)
```
BenchmarkDotNet v0.13.12, Windows 11 (10.0.26100.4946)
Unknown processor
.NET SDK 9.0.304
  [Host]     : .NET 9.0.8 (9.0.825.36511), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 9.0.8 (9.0.825.36511), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI


| Method | Sql                  | Iterations | Mean       | Error    | StdDev    | Gen0   | Allocated |
|------- |--------------------- |----------- |-----------:|---------:|----------:|-------:|----------:|
| Simple | SELEC(...)ls od[39]  | 1          |   445.2 ns |  4.32 ns |   3.83 ns | 0.0687 |     864 B |
| Simple | SELE(...)_id) [101]  | 1          | 1,349.6 ns | 46.81 ns | 136.55 ns | 0.0935 |    1184 B |
```

### After (with cache disabled)
```
BenchmarkDotNet v0.13.12, Windows 11 (10.0.26100.4946)
Unknown processor
.NET SDK 9.0.304
  [Host]     : .NET 9.0.8 (9.0.825.36511), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI
  DefaultJob : .NET 9.0.8 (9.0.825.36511), X64 RyuJIT AVX-512F+CD+BW+DQ+VL+VBMI


| Method | Sql                  | Iterations | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------- |--------------------- |----------- |---------:|---------:|---------:|-------:|----------:|
| Simple | SELEC(...)ls od [39] | 1          | 355.9 ns |  3.49 ns |  2.92 ns | 0.0420 |     528 B |
| Simple | SELE(...)_id) [101]  | 1          | 804.1 ns | 14.63 ns | 12.97 ns | 0.0725 |     912 B |
```

~I've left the summary benchmark results in SqlProcessorBenchmarks for comparison.~

We can save several hundred bytes per statement (exact saving depends on the input) by avoid the `StringBuilder`s. This is a larger change to switch to rented memory + span processing, at least for NET targets. I would expect execution time to reduce also. This is something I can investigate if we think the gain offsets the effort? I'd also look to remove the `SqlProcessorState` to save another 40B per uncached statement.

* [ ] ~Unit tests added/updated~


src/Shared/SqlProcessor.cs
```diff
@@ -1,15 +1,14 @@
...
-using System.Collections;
 using System.Text;
 
 namespace OpenTelemetry.Instrumentation;
 
 internal static class SqlProcessor
 {
     private const int CacheCapacity = 1000;
-    private static readonly Hashtable Cache = [];
+    private static readonly Dictionary<string, SqlStatementInfo> Cache = [];
```
martincostello, line 11: I wonder if it is worth pre-allocating some space for the cache?

test/OpenTelemetry.Contrib.Shared.Benchmarks/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    benchmarks.Simple();
+
+    MemoryProfiler.GetSnapshot();
+
+    // benchmarks.Sql = "SELECT order_date\nFROM   (SELECT *\nFROM   orders o\nJOIN customers c\nON o.customer_id = c.customer_id)";
+
+    // benchmarks.Simple();
+
+    // Console.ReadKey();
```
martincostello: Leftovers?

test/OpenTelemetry.Contrib.Shared.Benchmarks/SqlProcessorBenchmarks.cs
```diff
@@ -0,0 +1,141 @@
...
+
+| Method | Sql                  | Iterations | Mean     | Error    | StdDev   | Allocated |
+|------- |--------------------- |----------- |---------:|---------:|---------:|----------:|
+| Simple | SELEC(...)ls od [39] | 1          | 15.47 ns | 0.336 ns | 0.373 ns |         - |
+| Simple | SELE(...)_id) [101]  | 1          | 30.94 ns | 0.505 ns | 0.472 ns |         - |
+
+// * Hints *
+Outliers
+  SqlProcessorBenchmarks.Simple: Default -> 1 outlier  was  removed (18.57 ns)
+*/
```
martincostello: Maybe check the `.md` result files in initially with the old implementation, then another commit with the updated ones? Then they're easy to read using the Git diff (basically what I did just now with https://github.com/App-vNext/Polly/commit/cf50bc52196f3153ca8ca7b2f6900d8d451a4a92).

stevejgordon: Yeah, can do. I saw that in most other benchmarks one or more results where just commented at the top, so I copied that. We don't need the full trail here though.

test/OpenTelemetry.Contrib.Shared.Benchmarks/Program.cs
stevejgordon: I can remove. I had left them there in case anyone wanted to use them, but it's probably unnecessary.

src/Shared/SqlProcessor.cs
stevejgordon: This one is hard to say. Some apps may include the instrumentation (via autoinstrumentation) but not have SQL at all, others may have only a few statements. I'd lean towards just letting it grow as needed.

test/OpenTelemetry.Contrib.Shared.Benchmarks/SqlProcessorBenchmarks.cs
stevejgordon: @martincostello , I've updated the config and commited the before and after results.

martincostello (approved):
Nice.

It's much easier to grok the improvement by looking at [this](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3031/commits/9af19ebf29aaf31f2ff550f8374f6dbd851787c5#diff-fb0504062261995102778c200774fdb5a5a904bec6165d17582bc5908e071f88) 👍

Kielek:
Only some `infra` stuff comments.

I know that you are adding benchmarks to the `Shared` folder, but it is done mostly for SqlClient.

There are 2 options to fix it.

1. Move your code to SqlClient specific test project, as defined under https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c9913e8e36c741f85b22f3b1ce8377389fe26d1d/build/Projects/Component.proj#L14-L16. No more infrastructure related stuff is needed.
2. Keep it as Shared benchmarks, it is also fine, but requires some additional changes.
  * need to update detection changes: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c9913e8e36c741f85b22f3b1ce8377389fe26d1d/.github/workflows/ci.yml#L33
  * Extend this project definiotion https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c9913e8e36c741f85b22f3b1ce8377389fe26d1d/build/Projects/OpenTelemetry.Contrib.Shared.Tests.proj by benchmarks, similarly to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/c9913e8e36c741f85b22f3b1ce8377389fe26d1d/build/Projects/Component.proj#L14-L16


I have resolved conflicts with the main during review.


EDIT: The second option is probably better from the infrastructure perspective, the first one, for the SqlClient developers.

stevejgordon:
@Kielek Thanks. I went for a shared project as I assumed at some point their may be other benchmarks relating to shared content, but I'm happy with whatever the team think is best.

Kielek:
I think that, we can go with Option 2 and keep in share project. 
Could you please adjust infrastructure stuff?

Kielek (approved):

Kielek:
infra adjusted as agreed with @stevejgordon in private channel.

PR #3043, opened 2025-08-29
[Resources.AWS.Tests] Fix unit test result when running on EC2

lukeina2z:
When the unit test in AWSEC2DetectorTests runs on an EC2 machine, it reports a failure. The test assumes it will never run on an EC2 instance, which is not a safe assumption.

This fix adds a conditional check to detect if the current environment is EC2 and adjusts the assertion accordingly.

dotnet build: pass
dotnet test: pass


test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
```diff
@@ -1,20 +1,47 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Net.Http;
```
Kielek, line 5: Before creating PR, consider building your changes in `Release` mode.
It should detect the same issues as our CI.

One of example is redundant using.

```suggestion
```

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
```diff
@@ -1,20 +1,47 @@
...
 using Xunit;
 
 namespace OpenTelemetry.Resources.AWS.Tests;
 
 public class AWSEC2DetectorTests
 {
+    private static readonly HttpClient HttpClient = new HttpClient
```
Kielek:
```suggestion
    private static readonly HttpClient HttpClient = new()

```

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
```diff
@@ -1,20 +1,47 @@
...
+    {
+        Timeout = TimeSpan.FromSeconds(3),
+    };
+
+    public static bool IsRunningOnEC2()
+    {
+        try
+        {
+            var task = HttpClient.GetStringAsync(new Uri("http://169.254.169.254/latest/meta-data/instance-id"));
+            task.Wait();
```
Kielek: xUnit is working fine with async method. It is bad practice to utilize blocking methods (Wait) even within test scope.

cosnider something like

```csharp
    public static async Task<bool> IsRunningOnEC2()
    {
        try
        {
            var instanceId = await HttpClient.GetStringAsync(new Uri("http://169.254.169.254/latest/meta-data/instance-id"));
            return !string.IsNullOrEmpty(instanceId);
        }
        catch
        {
            return false;
        }
    }

    [Fact]
    public async Task TestDetect()
    {
        var awsEC2Detector = new AWSEC2Detector(
            new OpenTelemetry.AWS.AWSSemanticConventions(
                SemanticConventionVersion.Latest));

        if (!await IsRunningOnEC2())
        {
            Assert.Empty(awsEC2Detector.Detect().Attributes); // will be null as it's not in ec2 environment
        }
        else
        {
            Assert.NotEmpty(awsEC2Detector.Detect().Attributes);
        }
    }
```

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
```diff
@@ -1,20 +1,47 @@
...
+    private static readonly HttpClient HttpClient = new HttpClient
+    {
+        Timeout = TimeSpan.FromSeconds(3),
+    };
+
+    public static bool IsRunningOnEC2()
+    {
+        try
+        {
+            var task = HttpClient.GetStringAsync(new Uri("http://169.254.169.254/latest/meta-data/instance-id"));
```
Kielek: It will be great to put some reference to this magic string, even if it is already in this repository.

Kielek:
@normj, FYI

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
lukeina2z: Good to learn this.

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
lukeina2z: changed.

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
lukeina2z: netfx still needs this. Will add directive around it.

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
lukeina2z: great code. Like the async/await usage.

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
lukeina2z: changed magic string to local const string.

test/OpenTelemetry.Resources.AWS.Tests/AWSEC2DetectorTests.cs
```diff
@@ -1,20 +1,49 @@
...
 
-        Assert.Empty(awsEC2Detector.Detect().Attributes); // will be null as it's not in ec2 environment
+        if (!await IsRunningOnEC2())
+        {
+            Assert.Empty(awsEC2Detector.Detect().Attributes); // will be null as it's not in ec2 environment
+        }
+        else
+        {
+            Assert.NotEmpty(awsEC2Detector.Detect().Attributes);
+        }
```
martincostello: To avoid the duplication here, put the attributes into a local variable, then assert on it after the check on the environment.

lukeina2z: Good call.

Kielek (approved):

normj (approved):

Issue #3049, opened 2025-09-02
[bug] AWS SDK instrumentation library causes auto-instrumentation library to fail generating traces when app uses AWS SDK v3


lukeina2z:
### Component

OpenTelemetry.Instrumentation.AWS

### Package Version

Project 'OpenTelemetry.Instrumentation.AWS' has the following package references
   [net472]:
   Top-level Package                                 Requested              Resolved
   > AWSSDK.Core                                     [4.0.0, 5.0.0)         3.7.400
   > AWSSDK.SimpleNotificationService                [4.0.0, 5.0.0)         3.7.400
   > AWSSDK.SQS                                      [4.0.0, 5.0.0)         3.7.400
   > Microsoft.CodeAnalysis.PublicApiAnalyzers       [4.14.0]               4.14.0
   > Microsoft.NETFramework.ReferenceAssemblies      [1.0.3,2.0)            1.0.3
   > Microsoft.SourceLink.GitHub                     [8.0.0,9.0)            8.0.0
   > MinVer                                          [6.0.0,7.0)            6.0.0
   > StyleCop.Analyzers                              [1.2.0-beta.556,2.0)   1.2.0-beta.556

   [netstandard2.0]:
   Top-level Package                                    Requested              Resolved
   > AWSSDK.Core                                        [4.0.0, 5.0.0)         3.7.400
   > AWSSDK.SimpleNotificationService                   [4.0.0, 5.0.0)         3.7.400
   > AWSSDK.SQS                                         [4.0.0, 5.0.0)         3.7.400
   > Microsoft.CodeAnalysis.PublicApiAnalyzers          [4.14.0]               4.14.0
   > Microsoft.NETFramework.ReferenceAssemblies         [1.0.3,2.0)            1.0.3
   > Microsoft.SourceLink.GitHub                        [8.0.0,9.0)            8.0.0
   > MinVer                                             [6.0.0,7.0)            6.0.0
   > NETStandard.Library                          (A)   [2.0.3, )              2.0.3
   > StyleCop.Analyzers                                 [1.2.0-beta.556,2.0)   1.2.0-beta.556

   [net8.0]:
   Top-level Package                                 Requested              Resolved
   > AWSSDK.Core                                     [4.0.0, 5.0.0)         3.7.400
   > AWSSDK.SimpleNotificationService                [4.0.0, 5.0.0)         3.7.400
[...]

PR #2720, opened 2025-04-29
[Instrumentation.AWS] Update to AWSSDK v4

martincostello:
Fixes #2719

## Changes

Update OpenTelemetry.Instrumentation.AWS to depend on v4 of the AWS SDK for .NET.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

Issue #2719, opened 2025-04-29
[feature request] Support AWS SDK v4


martincostello:
### Component

OpenTelemetry.Instrumentation.AWS

### Is your feature request related to a problem?

Version 4 of the AWSSDK [has just been shipped](https://aws.amazon.com/blogs/developer/general-availability-of-aws-sdk-for-net-v4-0/).

> It's not possible to use both V3 and V4 packages within an application. Conflicts between the [AWSSDK.Core](https://www.nuget.org/packages/AWSSDK.Core) package and service packages like [AWSSDK.SQS](https://www.nuget.org/packages/AWSSDK.SQS) will occur if you attempt to mix V3 and V4 packages. When updating to V4, ensure that all references to AWSSDK.* packages are updated to 4.0.0 or greater.

The OpenTelemetry.Instrumentation.AWS package depends on v3.7.400 of [AWSSDK.Core](https://www.nuget.org/packages/AWSSDK.Core/), [AWSSDK.SimpleNotificationService](https://www.nuget.org/packages/AWSSDK.SimpleNotificationService/) and [AWSSDK.SQS](https://www.nuget.org/packages/AWSSDK.SQS/) which creates issues for applications wanting to use the v4 SDK that use these packages as v3 and v4 packages cannot be used with each other.

/cc @normj

### What is the expected behavior?

Applications using AWS-specific OpenTelemetry NuGet packages can use the v4 SDK.

### Which alternative solutions or features have you considered?

None.

ppittle:
Thinking aloud: Given that AWS SDK v4 is incompatible with v3, what kind of versioning bump should this require for the AWS OpenTelemetry packages?

normj:
My understanding from what @muhammad-othman told me when he was doing the testing the OpenTelemetry package is it will work with either V3 or V4. The SDK's observability API inside the SDK that this package uses haven't changed between V3 and V4.

I would assume users would be bringing in the AWS SDK dependencies first and then the OpenTelemetry package that would integrate with what is available. That being I'm sure there are friction points like the referenced service packages on the OpenTelemetry package like SQS. Users would need to be sure if they are using those services to pull in their own V4 version and not rely on the one in the SDK. Plus we will eventually add new api/features to the SDK's observability in V4 that won't be available in V4. So in the interest of encouraging V4 update I support doing a minor version update of the OpenTelemetry package to use V4.

muhammad-othman:
[...]

PR #3050, opened 2025-09-02
[fix] #3049 Support AWS SDK v3 in AWS SDK instrumentation library

lukeina2z:
This PR addresses issue #3049:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3049

The AWS SDK instrumentation library should support both v3 and v4 SDKs. While tests may primarily focus on v4, SDK v3 is not deprecated, and the instrumentation library code is compatible with both versions.

Validation / Testing:
dotnet build
dotnet test (all test failures were investigated and are not caused by this PR) 
Manual testing with instrumentation enabled on apps using AWS SDK v3
Manual testing with instrumentation enabled on apps using AWS SDK v4

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
+++ b/src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
@@ -23,9 +23,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="[4.0.0, 5.0.0)" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[4.0.0, 5.0.0)" />
-    <PackageReference Include="AWSSDK.SQS" Version="[4.0.0, 5.0.0)" />
+    <PackageReference Include="AWSSDK.Core" Version="[3.7.400, 5.0.0)" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[3.7.400, 5.0.0)" />
+    <PackageReference Include="AWSSDK.SQS" Version="[3.7.400, 5.0.0)" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWS/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Add support for AWS SDK v3 to address this issue.
+  ([#3049](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3049))
+
 ## 1.12.0
 
 Released 2025-May-06
--- a/test/OpenTelemetry.Instrumentation.AWS.Tests/OpenTelemetry.Instrumentation.AWS.Tests.csproj
+++ b/test/OpenTelemetry.Instrumentation.AWS.Tests/OpenTelemetry.Instrumentation.AWS.Tests.csproj
@@ -15,6 +15,8 @@
     <PackageReference Include="AWSSDK.BedrockAgentRuntime" Version="4.0.0" />
     <PackageReference Include="AWSSDK.BedrockRuntime" Version="4.0.0" />
     <PackageReference Include="AWSSDK.S3" Version="4.0.0" />
+    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="4.0.0" />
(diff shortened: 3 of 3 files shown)


Kielek:
@lukeina2z, what's about issue described #2719 and solved in #2720?

src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj
```diff
@@ -23,9 +23,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="AWSSDK.Core" Version="[4.0.0, 5.0.0)" />
-    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[4.0.0, 5.0.0)" />
-    <PackageReference Include="AWSSDK.SQS" Version="[4.0.0, 5.0.0)" />
+    <PackageReference Include="AWSSDK.Core" Version="[3.7.400, 5.0.0)" />
```
martincostello, line 26: I'm sceptical this will work, but I'm willing to be proven wrong. I can see what happens if I try to use the prerelease packages from this PR in an application of mine using v4 and see if it breaks.

When testing various applications using OpenTelemetry with the v4 SDK, there were various runtime (not compile time) issues that made the version bump necessary for them to be functional, mostly due to breaking changes in AWSSDK.Core.

If that is still the case, then it won't be possible to support both in the same packages. We'd have to introduce a new parallel set for v3 support.

martincostello: Confirmed - this change does not work. It breaks applications using the v4 SDK:

> /home/runner/work/alexa-london-travel/alexa-london-travel/src/LondonTravel.Skill/LondonTravel.Skill.csproj : error NU1608: Detected package version outside of dependency constraint: AWSSDK.SQS 3.7.400 requires AWSSDK.Core (>= 3.7.400 && < 4.0.0) but version AWSSDK.Core 4.0.0.2 was resolved.
/home/runner/work/alexa-london-travel/alexa-london-travel/src/LondonTravel.Skill/LondonTravel.Skill.csproj : error NU1608: Detected package version outside of dependency constraint: AWSSDK.SimpleNotificationService 3.7.400 requires AWSSDK.Core (>= 3.7.400 && < 4.0.0) but version AWSSDK.Core 4.0.0.2 was resolved.

Tested in https://github.com/martincostello/alexa-london-travel/pull/1943.

lukeina2z: In my PR the constrain has been changed to [3.7.400, 5.0.0). 

So if your application depends on 3.7.400, the auto-instrumentation library should use AWSSDK.Core/v3.7.400. 

If your application depends on AWSSDK.Core/v4, then the instrumentation library should use AWSSDK.Core/v4.

May I know who loads AWSSDK.SQS/v3.7.400 in your testing app? Who loads AWSSDK.Core/v4?

If you could provide me with access to your sample app and repro instruction. That would be very helpful.

martincostello: https://github.com/martincostello/alexa-london-travel

lukeina2z: Thanks for sharing the app. The problem is AWS SDK Instrumentation Library has two more dependencies SNS and SQS beside Core. If we can stick with Core only by removing  SNS and SQS. That will support both v3 and v4.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AWS/OpenTelemetry.Instrumentation.AWS.csproj#L27-L28

lukeina2z: As a workaround, explicitly add  SNS & SQS v4 in your project should address the build break. But, I understand this will cause unpleasant  dev experience.

    <PackageReference Include="AWSSDK.SimpleNotificationService" Version="[4.0.0, 5.0.0)" />
    <PackageReference Include="AWSSDK.SQS" Version="[4.0.0, 5.0.0)" />

martincostello: My project is just one datapoint, but it's just highlighting that this isn't a simple problem to fix.

PR #3064, opened 2025-09-03
[OpAMP.Client] WebSocket Transport

RassK:
## Changes

Adds WebSockets support for .NET.

Leaving TODO:
- WebSockets for .NET Framework
- WebSockets sending large packages (WsTransmitter.cs)
- Certification handling

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
@@ -1,0 +1,134 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NET
+
+using System.Buffers;
+using System.Net.WebSockets;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsReceiver : IDisposable
+{
+    private const int RentalBufferSize = 4096;
+    private const int ReceiveBufferSize = 8192;
+
+    private readonly ClientWebSocket ws;
+    private readonly Thread receiveThread;
+    private readonly FrameProcessor processor;
+
+    private readonly byte[] receiveBuffer = new byte[ReceiveBufferSize];
+
+    private CancellationToken token;
+
+    public WsReceiver(ClientWebSocket ws, FrameProcessor processor)
+    {
+        Guard.ThrowIfNull(ws, nameof(ws));
+        Guard.ThrowIfNull(processor, nameof(processor));
+
+        this.ws = ws;
+        this.processor = processor;
+        this.receiveThread = new Thread(this.ReceiveLoop);
+    }
+
+    public void Start(CancellationToken token = default)
+    {
+        this.token = token;
+        this.receiveThread.Start();
+    }
+
+    public void Dispose()
+    {
+        this.receiveThread?.Join();
+    }
+
+    private static void ReturnRentalBuffers(List<byte[]>? rentalBuffers)
+    {
+        if (rentalBuffers == null)
+        {
+            return;
+        }
+
+        foreach (var rental in rentalBuffers)
+        {
+            ArrayPool<byte>.Shared.Return(rental);
+        }
+    }
+
+    private async void ReceiveLoop()
+    {
+        while (true)
+        {
+            if (this.token.IsCancellationRequested)
+            {
+                this.token.ThrowIfCancellationRequested();
+            }
+
+            await this.ReceiveAsync(this.token).ConfigureAwait(false);
+        }
+    }
+
+    private async Task ReceiveAsync(CancellationToken token)
+    {
+        var totalCount = 0;
+        var workingCount = 0;
+        var isClosed = false;
+        WebSocketReceiveResult result;
+        byte[] workingBuffer = this.receiveBuffer;
+
+        List<byte[]>? rentalBuffers = null;
+
+        do
+        {
+            if (this.token.IsCancellationRequested)
+            {
+                this.token.ThrowIfCancellationRequested();
+            }
+
+            // out of space, need to rent more
+            if (workingBuffer.Length - workingCount == 0)
+            {
+                if (rentalBuffers == null)
+                {
+                    rentalBuffers = [this.receiveBuffer];
+                }
+
+                workingBuffer = ArrayPool<byte>.Shared.Rent(RentalBufferSize);
+                workingCount = 0;
+                rentalBuffers.Add(workingBuffer);
+            }
+
+            if (this.ws.State != WebSocketState.Open)
+            {
+                // Connection is closed, nothing more to read.
+                isClosed = true;
+                break;
+            }
+
+            var segment1 = new ArraySegment<byte>(workingBuffer, workingCount, workingBuffer.Length - workingCount);
+            result = await this.ws.ReceiveAsync(segment1, token).ConfigureAwait(false);
+
+            isClosed = result.CloseStatus != null;
+            workingCount += result.Count;
+            totalCount += result.Count;
+        }
+        while (!result.EndOfMessage);
+
+        if (!isClosed)
+        {
+            var sequence =
+                rentalBuffers?.Count > 1
+                    ? rentalBuffers.CreateSequenceFromBuffers(workingCount + 1)
+                    : new ReadOnlySequence<byte>(this.receiveBuffer);
+
+            this.processor.OnServerFrame(sequence, totalCount, verifyHeader: true);
+        }
+
+        ReturnRentalBuffers(rentalBuffers);
+    }
+}
+
+#endif
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
@@ -1,0 +1,54 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NET
+
+using System.Net.WebSockets;
+using Google.Protobuf;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsTransmitter
+{
+    private const int BufferSize = 4096;
+
+    private readonly byte[] buffer = new byte[BufferSize];
+
+    private readonly ClientWebSocket ws;
+
+    public WsTransmitter(ClientWebSocket ws)
+    {
+        Guard.ThrowIfNull(ws, nameof(ws));
+
+        this.ws = ws;
+    }
+
+    public async Task SendAsync(IMessage message, CancellationToken token = default)
+    {
+        var headerSize = OpAmpWsHeaderHelper.WriteHeader(this.buffer);
+        var size = message.CalculateSize();
+
+        // fits to the buffer, send completely
+        if (size + headerSize <= BufferSize)
+        {
+            var segment = new ArraySegment<byte>(this.buffer, headerSize, size);
+            message.WriteTo(segment);
+
+            // resegment to include the header byte
+            segment = new ArraySegment<byte>(this.buffer, 0, size + headerSize);
+
+            await this.ws.SendAsync(segment, WebSocketMessageType.Binary, true, token).ConfigureAwait(false);
+        }
+
+        // Does not fit, need to chunk the message
+        else
+        {
+            // TODO: Implement chunking logic for large messages
+            throw new NotImplementedException();
+        }
+    }
+}
+
+#endif
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
@@ -1,0 +1,70 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if NET
+
+using System.Net.WebSockets;
+using Google.Protobuf;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Transport;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsTransport : IOpAmpTransport, IDisposable
+{
+    private readonly Uri uri;
+    private readonly SocketsHttpHandler handler = new();
+    private readonly ClientWebSocket ws = new();
+    private readonly WsReceiver receiver;
+    private readonly WsTransmitter transmitter;
+    private readonly FrameProcessor processor;
+
+    public WsTransport(Uri serverUrl, FrameProcessor processor)
+    {
+        Guard.ThrowIfNull(serverUrl, nameof(serverUrl));
+        Guard.ThrowIfNull(processor, nameof(processor));
+
+        // TODO: fix trust all certificates
+#pragma warning disable CA5359 // Do Not Disable Certificate Validation
+        this.handler.SslOptions.RemoteCertificateValidationCallback = (sender, certificate, chain, sslPolicyErrors) => true;
+#pragma warning restore CA5359 // Do Not Disable Certificate Validation
+        this.uri = serverUrl;
+        this.processor = processor;
+        this.receiver = new WsReceiver(this.ws, this.processor);
+        this.transmitter = new WsTransmitter(this.ws);
+    }
+
+    public async Task StartAsync(CancellationToken token = default)
+    {
+        using var invoker = new HttpMessageInvoker(this.handler);
+
+        await this.ws
+            .ConnectAsync(this.uri, invoker, token)
+            .ConfigureAwait(false);
+
+        this.receiver.Start(token);
+    }
+
+    public async Task StopAsync(CancellationToken token = default)
+    {
+        await this.ws
+            .CloseAsync(WebSocketCloseStatus.NormalClosure, "Client closed connection", token)
+            .ConfigureAwait(false);
+    }
+
+    public Task SendAsync<T>(T message, CancellationToken token = default)
+        where T : IMessage<T>
+    {
+        return this.transmitter.SendAsync(message, token);
+    }
+
+    public void Dispose()
+    {
+        this.handler.Dispose();
+        this.ws.Dispose();
+        this.receiver.Dispose();
+    }
+}
+
+#endif
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Utils/BufferSegment.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Utils/BufferSegment.cs
@@ -1,0 +1,21 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Buffers;
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal class BufferSegment : ReadOnlySequenceSegment<byte>
+{
+    public BufferSegment(byte[] buffer)
+    {
+        this.Memory = new ReadOnlyMemory<byte>(buffer);
+    }
+
+    public void SetNext(BufferSegment segment)
+    {
+        this.Next = segment;
+
+        segment.RunningIndex = this.RunningIndex + this.Memory.Length;
+    }
+}
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
@@ -1,0 +1,89 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal class OpAmpWsHeaderHelper
+{
+    public const int OpAmpProtocolHeader = 0x00;
+    public const int MaxHeaderLength = 10; // Maximum length for varint64 encoding
+
+    public static int WriteHeader(ArraySegment<byte> buffer, ulong header = 0)
+    {
+        if (buffer.Count < MaxHeaderLength)
+        {
+            throw new InvalidOperationException("Ensure 10 bytes for buffer.");
+        }
+
+        return EncodeVarint64(buffer, header);
+    }
+
+    public static bool TryVerifyHeader(ReadOnlyMemory<byte> buffer, out int headerSize)
+    {
+        try
+        {
+            var header = DecodeVarint64(buffer.Span, out headerSize);
+            return header == OpAmpProtocolHeader;
+        }
+        catch (Exception)
+        {
+            headerSize = -1;
+            return false;
+        }
+    }
+
+    public static ulong DecodeVarint64(ReadOnlySpan<byte> buffer, out int bytesRead)
+    {
+        ulong result = 0;
+        int shift = 0;
+        bytesRead = 0;
+
+        foreach (byte b in buffer)
+        {
+            ulong value = (ulong)(b & 0x7F);
+            result |= value << shift;
+            bytesRead++;
+
+            if ((b & 0x80) == 0)
+            {
+                return result;
+            }
+
+            shift += 7;
+
+            // 64 bits max + buffer
+            if (shift >= 70)
+            {
+                throw new OverflowException("Varint is too long for 64-bit integer.");
+            }
+        }
+
+        throw new ArgumentException("Incomplete varint data.");
+    }
+
+    public static int EncodeVarint64(ArraySegment<byte> buffer, ulong value)
+    {
+        int bytesWritten = 0;
+        int offset = buffer.Offset;
+
+        while (value > 0x7F)
+        {
+            if (bytesWritten >= buffer.Count)
+            {
+                throw new ArgumentException("Buffer is too small for varint encoding.");
+            }
+
+            buffer.Array![offset + bytesWritten++] = (byte)((value & 0x7F) | 0x80);
+            value >>= 7;
+        }
+
+        if (bytesWritten >= buffer.Count)
+        {
+            throw new ArgumentException("Buffer is too small for varint encoding.");
+        }
+
+        buffer.Array![offset + bytesWritten++] = (byte)value; // Last byte without continuation bit
+
+        return bytesWritten;
+    }
+}
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
@@ -7,6 +7,21 @@ namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
 
 internal static class SequenceHelper
 {
+    public static ReadOnlyMemory<byte> GetHeaderSegment(ReadOnlySequence<byte> sequence)
+    {
+        if (sequence.IsEmpty)
+        {
+            return default;
+        }
+
+        if (sequence.First.Length < OpAmpWsHeaderHelper.MaxHeaderLength)
+        {
+            throw new InvalidOperationException("Sequence is too small to contain a valid header.");
+        }
+
+        return sequence.First.Slice(0, OpAmpWsHeaderHelper.MaxHeaderLength);
+    }
+
     public static ReadOnlySequence<byte> AsSequence(this byte[] message)
     {
         if (message == null || message.Length == 0)
@@ -15,5 +30,26 @@ internal static class SequenceHelper
         }
 
         return new ReadOnlySequence<byte>(message);
+    }
+
+    public static ReadOnlySequence<byte> CreateSequenceFromBuffers(this IReadOnlyList<byte[]> buffers, int endIndex)
+    {
+        if (buffers == null || buffers.Count == 0)
+        {
+            return ReadOnlySequence<byte>.Empty;
+        }
+
+        BufferSegment segment1 = new BufferSegment(buffers[0]);
+        BufferSegment currentSegment = segment1;
+
+        for (int i = 1; i < buffers.Count; i++)
+        {
+            BufferSegment nextSegment = new BufferSegment(buffers[i]);
+
+            currentSegment.SetNext(nextSegment);
+            currentSegment = nextSegment;
+        }
+
+        return new ReadOnlySequence<byte>(segment1, 0, currentSegment, endIndex);
     }
 }
--- a/test/OpenTelemetry.OpAmp.Client.Tests/Mocks/MockListener.cs
+++ b/test/OpenTelemetry.OpAmp.Client.Tests/Mocks/MockListener.cs
@@ -9,7 +9,18 @@ namespace OpenTelemetry.OpAmp.Client.Tests.Mocks;
 
 internal class MockListener : IOpAmpListener<CustomMessageMessage>
 {
+    private AutoResetEvent messageEvent = new(false);
+
     public ConcurrentBag<CustomMessageMessage> Messages { get; private set; } = [];
 
-    public void HandleMessage(CustomMessageMessage message) => this.Messages.Add(message);
+    public void HandleMessage(CustomMessageMessage message)
+    {
+        this.Messages.Add(message);
+        this.messageEvent.Set();
+    }
+
+    public void WaitForMessages(TimeSpan timeout)
+    {
+        this.messageEvent.WaitOne(timeout);
+    }
 }
--- a/test/OpenTelemetry.OpAmp.Client.Tests/SequenceHelperTests.cs
+++ b/test/OpenTelemetry.OpAmp.Client.Tests/SequenceHelperTests.cs
@@ -1,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Buffers;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+using Xunit;
+
+namespace OpenTelemetry.OpAmp.Client.Tests;
+
+public class SequenceHelperTests
+{
+    [Fact]
+    public void SequenceHelperTests_GetHeaderSegment()
+    {
+        byte[] data = Enumerable.Range(0, 12).Select(i => (byte)i).ToArray();
+        byte[] expexted = Enumerable.Range(0, OpAmpWsHeaderHelper.MaxHeaderLength).Select(i => (byte)i).ToArray();
+
+        var header = SequenceHelper.GetHeaderSegment(data.AsSequence());
+
+        Assert.Equal(expexted, header.ToArray());
+    }
+
+    [Fact]
+    public void SequenceHelperTests_GetHeaderSegment_TooSmallFirstSegment()
+    {
+        byte[] data = Enumerable.Range(0, 5).Select(i => (byte)i).ToArray();
+
+        Assert.Throws<InvalidOperationException>(() => SequenceHelper.GetHeaderSegment(data.AsSequence()));
+    }
+
+    [Fact]
+    public void SequenceHelperTests_AsSequence()
+    {
+        var buffer = new byte[] { 0x04, 0x05, 0x06, 0x07 };
+        var sequence = buffer.AsSequence();
+
+        Assert.Equal(buffer, sequence.ToArray());
+    }
+
+    [Fact]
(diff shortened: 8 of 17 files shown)


src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Net.WebSockets;
+using Google.Protobuf;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsTransmitter
+{
+    private const int BufferSize = 4096;
```
RassK, line 15: Need to choose realistic buffer size, anyone suggests what are the usuals?

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Net.WebSockets;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsReceiver : IDisposable
+{
+    private const int RentalBufferSize = 4096;
```
RassK: Need to choose realistic buffer size, anyone suggests what are the usuals?

test/OpenTelemetry.OpAmp.Client.Tests/Mocks/MockListener.cs
```diff
@@ -9,7 +9,18 @@ namespace OpenTelemetry.OpAmp.Client.Tests.Mocks;
 
 internal class MockListener : IOpAmpListener<CustomMessageMessage>
 {
+    private AutoResetEvent messageEvent = new(false);
```
martincostello, line 12: This implements `IDisposable` so this type should implement it too to tidy up.

test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTest.cs
```diff
@@ -13,11 +11,13 @@ namespace OpenTelemetry.OpAmp.Client.Tests;
 
 public class PlainHttpTransportTest
 {
-    [Fact]
-    public async Task PlainHttpTransport_SendReceiveCommunication()
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task PlainHttpTransport_SendReceiveCommunication(bool useSmallPackages)
```
martincostello: Just curious, should it be `packets` not `packages`?

test/OpenTelemetry.OpAmp.Client.Tests/SequenceHelperTests.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        byte[] data = Enumerable.Range(0, 5).Select(i => (byte)i).ToArray();
+
+        Assert.Throws<InvalidOperationException>(() => SequenceHelper.GetHeaderSegment(data.AsSequence()));
+    }
+
+    [Fact]
+    public void SequenceHelperTests_AsSequence()
+    {
+        var buffer = new byte[] { 0x04, 0x05, 0x06, 0x07 };
```
martincostello, line 15: Do we need to worry about/test for endianness anywhere for this stuff to make sure the byte order is correct?

test/OpenTelemetry.OpAmp.Client.Tests/Tools/FrameGenerator.cs
```diff
@@ -1,32 +1,83 @@
...
+        return new MockAgentFrame(uid, frame);
+    }
+
+    public static MockServerFrame GenerateMockServerFrame(ByteString? instanceUid = null, bool isSmall = true, bool addHeader = false)
+    {
+        var content = "This is a mock server frame for testing purposes.";
+        if (!isSmall)
+        {
+            var padding = new string('A', 100_000);
+            content = string.Concat(content, Environment.NewLine, padding);
```
martincostello: Is it correct for the new line for the server frames to be OS specific, rather than always `\n`?

test/OpenTelemetry.OpAmp.Client.Tests/Tools/OpAmpFakeWebSocketServer.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        this.httpServer = TestWebSocketServer.RunServer(
+            async socket =>
+            {
+                var buffer = new byte[8 * 1024];
+                var ms = new MemoryStream();
+
+                try
+                {
+                    while (socket.State == WebSocketState.Open)
```
martincostello, line 29: Feels like this server should use a `CancellationTokenSource` which `Dispose()` can signal, and only run while it's not signalled. Then if the class is disposed without the socket closing the loop would exit more gracefully?

test/OpenTelemetry.OpAmp.Client.Tests/WsTransportTest.cs
```diff
@@ -0,0 +1,60 @@
...
+        frameProcessor.Subscribe(mockListener);
+
+        using var wsTransport = new WsTransport(opAmpEndpoint, frameProcessor);
+        await wsTransport.StartAsync(CancellationToken.None);
+
+        // Send only small packages, currently sending large package is not supported in WsTransport
+        var mockFrame = FrameGenerator.GenerateMockAgentFrame(isSmall: true);
+
+        // Act
+        var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```
martincostello:
```suggestion
        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
```

test/Shared/TestWebSocketServer.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            try
+            {
+#if NET
+                port = RandomNumberGenerator.GetInt32(2000, 5000);
+#else
+#pragma warning disable CA5394 // Do not use insecure randomness
+                port = GlobalRandom.Next(2000, 5000);
+#pragma warning restore CA5394 // Do not use insecure randomness
+#endif
```
martincostello, line 35: Does this support using port `0` and the code will just find a free port automatically? Kestrel supports doing that, for example.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
```diff
@@ -7,6 +7,21 @@ namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
...
+    public static ReadOnlyMemory<byte> GetHeaderSegment(ReadOnlySequence<byte> sequence)
+    {
+        if (sequence.IsEmpty)
+        {
+            return default;
+        }
+
+        if (sequence.First.Length < OpAmpWsHeaderHelper.MaxHeaderLength)
+        {
+            throw new InvalidOperationException("Sequence is too small to contain a valid header.");
```
martincostello: Include what that too-small size is in the exception message?

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
```diff
@@ -16,4 +31,25 @@ public static ReadOnlySequence<byte> AsSequence(this byte[] message)
...
     }
+
+    public static ReadOnlySequence<byte> CreateSequenceFromBuffers(this IReadOnlyList<byte[]> buffers, int endIndex)
+    {
+        if (buffers == null || buffers.Count == 0)
+        {
+            return ReadOnlySequence<byte>.Empty;
+        }
+
+        BufferSegment segment1 = new BufferSegment(buffers[0]);
```
martincostello: Is there any benefit in special-casing a length of 1 and just returning it without the loop and wrapping it with a segment?

src/OpenTelemetry.OpAmp.Client/Internal/Utils/BufferSegment.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Buffers;
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal class BufferSegment : ReadOnlySequenceSegment<byte>
```
martincostello:
```suggestion
internal sealed class BufferSegment : ReadOnlySequenceSegment<byte>
```

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
```diff
@@ -0,0 +1,89 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal class OpAmpWsHeaderHelper
```
martincostello:
```suggestion
internal static class OpAmpWsHeaderHelper
```

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public const int OpAmpProtocolHeader = 0x00;
+    public const int MaxHeaderLength = 10; // Maximum length for varint64 encoding
+
+    public static int WriteHeader(ArraySegment<byte> buffer, ulong header = 0)
+    {
+        if (buffer.Count < MaxHeaderLength)
+        {
+            throw new InvalidOperationException("Ensure 10 bytes for buffer.");
+        }
```
martincostello: If this is to trap an bug and not something the use could cause, maybe it should be a `Debug.Assert()` rather than an exception?

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+        try
+        {
+            var header = DecodeVarint64(buffer.Span, out headerSize);
+            return header == OpAmpProtocolHeader;
+        }
+        catch (Exception)
+        {
+            headerSize = -1;
+            return false;
+        }
```
martincostello: Would be more performant to use the `bool Try...(, out var ...)` pattern and avoid throwing and catching exceptions for invalid data.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            if ((b & 0x80) == 0)
+            {
+                return result;
+            }
+
+            shift += 7;
+
+            // 64 bits max + buffer
+            if (shift >= 70)
```
martincostello: Personally I think this is better of `0x7F`, `7` and `70` weren't magic numbers but named constants reused between the methods.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+            buffer.Array![offset + bytesWritten++] = (byte)((value & 0x7F) | 0x80);
+            value >>= 7;
+        }
+
+        if (bytesWritten >= buffer.Count)
+        {
+            throw new ArgumentException("Buffer is too small for varint encoding.");
+        }
+
+        buffer.Array![offset + bytesWritten++] = (byte)value; // Last byte without continuation bit
```
martincostello: This _might_ be able to be optimised like was done in https://github.com/open-telemetry/opentelemetry-dotnet/pull/6367.

Just flagging it as the code looks similar to how it was there before.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -0,0 +1,70 @@
...
+
+using System.Net.WebSockets;
+using Google.Protobuf;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Transport;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsTransport : IOpAmpTransport, IDisposable
```
martincostello: `sealed`? Otherwise it should implement a finalizer and call `Dispose(false)`.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -0,0 +1,70 @@
...
+    private readonly FrameProcessor processor;
+
+    public WsTransport(Uri serverUrl, FrameProcessor processor)
+    {
+        Guard.ThrowIfNull(serverUrl, nameof(serverUrl));
+        Guard.ThrowIfNull(processor, nameof(processor));
+
+        // TODO: fix trust all certificates
+#pragma warning disable CA5359 // Do Not Disable Certificate Validation
+        this.handler.SslOptions.RemoteCertificateValidationCallback = (sender, certificate, chain, sslPolicyErrors) => true;
```
martincostello: Using `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` is a better way to disable this where available as it doesn't flag the analyzer error.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
martincostello: 4096 is common in things like `Stream` - you could benchmark to find if there's a better value.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,140 @@
...
+
+using System.Buffers;
+using System.Net.WebSockets;
+using OpenTelemetry.Internal;
+using OpenTelemetry.OpAmp.Client.Internal;
+using OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+namespace OpenTelemetry.OpAmp.Client.Transport.WebSocket;
+
+internal class WsReceiver : IDisposable
```
martincostello: `sealed` or implement finalizer that disposes.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,140 @@
...
+        byte[] workingBuffer = this.receiveBuffer;
+
+        List<byte[]>? rentalBuffers = null;
+
+        do
+        {
+            if (this.token.IsCancellationRequested)
+            {
+                this.token.ThrowIfCancellationRequested();
+            }
```
martincostello: This should be the same?

```suggestion
            this.token.ThrowIfCancellationRequested();
```

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private async void ReceiveLoop()
+    {
+        while (true)
+        {
+            if (this.token.IsCancellationRequested)
+            {
+                this.token.ThrowIfCancellationRequested();
+            }
```
martincostello:
```suggestion
            this.token.ThrowIfCancellationRequested();
```

test/OpenTelemetry.OpAmp.Client.Tests/Mocks/MockListener.cs
RassK: interestingly well hidden so static analyser does not complain.

test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTest.cs
RassK: thanks, adjusting language https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/a8802db3cc8c03ca6eb9f99f2dc68ed070355c11

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
RassK: seems `SocketsHttpHandler` doesn't have that validator.

test/OpenTelemetry.OpAmp.Client.Tests/Tools/FrameGenerator.cs
RassK: I'll add just in case `__`, it's more static. Currently nothing is looking for it, since `Assert.StartsWith` does all.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
RassK: May need to tune later then. Prob need to get full session up and have some stats based on different message sizes.
My theory is that the most common message size should be default. 4kB for small messages, maybe increase a new block size to 8kB if there are already 2-3 4kB blocks requested 🤔

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
RassK: Note to myself: I need to check the minimal size of the message. If header + message will be below 10B then there is a bug.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
martincostello: Ah the first argument is different (`HttpMessageHandler` vs. `object`).

Any reason you need to specifically use `SocketsHttpHandler` rather than `HttpClientHandler` - IIRC, the latter just uses the former internally anyway.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
RassK: refactored https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/75e706a6685a9b5ada755c89dd5beed0002c003f

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
RassK: I was looking into `bool Try...(, out var ...)` pattern while refactoring. I was thinking if this part goes wrong and there is any exception then the whole send cycle should fail completely. If I just rise the exception in a level higher, does it save any performance?

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
RassK: Seems correct and works with `HttpClientHandler` also https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/59d9314d7e807f5df6d944a024c599e18fbc4f9c

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
martincostello: I think its more performant in the success case due to the lack of `catch` blocks.

If there's an issue it would instead return false, and then the caller would react to that, so it would still stop the cycle. The top of the call chain that cares if the entire operation was successful or not can do the throwing.

test/OpenTelemetry.OpAmp.Client.Tests/Tools/OpAmpFakeWebSocketServer.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/7bac1e79983b538a2a4b8aa3082e8fe218a3b952

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
RassK: I think it could be further optimized than always computing varint. Since the header is stable and fixed now, just compute it once and then block copy always to the destination.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/24248094df0eed960433b9e45fe5e60adfb8a289

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
RassK: seems true https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/24248094df0eed960433b9e45fe5e60adfb8a289

src/OpenTelemetry.OpAmp.Client/Internal/Utils/OpAmpWsHeaderHelper.cs
RassK: Was looking into other issue while this refactor worked here https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/de41721bd1476c573b9e8dbeececb60dcdefd2d4

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
RassK: Refactored https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/de41721bd1476c573b9e8dbeececb60dcdefd2d4. The original was invalid concept although it worked.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
RassK: Wouldn't `i=1` produce similar jump as if statement anyway? Don't see extra allocations either.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,134 @@
...
+    private CancellationToken token;
+
+    public WsReceiver(ClientWebSocket ws, FrameProcessor processor)
+    {
+        Guard.ThrowIfNull(ws, nameof(ws));
+        Guard.ThrowIfNull(processor, nameof(processor));
+
+        this.ws = ws;
+        this.processor = processor;
+        this.receiveThread = new Thread(this.ReceiveLoop);
```
martincostello: Would be useful to give the thread a name for viewing when debugging.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            ArrayPool<byte>.Shared.Return(rental);
+        }
+    }
+
+    private async void ReceiveLoop()
+    {
+        while (true)
+        {
+            this.token.ThrowIfCancellationRequested();
```
martincostello: Just noticed that this is running on a background thread - maybe it would be safer to do the following instead of having the thread throw:

```suggestion
        while (!this.token.IsCancellationRequested)
        {
```

Similarly, should this method have a `catch` in it in case `ReceiveAsync()` throws?

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -0,0 +1,134 @@
...
+            if (workingBuffer.Length - workingCount == 0)
+            {
+                if (rentalBuffers == null)
+                {
+                    rentalBuffers = [this.receiveBuffer];
+                }
+
+                workingBuffer = ArrayPool<byte>.Shared.Rent(RentalBufferSize);
+                workingCount = 0;
+                rentalBuffers.Add(workingBuffer);
```
martincostello, line 96: Should we put a limit/ceiling on how big this can grow?

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
martincostello: Wouldn't the allocation be the `BufferSegment` class wrapping the single buffer?

Couldn't that be avoided by doing this?

```csharp
if (buffers.Count == 1)
{
    return new ReadOnlySequence<byte>(buffers[0]);
}
```

src/OpenTelemetry.OpAmp.Client/Internal/Utils/Varint64.cs
```diff
@@ -0,0 +1,65 @@
...
+        while (value > DataMask)
+        {
+            bytes.Add((byte)((value & DataMask) | ContinuationBit));
+            value >>= BitsPerByte;
+        }
+
+        bytes.Add((byte)value); // Last byte without continuation bit
+
+        return bytes.ToArray();
+    }
```
martincostello: Can the `List<T>` allocation be avoided by just allocating an array of size `MaxEncodedBytes` then slicing it at the return by the number of bytes writter?

test/OpenTelemetry.OpAmp.Client.Tests/Tools/OpAmpFakeWebSocketServer.cs
```diff
@@ -0,0 +1,109 @@
...
+    private readonly BlockingCollection<AgentToServer> frames = [];
+    private readonly CancellationTokenSource cts = new();
+
+    public OpAmpFakeWebSocketServer(bool useSmallPackets)
+    {
+        this.httpServer = TestWebSocketServer.RunServer(
+            async socket =>
+            {
+                var buffer = new byte[8 * 1024];
+                var ms = new MemoryStream();
```
martincostello:
```suggestion
                using var ms = new MemoryStream();
```

test/OpenTelemetry.OpAmp.Client.Tests/Tools/OpAmpFakeWebSocketServer.cs
```diff
@@ -0,0 +1,109 @@
...
+                            var response = GenerateResponse(frame, useSmallPackets);
+                            await socket.SendAsync(response, WebSocketMessageType.Binary, true, this.cts.Token).ConfigureAwait(false);
+                        }
+                    }
+                }
+                finally
+                {
+                    ms.Dispose();
+
+                    if (socket.State == WebSocketState.Open || socket.State == WebSocketState.CloseReceived)
```
martincostello:
```suggestion
                {
                    if (socket.State == WebSocketState.Open || socket.State == WebSocketState.CloseReceived)
```

test/OpenTelemetry.OpAmp.Client.Tests/WsTransportTest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        using var mockListener = new MockListener();
+        var frameProcessor = new FrameProcessor();
+        frameProcessor.Subscribe(mockListener);
+
+        using var cts = new CancellationTokenSource();
+        using var wsTransport = new WsTransport(opAmpEndpoint, frameProcessor);
+        await wsTransport.StartAsync(cts.Token);
+
+        // Send only small packages, currently sending large package is not supported in WsTransport
```
martincostello:
```suggestion
        // Send only small packets, currently sending large package is not supported in WsTransport
```

src/OpenTelemetry.OpAmp.Client/Internal/Utils/SequenceHelper.cs
RassK: You are correct, forgot that ReadOnlySequence can accept byte[] also, not just segments. That saves a wrapper allocation for sure.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/4cc4f0e64d75167cc9507257a8c2159cf2b91987

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/45f06167be7b0c39bcff47144de83dabe4bc2ab3

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/9729ffd2518940de74162513c803eca3ad3183f2

src/OpenTelemetry.OpAmp.Client/Internal/Utils/Varint64.cs
RassK: Faster and less allocations 👍 
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/1e50ffce091acc781a28ad3f8c694f87ddbe6f13

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
RassK: Quite possibly true to prevent it eating all the memory. I'll add constant here for this step, later when configuration is re-worked, then it should move there with all of those defaults.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/75042e75b26d19dfc448216ffc3801d205177074

test/Shared/TestWebSocketServer.cs
RassK: `HttpListenerPrefixCollection` seems not to support port 0. Instead of looping random ports in `TestWebSocketServer` and `TestHttpServer`, seems you may start a quick tcp server with port 0 and then use the provided port 🤔

test/OpenTelemetry.OpAmp.Client.Tests/SequenceHelperTests.cs
RassK: `Assert.Equal` should use `Span<T>.SequenceEqual` if you referred to that?

martincostello: I meant more does the byte order in the messages change based on the endianness of the current machine. I'm guessing probably not, but this line made me wonder that.

RassK: Ok, I was thinking if it was truly endianness topic or not 😄
I'm not truly sure it will be problem free. Protobuf should be handling all it knows about, but custom data fields, possibly not.

src/OpenTelemetry.OpAmp.Client/Internal/Utils/Varint64.cs
```diff
@@ -0,0 +1,72 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Buffers;
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal static class Varint64
+{
+    private const byte DataMask = 0x7F;                         // 0111 1111 - lower 7 bits of a byte
+    private const byte ContinuationBit = 0x80;                  // 1000 0000 - signals more bytes follow
```
Kielek: Binary literals? https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types#integer-literals

test/OpenTelemetry.OpAmp.Client.Tests/Varint64Tests.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    [Theory]
+    [InlineData(0, 1, new byte[] { 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
+    [InlineData(1, 1, new byte[] { 0x01, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
+    [InlineData(127, 1, new byte[] { 0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
+    [InlineData(128, 2, new byte[] { 0x80, 0x01, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
+    [InlineData(987654321, 5, new byte[] { 0xB1, 0xD1, 0xF9, 0xD6, 0x03, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF })]
+    [InlineData(ulong.MaxValue, 10, new byte[] { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01 })]
+    public void OpAmpHeaderHelper_Decode(ulong value, int verifiedLength, byte[] buffer)
```
Kielek:
```suggestion
    public void OpAmpHeaderHelper_Decode(ulong expectedValue, int  expectedLength, byte[] buffer)
```

test/OpenTelemetry.OpAmp.Client.Tests/Varint64Tests.cs
```diff
@@ -0,0 +1,44 @@
...
+public class Varint64Tests
+{
+    [Theory]
+    [InlineData(0, new byte[] { 0x00 })]
+    [InlineData(1, new byte[] { 0x01 })]
+    [InlineData(127, new byte[] { 0x7F })]
+    [InlineData(128, new byte[] { 0x80, 0x01 })]
+    [InlineData(987654321, new byte[] { 0xB1, 0xD1, 0xF9, 0xD6, 0x03 })]
+    [InlineData(ulong.MaxValue, new byte[] { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x01 })]
+    public void OpAmpHeaderHelper_Encode(ulong value, byte[] verifiedEncoded)
```
Kielek:
```suggestion
    public void OpAmpHeaderHelper_Encode(ulong value, byte[] expectedEncoded)
```

Kielek (commented):
Some NITs

test/OpenTelemetry.OpAmp.Client.Tests/Varint64Tests.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/73491d0cacb53781d545bc8bf42bdfc582097f30

test/OpenTelemetry.OpAmp.Client.Tests/Varint64Tests.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/73491d0cacb53781d545bc8bf42bdfc582097f30

src/OpenTelemetry.OpAmp.Client/Internal/Utils/Varint64.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3064/commits/eadbb0946c9beecc0574c7af6fc25f9c3b2868f7

src/OpenTelemetry.OpAmp.Client/Internal/Utils/Varint64.cs
```diff
@@ -0,0 +1,72 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Buffers;
+
+namespace OpenTelemetry.OpAmp.Client.Internal.Utils;
+
+internal static class Varint64
+{
+    private const byte DataMask = 0b_0111_1111;                 // 0111 1111 - lower 7 bits of a byte
+    private const byte ContinuationBit = 0b_1000_0000;          // 1000 0000 - signals more bytes follow
```
martincostello:
```suggestion
    private const byte DataMask = 0b_0111_1111;                 // Lower 7 bits of a byte
    private const byte ContinuationBit = 0b_1000_0000;          // Signals more bytes follow
```

Issue #2909, opened 2025-07-14
[Instrumentation.AspNet] Stable release


Kielek:
### Component

OpenTelemetry.Instrumentation.AspNet

### Is your feature request related to a problem?

Issue to track the effort to stabilize OpenTelemetry.Instrumentation.AspNet and OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule.

### What is the expected behavior?

.

### Which alternative solutions or features have you considered?

.

### Additional context

.

Kielek:
Metrics status, based on: https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md
Implementation needs to be adjusted. It should be covered by: #2916 

This package is supporting only
[http.server.request.duration](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md#metric-httpserverrequestduration).
Following metrics are not supported. Stable release should be not blocked by this. Introducing them lates should not be treated as a breaking change.
* [http.server.active_requests](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md#metric-httpserveractive_requests)
* [http.server.request.body.size](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md#metric-httpserverrequestbodysize)
* [http.server.response.body.size](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md#metric-httpserverresponsebodysize).


[http.server.request.duration](https://github.com/open-telemetry/semantic-conventions/blob/v1.36.0/docs/http/http-metrics.md#metric-httpserverrequestduration) status:
* [x] name
* [x] type
* [x] unit
* [x] description
* [ ] attributes
    * [x] http.request.method
    * [x] url.scheme
    * [x] error.type - will be handled by #2921
    * [x] http.response.status_code
    * [x] http.route - added, when available
    * [ ] network.protocol.name - not supported, not needed if http is used
    * [ ] network.protocol.version - not supported, it is fine
    * [ ] server.address - opt-in, hidden by internal property, officially not supported
    * [ ] server.port - opt-in, hidden by internal property, officially not supported
    * [ ] user_agent.synthetic.type - development state, not supported

Kielek:
[...]

PR #3071, opened 2025-09-05
[Instrumentation.AspNet.TelemetryHttpModule] Rename activity source + trim public API

Kielek:
Towards #2909 

## Changes

Rename Activity source name to the full library name
Remove constants from the public API.


martincostello (approved):

src/OpenTelemetry.Instrumentation.AspNet/AspNetInstrumentationTracerProviderBuilderExtensions.cs
```diff
@@ -52,7 +52,7 @@ public static TracerProviderBuilder AddAspNetInstrumentation(
                 {
                     return AspNetInstrumentation.Instance.HandleManager.AddTracingHandle();
                 });
-                tracerProviderBuilder.AddSource(TelemetryHttpModule.AspNetSourceName);
+                tracerProviderBuilder.AddSource("OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule");
```
alanwest: Bit of a nit here...

I know this is _technically_ the correct name based on our prior patterns since the `OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule` is the actual package containing the code that produces the span, but it also kind of feels like it exposes an implementation/code organization detail.

I don't feel strongly about this but `OpenTelemetry.Instrumentation.AspNet` "feels" better since it is the package the end user will be more familiar with because it's likely the one they'll take a dependency on.

Approaching this from a different angle, I guess maybe I should be asking someone to remind me why we have this instrumentation separated into two packages? Is it possible for us to merge the two packages?

Kielek: In general, there were `Microsoft.AspNet.TelemetryCorrelation` package donated to OTel and it was discussed under the SIG. There were no requests to merge them.

Technically it should be possible to merge them, but I am not convinced that we should do it.
Telemetry module now has clear boundaries, and I do not see any better option to start Activities without this module.

@rajkumar-rangaraj, any thoughts?

-------

Archeological research:
 - instrumentation introduced in 
https://github.com/open-telemetry/opentelemetry-dotnet/pull/543/files#diff-018a882bc47a0ab03a46f7b33ab9508c34160f52885e2061ed9f037a04f84eb6. It based on external Microsoft.AspNet.TelemetryCorrelation package
 - telemetry module donated to OTel in scope of https://github.com/open-telemetry/opentelemetry-dotnet/issues/2222
 - Renamed to current name in https://github.com/open-telemetry/opentelemetry-dotnet/pull/2238

But most important are notes from the SIG 2021-08-03
 [reyang] Microsoft is willing to donate https://github.com/aspnet/Microsoft.AspNet.TelemetryCorrelation to OpenTelemetry .NET project.

* The project is under Apache 2 license, well aligned with OpenTelemetry.
* It looks like a win-win - OpenTelemetry .NETinstrumentation libs  could move faster, the component will be more vendor/backend agnostic.
* Reiley already talked to the ASP.NET folks and got their okay.
* Already gave heads up during the maintainers meeting, and got good-to-go
* https://docs.google.com/document/d/1-tGqaT9fn9Bpuy4roLS1YDfLzIWHodUFfyrBlyJVz9g/edit#
* Questions:
  * Should this be in the OpenTelemetry .NET repo, or OpenTelemetry .NET contrib repo? (proposal from Reiley: OpenTelemetry .NET, since it is used by ASP.NET instrumentation library which is in the OpenTelemetry .NET repo)
    * If later we move the instrumentation libraries to the contrib repo, we will move it together
  * Should this be part of the instrumentation library, or as a separate library which has no dependency on OpenTelemetry .NET API/SDK? (proposal from Reiley: a separate package, and the instrumentation library can take dependency on it)
 * How should we migrate the code?
  *  Proposal from Reiley:
     * Fork a “aspnet-telemetrycorrelation” branch from main
     * Each step would be a separate PR
     * Just copy the code, with no change
     * Make minimum change to clean up Microsoft specific stuff (comments, types, etc.)
     * Fix the styling/sanity check
     * Fix the project/solution/build
     * Integrate it with the instrumentation library, major change could happen at this stage
     * Update test cases
     * Merge the “aspnet-telemetrycorrelation” branch back to main, keep individual commit history rather than squash the commits
     * Delete the “aspnet-telemetrycorrelation” branch
     * Do a test release (most likely we'll just catch the 1.2.0 alpha train)

rajkumar-rangaraj: The separation of these modules clearly defines their boundaries. I support maintaining it as-is.

Regarding the naming discussion, I support using `OpenTelemetry.Instrumentation.AspNet` as it provides a cleaner, more intuitive experience for end users. While `OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule` may be technically more precise in reflecting the internal implementation, it exposes unnecessary implementation details that users don't need to be concerned with.

From a developer experience perspective, `OpenTelemetry.Instrumentation.AspNet` is:

- Easier to remember and reference
- More aligned with what users actually interact with (the main instrumentation package)
- Consistent with the principle of hiding implementation complexity behind simpler interfaces

Kielek: SourceName adjusted

Kielek:
@alanwest, @rajkumar-rangaraj, I am merging this PR as ypur comments were adressed. If you find anything else, I will be happy to fix in follow ups.

PR #3075, opened 2025-09-05
[EFCore] Remove non-defined DB system names

### LINKED ISSUE #3035: [EFCore] non-semantic convention db-providers
There is some custom set of non-semantic convention compliant db-names.
We should either remove them or update semantic convention.

Ref: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/e6c4e9dc420cb7d89a9c0244ff4a4eb8533ec431/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs#L310-L326

martincostello:
Fixes #3035

## Changes

Remove any `db.system.name` values that are not defined in the Semantic Conventions.


src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -286,7 +286,7 @@ public override void OnEventWritten(string name, object? payload)
     /// A tuple containing the respective <c>db.system</c> and <c>db.system.name</c> values.
     /// </returns>
     internal static (string Old, string New) GetDbSystemNames(string? providerOrCommandName) =>
-        //// "${Attribute} has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used."
```
alanwest, line 281: Nothing that I think needs to be addressed in this PR but just wondering...

The spec does still allow for a custom value to be used when no well-known value applies. Do you see this as being applicable to the EFCore instrumentation? Would you envision any need for an end user to explicitly supply a custom value for `db.system.name`?

Kielek: There is a Enrich method, it allow users do basically anything with tags. I do not think that we need to add specific property/env. var. to set this value in any other way.

PR #3110, opened 2025-09-11
[AspNet] Use Http*Base classes in public API

### LINKED ISSUE #1438: Rework HTTP telemetry modules to enable easier testing

* `OpenTelemetry.Instrumentation.AspNet 1.6.0-beta.2`
* `OpenTelemetry.Instrumentation.AspNet.TelemetryModule 1.6.0-beta.2`

* `net48`

* [x] Feature Request
* [ ] Bug

What do you expect to see?

Classes such as `TelemetryHttpModule` and `ActivityHelper` should use `HttpContextBase` and `HttpRequestBase` in order to be testable.

All http telemetry modules make use of `HttpContext` and `HttpRequest` which makes testing harder. As it stands, I'm unable to mock `HttpRequest.ServerVariables` because it's a readonly field with a backing field of an internal type.

## Additional Context

The issue recently sprang up in #1429 where I was unable to test `HttpRequest.ServerVariables` because it is read only. Any attempt of mocking it also failed because the classes are sealed. The specific case was to be able to unit test for `HttpRequest.ServerVariables["SERVER_PROTOCOL"]`.

Kielek:
Fixes #1438

## Changes

All usages of `HttpRequest`, `HttpResponse` and `HttpContext` replaced by `HttpRequestBase`, `HttpResponseBase` and `HttpContextBase` respectively.

Added simple test to utilize new (internal) API.

The only drawback I see is `src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/TelemetryHttpModule.cs` where there is couple of calls of (`new HttpContextWrapper(realContextHere)`.

It is the last moment where we can consider such breaking change without rising major version.


Kielek:
@qhris, it will be great if you can check this PR. You have requested these changes long time ago.

qhris:
I'm not in a situation to check this out in much detail at the moment, but it looks good to me.

From what I can remember, setting `ServerVariables` was especially hard without this PR.

Last time I looked into this, I found it very hard to actually test functionality without having to spin up a local server each time which led to very slow development and testing. But it has been almost 2 years since I made the issue and my memory is a bit fuzzy at this point. :smile: @alanwest

Kielek:
The main drawback I see is `src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/TelemetryHttpModule.cs` where there is couple of calls of (`new HttpContextWrapper(realContextHere)`.

Based on additional research, the `HttpContext` classes was introduced into the framework in 1.0 version. There were no possibility to add Based classes to the inheritance without introducing breaking changes. Ref: thttps://stackoverflow.com/a/1973727

As Base classes are _modern_, I would go with this implementation.

PR #3116, opened 2025-09-15
[SqlClient/EFCore] summary optimization phase 2

stevejgordon:
Contributes to #2238

## Changes

This is a rather large refactor of the `SqlProcessor`. The initial work focused on removing all allocation overhead. Further work optimised the non-cached execution time with `Span`-based code. Several iterations have led to the code in this PR which is a balance between readability and performance. In all cases, the execution time is reduced compared to the existing code.

Further work has been introduced to support additional SQL keywords and less common statements. This is not exhaustive but builds upon those previously supported.

@alanwest I'll add some PR comments around some specific items that may need discussion. I've included `SqlProcessorAdditionalTestCases.json` with more statements. We could eventually contribute these into the semantic conventions. These will need a careful review to check we are happy the expected summaries align with the intent from the semantic conventions.

### Before
```
| Method          | Sql                  | Mean     | Error    | StdDev    | Median   | Allocated |
|---------------- |--------------------- |---------:|---------:|----------:|---------:|----------:|
| GetSanitizedSql | CREAT(...)s(Id) [56] | 420.6 ns | 22.18 ns |  60.71 ns | 428.3 ns |     584 B |
| GetSanitizedSql | DELET(...) = 42 [32] | 255.7 ns |  7.11 ns |  20.18 ns | 252.0 ns |     448 B |
| GetSanitizedSql | INSER(...)3e-5) [76] | 514.4 ns | 16.39 ns |  48.33 ns | 513.3 ns |     680 B |
| GetSanitizedSql | SELEC(...)ls od [39] | 439.8 ns | 39.56 ns | 116.65 ns | 374.5 ns |     528 B |
| GetSanitizedSql | SELE(...)tory [111]  | 584.5 ns | 11.55 ns |  24.36 ns | 583.5 ns |    1056 B |
| GetSanitizedSql | SELEC(...)table [69] | 269.9 ns |  8.34 ns |  23.79 ns | 271.8 ns |     600 B |
| GetSanitizedSql | SELEC(...) c.Id [74] | 563.1 ns | 22.71 ns |  66.25 ns | 553.4 ns |     736 B |
| GetSanitizedSql | SELE(...)_id) [101]  | 691.8 ns | 19.53 ns |  56.02 ns | 693.5 ns |     912 B |
| GetSanitizedSql | UPDAT(...) = 42 [44] | 328.1 ns | 11.58 ns |  33.40 ns | 329.8 ns |     504 B |
```

### After
```
| Method          | Sql                  | Mean     | Error   | StdDev  | Median   | Allocated |
|---------------- |--------------------- |---------:|--------:|--------:|---------:|----------:|
| GetSanitizedSql | CREAT(...)s(Id) [56] | 207.9 ns | 4.12 ns | 7.54 ns | 205.0 ns |     248 B |
| GetSanitizedSql | DELET(...) = 42 [32] | 161.2 ns | 1.90 ns | 1.59 ns | 161.2 ns |     128 B |
| GetSanitizedSql | INSER(...)3e-5) [76] | 254.1 ns | 2.37 ns | 2.22 ns | 253.8 ns |     192 B |
| GetSanitizedSql | SELEC(...)ls od [39] | 146.3 ns | 1.98 ns | 1.65 ns | 145.8 ns |     184 B |
| GetSanitizedSql | SELE(...)tory [111]  | 267.8 ns | 1.41 ns | 1.32 ns | 268.0 ns |     424 B |
| GetSanitizedSql | SELEC(...)table [69] | 106.5 ns | 2.13 ns | 3.45 ns | 104.7 ns |     128 B |
| GetSanitizedSql | SELEC(...) c.Id [74] | 246.1 ns | 1.65 ns | 1.29 ns | 246.3 ns |     264 B |
| GetSanitizedSql | SELE(...)_id) [101]  | 294.1 ns | 2.04 ns | 1.81 ns | 294.1 ns |     312 B |
| GetSanitizedSql | UPDAT(...) = 42 [44] | 196.9 ns | 0.97 ns | 0.86 ns | 197.2 ns |     144 B |
```

### DIFF
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -689,9 +689,17 @@ internal static class SqlProcessor
         // In this case, we fast-path to the closing parenthesis and replace the entire contents with a single '?'.
 
         if (state.SanitizedPosition > 0 && buffer[state.SanitizedPosition - 1] == OpenParenChar
-            && state.PreviousTokenEndPosition - state.PreviousTokenStartPosition == 2
-            && sql.Slice(state.PreviousTokenStartPosition, 2).Equals("in", StringComparison.OrdinalIgnoreCase))
+            && state.PreviousTokenEndPosition - state.PreviousTokenStartPosition == 2)
         {
+            // Check the token is actually "IN" (case-insensitive) to avoid false positives.
+            var firstChar = sql[state.PreviousTokenStartPosition];
+            var secondChar = sql[state.PreviousTokenStartPosition + 1];
+
+            if (!((firstChar == 'i' || firstChar == 'I') && (secondChar == 'n' || secondChar == 'N')))
+            {
+                return false;
+            }
+
             var closingParenIndex = sql.Slice(parsePosition).IndexOf(CloseParenChar);
 
             if (closingParenIndex > 0)
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -12,10 +12,25 @@ internal static class SqlProcessor
     private const int MaxSummaryLength = 255;
     private const int CacheCapacity = 1000;
 
+    private const char SanitizationPlaceholder = '?';
+    private const char SpaceChar = ' ';
+    private const char CommaChar = ',';
+    private const char OpenParenChar = '(';
+    private const char CloseParenChar = ')';
+    private const char DashChar = '-';
+    private const char ForwardSlashChar = '/';
+    private const char SingleQuoteChar = '\'';
+    private const char AsteriskChar = '*';
+    private const char UnderscoreChar = '_';
+    private const char DotChar = '.';
+    private const char NewLineChar = '\n';
+    private const char CarriageReturnChar = '\r';
+    private const char TabChar = '\t';
+
     private static readonly ConcurrentDictionary<string, SqlStatementInfo> Cache = new();
 
 #if NET
-    private static readonly SearchValues<char> WhitespaceSearchValues = SearchValues.Create([' ', '\t', '\r', '\n']);
+    private static readonly SearchValues<char> WhitespaceSearchValues = SearchValues.Create([SpaceChar, TabChar, CarriageReturnChar, NewLineChar]);
 #endif
 
     // We can extend this in the future to include more keywords if needed.
@@ -129,9 +144,9 @@ internal static class SqlProcessor
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static bool IsAsciiIdentifierChar(char c) =>
 #if NET
-        char.IsAsciiLetter(c) || char.IsAsciiDigit(c) || c == '_' || c == '.';
+        char.IsAsciiLetter(c) || char.IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
 #else
-        IsAsciiLetter(c) || IsAsciiDigit(c) || c == '_' || c == '.';
+        IsAsciiLetter(c) || IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
 #endif
 
     private static SqlStatementInfo SanitizeSql(ReadOnlySpan<char> sql)
@@ -180,7 +195,7 @@ internal static class SqlProcessor
         var summaryLength = Math.Min(state.SummaryPosition, MaxSummaryLength);
 
         // Trim trailing space
-        if (summaryLength > 0 && state.SummaryBuffer[summaryLength - 1] == ' ')
+        if (summaryLength > 0 && state.SummaryBuffer[summaryLength - 1] == SpaceChar)
         {
             summaryLength -= 1;
         }
@@ -205,7 +220,7 @@ internal static class SqlProcessor
 #if NET
         var indexOfNextWhitespace = sql.Slice(start).IndexOfAny(WhitespaceSearchValues);
 #else
-        var indexOfNextWhitespace = sql.Slice(start).IndexOfAny([' ', '\t', '\r', '\n']);
+        var indexOfNextWhitespace = sql.Slice(start).IndexOfAny([SpaceChar, TabChar, CarriageReturnChar, NewLineChar]);
 #endif
 
         var length = indexOfNextWhitespace >= 0 ? indexOfNextWhitespace : sql.Length - start;
@@ -261,7 +276,7 @@ internal static class SqlProcessor
             // Check if the previous character is '(', in which case, we only check against the SELECT keyword.
             // Otherwise, check if the previous keyword may be the start of a keyword chain so we can limit the
             // number of keyword comparisons we need to do by only comparing for tokens we expect to appear next.
-            if (state.ParsePosition > 0 && sql[state.ParsePosition - 1] == '(')
+            if (state.ParsePosition > 0 && sql[state.ParsePosition - 1] == OpenParenChar)
             {
                 keywordsToCheck = SelectOnlyKeywordArray;
             }
@@ -335,7 +350,7 @@ internal static class SqlProcessor
 
         // If we get this far, we have not matched a keyword, so we copy the token as-is.
 
-        if (char.IsLetter(currentChar) || currentChar == '_')
+        if (char.IsLetter(currentChar) || currentChar == UnderscoreChar)
         {
             // This first block handles identifiers (which start with a letter or underscore).
 
@@ -367,7 +382,7 @@ internal static class SqlProcessor
                     state.SummaryPosition += length;
 
                     // Add a space after the identifier. The trailing space will be trimmed later.
-                    state.SummaryBuffer[state.SummaryPosition++] = ' ';
+                    state.SummaryBuffer[state.SummaryPosition++] = SpaceChar;
                 }
             }
 
@@ -379,7 +394,7 @@ internal static class SqlProcessor
             // Special case: if the token is a comma and follows a FROM keyword,
             // we are in an old style implicit join and we want to capture the next valid identifier in the summary.
             var prevKeyword = state.PreviousParsedKeyword?.SqlKeyword ?? SqlKeyword.Unknown;
-            state.CaptureNextTokenInSummary = prevKeyword == SqlKeyword.From && currentChar == ',';
+            state.CaptureNextTokenInSummary = prevKeyword == SqlKeyword.From && currentChar == CommaChar;
 
             buffer[state.SanitizedPosition++] = currentChar;
             state.ParsePosition++;
@@ -397,7 +412,7 @@ internal static class SqlProcessor
 #if NET
             if (WhitespaceSearchValues.Contains(currentChar))
 #else
-            if (currentChar == ' ' || currentChar == '\t' || currentChar == '\r' || currentChar == '\n')
+            if (currentChar == SpaceChar || currentChar == TabChar || currentChar == CarriageReturnChar || currentChar == NewLineChar)
 #endif
             {
                 foundWhitespace = true;
@@ -422,13 +437,12 @@ internal static class SqlProcessor
         var iPlusTwo = i + 2;
 
         // Scan past multi-line comment
-        if (ch == '/' && iPlusOne < length && sql[iPlusOne] == '*')
+        if (ch == '/' && iPlusOne < length && sql[iPlusOne] == AsteriskChar)
         {
-#if NET
             var rest = sql.Slice(iPlusTwo);
             while (!rest.IsEmpty)
             {
-                var starIdx = rest.IndexOf('*');
+                var starIdx = rest.IndexOf(AsteriskChar);
                 if (starIdx < 0)
                 {
                     // Unterminated comment, consume to end
@@ -437,7 +451,7 @@ internal static class SqlProcessor
                 }
 
                 // Check for closing */
-                if (starIdx + 1 < rest.Length && rest[starIdx + 1] == '/')
+                if (starIdx + 1 < rest.Length && rest[starIdx + 1] == ForwardSlashChar)
                 {
                     state.ParsePosition = iPlusTwo + starIdx + 2; // position after */
                     return true;
@@ -449,29 +463,14 @@ internal static class SqlProcessor
 
             state.ParsePosition = length;
             return true;
-#else
-            for (i += 2; i < length; ++i)
-            {
-                ch = sql[i];
-                if (ch == '*' && iPlusOne < length && sql[iPlusOne] == '/')
-                {
-                    i += 1;
-                    break;
-                }
-            }
-
-            state.ParsePosition = ++i;
-            return true;
-#endif
         }
 
         // Scan past single-line comment
-        if (ch == '-' && iPlusOne < length && sql[iPlusOne] == '-')
+        if (ch == DashChar && iPlusOne < length && sql[iPlusOne] == DashChar)
         {
-#if NET
             // Find next line break efficiently and preserve the newline for whitespace handling
             var rest = sql.Slice(iPlusTwo);
-            var idx = rest.IndexOfAny('\r', '\n');
+            var idx = rest.IndexOfAny(CarriageReturnChar, NewLineChar);
             if (idx >= 0)
             {
                 // Position at the newline so ParseWhitespace can copy it
@@ -483,20 +482,6 @@ internal static class SqlProcessor
             }
 
             return true;
-#else
-            for (i += 2; i < length; ++i)
-            {
-                ch = sql[i];
-                if (ch == '\r' || ch == '\n')
-                {
-                    i -= 1;
-                    break;
-                }
-            }
-
-            state.ParsePosition = ++i;
-            return true;
-#endif
         }
 
         return false;
@@ -505,13 +490,12 @@ internal static class SqlProcessor
     private static bool SanitizeStringLiteral(ReadOnlySpan<char> sql, Span<char> buffer, ref ParseState state)
     {
         var currentChar = sql[state.ParsePosition];
-        if (currentChar == '\'')
+        if (currentChar == SingleQuoteChar)
         {
-#if NET
             var rest = sql.Slice(state.ParsePosition + 1);
             while (!rest.IsEmpty)
             {
-                var idx = rest.IndexOf('\'');
+                var idx = rest.IndexOf(SingleQuoteChar);
                 var idxPlusOne = idx + 1;
                 if (idx < 0)
                 {
@@ -519,7 +503,7 @@ internal static class SqlProcessor
                     return true;
                 }
 
-                if (idxPlusOne < rest.Length && rest[idxPlusOne] == '\'')
+                if (idxPlusOne < rest.Length && rest[idxPlusOne] == SingleQuoteChar)
                 {
                     // Skip escaped quote ('')
                     rest = rest.Slice(idx + 2);
@@ -529,36 +513,12 @@ internal static class SqlProcessor
                 // Found terminating quote
                 state.ParsePosition = sql.Length - rest.Length + idxPlusOne;
 
-                buffer[state.SanitizedPosition++] = '?';
+                buffer[state.SanitizedPosition++] = SanitizationPlaceholder;
                 return true;
             }
 
-            buffer[state.SanitizedPosition++] = '?';
+            buffer[state.SanitizedPosition++] = SanitizationPlaceholder;
             return true;
-#else
-            var i = state.ParsePosition + 1;
-            var length = sql.Length;
-            var iPlusOne = i + 1;
-            for (; i < length; ++i)
-            {
-                currentChar = sql[i];
-                if (currentChar == '\'' && iPlusOne < length && sql[iPlusOne] == '\'')
-                {
-                    ++i;
-                    continue;
-                }
-
-                if (currentChar == '\'')
-                {
-                    break;
-                }
-            }
-
-            state.ParsePosition = ++i;
-
-            buffer[state.SanitizedPosition++] = '?';
-            return true;
-#endif
         }
 
         return false;
@@ -600,7 +560,7 @@ internal static class SqlProcessor
 
             state.ParsePosition = ++i;
 
-            buffer[state.SanitizedPosition++] = '?';
+            buffer[state.SanitizedPosition++] = SanitizationPlaceholder;
             return true;
         }
 
@@ -615,7 +575,7 @@ internal static class SqlProcessor
         var iPlusOne = i + 1;
 
         // If the digit follows an open bracket, check for a parenthesized digit sequence
-        if (i > 0 && sql[i - 1] == '('
+        if (i > 0 && sql[i - 1] == OpenParenChar
             && IsAsciiDigit(currentChar))
         {
             int start = i;
@@ -627,7 +587,7 @@ internal static class SqlProcessor
                 j++;
             }
 
-            if (j < length && sql[j] == ')')
+            if (j < length && sql[j] == CloseParenChar)
             {
                 // Copy the digits and the closing bracket to the buffer
                 sql.Slice(start, j - start + 1).CopyTo(buffer.Slice(state.SanitizedPosition));
@@ -640,7 +600,7 @@ internal static class SqlProcessor
         }
 
         // Scan past leading sign
-        if ((currentChar == '-' || currentChar == '+') && iPlusOne < length && (IsAsciiDigit(sql[iPlusOne]) || sql[iPlusOne] == '.'))
+        if ((currentChar == '-' || currentChar == '+') && iPlusOne < length && (IsAsciiDigit(sql[iPlusOne]) || sql[iPlusOne] == DotChar))
         {
             i += 1;
             iPlusOne = i + 1;
@@ -691,7 +651,7 @@ internal static class SqlProcessor
 
             state.ParsePosition = ++i;
 
-            buffer[state.SanitizedPosition++] = '?';
+            buffer[state.SanitizedPosition++] = SanitizationPlaceholder;
             return true;
         }
 
@@ -774,22 +734,24 @@ internal static class SqlProcessor
             ViewKeyword = new("view", SqlKeyword.View, DdlKeywords);
 
             // Phase 2: Build arrays that depend on instances
+            // NOTE: This array is sorted by an estimation of the most likely
+            // keywords first to optimise the comparison loop.
             DdlSubKeywords =
             [
                 TableKeyword,
                 IndexKeyword,
+                UniqueKeyword,
+                ClusteredKeyword,
+                NonClusteredKeyword,
                 ViewKeyword,
                 ProcedureKeyword,
                 TriggerKeyword,
                 DatabaseKeyword,
-                SchemaKeyword,
-                FunctionKeyword,
                 UserKeyword,
                 RoleKeyword,
                 SequenceKeyword,
-                UniqueKeyword,
-                ClusteredKeyword,
-                NonClusteredKeyword,
+                SchemaKeyword,
+                FunctionKeyword,
             ];
 
             // Phase 3: Wire follow relationships
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -765,10 +765,22 @@ internal static class SqlProcessor
             ViewKeyword = new("view", SqlKeyword.View, DdlKeywords);
 
             // Phase 2: Build arrays that depend on instances
-            DdlSubKeywords = [
-                TableKeyword, IndexKeyword, ViewKeyword, ProcedureKeyword, TriggerKeyword,
-                DatabaseKeyword, SchemaKeyword, FunctionKeyword, UserKeyword, RoleKeyword, SequenceKeyword, UniqueKeyword,
-                ClusteredKeyword, NonClusteredKeyword
+            DdlSubKeywords =
+            [
+                TableKeyword,
+                IndexKeyword,
+                ViewKeyword,
+                ProcedureKeyword,
+                TriggerKeyword,
+                DatabaseKeyword,
+                SchemaKeyword,
+                FunctionKeyword,
+                UserKeyword,
+                RoleKeyword,
+                SequenceKeyword,
+                UniqueKeyword,
+                ClusteredKeyword,
+                NonClusteredKeyword,
             ];
 
             // Phase 3: Wire follow relationships
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -16,6 +16,7 @@ internal static class SqlProcessor
     private const char SpaceChar = ' ';
     private const char CommaChar = ',';
     private const char OpenParenChar = '(';
+    private const char CloseParenChar = ')';
     private const char DashChar = '-';
     private const char ForwardSlashChar = '/';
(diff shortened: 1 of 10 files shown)


build/Common.nonprod.props
```diff
@@ -24,7 +24,7 @@
       Please sort alphabetically.
       Refer to https://docs.microsoft.com/nuget/concepts/package-versioning for semver syntax.
     -->
-    <BenchmarkDotNetPkgVer>[0.13.12,0.14)</BenchmarkDotNetPkgVer>
+    <BenchmarkDotNetPkgVer>0.15.2</BenchmarkDotNetPkgVer>
```
stevejgordon: @Kielek Not sure if this has side effects. I had issues running the benchmarks on my machine which were resolved by updating to the latest.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
```diff
@@ -21,12 +21,50 @@ public static class SqlProcessorTestCases
...
+                if (DbSystemTestCasesToExecute.Contains(testCase.Input.DbSystemName))
+                {
+                    data.Add(testCase);
+                }
+            }
+        }
+
+        // Reintroduce "alter_table" with a revised expected output.
+        // We don't expect the number '255' to be redacted when defining the size of a column.
+        data.Add(new()
```
stevejgordon, line 50: @alanwest This specific test in the semantic conventions expects the `255` literal to be sanitized. My current implementation handles column definitions such that we retain the value. If you're happy with this, should I update this in the semantic conventions?

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
```diff
@@ -21,12 +21,50 @@ public static class SqlProcessorTestCases
...
+            {
+                DbSystemName = "other_sql",
+                Query = "ALTER  TABLE MyTable ADD Name varchar(255)",
+            },
+            Expected = new()
+            {
+                SanitizedQueryText = ["ALTER  TABLE MyTable ADD Name varchar(255)"],
+                Summary = "ALTER TABLE MyTable",
+            },
+        });
```
stevejgordon, line 63: @alanwest This specific test in the semantic conventions expects the 255 literal to be sanitized. My current implementation handles column definitions such that we retain the value. If you're happy with this, should I update this in the semantic conventions?

build/Common.nonprod.props
martincostello: I think this is fine - it's only for tests, so there shouldn't be any need to constrain the range.

Once we move from dependabot to renovate we'll get updates for packages such as this going forward.

test/OpenTelemetry.Contrib.Shared.Tests/OpenTelemetry.Contrib.Shared.Tests.csproj
```diff
@@ -31,6 +31,9 @@
   </ItemGroup>
 
   <ItemGroup>
+    <EmbeddedResource Include="SqlProcessorAdditionalTestCases.json">
+      <LogicalName>SqlProcessorAdditionalTestCases.json</LogicalName>
+    </EmbeddedResource>
     <EmbeddedResource Include="SqlProcessorTestCases.json">
       <LogicalName>SqlProcessorTestCases.json</LogicalName>
     </EmbeddedResource>
```
martincostello: These could be simplified down to:

```suggestion
    <EmbeddedResource Include="*.json" LogicalName="%(FileName)%(Extension)" />
```

src/Shared/SqlProcessor.cs
```diff
@@ -1,14 +1,81 @@
...
+using System.Runtime.CompilerServices;
 
 namespace OpenTelemetry.Instrumentation;
 
 internal static class SqlProcessor
 {
     private const int CacheCapacity = 1000;
-    private static readonly Dictionary<string, SqlStatementInfo> Cache = [];
+
+    private static readonly ConcurrentDictionary<string, SqlStatementInfo> Cache = new();
```
martincostello, line 30: For the TFMs that support it, would it be worth looking into whether `AlternateLookup` can be used to lookup the cache entries as a span?

src/Shared/SqlProcessor.cs
```diff
@@ -17,66 +84,368 @@ public static SqlStatementInfo GetSanitizedSql(string? sql)
...
+            {
+                ParseNextTokenFast(sql, buffer, ref state);
+            }
+            else
+            {
+                ParseNextToken(sql, buffer, ref state);
+            }
+        }
+
+        var summaryLength = Math.Min(state.SummaryPosition, 255);
```
martincostello: Put the `255`s into a named constant?

src/Shared/SqlProcessor.cs
```diff
@@ -87,62 +456,111 @@ private static bool SkipComment(string sql, ref int index)
...
         if (ch == '-' && i + 1 < length && sql[i + 1] == '-')
         {
+#if NET
+            // Find next line break efficiently and preserve the newline for whitespace handling
+            var rest = sql.Slice(i + 2);
+            var idx = rest.IndexOfAny('\r', '\n');
+            if (idx >= 0)
+            {
+                // Position at the newline so ParseWhitespace can copy it
+                state.ParsePosition = i + 2 + idx;
```
martincostello: Is there any benefit to caching the `i+2` into a variable?

src/Shared/SqlProcessor.cs
```diff
@@ -87,62 +456,111 @@ private static bool SkipComment(string sql, ref int index)
...
+                state.ParsePosition = sql.Length;
+            }
+
+            return true;
+#else
             for (i += 2; i < length; ++i)
             {
                 ch = sql[i];
-                if (ch is '\r' or '\n')
+                if (ch == '\r' || ch == '\n')
```
martincostello: Is this faster than `is` and `or`?

src/Shared/SqlProcessor.cs
```diff
@@ -87,62 +456,111 @@ private static bool SkipComment(string sql, ref int index)
...
+
+                if (idx + 1 < rest.Length && rest[idx + 1] == '\'')
+                {
+                    // Skip escaped quote ('')
+                    rest = rest.Slice(idx + 2);
+                    continue;
+                }
+
+                // Found terminating quote
+                state.ParsePosition = sql.Length - rest.Length + idx + 1;
```
martincostello: Same question here regarding storing `i+1` into a variable.

src/Shared/SqlProcessor.cs
```diff
@@ -87,62 +456,111 @@ private static bool SkipComment(string sql, ref int index)
...
+
+                buffer[state.SanitizedPosition++] = '?';
```
martincostello: Shall we make `'?'` a named constant like `RedactedChar` or similar?

src/Shared/SqlProcessor.cs
```diff
@@ -161,58 +585,86 @@ private static bool SanitizeHexLiteral(string sql, ref int index)
...
         return false;
     }
 
-    private static bool SanitizeNumericLiteral(string sql, ref int index)
+    private static bool SanitizeNumericLiteral(ReadOnlySpan<char> sql, Span<char> buffer, ref ParseState state)
     {
-        var i = index;
-        var ch = sql[i];
+        var i = state.ParsePosition;
+        var nextChar = sql[i];
```
martincostello: `nextChar` kind of reads to me like not the one we're looking at now, but the one to come after. Then there's access to `i + 1` which feels to me like next, but we're already calling `i` next, so that would be `next next`.

Maybe this could be renamed to `thisChar`/`currentChar`?

src/Shared/SqlProcessor.cs
```diff
@@ -228,177 +680,241 @@ private static bool SanitizeNumericLiteral(string sql, ref int index)
...
-#else
-                var collection = sql.Substring(index, i - index);
-                state.DbQuerySummary.Append(' ').Append(collection);
-#endif
-            }
+        public SqlKeywordInfo? PreviousParsedKeyword; // 8 bytes (reference type)
 
-            i -= 1;
+        public SqlKeyword FirstSummaryKeyword; // 4 bytes (enum, underlying int)
+        public SqlKeyword PreviousSummaryKeyword; // 4 bytes (enum, underlying int)
```
martincostello, line 730: Is there any advantage to using a smaller type like `short` (or `byte`) for the enum given the number of possible values is relatively small?

src/Shared/SqlProcessor.cs
```diff
@@ -228,177 +680,241 @@ private static bool SanitizeNumericLiteral(string sql, ref int index)
...
-    private static void AppendNormalized(StringBuilder builder, char ch)
+    private sealed class SqlKeywordInfo
     {
-        if (!char.IsWhiteSpace(ch) || builder.Length == 0 || !char.IsWhiteSpace(builder[builder.Length - 1]))
+        // Used on keywords that are only included in the summary if they are the first keyword in the statement.
+        private static readonly SqlKeyword[] Unknown = [SqlKeyword.Unknown];
+
+        private static readonly SqlKeyword[] DdlKeywords = [
+            SqlKeyword.Create, SqlKeyword.Drop, SqlKeyword.Alter
+        ];
```
martincostello:
```suggestion
        private static readonly SqlKeyword[] DdlKeywords =
        [
            SqlKeyword.Create,
            SqlKeyword.Drop,
            SqlKeyword.Alter,
        ];
```

src/Shared/SqlProcessor.cs
```diff
@@ -228,177 +680,241 @@ private static bool SanitizeNumericLiteral(string sql, ref int index)
...
+            UpdateKeyword = new("update", SqlKeyword.Update, Unknown);
+            UserKeyword = new("user", SqlKeyword.User, DdlKeywords);
+            ViewKeyword = new("view", SqlKeyword.View, DdlKeywords);
+
+            // Phase 2: Build arrays that depend on instances
+            DdlSubKeywords = [
+                TableKeyword, IndexKeyword, ViewKeyword, ProcedureKeyword, TriggerKeyword,
+                DatabaseKeyword, SchemaKeyword, FunctionKeyword, UserKeyword, RoleKeyword, SequenceKeyword, UniqueKeyword,
+                ClusteredKeyword, NonClusteredKeyword
+            ];
```
martincostello: Do the order have significance? If not, can we sort them A-Z?

```suggestion
            DdlSubKeywords =
            [
                TableKeyword,
                IndexKeyword,
                ViewKeyword,
                ProcedureKeyword,
                TriggerKeyword,
                DatabaseKeyword,
                SchemaKeyword,
                FunctionKeyword,
                UserKeyword,
                RoleKeyword,
                SequenceKeyword,
                UniqueKeyword,
                ClusteredKeyword,
                NonClusteredKeyword,
            ];
```

src/Shared/SqlProcessor.cs
stevejgordon: I played with various variables of `AlternateLookup` and also `FrozenDictionary` for some of the other lookups. The main issue is it is .NET 9+ so we end up with more complexity to manage across TFMs. I didn't specifically try it here bt the input is a string for this particular dictionary so I don't think it's would make much difference.

src/Shared/SqlProcessor.cs
stevejgordon: Maybe a nanosecond or two I guess. Can do though. It won't hurt.

src/Shared/SqlProcessor.cs
stevejgordon: Will do

src/Shared/SqlProcessor.cs
stevejgordon: Yeah, can do.

src/Shared/SqlProcessor.cs
stevejgordon: I don't believe so, but I haven't tested. I think they should end up compiling to the same IL.

src/Shared/SqlProcessor.cs
stevejgordon: Yeah `currentChar` is clearer. I think I even had that in places at one point!

src/Shared/SqlProcessor.cs
stevejgordon: I can try it and test. As we're passinf by-ref I don't think it should make much difference, if any.

src/Shared/SqlProcessor.cs
martincostello: In that case could we leave it as it was? I'm ~sure there's an analyzer that suggests doing this, so it'll come back and re-suggest the change if that's the case.

src/Shared/SqlProcessor.cs
stevejgordon: Yeah, order matters. I've tried to sort by the most likely although only based on my best guess. Arguably unique, clustered, nonclustered could probably move under procedure. Doing stuff with users, roles etc is likely rare is most application, whereas working with indexes may be more common. All are likely relatively rare scenarios TBH.

src/Shared/SqlProcessor.cs
martincostello: Just suggested as the comments suggest field size is important.

stevejgordon:
A couple of the tests are failin gon net462 - I'll look into those tomorrow.

build/Common.nonprod.props
Kielek: I am fine with merging this in this PR/separate. It is up to you, if you want cleaner PR, not affected by package bumps.

stevejgordon: I'll include it in this one. Thanks for the input.

src/Shared/SqlProcessor.cs
stevejgordon: Tested and it made things slightly worse, although as these small measurements it's not exact. As we're by-ref, I think we're fine here.

src/Shared/SqlProcessor.cs
stevejgordon: Done

src/Shared/SqlProcessor.cs
stevejgordon: Done

src/Shared/SqlProcessor.cs
stevejgordon: Done

src/Shared/SqlProcessor.cs
stevejgordon: Done

stevejgordon:
@martincostello There seem to be a bunch of EFCore failures now. I'll take a look into those, hopefully tomorrow and see why this breaks those.

stevejgordon:
@martincostello The EF Core tests pass on my machine. Are you able to rerun these please?

stevejgordon:
@alanwest What are you thoughts on the revised test (see https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116#discussion_r2349205949) and additional test cases? I'm on leave next week, so will pick up any changes when I'm back.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
alanwest: The intent of the original implementation was to aggressively sanitize since I figured any number or string literal in the query added little value to `db.query.text`. That said, I think this is ok.

The reason `SanitizedQueryText` is an array is to allow alternate acceptable forms of sanitization. I don't feel strongly that we should keep the previous form as an option, but it's a question we can raise when we update the test suite upstream in case others have a preference.

We just want to be extra cautious not to expose sensitive data. One thing the test suite from the semantic conventions repo does not cover are pathological SQL statements. For example, SQL that is not valid. Something like `ALTER TABLE MyTable SELECT * FROM Users WHERE (something sensitive here)`. I don't think pathological test cases belong in the test suite from the semantic conventions, but I think we may want to add some coverage separately. I guess ideally it would be nice to not set `db.query.text` at all in case of invalid SQL, but I think it may be challenging to detect without implementing a full SQL parser which would be silly. Maybe we could consider unsetting `db.query.text` in the case of an exception indicating invalid SQL?

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -0,0 +1,600 @@
...
+  },
+  {
+    "name": "summary_truncated",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "SELECT * FROM Vecnzotjejucwitzgrfifscuevittljlnrlpbruvkezeptqciyvbjhsytmbucbhwttidayecnthaztxbyppbyztcqccedeirgkxzrfezjxfwbtuqxeusroqgbvulgmsvnelovkxqsaqmlogomkhtjuirzhaocxlrmerihnmwaelullionarkmxwdamhduwrbooknqsnilurutgyerxphokeqnnoumbpcfjtmqrbpukjllofiwaltyoawkp o, OrderDetails od"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT * FROM Vecnzotjejucwitzgrfifscuevittljlnrlpbruvkezeptqciyvbjhsytmbucbhwttidayecnthaztxbyppbyztcqccedeirgkxzrfezjxfwbtuqxeusroqgbvulgmsvnelovkxqsaqmlogomkhtjuirzhaocxlrmerihnmwaelullionarkmxwdamhduwrbooknqsnilurutgyerxphokeqnnoumbpcfjtmqrbpukjllofiwaltyoawkp o, OrderDetails od"
```
alanwest, line 23: The spec for [generating db.query.summary](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-spans.md#generating-a-summary-of-the-query) states that truncation should not occur within an operation or target.

> Instrumentations that parse the query to set db.query.summary SHOULD truncate the summary to 255 characters (ensuring truncation does not occur within an operation name or target).

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -0,0 +1,600 @@
...
+    "name": "create_unique_clustered_index",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "CREATE UNIQUE CLUSTERED INDEX IX_Employee_Email ON Employees (Email);"
+    },
+    "expected": {
+      "db.query.text": [
+        "CREATE UNIQUE CLUSTERED INDEX IX_Employee_Email ON Employees (Email);"
+      ],
+      "db.query.summary": "CREATE UNIQUE CLUSTERED INDEX IX_Employee_Email"
```
alanwest: There is not a strict specification for this, but I think I'd lean towards trying to exclude qualifiers like `UNIQUE` and `CLUSTERED` when parsing the operation.

Also this appears to be a test that should specifically target MS SQL Server, so `db.system.name` should equal `microsoft.sql_server`. This is so that the test suite can be filtered down to tests only applicable to certain database systems if necessary. The list of known `db.system.name` values are listed [here](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/db.md).

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -0,0 +1,600 @@
...
+    "name": "select_distinct",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "SELECT DISTINCT Country FROM Countries"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT DISTINCT Country FROM Countries"
+      ],
+      "db.query.summary": "SELECT DISTINCT Countries"
```
alanwest: Same comment here regarding optional qualifiers like `DISTINCT`. I'm open to discussion, but I'm inclined to exclude them.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -0,0 +1,600 @@
...
+  },
+  {
+    "name": "exec_sp_rename_table",
+    "input": {
+      "db.system.name": "mssql",
+      "query": "EXEC sp_rename 'OldTable', 'NewTable';"
+    },
+    "expected": {
+      "db.query.text": [
+        "EXEC sp_rename 'OldTable', 'NewTable';"
```
alanwest: Is this special handling for `sp_rename`? I'd expect parameter values to be sanitized.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Ah yes, I must have scanned over that. I'll update to handle this. Do you think we should a) truncate at the end of the last token that falls before 255, or b) all completion of any tokens starting before 255 chars, so potentially going over the 255 recommended limit? The later is only likely an issue for unusually long target names, but I doubt that's a real issue.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
stevejgordon: I'll open a PR upstream for the test and we can discuss there. I think both forms are acceptable if its hard to achieve without overhead in other languages.

Regarding invalid SQL statements. I'm unsure if we want to avoid setting `db.query.text`. Firstly, as you say, we end up building a full parser which is complex. Also, if an invalid statement is included, that will fail the DB command and in that case, the user likely wants to know what the bad SQL was. My only real concern would be SQL built from unrestricted/invalidated user input. In that case, there could be a small vector to make something nasty that hangs up our parsing. Realistically though, if the customer code accept unvalidated user input, they have bigger problems!

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Yes I was unsure whether we wanted those keywords capture in the summary. It adds context to what the statement does, directly in the span name, but perhaps there's little value in that. I doubt many instrumented apps create indices. I'm happy to update to exclude these for now. I can add the test upstream and perhaps a discussion can occur there on whether to include these (or not). My only concern would be if that discussion delays getting this instrumentation to a stable state.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Same comment as https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116#discussion_r2390811394

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Actually this test wasn't running at all! I'll update to make sure it does. I expect these will be sanitised. We could special case here as we know these to be table names, and that might be useful to know, but I doubt it's worth the effort as this is likely to be rarely used.

stevejgordon: Fixed in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116/commits/ab848e12090f01fe33c9df4b3bdb16fa2edeec01

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Fixed in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116/commits/007f300cb0e83090e392e1f0d49bcf3d0d5dfd76

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Fixed in https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116/commits/007f300cb0e83090e392e1f0d49bcf3d0d5dfd76

martincostello (approved):

stevejgordon:
@martincostello I have one more (hopefully) thing to tackle which is the truncation once we decide on option a) or b).

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
alanwest: I think always staying <= 255 would be more true to the spec. So in this example `db.query.summary` would just be `SELECT`.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
alanwest: Sounds good. We can always decide they should be included later if there is demand for it.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
alanwest: > I'm unsure if we want to avoid setting db.query.text.

This is kind of a secondary concern to me. Primary concern is that the literal values are sanitized even if the SQL is malformed. I think we should err on the side of being extra aggressive.

So this test case does bring me a little bit of pause. Are you confident that any literal values that might come after the statement will be sanitized? Few examples

1. `ALTER  TABLE MyTable ADD Name varchar(255) 'mypassword' 1234` would become `ALTER  TABLE MyTable ADD Name varchar(255) ? ?`
2. `ALTER  TABLE MyTable 'mypassword' ADD 123 Name varchar(255)` would become `ALTER  TABLE MyTable ? ADD ? Name varchar(255)`

stevejgordon: I've done some deeper review of this. Those examples seem to be okay, but there are other cases where the current logic is too open and we'd skip required sanitisation.

We could limit this more specifically to only after known type names (e.g. `VARCHAR`) by looking back to check a known type keyword preceeds the parameterised numeric value. I did discover there's no real common list of type names though across dialects though. If we wanted to add tests, we'd either need to repeat some test cases for each `db.system.name` where they apply, or update the test JSON so that `db.system.name` accepts an array for the DB systems the test case applies to. 

Perhaps this is more effort than is neccesary for this case, and we just allow them to be sanitized. I'm on the fence, so happy to go with the concensus here. @martincostello do you have any feelings towards this?

src/Shared/SqlProcessor.cs
```diff
@@ -17,141 +101,470 @@ public static SqlStatementInfo GetSanitizedSql(string? sql)
...
+        var summary = state.SummaryBuffer.Slice(0, state.SummaryPosition);
+
+        // If we have exceeded the max length for the summary, find the index of the last whitespace
+        // and trim the summary to that position. This avoids truncating within an operation name or target.
+        if (state.SummaryPosition > MaxSummaryLength)
+        {
+#if NET
+            var indexOfLastWhitespace = summary.Slice(0, MaxSummaryLength).LastIndexOfAny(WhitespaceSearchValues);
+#else
+            var indexOfLastWhitespace = summary.Slice(0, MaxSummaryLength).LastIndexOfAny([SpaceChar, TabChar, CarriageReturnChar, NewLineChar]);
```
martincostello: Is the compiler smart enough here to not allocate a new array here every time, or do we need to create an equivalent of `WhitespaceSearchValues` that's a `char[]`?

src/Shared/SqlProcessor.cs
```diff
@@ -17,141 +101,470 @@ public static SqlStatementInfo GetSanitizedSql(string? sql)
...
+        ReadOnlySpan<char> sql,
+        Span<char> buffer,
+        ref ParseState state)
+    {
+        var start = state.ParsePosition;
+
+#if NET
+        var indexOfNextWhitespace = sql.Slice(start).IndexOfAny(WhitespaceSearchValues);
+#else
+        var indexOfNextWhitespace = sql.Slice(start).IndexOfAny([SpaceChar, TabChar, CarriageReturnChar, NewLineChar]);
```
martincostello: +1

src/Shared/SqlProcessor.cs
stevejgordon: Good question and catch! I've made a static array for these just to be sure.

src/Shared/SqlProcessor.cs
stevejgordon: Good question and catch! I've made a static array for these just to be sure.

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
stevejgordon: Implementation and test case updated to apply this behaviour

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
stevejgordon: For now, I've removed the special handling and we simply sanitize (as before) any literal, regardless of it following a type name. I can add it back, with some more controls if we want to. I've also added these malformed examples to the additional tests so we will detect any such issues if we reintroduce support for displaying the values.

I've added a mechanism to track the previous token which supports the simplified IN clauses that @alanwest and I discussed offline. We could reuse that logic to check for type names like `VARCHAR` and more carefully skip sanitization in those cases. The only slight downside would be slightly slower execution time on the first literal after parentheses because we'd need to then check for known type names as part of that specialised path. I can benchmark the difference if we're interested to pursue this?

martincostello: No strong feelings on this from me.

stevejgordon: In that case, I propose we let the values be sanitized and we can revisit if there's a strong need from users for this. I suspect most applications won't be running such commands and the value would be low.

stevejgordon:
@alanwest / @martincostello - I think this PR is "done" then unless we have any more items to address?

martincostello (approved):

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorTestCases.cs
alanwest: > I've added a mechanism to track the previous token which supports the simplified IN clauses

🎉 

> For now, I've removed the special handling and we simply sanitize (as before) any literal, regardless of it following a type name.

👍 sounds good. I think this is the safest option for now. We can always circle back to this later.

alanwest (approved):
LGTM!

PR #3121, opened 2025-09-17
[OpAMP.Client] Support large WebSocket messages

RassK:
## Changes

Supports large websocket messages sent by the client.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
@@ -54,7 +54,7 @@ internal sealed class WsTransmitter
             // Copy the message
             Buffer.BlockCopy(messageBuffer, 0, buffer, headerSize, size);
 
-            while (true)
+            while (!token.IsCancellationRequested)
             {
                 var count = buffer.Length - offset < BufferSize
                     ? buffer.Length - offset
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
@@ -43,8 +43,34 @@ internal sealed class WsTransmitter
         // Does not fit, need to chunk the message
         else
         {
-            // TODO: Implement chunking logic for large messages
-            throw new NotImplementedException();
+            // It's expected that large messages are created very rarely by the client.
+            var messageBuffer = message.ToByteArray();
+            var buffer = new byte[headerSize + size];
+            var offset = 0;
+
+            // Copy the already written header
+            Buffer.BlockCopy(this.buffer, 0, buffer, 0, headerSize);
+
+            // Copy the message
+            Buffer.BlockCopy(messageBuffer, 0, buffer, headerSize, size);
+
+            while (true)
+            {
+                var count = buffer.Length - offset < BufferSize
+                    ? buffer.Length - offset
+                    : BufferSize;
+
+                var segment = new ArraySegment<byte>(buffer, offset, count);
+                var isEnd = (offset + count) == buffer.Length;
+                await this.ws.SendAsync(segment, WebSocketMessageType.Binary, isEnd, token).ConfigureAwait(false);
+
+                offset += count;
+
+                if (offset >= buffer.Length)
+                {
+                    break;
+                }
+            }
         }
     }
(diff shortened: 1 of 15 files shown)


src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
```diff
@@ -43,8 +43,34 @@ public async Task SendAsync(IMessage message, CancellationToken token = default)
...
+            var buffer = new byte[headerSize + size];
+            var offset = 0;
+
+            // Copy the already written header
+            Buffer.BlockCopy(this.buffer, 0, buffer, 0, headerSize);
+
+            // Copy the message
+            Buffer.BlockCopy(messageBuffer, 0, buffer, headerSize, size);
+
+            while (true)
```
martincostello, line 57: Could check for `token.IsCancellationRequested` and add a `ThrowIfCancellationRequested()` after the end of the block? I'm always a bit wary of `while(true)` 😄

RassK: I can see the missed opportunity for `token.IsCancellationRequested` check but couldn't see any good place for `ThrowIfCancellationRequested()`? In the middle there is already `this.ws.SendAsync(..., token)` which would throw, then a few hops to check again `token.IsCancellationRequested` with no additional logic to continue.

martincostello: I think we'd still want a call after line 73 before the method returns, otherwise if the loop exits because cancellation was signalled before the buffer completes copying the method will return success when it didn't complete the full copy/send.

RassK: But then you would want to have no `token.IsCancellationRequested` checks and both `ThrowIfCancellationRequested()`. There is a very unlikely but still possible successful return. Although personally I don't see it as an issue, since whatever way a method returns the highest level (which calls for cancel) always knowns the reason and can act accordingly + catching an exception in the middle also nullifies the cancellation throw.

it's possible to generate the exception in the higher level again to make sure it bubbles to the caller. So that leaves the internal loop more performant.

martincostello: I think we want either:

1. `while (true)` with `token.ThrowIfCancellationRequested()` on the first line of the loop
2. `while (!token.IsCancellationRequested)` with a single `token.ThrowIfCancellationRequested()` immediately after the loop

RassK: woun't `while (!token.IsCancellationRequested)` cause issues in any way?

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
```diff
@@ -43,8 +43,34 @@ public async Task SendAsync(IMessage message, CancellationToken token = default)
         // Does not fit, need to chunk the message
         else
         {
-            // TODO: Implement chunking logic for large messages
-            throw new NotImplementedException();
+            // It's expected that large messages are created very rarely by the client.
+            var messageBuffer = message.ToByteArray();
+            var buffer = new byte[headerSize + size];
```
Kielek: local variable hides the filed from the class.
If it is intentional, please change the name

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransmitter.cs
Kielek: This construction will not rise the exception when it was really cancelled.
I would go with option 1. While true + ThrowIwCancellationRequested.

martincostello: `while (!token.IsCancellationRequested)` by itself would mean it's possible (if the timing is just right) for the method to exit when cancelled without draining the buffer completely.

RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3121/commits/525d60341829727fc64fc6ce286b2923dc9889e5

PR #3145, opened 2025-09-23
[Extensions.AWS] ensure aws-xray-propagator maintains trace state

majanjua-amzn:
## Fixes: N/A
AWSXRayPropagator up until now overwrites the trace state to the default as it does not check if this has been set previously by, for example, the W3CTraceContextPropagator.

The desired behaviour is that the AWS X-Ray Propagator does not fill it in itself but also does not overwrite the information by a previously set propagator, e.g. `OTEL_PROPAGATORS=tracecontext,xray` and `OTEL_PROPAGATORS=xray,tracecontext` should lead to the same behaviour.

## Changes
- Updated the parsing function to take in the context and propagate the value of the trace state from the context as part of the newly created context
- Updated unit test and changelog accordingly

* [x] Changes in public API reviewed (if applicable): N/A

### DIFF
--- a/test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
+++ b/test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
@@ -261,4 +261,20 @@ public class AWSXRayPropagatorTests
 
         Assert.Equal(new PropagationContext(activityContext, default), this.awsXRayPropagator.Extract(default, carrier, Getter));
     }
+
+    [Fact]
+    public void TestExtractTraceHeaderFromUpstreamPropagator()
+    {
+        var carrier = new Dictionary<string, string>()
+        {
+            { AWSXRayTraceHeaderKey, "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1" },
+        };
+        var traceId = ActivityTraceId.CreateFromString(TraceId.AsSpan());
+        var parentId = ActivitySpanId.CreateFromString(ParentId.AsSpan());
+        var traceState = "key=val,foo=bar";
+        var context = new PropagationContext(new ActivityContext(traceId, parentId, ActivityTraceFlags.Recorded, traceState), default);
+
+        // Extraction is expected to fast-exit, returning the original context
+        Assert.Equal(context, this.awsXRayPropagator.Extract(context, carrier, Getter));
+    }
 }
--- a/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* test: ensure aws xray propagator maintains context from previous propagator
+  ([#3145](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3145))
+
 ## 1.12.1
 
 Released 2025-Sep-03


Kielek:
@srprash, @normj,

could you please review?

test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
```diff
@@ -261,4 +261,19 @@ public void TestExtractTraceHeaderWithAdditionalField()
...
+    public void TestExtractTraceHeaderWithTraceState()
+    {
+        var carrier = new Dictionary<string, string>()
+        {
+            { AWSXRayTraceHeaderKey, "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1" },
+        };
+        var traceState = "key=val,foo=bar";
+        var inputContext = new PropagationContext(new ActivityContext(default, default, ActivityTraceFlags.None, traceState), default);
+
+        var extractedContext = this.awsXRayPropagator.Extract(inputContext, carrier, Getter);
```
normj: The test doesn't really test anything because it goes into `Extract` and quickly returns because the following condition at the start of the extract is true.
```
        if (context.ActivityContext.IsValid())
        {
            return context;
        }
```

src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
```diff
@@ -223,7 +223,8 @@ internal static bool TryParseXRayTraceHeader(string rawHeader, out ActivityConte
         var activityParentId = ActivitySpanId.CreateFromString(parentId);
         var activityTraceOptions = traceOptions == SampledValue ? ActivityTraceFlags.Recorded : ActivityTraceFlags.None;
 
-        activityContext = new ActivityContext(activityTraceId, activityParentId, activityTraceOptions, isRemote: true);
+        // Include trace state previously set by other propagators
+        activityContext = new ActivityContext(activityTraceId, activityParentId, activityTraceOptions, context.ActivityContext.TraceState, isRemote: true);
```
normj: I was trying to find what other propagators do for this since this doesn't seem like an X-Ray specific pattern that must be followed. I saw in [TraceContextPropagator](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/Context/Propagation/TraceContextPropagator.cs?utm_source=chatgpt.com#L76) it is using the `getter` to get the value with a lot of extra code for parsing that I don't think we would want to duplicate here. In the [JaegerPropagator](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Extensions.Propagators/JaegerPropagator.cs#L71) and [B3Propagator](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Extensions.Propagators/B3Propagator.cs#L176) I don't see them passing along the tracestate.

@Kielek Is there an expectation on propagators caring forward trace state set by previous propagators?

normj (changes requested):

src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
Kielek: Sorry for late response.

By https://www.w3.org/TR/trace-context/ tracestate should be propagated only when you have W3C propagator context enabled.

There is a possibility to have more than one propagator enabled.

It is about AWS XRay propagator specification (if you have it) making this required/optional when working with only this propagator.

BTW if you fully work with the W3C now, I would consider even obsoleting this functionality here.

majanjua-amzn: You're correct to point out that B3 and Jaeger don't do this. This is in fact how we found the issue of certain propagators hindering the W3C propagator from propagating the trace state, eg. by setting `OTEL_PROPAGATORS=tracecontext,b3`, B3 would overwrite the trace state set by the W3C propagator.

X-Ray team found the same issue in our propagator, so we set out to fix this with this PR.

majanjua-amzn:
Closing the PR as thanks to the comments I've confirmed the fast-exit done through the following code block suffices to maintain the context and the associated trace state:
```
        if (context.ActivityContext.IsValid())
        {
            return context;
        }
```

Issue #1516, opened 2023-12-21
ResourceDetector.Host- implement other resources


Kielek:
# Issue with OpenTelemetry.ResourceDetectors.Host

List of [all OpenTelemetry NuGet
packages](https://www.nuget.org/profiles/OpenTelemetry) and version that you are
using (e.g. `OpenTelemetry 1.3.2`):

* `OpenTelemetry.ResourceDetectors.Host` `0.1.0-aplha.1`

Runtime version (e.g. `net462`, `net48`, `net6.0`, `net7.0` etc. You can
find this information from the `*.csproj` file):

* N/A

**Is this a feature request or a bug?**

* [x] Feature Request
* [ ] Bug

**What is the expected behavior?**

Implement all possible resources from https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/host.md

**What is the actual behavior?**

v0.1.0-alpha.1 implements only `host.name`.

thompson-tomo:
As of [v1.42.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/resource/host.md):

As a starting point we should implement:

- ~~host.arch~~

Recommended:

- host.image.id
- host.image.name
- host.image.version
- host.type

Opt-in:

- host.ip
- host.mac
- host.cpu.cache.l2.size
- host.cpu.family
- host.cpu.model.id
- host.cpu.model.name
- host.cpu.vendor.id
- host.cpu.stepping
- host.cpu.vendor.id

martincostello:
Updated the list as of v1.44.0 for what's missing.

PetarJerinic:
I volunteer to take this one. My thoughts/proposal are below. If maintainers agree with the direction, I'll create a PR.

Based on the attributes in v1.44.0 and the packages that exist here, I would propose the following split: 
- `host.type`, `host.image.{id,name,version}`. Add them to the cloud packages (`OpenTelemetry.Resources.AWS`, `OpenTelemetry.Resources.Azure`, `OpenTelemetry.Resources.Gcp`), not `OpenTelemetry.Resources.Host` since they come from the providers. #5110 is an example for `host.image.id`. Since those are recommended, I would have them on by default (just like `host.name`, `host.id` and `host.arch`)
[...]

PR #3147, opened 2025-09-25
Add host.arch attribute to HostDetector and update tests

AguirreMoy:
Partially implements #1516

## Changes

Adds detection and reporting of the host architecture (host.arch) using `System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture` in the `HostDetector`.

The host.arch attribute is now included in the detected resource attributes on supported platforms. The README has been updated to document the new attribute, and unit tests have been enhanced to verify that host.arch is present and correctly set. 

This improves observability by providing architecture information as part of the host resource metadata.

CleanShot 2025-09-24 at 20 04 16@2x

### DIFF
--- a/src/OpenTelemetry.Resources.Host/CHANGELOG.md
+++ b/src/OpenTelemetry.Resources.Host/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added support for the `host.arch` resource attribute in `HostDetector`, reporting the host architecture (e.g., `X64`, `Arm64`) using `System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture`.
+
 ## 1.12.0-beta.1
 
 Released 2025-May-06
--- a/src/OpenTelemetry.Resources.Host/HostDetector.cs
+++ b/src/OpenTelemetry.Resources.Host/HostDetector.cs
@@ -94,6 +94,14 @@ internal sealed class HostDetector : IResourceDetector
             {
                 attributes.Add(new(HostSemanticConventions.AttributeHostId, machineId));
             }
+#if !NETFRAMEWORK
+            // Architecture is only supported in .NET 5+
+            var arch = RuntimeInformation.ProcessArchitecture.ToString();
+            if (arch != null && !string.IsNullOrEmpty(arch))
+            {
+                attributes.Add(new(HostSemanticConventions.AttributeHostArch, arch));
+            }
+#endif
 
             return new Resource(attributes);
         }
--- a/src/OpenTelemetry.Resources.Host/README.md
+++ b/src/OpenTelemetry.Resources.Host/README.md
@@ -2,7 +2,7 @@
 
 | Status      |           |
 | ----------- | --------- |
-| Stability   | [Beta](../../README.md#beta) |                                                 |
+| Stability   | [Beta](../../README.md#beta) |
 | Code Owners | [@Kielek](https://github.com/Kielek), [@lachmatt](https://github.com/lachmatt) |
 
 [![NuGet version badge](https://img.shields.io/nuget/v/OpenTelemetry.Resources.Host)](https://www.nuget.org/packages/OpenTelemetry.Resources.Host)
--- a/src/OpenTelemetry.Resources.Host/README.md
+++ b/src/OpenTelemetry.Resources.Host/README.md
@@ -1,8 +1,8 @@
 # Host Resource Detectors
 
-| Status      |                                                                                |
-| ----------- | ------------------------------------------------------------------------------ |
-| Stability   | [Beta](../../README.md#beta)                                                   |
+| Status      |           |
+| ----------- | --------- |
+| Stability   | [Beta](../../README.md#beta) |                                                 |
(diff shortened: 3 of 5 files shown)


src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -94,6 +94,14 @@ public Resource Detect()
             {
                 attributes.Add(new(HostSemanticConventions.AttributeHostId, machineId));
             }
+#if !NETFRAMEWORK
+            // Architecture is only supported in .NET 5+
+            var arch = RuntimeInformation.ProcessArchitecture.ToString();
```
Kielek: I think that you need to use some custom mapping for the expected values. From: https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/resource/host.md

host.arch has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

For the values not defined in the table, probbly ToLowerInvariant() will be better option.

src/OpenTelemetry.Resources.Host/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Added support for the `host.arch` resource attribute in `HostDetector`, reporting the host architecture (e.g., `X64`, `Arm64`) using `System.Runtime.InteropServices.RuntimeInformation.ProcessArchitecture`.
```
Kielek: Please add link to this PR, as for other entries.

Kielek: Please include information that it is supported only on .NET.
Details about the source is not so important.

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -54,8 +54,11 @@ using var loggerFactory = LoggerFactory.Create(builder =>
 The resource detectors will record the following metadata based on where
 your application is running:
 
-- **HostDetector**: `host.id` (when running on non-containerized systems), `host.name`.
+- **HostDetector**:
+  - `host.id` (when running on non-containerized systems)
+  - `host.name`
+  - `host.arch` (the host architecture, e.g., `X64`, `Arm64`)
```
Kielek: Missing information that it is supported only on .NET

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -94,6 +94,14 @@ public Resource Detect()
...
+            if (arch != null && !string.IsNullOrEmpty(arch))
+            {
+                attributes.Add(new(HostSemanticConventions.AttributeHostArch, arch));
```
Kielek, line 126: You should conditionally update line 87 with pre-alocated buffer for strings.

Kielek (commented):
Thanks for your contribution. Please fix some issues I have found.

src/OpenTelemetry.Resources.Host/CHANGELOG.md
AguirreMoy: Added

AguirreMoy: Added details that it's only supported on .net 5+ and not framework

src/OpenTelemetry.Resources.Host/HostDetector.cs
AguirreMoy: Good catch, updated

src/OpenTelemetry.Resources.Host/README.md
AguirreMoy: Updated

src/OpenTelemetry.Resources.Host/HostDetector.cs
AguirreMoy: Alrighty, mapped to the existing ones on table and the ones not covered will use tolowerinvariant

AguirreMoy:
Hi @Kielek , thanks for the quick review. I've addressed your comments. I marked the easy ones as resolved and left the remainder open for further feedback. Thanks!

src/OpenTelemetry.Resources.Host/HostDetector.cs
Kielek: I changed a bit behavior, when it is not well known mapped version we just omitting this attribute.
It is more strict behavior, but current implementation covers majority of supported runtimes.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -76,6 +76,29 @@ internal HostDetector(
...
+            Architecture.Ppc64le => "ppc64",
+
+            // following architectures do not have a mapping in OTEL spec https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/resource/host.md
+            Architecture.Wasm => null,
+            Architecture.LoongArch64 => null,
+#endif
+            _ => null,
+        };
+    }
+#endif
```
ysolomchenko, line 100: Why not make all the code #if NET instead of #if !NETFRAMEWORK?

Kielek: NetStandard targets can compile with this 4 cases

PR #3214, opened 2025-10-09
Support for specifying resouce attributes in Geneva traces exporter

### LINKED ISSUE #2999: [feature request] GenevaTraceExporter: Add ResourceBuilder support
### Component

OpenTelemetry.Exporter.Geneva

It is [recommended](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/docs/trace/README.md#modelling-static-tags-as-resource) to model static tags as Resource, but it is useless because Geneva Trace Exporter ignores Resource

GenevaTraceExporter respects the Resource builder and adds tags from Resource to the exported traces.

There are plenty of workarounds such as custom processors, OpenTelemetry.Extensions.Enrichment library, etc, but each of them adds tags to every single Activity, which wastes performance for static tags which are always the same for the app lifecycle.

### Additional context

_No response_

### Tip

mattsains:
Fixes #2999

## Changes

Copies resource attributes to exported spans in part C of common schema, and also maps the following resource attributes:

* service.name -> Geneva cloud.role
* service.instanceId -> Geneva cloud.roleInstance

I also made substantial changes to the unit test to make sure all spans generated in the tests are verified correctly.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.13.0-rc.1
+
+Released 2025-Oct-28
+
 ## 1.13.0-beta.2
 
 Released 2025-Oct-22
--- a/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/README.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/README.md
@@ -2,7 +2,7 @@
 
 | Status      |           |
 | ----------- | --------- |
-| Stability   | [Beta](../../README.md#beta) |
+| Stability   | [Release candidate](../../README.md#release-candidate) |
 | Code Owners | [@open-telemetry/dotnet-contrib-maintainers](https://github.com/orgs/open-telemetry/teams/dotnet-contrib-maintainers) |
 
 [![NuGet version badge](https://img.shields.io/nuget/v/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule)](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNet.TelemetryHttpModule/)
--- a/src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 1.13.0-rc.1
+
+Released 2025-Oct-28
+
 ## 1.13.0-beta.2
 
 Released 2025-Oct-22
--- a/src/OpenTelemetry.Instrumentation.AspNet/README.md
+++ b/src/OpenTelemetry.Instrumentation.AspNet/README.md
@@ -2,7 +2,7 @@
 
 | Status      |           |
 | ----------- | --------- |
-| Stability   | [Beta](../../README.md#beta) |
+| Stability   | [Release candidate](../../README.md#release-candidate) |
 | Code Owners | [@open-telemetry/dotnet-contrib-maintainers](https://github.com/orgs/open-telemetry/teams/dotnet-contrib-maintainers) |
 
 [![NuGet version badge](https://img.shields.io/nuget/v/OpenTelemetry.Instrumentation.AspNet)](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.AspNet)


src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -328,28 +331,75 @@ internal ArraySegment<byte> SerializeActivity(Activity activity)
         var dtBegin = activity.StartTimeUtc;
         var tsBegin = dtBegin.Ticks;
         var tsEnd = tsBegin + activity.Duration.Ticks;
-        var dtEnd = new DateTime(tsEnd);
+        var dtEnd = new DateTime(tsEnd, DateTimeKind.Utc);
+
+        string? serviceName = null;
```
rajkumar-rangaraj: Since the Resource is typically static and doesn't change during the exporter's lifetime, consider capturing it as a field during exporter initialization rather than passing it as a parameter on every call. This would reduce overhead and make it clear that resource attributes are a configuration-time concern. This is closely related to `TableNameMappings`, may also need to address what would happen if attribute is set in both TableNameMappings  and Resource.

mattsains-msft: thanks, I will do, I considered this but wasn't entirely sure the lifetime of resource. I wasn't sure if it was static. Now that I know, I feel more confident making that change.

mattsains-msft: Could you explain a bit more what you meant regarding TableNameMappings? It seems to me that the table name mapping only translates "Span" into a custom table name if necessary. It shouldn't affect resource attributes or actually any common schema field other than the part A name

rajkumar-rangaraj: sorry, meant to say PrepopulatedFields - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/e343281bbc2126961bb7c535f2e8b0ba071bf55e/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs#L99

mattsains-msft: I've made this change.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Released 2025-Oct-13
 
 * Updated OpenTelemetry core component version(s) to `1.13.1`.
   ([#3218](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3218))
+* Support for specifying resource attributes, including
```
rajkumar-rangaraj, line 11: - 
```suggestion
* Support for specifying resource attributes, including
  `service.name`, `service.instanceId`, and custom attributes.
```

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -146,8 +147,6 @@ public void GenevaTraceExporter_Success_Windows()
                 options.ConnectionString = "EtwSession=OpenTelemetry";
                 options.PrepopulatedFields = new Dictionary<string, object>
                 {
-                    ["cloud.role"] = "BusyWorker",
```
rajkumar-rangaraj, line 149: Can we have one more test to verify what happens when both cloud.role and service.name are present? The same applies to cloud.roleInstance as well.

rajkumar-rangaraj (approved):

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
mattsains: I've added this test.

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -313,40 +320,90 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
         return urlStringBuilder.ToString();
     }
 
+    internal int AppendResourceAttributesToBuffer(byte[] buffer, int cursor)
```
cijothomas, line 323: doing this in hot path (export) is going to affect performance. Resource information is statically available, we need to leverage that, and augment the prologue *once* and re-use in hot path.

mattsains-msft: This function is only called for the first Export call, and never again for the life of the MsgPackTraceExporter object. I was told by @rajkumar-rangaraj that the Resource is not available during the constructor, which is why I do it on the first export call

cijothomas: Got it. If this is done during 1st export only, then this is the best we can do. Raj is right, resource is not obtainable at ctor.

cijothomas: (feel free to run a benchmarks to confirm there is major perf hit)

PR #3258, opened 2025-10-21
[Instrumentation.Hangfire] Add metrics support

### LINKED ISSUE #2075: Support for metrics in OpenTelemetry.Instrumentation.Hangfire
### Component

OpenTelemetry.Instrumentation.Hangfire

Currently to use hangfire metrics you have either:
1. Install separate exporter
2. Use prometheus directly with hangfire to prometheus exporter

I want to install OpenTelemetry.Instrumentation.Hangfire and have all the metrics from nuget `Hangfire.Prometheus.NetCore`

nuget `Hangfire.Prometheus.NetCore`

### Additional context

_No response_

gorbach:
Fixes #2075

## Changes

Adds OpenTelemetry metrics support to Hangfire instrumentation following workflow semantic conventions.

Implemented metrics:
- `workflow.execution.count` - Counter for task executions
- `workflow.execution.duration` - Histogram for execution duration
- `workflow.execution.status` - UpDownCounter for state transitions
- `workflow.execution.errors` - Counter for execution errors
- `hangfire.queue.latency` - Optional histogram for queue wait time

* [x] CONTRIBUTING guidelines followed
* [x] CHANGELOG.md updated

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Hangfire/README.md
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/README.md
@@ -289,12 +289,12 @@ The number of task executions which have been initiated.
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
-| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
-| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`   |
-| `workflow.execution.result`   | string | The result of executing the task              | Required                 | `success`, `failure`    |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
-| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
+| Attribute                   | Type   | Description                            | Requirement Level         | Values                |
+| --------------------------- | ------ | -------------------------------------- | ------------------------- | --------------------- |
+| `workflow.task.name`        | string | Name of the task (Hangfire job method) | Required                  | e.g., `MyJob.Execute` |
+| `workflow.execution.result` | string | The result of executing the task       | Required                  | `success`, `failure`  |
+| `workflow.platform.name`    | string | The workflow platform being used       | Recommended               | `hangfire`            |
+| `error.type`                | string | The type of error that occurred        | Conditionally Required[1] | Exception type name   |
 
 [1]: Required if and only if the task execution failed.
 
@@ -319,13 +319,13 @@ for different execution phases:
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
-| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
-| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`   |
-| `workflow.execution.result`   | string | The result of executing the task              | Required                 | `success`, `failure`    |
-| `workflow.execution.state`    | string | The execution phase being measured            | Required                 | `pending`, `executing`  |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
-| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
+| Attribute                   | Type   | Description                            | Requirement Level         | Values                 |
+| --------------------------- | ------ | -------------------------------------- | ------------------------- | ---------------------- |
+| `workflow.task.name`        | string | Name of the task (Hangfire job method) | Required                  | e.g., `MyJob.Execute`  |
+| `workflow.execution.result` | string | The result of executing the task       | Required                  | `success`, `failure`   |
+| `workflow.execution.state`  | string | The execution phase being measured     | Required                  | `pending`, `executing` |
+| `workflow.platform.name`    | string | The workflow platform being used       | Recommended               | `hangfire`             |
+| `error.type`                | string | The type of error that occurred        | Conditionally Required[1] | Exception type name    |
 
 [1]: Required if and only if the task execution failed.
 
@@ -333,9 +333,9 @@ for different execution phases:
 
 The number of actively running tasks grouped by task and the current state.
 
-| Units          | Instrument Type  | Value Type |
-|----------------|------------------|------------|
-| `{executions}` | UpDownCounter    | `Int64`    |
+| Units          | Instrument Type | Value Type |
+| -------------- | --------------- | ---------- |
+| `{executions}` | UpDownCounter   | `Int64`    |
 
 > [!NOTE]
 > This metric tracks the current state of job executions. When a job
@@ -344,12 +344,12 @@ The number of actively running tasks grouped by task and the current state.
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
-|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------------------------|
-| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`                     |
-| `workflow.execution.state`    | string | Current state of the execution                | Required                 | `pending`, `executing`, `completed`       |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`                                |
-| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name                       |
+| Attribute                  | Type   | Description                            | Requirement Level         | Values                              |
+| -------------------------- | ------ | -------------------------------------- | ------------------------- | ----------------------------------- |
+| `workflow.task.name`       | string | Name of the task (Hangfire job method) | Required                  | e.g., `MyJob.Execute`               |
+| `workflow.execution.state` | string | Current state of the execution         | Required                  | `pending`, `executing`, `completed` |
+| `workflow.platform.name`   | string | The workflow platform being used       | Recommended               | `hangfire`                          |
+| `error.type`               | string | The type of error that occurred        | Conditionally Required[1] | Exception type name                 |
 
 [1]: Required if and only if the task execution failed.
 
@@ -357,11 +357,11 @@ The number of actively running tasks grouped by task and the current state.
 
 Hangfire job states are mapped to workflow semantic convention states as follows:
 
-| Hangfire State                     | Workflow State |
-| ---------------------------------- | -------------- |
-| Scheduled, Enqueued, Awaiting      | `pending`      |
-| Processing                         | `executing`    |
-| Succeeded, Failed, Deleted         | `completed`    |
+| Hangfire State                | Workflow State |
+| ----------------------------- | -------------- |
+| Scheduled, Enqueued, Awaiting | `pending`      |
+| Processing                    | `executing`    |
+| Succeeded, Failed, Deleted    | `completed`    |
 
 #### workflow.execution.errors
 
@@ -373,30 +373,30 @@ The number of errors encountered in task executions.
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level | Values                  |
-|-------------------------------|--------|-----------------------------------------------|-------------------|-------------------------|
-| `error.type`                  | string | The type of error that occurred               | Required          | Exception type name     |
-| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required          | e.g., `MyJob.Execute`   |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended       | `hangfire`              |
+| Attribute                | Type   | Description                            | Requirement Level | Values                |
+| ------------------------ | ------ | -------------------------------------- | ----------------- | --------------------- |
+| `error.type`             | string | The type of error that occurred        | Required          | Exception type name   |
+| `workflow.task.name`     | string | Name of the task (Hangfire job method) | Required          | e.g., `MyJob.Execute` |
+| `workflow.platform.name` | string | The workflow platform being used       | Recommended       | `hangfire`            |
 
 #### workflow.outcome
 
 The number of workflow instances which have been initiated. In Hangfire, this tracks
 individual job completions.
 
-| Units          | Instrument Type | Value Type |
-| -------------- | --------------- | ---------- |
-| `{workflows}`  | Counter         | `Int64`    |
+| Units         | Instrument Type | Value Type |
+| ------------- | --------------- | ---------- |
+| `{workflows}` | Counter         | `Int64`    |
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
-| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
-| `workflow.definition.name`    | string | Name of the workflow (Hangfire job method)    | Required                 | e.g., `MyJob.Execute`   |
-| `workflow.result`             | string | The result of the workflow                    | Required                 | `success`, `failure`    |
-| `workflow.trigger.type`       | string | Type of trigger that initiated the workflow   | Required                 | `api`, `cron`, `schedule` |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
-| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
+| Attribute                  | Type   | Description                                 | Requirement Level         | Values                    |
+| -------------------------- | ------ | ------------------------------------------- | ------------------------- | ------------------------- |
+| `workflow.definition.name` | string | Name of the workflow (Hangfire job method)  | Required                  | e.g., `MyJob.Execute`     |
+| `workflow.result`          | string | The result of the workflow                  | Required                  | `success`, `failure`      |
+| `workflow.trigger.type`    | string | Type of trigger that initiated the workflow | Required                  | `api`, `cron`, `schedule` |
+| `workflow.platform.name`   | string | The workflow platform being used            | Recommended               | `hangfire`                |
+| `error.type`               | string | The type of error that occurred             | Conditionally Required[1] | Exception type name       |
 
 [1]: Required if and only if the workflow execution failed.
 
@@ -404,9 +404,9 @@ individual job completions.
 
 The number of actively running workflows grouped by definition and the current state.
 
-| Units          | Instrument Type  | Value Type |
-| -------------- | ---------------- | ---------- |
-| `{workflows}`  | UpDownCounter    | `Int64`    |
+| Units         | Instrument Type | Value Type |
+| ------------- | --------------- | ---------- |
+| `{workflows}` | UpDownCounter   | `Int64`    |
 
 > [!NOTE]
 > This metric tracks the workflow lifecycle including jobs that haven't
@@ -416,13 +416,13 @@ The number of actively running workflows grouped by definition and the current s
 
 **Attributes:**
 
-| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
-| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------------------------- |
-| `workflow.definition.name`    | string | Name of the workflow (Hangfire job method)    | Required                 | e.g., `MyJob.Execute`                     |
-| `workflow.state`              | string | Current state of the workflow                 | Required                 | `pending`, `executing`, `completed`       |
-| `workflow.trigger.type`       | string | Type of trigger that initiated the workflow   | Required                 | `api`, `cron`, `schedule`                 |
-| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`                                |
-| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name                       |
+| Attribute                  | Type   | Description                                 | Requirement Level         | Values                              |
+| -------------------------- | ------ | ------------------------------------------- | ------------------------- | ----------------------------------- |
+| `workflow.definition.name` | string | Name of the workflow (Hangfire job method)  | Required                  | e.g., `MyJob.Execute`               |
+| `workflow.state`           | string | Current state of the workflow               | Required                  | `pending`, `executing`, `completed` |
+| `workflow.trigger.type`    | string | Type of trigger that initiated the workflow | Required                  | `api`, `cron`, `schedule`           |
+| `workflow.platform.name`   | string | The workflow platform being used            | Recommended               | `hangfire`                          |
+| `error.type`               | string | The type of error that occurred             | Conditionally Required[1] | Exception type name                 |
 
 [1]: Required if and only if the workflow execution failed.
 
@@ -430,13 +430,13 @@ The number of actively running workflows grouped by definition and the current s
 
 Hangfire job states are mapped to workflow semantic convention states as follows:
 
-| Hangfire State                     | Workflow State | Workflow Trigger Type           |
-| ---------------------------------- | -------------- | ------------------------------- |
-| Scheduled                          | `pending`      | `schedule`                      |
-| Enqueued, Awaiting (from cron)     | `pending`      | `cron`                          |
-| Enqueued, Awaiting (fire-and-forget)| `pending`     | `api`                           |
-| Processing                         | `executing`    | (inherited from previous state) |
-| Succeeded, Failed, Deleted         | `completed`    | (inherited from previous state) |
+| Hangfire State                       | Workflow State | Workflow Trigger Type           |
+| ------------------------------------ | -------------- | ------------------------------- |
+| Scheduled                            | `pending`      | `schedule`                      |
+| Enqueued, Awaiting (from cron)       | `pending`      | `cron`                          |
+| Enqueued, Awaiting (fire-and-forget) | `pending`      | `api`                           |
+| Processing                           | `executing`    | (inherited from previous state) |
+| Succeeded, Failed, Deleted           | `completed`    | (inherited from previous state) |
 
 > [!IMPORTANT]
 > **Difference between `workflow.status` and `workflow.execution.status`:**


src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetrics.cs
```diff
@@ -0,0 +1,72 @@
...
+    internal static readonly string MeterName = AssemblyName.Name!;
+    private static readonly string InstrumentationVersion = Assembly.GetPackageVersion();
+
+    // Metric name constants
+    internal const string ExecutionCountMetricName = "workflow.execution.count";
+    internal const string ExecutionDurationMetricName = "workflow.execution.duration";
+    internal const string ExecutionStatusMetricName = "workflow.execution.status";
+    internal const string ExecutionErrorsMetricName = "workflow.execution.errors";
+
+    internal const string QueueLatencyMetricName = "hangfire.queue.latency";
```
thompson-tomo: Let's use the duration metric for this. I have commented on the sem conv pr of changes to support this use case.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
```diff
@@ -0,0 +1,201 @@
...
+    /// <param name="recurringJobId">Optional recurring job ID if this job was triggered by a recurring job.</param>
+    /// <returns>Tag list suitable for workflow.execution.status metric.</returns>
+    public static TagList BuildStateTags(Job job, string workflowState, string? errorType, string? recurringJobId)
+    {
+        var tags = new TagList
+        {
+            GetTaskName(job),
+            GetPlatformName(),
+            GetState(workflowState),
+            GetTriggerType(recurringJobId),
```
thompson-tomo: Let's leave this off the Workflow.execution metrics and just have it on the workflow metrics.

A Hangfire scenario would be if you are using Hangfire batches the Workflow would represent the complete running of the batch aka multiple tasks/job.

As such I would recommend including the Workflow.count metric to capture the triggers.

Kielek:
@gorbach, please check this comment https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2075#issuecomment-3420588992

The second point is technical requirements to separate metrics/traces signal. It will be great to rename classes in the separate PR and then merge it/rebase here.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /// <summary>
+    /// Maps Hangfire state names to workflow semantic convention state values.
+    /// </summary>
+    /// <param name="hangfireState">Hangfire state name.</param>
+    /// <returns>Mapped workflow state value, or <see langword="null"/> if the state is not recognized.</returns>
+    public static string? MapWorkflowState(string? hangfireState)
+    {
+        if (hangfireState == ScheduledState.StateName ||
```
thompson-tomo: Hmm, I am not sure scheduled would be considered as pending. I see a couple of options:

- introduce a scheduled option for workflow.execution.state
- not capture this state as part of the workflow.execution.status metric and simply ignore it
- not capture this state as part of the workflow.execution.status metric but instead introduce it as an option on the workflow.status metric.
- not capture this state as part of the workflow.execution.status metric but instead use pending state option on the workflow.status metric with the trigger type being schedule

I like option 4 as it naturally scales to support batches, you see all workflows and their state with the ability to see what is pending and if it is due to scheduling. In case of hangfire I don't foresee pending being used for anything other than scheduled.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,102 @@
...
+            this.options.DisplayNameFunc,
+            workflowState,
+            errorType);
+
+        HangfireMetrics.ExecutionStatus.Add(1, tags);
+    }
+
+    public void OnStateUnapplied(ApplyStateContext context, IWriteOnlyTransaction transaction)
+    {
+        var workflowState = HangfireTagBuilder.MapWorkflowState(context.OldStateName);
```
thompson-tomo: Should we also be checking old state != new state?

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,102 @@
...
+    public HangfireMetricsStateFilter(HangfireMetricsInstrumentationOptions options)
+#pragma warning restore CA1019 // Define accessors for attribute arguments
+    {
+        this.options = options ?? throw new ArgumentNullException(nameof(options));
+    }
+
+    public void OnStateApplied(ApplyStateContext context, IWriteOnlyTransaction transaction)
+    {
+        var workflowState = HangfireTagBuilder.MapWorkflowState(context.NewState.Name);
+        if (workflowState == null)
```
thompson-tomo: Should we also be checking new state != old state

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            return;
+        }
+
+        var errorType = GetErrorTypeFromNewState(context.NewState);
+        var tags = HangfireTagBuilder.BuildStateTags(
+            context.BackgroundJob,
+            this.options.DisplayNameFunc,
```
thompson-tomo: I don't think we should be reusing the trace display name function here. Instead We should use the Name of the job method.

gorbach: That's `DisplayNameFunc` from `HangfireMetricsInstrumentationOptions`. I added it becuase in hangfire 1.8 code `backgroundJob.Job?.ToString()` returns job name with a queue:

```CSharp
        public override string ToString()
        {
            return ToString(includeQueue: true);
        }

        public string ToString(bool includeQueue)
        {
            var sb = new StringBuilder()
                .Append(Type.ToGenericTypeString())
                .Append('.')
                .Append(Method.Name);

            if (includeQueue && Queue != null)
            {
                sb.Append(" (").Append(Queue).Append(')');
            }

            return sb.ToString();
        }
```

But `OpenTelemetry.Instrumentation.Hangfire` references Hangfire 1.7. Actually, I think perfect option would be to reference newer hangfire :)

gorbach: The current implementation of `HangfireInstrumentation.DefaultDisplayNameFunc` has a flaw: if the class containing the method implementation is generic, the job name becomes unreadable — for example, ``JOB TestJobWithGenerics\1.ProcessSingle``. However, using Hangfire’s `ToString()` method produces a nicely formatted name: `TestJobWithGenerics<String>.ProcessSingle`.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
gorbach: Thanks for the feedback! I want to clarify the scope of the workflow.status metric for Hangfire:

  Should `workflow.status` track:

**Option A** : Only scheduled jobs
  - `ScheduledState` → `workflow.status{state=pending, trigger.type=schedule}`
  - Once enqueued, only track in `workflow.execution.status{state=pending}`
  - Other job types (fire-and-forget, recurring) never appear in `workflow.status`

**Option B**: All jobs throughout their lifecycle
  - All jobs tracked in both `workflow.status` AND `workflow.execution.status`
  - `workflow.status` shows the complete workflow state
  - `workflow.execution.status` shows execution-specific states (excluding scheduled)

**Option C**: Something else?

In Hangfire's simple case (1 workflow = 1 task), Option A seems sufficient and avoids duplication. But I want to ensure this aligns with your vision for how the metrics should support batches in the future.

thompson-tomo: I like option b as I see them having different values. The Workflow moves from pending to active once the job moves out of scheduled. It only stops being active when the is final flag is set. 
As such if the execution state is pending it doesn't mean pending the Workflow state is pending.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,102 @@
...
+            workflowState,
+            errorType);
+
+        HangfireMetrics.ExecutionStatus.Add(1, tags);
```
thompson-tomo: We should only be adding to the status metric when the state is not final otherwise the counter will continuously grow for some states.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
thompson-tomo: So for me, I see:

- Name: Method.Name
- Id: Type.ToGenericTypeString() + '.' + Method.Name

I don't see the queue name as being a part of the task name as my assumption is that a re-occuring job would likely end up with a different name each execution. This would then make it a description for the execution and not the task/Workflow definition.

gorbach: The problem is that we cannot call `ToGenericTypeString` it is internal to Hangfire. We can use ToString()/ToString(false) in 1.8+ or replicate ToGenericTypeString

thompson-tomo: Sorry I Didn't realise it was a hangfire method. I would actually use Type.FullName as I don't see any issue with TestJobWithGenerics\1.ProcessSingle as the name/id as it shouldn't change based on arguments.

We could think about capturing on the span details about the arguments but that would be a seperate discussion to metrics.

gorbach: Yes, ``TestJobWithGenerics\1.ProcessSingle`` is "stable" but not really readable, I would prefer the formatted options from hangfire. Maybe we can copy that code from hangfire sources ?

thompson-tomo: Is there a way we could end up with TestJobWithGenerics.ProcessSingle as the name. Reason is I don't feel that we should be splitting tasks based on generic type used. Otherwise sure we could incorporate the logic from hangfire.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
gorbach: Do you mean `workflowState`, i.e. mapped hangfire state? I don't think that we need to add that check. If different hangfire states are mapped to the same `workflowState`, it means that we just decrement it and increment in the same instant. I beleive it is ok.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
gorbach: The problem is that generic argument could be more important than type. I used to work with a project, that had "base" class for all jobs - "HangfireBaseJob<T>" and was parametrized by real job of type "IJob". So in such system all jobs were displayed like "HangfireBaseJob<DoSomeUsefuleStuff>". While design is questionable, I beleive we should support such use cases.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
gorbach: I've implemented `workflow.status`

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
gorbach: I've added a check that prevents recording when the mapped state doesn't change. So transitions like **Succeeded** → **Deleted** (both map to completed) are now skipped.

  I'm confused about tracking final states:

  1. The semantic conventions list **completed** as a valid value for `workflow.execution.state`. Doesn't this mean we should track it in the status metric?
  2. If we don't track final states in the status metric, how would users know how many jobs are currently in `completed`/`succeeded`/`failed `states? Or should they use a different metric for this?
  3. Could you give me an example of when "the counter will continuously grow"? I want to make sure I understand the scenario you're concerned about.

  The documentation says the status metric tracks "actively running tasks" - does "completed" count as "actively running", or should we only track pending, executing, and finalizing?

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
thompson-tomo: Hmm, let's incorporate the logic from hangfire then to cater for such use cases but avoid it being defined in options.

gorbach: done

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsErrorFilterAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    private readonly HangfireMetricsInstrumentationOptions options;
+
+#pragma warning disable CA1019 // Define accessors for attribute arguments
+    public HangfireMetricsErrorFilterAttribute(HangfireMetricsInstrumentationOptions options)
+#pragma warning restore CA1019 // Define accessors for attribute arguments
+    {
+        this.options = options ?? throw new ArgumentNullException(nameof(options));
+    }
+
```
thompson-tomo:
```suggestion
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsInstrumentation.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+internal sealed class HangfireMetricsInstrumentation : IDisposable
+{
+    private readonly List<object> filters = new();
+
+    public HangfireMetricsInstrumentation(HangfireMetricsInstrumentationOptions options)
+    {
+        this.AddFilter(new HangfireMetricsJobFilterAttribute());
+        this.AddFilter(new HangfireMetricsStateFilter());
+        this.AddFilter(new HangfireMetricsErrorFilterAttribute(options));
```
thompson-tomo:
```suggestion
        this.AddFilter(new HangfireMetricsErrorFilterAttribute());
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+        catch
+        {
+            return null;
+        }
+    }
+
+    public void OnStateApplied(ApplyStateContext context, IWriteOnlyTransaction transaction)
+    {
+        this.RecordWorkflowStateChange(context, context.OldStateName, context.NewState.Name, StateTransitionDirection.Applied);
+        this.RecordExecutionStateChange(context, context.OldStateName, context.NewState.Name, StateTransitionDirection.Applied);
```
thompson-tomo:
```suggestion
        this.RecordWorkflowStateChange(context, StateTransitionDirection.Applied);
        this.RecordExecutionStateChange(context, StateTransitionDirection.Applied);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public void OnStateUnapplied(ApplyStateContext context, IWriteOnlyTransaction transaction)
+    {
+        this.RecordWorkflowStateChange(context, context.OldStateName, context.NewState.Name, StateTransitionDirection.Unapplied);
+        this.RecordExecutionStateChange(context, context.OldStateName, context.NewState.Name, StateTransitionDirection.Unapplied);
```
thompson-tomo:
```suggestion
        this.RecordWorkflowStateChange(context, StateTransitionDirection.Unapplied);
        this.RecordExecutionStateChange(context, StateTransitionDirection.Unapplied);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private void RecordWorkflowStateChange(
+        ApplyStateContext context,
+        string? oldStateName,
+        string? newStateName,
+        StateTransitionDirection direction)
+    {
+        var oldState = HangfireStateMapper.MapWorkflowState(oldStateName);
+        var newState = HangfireStateMapper.MapWorkflowState(newStateName);
```
thompson-tomo:
```suggestion
    private void RecordWorkflowStateChange(
        ApplyStateContext context,
        StateTransitionDirection direction)
    {
        var oldState = HangfireStateMapper.MapWorkflowState(context.OldStateName);
        var newState = HangfireStateMapper.MapWorkflowState(context.NewState.Name);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private void RecordExecutionStateChange(
+        ApplyStateContext context,
+        string? oldStateName,
+        string? newStateName,
+        StateTransitionDirection direction)
+    {
+        var oldState = HangfireStateMapper.MapExecutionState(oldStateName);
+        var newState = HangfireStateMapper.MapExecutionState(newStateName);
```
thompson-tomo:
```suggestion
    private void RecordExecutionStateChange(
        ApplyStateContext context,
        StateTransitionDirection direction)
    {
        var oldState = HangfireStateMapper.MapExecutionState(context.OldStateName);
        var newState = HangfireStateMapper.MapExecutionState(context.NewState.Name);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsJobFilterAttribute.cs
```diff
@@ -0,0 +1,68 @@
...
+        // Record execution duration (with state="executing" to differentiate from pending phase)
+        if (performedContext.Items.TryGetValue(StopwatchKey, out var stopwatchObj) && stopwatchObj is Stopwatch stopwatch)
+        {
+            stopwatch.Stop();
+            var duration = stopwatch.Elapsed.TotalSeconds;
+
+            var durationTags = HangfireTagBuilder.BuildExecutionTags(
+                backgroundJob,
+                performedContext.Exception,
+                workflowState: WorkflowAttributes.WorkflowStateValues.Executing);
```
thompson-tomo:
```suggestion
            var durationTags = HangfireTagBuilder.BuildExecutionTags(
                performedContext,
                workflowState: WorkflowAttributes.WorkflowStateValues.Executing);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsErrorFilterAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+    }
+
+    public void OnPerformed(PerformedContext performedContext)
+    {
+        if (performedContext.Exception != null)
+        {
+            var errorTags = HangfireTagBuilder.BuildErrorTags(
+                performedContext.BackgroundJob,
+                performedContext.Exception);
```
thompson-tomo:
```suggestion
            var errorTags = HangfireTagBuilder.BuildErrorTags(performedContext);
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
```diff
@@ -0,0 +1,254 @@
...
+        string workflowState,
+        string? recurringJobId,
+        string? errorType)
+    {
+        var tags = new TagList
+        {
+            GetDefinitionName(backgroundJob),
+            GetPlatformName(),
+            new KeyValuePair<string, object?>(WorkflowAttributes.AttributeWorkflowState, workflowState),
+            GetTriggerType(recurringJobId, isScheduled: workflowState == WorkflowAttributes.WorkflowStateValues.Pending && string.IsNullOrEmpty(recurringJobId)),
```
thompson-tomo:
```suggestion
            GetTriggerType(recurringJobId, workflowState),
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireTagBuilder.cs
```diff
@@ -0,0 +1,254 @@
...
+        if (!string.IsNullOrEmpty(recurringJobId))
+        {
+            return new(WorkflowAttributes.AttributeWorkflowTriggerType, WorkflowAttributes.WorkflowTriggerTypeValues.Cron);
+        }
+
+        // Check if job is scheduled for future execution
+        if (isScheduled)
+        {
+            return new(WorkflowAttributes.AttributeWorkflowTriggerType, WorkflowAttributes.WorkflowTriggerTypeValues.Schedule);
+        }
```
thompson-tomo:
```suggestion
    private static KeyValuePair<string, object?> GetTriggerType(string? recurringJobId, WorkflowAttributes.WorkflowStateValues? workflowState = null)
    {
        // Check if job was triggered by a recurring job (cron)
        if (!string.IsNullOrEmpty(recurringJobId))
        {
            return new(WorkflowAttributes.AttributeWorkflowTriggerType, WorkflowAttributes.WorkflowTriggerTypeValues.Cron);
        }

        // Check if job is scheduled for future execution
        if (workflowState == WorkflowAttributes.WorkflowStateValues.Pending)
        {
            return new(WorkflowAttributes.AttributeWorkflowTriggerType, WorkflowAttributes.WorkflowTriggerTypeValues.Schedule);
        }
```

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
thompson-tomo: Thanks, I noticed the descriptions were not clear and on reflection completed should be removed from the status metric as it is a final state where it can no longer change states. 

A user has the count metric for knowing the qty of an outcome ie failed vs succeeded. On reflection this metric should be renamed to *.result.

In effect status tracks what is happening at an instance in time (value goes up & down) whereas result summarises where it ends up (value only goes up)

I will make this clearer in the docs and I appreciate your feedback in helping to refine the docs.

thompson-tomo: So I have gone and updated the docs to help remove some of the ambguity and hopefully make things clearer. The one major thing of interest is the rename of the *.count metric to *.outcome and the *.outcome attribute renamed to *.result. Sorry about that but i did it as *.result is already a common attribute suffix.

The new link is https://github.com/thompson-tomo/semantic-conventions/pull/146.

src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -7,6 +8,15 @@ OpenTelemetry.Trace.HangfireInstrumentationOptions.HangfireInstrumentationOption
...
 OpenTelemetry.Trace.HangfireInstrumentationOptions.RecordException.set -> void
 OpenTelemetry.Trace.TracerProviderBuilderExtensions
+static OpenTelemetry.Metrics.MeterProviderBuilderExtensions.AddHangfireInstrumentation(this OpenTelemetry.Metrics.MeterProviderBuilder! builder) -> OpenTelemetry.Metrics.MeterProviderBuilder!
+static OpenTelemetry.Metrics.MeterProviderBuilderExtensions.AddHangfireInstrumentation(this OpenTelemetry.Metrics.MeterProviderBuilder! builder, string? name, System.Action<OpenTelemetry.Metrics.HangfireMetricsInstrumentationOptions!>? configure) -> OpenTelemetry.Metrics.MeterProviderBuilder!
+static OpenTelemetry.Metrics.MeterProviderBuilderExtensions.AddHangfireInstrumentation(this OpenTelemetry.Metrics.MeterProviderBuilder! builder, System.Action<OpenTelemetry.Metrics.HangfireMetricsInstrumentationOptions!>? configure) -> OpenTelemetry.Metrics.MeterProviderBuilder!
+OpenTelemetry.Metrics.HangfireMetricsInstrumentationOptions
+OpenTelemetry.Metrics.HangfireMetricsInstrumentationOptions.DisplayNameFunc.get -> System.Func<Hangfire.BackgroundJob!, string!>!
+OpenTelemetry.Metrics.HangfireMetricsInstrumentationOptions.DisplayNameFunc.set -> void
```
thompson-tomo: This should be able to be removed

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetricsStateFilter.cs
gorbach: Done, metrics/attributes renamed and final states are not tracked.

src/OpenTelemetry.Instrumentation.Hangfire/.publicApi/PublicAPI.Unshipped.txt
gorbach: Yep. fixed. And updated readme

thompson-tomo (commented):
Looks good to me & thanks for adapting based on feedback.

thompson-tomo (approved):

fred2u:
Thanks for the contribution, don't forget to update the CHANGELOG.md

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,317 @@ using var tracerProvider = Sdk
...
+- Track only jobs that have entered the execution pipeline (enqueued or later)
+- Use `workflow.task.name` and `workflow.execution.state`
+- Do NOT include `workflow.trigger.type` (trigger is a workflow-level concept)
+- Provide detailed execution performance metrics
+
 ## References
 
 * [OpenTelemetry Project](https://opentelemetry.io/)
 * [Hangfire Project](https://www.hangfire.io/)
+* [OpenTelemetry Workflow Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/workflow/workflow-metrics.md)
```
Kielek: I am curious how you defined metrics in this package. For sure, there is nothing under this link at the time of writhing this comment.

gorbach:
> Thanks for the contribution, don't forget to update the CHANGELOG.md

Done, thanks!

src/OpenTelemetry.Instrumentation.Hangfire/README.md
Kielek: This is still valid, even if you removed the lonk

gorbach: Sorry, I didn't get it. Should I keep the worong link ?

Kielek: No, removing invalid links is fine. I am asking what is the source of the truth for definition for metrics/attributes?
Did you have tried to create semantic convention for them?

fred2u (approved):

src/OpenTelemetry.Instrumentation.Hangfire/README.md
gorbach: I used this document - https://github.com/thompson-tomo/semantic-conventions/blob/feature/%231688_Workflows/docs/workflow/workflow-metrics.md

fred2u:
@Kielek Is there still something blocking this PR?

Kielek:
@fred2u, I am not sure if there is any blocker. The metrics names suggests that all of them came from the opentelemetry-specification, but in fact, all of most of them are custom. Need to check with othet @open-telemetry/dotnet-contrib-approvers what to do with this PR.

For sure, in the same time, you can clean up public contract for spans/traces in the separate PR. It will be great to align to standard conventions from this repository: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2075#issuecomment-3420588992

src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Add metrics instrumentation following workflow semantic conventions.
```
Kielek: It is not true. There is nothing like workflow semantic convention.

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
     .Build();
 ```
 
+## Metrics
+
+This instrumentation library collects metrics following the OpenTelemetry
+[workflow semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/workflow/workflow-metrics.md).
```
Kielek: Also here.

Kielek (commented):
I think that you need put clear information, that you are providing metrics, but all of them are hangfire specific, and it will be subject to change if there occurs any recommendations from the semantic convetion. When the statement will be clear, I think that we should be fine to merge.

Do you have any plan to work towards this definition?

src/OpenTelemetry.Instrumentation.Hangfire/README.md
thompson-tomo: Howabout something like
```suggestion
This instrumentation library collects metrics following a POC/draft definition of workflow metrics defined as part of [semantic-conventions/#1688](https://github.com/open-telemetry/semantic-conventions/issues/1688). 
As those definitions evolve, changes including breaking ones will flow back to this implementation.
```

thompson-tomo (approved):
So the work to get these definitions into semconv is covered by https://github.com/open-telemetry/semantic-conventions/issues/1688 with very recent interest from cicd sig to be involved, so hopefully this topic can now get traction.

I have added a comment on the readme with proposed text to make it clearer the signal definitions are subject to change.

thompson-tomo:
@gorbach could you look at the conflicts?

gorbach:
I've solved conflicts.

Kielek:
@gorbach, please check all errors in the CI.
For C# part it is worth to execute build locally as a Release.

gorbach:
I've fixed all errors in Release locally, but I'm not sure how can check CI.

Kielek:
> I've fixed all errors in Release locally, but I'm not sure how can check CI.

It is your first contribution here, all CI execution needs to be approved. I still see a lot of errors (including comoialtion, markdown, etc).

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+The instrumentation distinguishes between **workflow-level** and **execution-level** metrics:
+
+**Workflow-level metrics** (`workflow.outcome`, `workflow.status`):
+- Track the complete lifecycle of jobs, including pre-execution states (e.g., scheduled)
+- Use `workflow.definition.name`, `workflow.state`, and `workflow.trigger.type`
+- Provide visibility into all jobs regardless of execution status
+
+**Execution-level metrics** (`workflow.execution.*`):
+- Track only jobs that have entered the execution pipeline (enqueued or later)
```
thompson-tomo, line 503:
```suggestion
**Execution-level metrics** (`workflow.execution.*`):

- Track only jobs that have entered the execution pipeline (enqueued or later)
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+  - `workflow.state` tracks complete workflow lifecycle (used in `workflow.status`)
+  - See state mapping tables above for details
+- **error.type**: Set to the full type name of the exception when a job fails
+
+#### Workflow vs Execution Metrics
+
+The instrumentation distinguishes between **workflow-level** and **execution-level** metrics:
+
+**Workflow-level metrics** (`workflow.outcome`, `workflow.status`):
+- Track the complete lifecycle of jobs, including pre-execution states (e.g., scheduled)
```
thompson-tomo:
```suggestion
**Workflow-level metrics** (`workflow.outcome`, `workflow.status`):

- Track the complete lifecycle of jobs, including pre-execution states (e.g., scheduled)
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name     |
+
+¹ Required if and only if the task execution failed.
+
+#### workflow.execution.duration
+
+Duration of an execution grouped by task, state, and result. Records duration for different
+execution phases:
+- **state=pending**: Time spent waiting in queue (only when `RecordQueueLatency` is enabled)
```
thompson-tomo:
```suggestion
execution phases:

- **state=pending**: Time spent waiting in queue (only when `RecordQueueLatency` is enabled)
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+The following metrics are emitted by this instrumentation:
+
+**Execution-level metrics** track jobs that have entered the execution pipeline:
+- `workflow.execution.outcome` - Counter for completed executions
+- `workflow.execution.duration` - Histogram for execution duration (pending and executing phases)
+- `workflow.execution.status` - UpDownCounter for current execution state
+- `workflow.execution.errors` - Counter for execution errors
+
+**Workflow-level metrics** track the complete job lifecycle:
+- `workflow.outcome` - Counter for completed workflows
```
thompson-tomo, line 272:
```suggestion
**Workflow-level metrics** track the complete job lifecycle:

- `workflow.outcome` - Counter for completed workflows
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+> [!NOTE]
+> Add `using OpenTelemetry.Metrics;` to access `HangfireMetricsInstrumentationOptions`.
+
+### Available Metrics
+
+The following metrics are emitted by this instrumentation:
+
+**Execution-level metrics** track jobs that have entered the execution pipeline:
+- `workflow.execution.outcome` - Counter for completed executions
```
thompson-tomo:
```suggestion
**Execution-level metrics** track jobs that have entered the execution pipeline:

- `workflow.execution.outcome` - Counter for completed executions
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
     .Build();
 ```
 
+## Metrics
+
+This instrumentation library collects metrics following a POC/draft definition of workflow metrics defined as part of [semantic-conventions/#1688](https://github.com/open-telemetry/semantic-conventions/issues/1688).
```
thompson-tomo:
```suggestion
This instrumentation library collects metrics following a POC/draft definition of workflow metrics defined as part of [semantic-conventions/#1688](https://github.com/open-telemetry/semantic-conventions/issues/1688).
```

src/OpenTelemetry.Instrumentation.Hangfire/OpenTelemetry.Instrumentation.Hangfire.csproj
```diff
@@ -5,7 +5,6 @@
     <Description>OpenTelemetry Hangfire Instrumentation.</Description>
     <PackageTags>$(PackageTags);Hangfire</PackageTags>
     <MinVerTagPrefix>Instrumentation.Hangfire-</MinVerTagPrefix>
-    <SignAssembly>false</SignAssembly>
```
thompson-tomo, line 8: Revert this removal

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+> is not needed for counting completed executions.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
+| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`   |
+| `workflow.execution.result`   | string | The result of executing the task              | Required                 | `success`, `failure`    |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name     |
```
thompson-tomo:
```suggestion
| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+¹ Required if and only if the task execution failed.
```
thompson-tomo:
```suggestion
[1]: Required if and only if the task execution failed.
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
+| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`   |
+| `workflow.execution.result`   | string | The result of executing the task              | Required                 | `success`, `failure`    |
+| `workflow.execution.state`    | string | The execution phase being measured            | Required                 | `pending`, `executing`  |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name     |
```
thompson-tomo:
```suggestion
| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name     |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+¹ Required if and only if the task execution failed.
```
thompson-tomo:
```suggestion
[1]: Required if and only if the task execution failed.
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+> to a new state, the previous state is decremented and the new state is incremented.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------------------------|
+| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required                 | e.g., `MyJob.Execute`                     |
+| `workflow.execution.state`    | string | Current state of the execution                | Required                 | `pending`, `executing`, `completed`       |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`                                |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name                       |
```
thompson-tomo:
```suggestion
| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name                       |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+¹ Required if and only if the task execution failed.
```
thompson-tomo:
```suggestion
[1]: Required if and only if the task execution failed.
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
+| `workflow.definition.name`    | string | Name of the workflow (Hangfire job method)    | Required                 | e.g., `MyJob.Execute`   |
+| `workflow.result`             | string | The result of the workflow                    | Required                 | `success`, `failure`    |
+| `workflow.trigger.type`       | string | Type of trigger that initiated the workflow   | Required                 | `api`, `cron`, `schedule` |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`              |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name     |
```
thompson-tomo:
```suggestion
| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name     |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+¹ Required if and only if the workflow execution failed.
```
thompson-tomo:
```suggestion
[1]: Required if and only if the workflow execution failed.
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------------------------|
+| `workflow.definition.name`    | string | Name of the workflow (Hangfire job method)    | Required                 | e.g., `MyJob.Execute`                     |
+| `workflow.state`              | string | Current state of the workflow                 | Required                 | `pending`, `executing`, `completed`       |
+| `workflow.trigger.type`       | string | Type of trigger that initiated the workflow   | Required                 | `api`, `cron`, `schedule`                 |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended              | `hangfire`                                |
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required¹  | Exception type name                       |
```
thompson-tomo:
```suggestion
| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name                       |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,6 +164,315 @@ using var tracerProvider = Sdk
...
+
+¹ Required if and only if the workflow execution failed.
```
thompson-tomo:
```suggestion
[1]: Required if and only if the workflow execution failed.
```

test/OpenTelemetry.Instrumentation.Hangfire.Tests/HangfireMetricsTests.cs
```diff
@@ -0,0 +1,536 @@
...
+
+        // Act - Schedule a job for immediate execution (1 second delay)
+        var jobId = BackgroundJob.Schedule<TestJob>(x => x.Execute(), TimeSpan.FromSeconds(1));
+
+        // Wait for job to complete full lifecycle
+        await this.hangfireFixture.WaitJobProcessedAsync(jobId, 10);
+
+        meterProvider.ForceFlush();
+
+        // Assert - workflow.status should track full lifecycle (scheduled → enqueued → executing → completed)
```
thompson-tomo:
```suggestion
        // Assert - workflow.status should track full lifecycle (scheduled -> enqueued -> executing -> completed)
```

test/OpenTelemetry.Instrumentation.Hangfire.Tests/HangfireMetricsTests.cs
```diff
@@ -0,0 +1,536 @@
...
+
+        // Final workflow state - all states should be 0 as job exits status tracking when completed
+        // (Completed state no longer exists in semantic conventions)
+        var workflowExecutingPoint = workflowPoints.FindFirstWithTag(WorkflowAttributes.AttributeWorkflowState, WorkflowAttributes.WorkflowStateValues.Executing);
+        if (workflowExecutingPoint != null)
+        {
+            Assert.Equal(0, workflowExecutingPoint.Value.GetSumLong());
+        }
+
+        // Assert - workflow.execution.status should also track lifecycle (enqueued → executing)
```
thompson-tomo:
```suggestion
        // Assert - workflow.execution.status should also track lifecycle (enqueued -> executing)
```

thompson-tomo (approved):
Bunch of fixes to improve the ci/cd pipeline.

Can you take a look at the markdownlint check and go through splitting up the long lines and the list symbol isdue. Also run dotnet format from the cli.

gorbach:
I beleive I've fixed all linter warnings, compilation errors (in release) and ran `dotnet format` for tests and code.

src/OpenTelemetry.Instrumentation.Hangfire/OpenTelemetry.Instrumentation.Hangfire.csproj
gorbach: I hade to make Hangfire.Tests assembly not signed, and couple for other fixes in proj. files, because Hangfire.Tests references internal fields from Hangfire.

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+  scheduled jobs)
+
+---
+
+#### workflow.execution.outcome
+
+The number of task executions which have been initiated.
+
+| Units          | Instrument Type | Value Type |
+|----------------|-----------------|------------|
```
thompson-tomo:
```suggestion
| -------------- | --------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `{executions}` | Counter         | `Int64`    |
+
+> [!NOTE]
+> This metric does NOT include the `workflow.execution.state` attribute, as
+> state differentiation is not needed for counting completed executions.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+
+Duration of an execution grouped by task, state, and result. Records duration
+for different execution phases:
+
+- **state=pending**: Time spent waiting in queue (only when
+  `RecordQueueLatency` is enabled)
+- **state=executing**: Time spent in actual execution
+
+| Units | Instrument Type | Value Type |
+|-------|-----------------|------------|
```
thompson-tomo:
```suggestion
| ----- | --------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+> [!NOTE]
+> The `workflow.execution.state` attribute is an extension to the current
+> semantic conventions to support measuring different execution phases
+> (pending vs executing). This attribute is being proposed for inclusion in
+> the official semantic conventions.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name                       |
+
+[1]: Required if and only if the task execution failed.
+
+**Hangfire State Mapping:**
+
+Hangfire job states are mapped to workflow semantic convention states as follows:
+
+| Hangfire State                     | Workflow State |
+|------------------------------------|----------------|
```
thompson-tomo:
```suggestion
| ---------------------------------- | -------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
+
+[1]: Required if and only if the task execution failed.
+
+#### workflow.execution.status
+
+The number of actively running tasks grouped by task and the current state.
+
+| Units          | Instrument Type  | Value Type |
+|----------------|------------------|------------|
```
thompson-tomo:
```suggestion
| -------------- | ---------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+
+> [!NOTE]
+> This metric tracks the current state of job executions. When a job
+> transitions to a new state, the previous state is decremented and the new
+> state is incremented.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| Scheduled, Enqueued, Awaiting      | `pending`      |
+| Processing                         | `executing`    |
+| Succeeded, Failed, Deleted         | `completed`    |
+
+#### workflow.execution.errors
+
+The number of errors encountered in task executions.
+
+| Units     | Instrument Type | Value Type |
+|-----------|-----------------|------------|
```
thompson-tomo, line 371:
```suggestion
| --------- | --------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `{error}` | Counter         | `Int64`    |
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level | Values                  |
+|-------------------------------|--------|-----------------------------------------------|-------------------|-------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ----------------- | ----------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `workflow.task.name`          | string | Name of the task (Hangfire job method)        | Required          | e.g., `MyJob.Execute`   |
+| `workflow.platform.name`      | string | The workflow platform being used              | Recommended       | `hangfire`              |
+
+#### workflow.outcome
+
+The number of workflow instances which have been initiated. In Hangfire, this tracks
+individual job completions.
+
+| Units          | Instrument Type | Value Type |
+|----------------|-----------------|------------|
```
thompson-tomo:
```suggestion
| -------------- | --------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1]  | Exception type name     |
+
+[1]: Required if and only if the workflow execution failed.
+
+#### workflow.status
+
+The number of actively running workflows grouped by definition and the current state.
+
+| Units          | Instrument Type  | Value Type |
+|----------------|------------------|------------|
```
thompson-tomo:
```suggestion
| -------------- | ---------------- | ---------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+individual job completions.
+
+| Units          | Instrument Type | Value Type |
+|----------------|-----------------|------------|
+| `{workflows}`  | Counter         | `Int64`    |
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                  |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+> [!NOTE]
+> This metric tracks the workflow lifecycle including jobs that haven't
+> entered the execution pipeline yet (e.g., scheduled jobs waiting for their
+> trigger time). When a workflow transitions to a new state, the previous
+> state is decremented and the new state is incremented.
+
+**Attributes:**
+
+| Attribute                     | Type   | Description                                   | Requirement Level        | Values                                    |
+|-------------------------------|--------|-----------------------------------------------|--------------------------|-------------------------------------------|
```
thompson-tomo:
```suggestion
| ----------------------------- | ------ | --------------------------------------------- | ------------------------ | ----------------------------------------- |
```

src/OpenTelemetry.Instrumentation.Hangfire/README.md
```diff
@@ -160,7 +164,348 @@ using var tracerProvider = Sdk
...
+| `error.type`                  | string | The type of error that occurred               | Conditionally Required[1] | Exception type name                       |
+
+[1]: Required if and only if the workflow execution failed.
+
+**Hangfire State Mapping:**
+
+Hangfire job states are mapped to workflow semantic convention states as follows:
+
+| Hangfire State                     | Workflow State | Workflow Trigger Type           |
+|------------------------------------|----------------|---------------------------------|
```
thompson-tomo:
```suggestion
| ---------------------------------- | -------------- | ------------------------------- |
```

problesal-sacyr:
Hi, when's this hitting NuGet? Thanks!

Kielek:
Someone needs to create https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/new?template=release_request.yml. Probably 1.14.0-beta.2 version.

PR #3160, opened 2025-10-01
[Instrumentation.AspNet] Normalize route template

RassK:
Linked https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1730
Linked https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2580
Linked https://github.com/open-telemetry/semantic-conventions/issues/2616

## Changes

Normalizes route template according to spec.
Replaces controller and action tokens. Area token is already replaced by the framework in the template.

image

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #3338, opened 2025-10-23
[Instrumentation.AspNetCore] Fix http.route for HTTP server spans

RassK:
Follow up to #3160

## Changes

Normalizes route template according to spec.
* Replaces controller, action, area tokens. 
* Removes constraints. 
* Removes unused optional parameters

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -9,7 +9,7 @@
   template for some routing options. Therefore, the HTTP span display name
   has been also changed to a more meaningful value.
   ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
-  
+
 ## 1.13.0
 
 Released 2025-Oct-22
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Fixes an issue where `http.route` attribute reported only a plain route template for some routing options.
+  Therefore, the HTTP span display name has been also changed to a more meaningful value.
+  ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
+  
+
 ## 1.13.0
 
 Released 2025-Oct-22
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -6,8 +6,8 @@
   ([#2822](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2822))
 
 * Fixes an issue where the http.route attribute on HTTP server spans reported
-  only a plain route template for some routing options. Therefore, the HTTP server span
-  display name has also been changed to a more meaningful value.
+  only a plain route template for some routing options. Therefore, the HTTP
+  server span display name has also been changed to a more meaningful value.
   ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
 
 ## 1.13.0
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -5,8 +5,9 @@
 * Add support for .NET 10.0.
   ([#2822](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2822))
 
-* Fixes an issue where `http.route` attribute reported only a plain route template for some routing options.
-  Therefore, the HTTP span display name has been also changed to a more meaningful value.
+* Fixes an issue where `http.route` attribute reported only a plain route
+  template for some routing options. Therefore, the HTTP span display name
+  has been also changed to a more meaningful value.
   ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
   
 
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -5,9 +5,9 @@
 * Add support for .NET 10.0.
   ([#2822](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2822))
 
-* Fixes an issue where `http.route` attribute reported only a plain route
-  template for some routing options. Therefore, the HTTP span display name
-  has been also changed to a more meaningful value.
+* Fixes an issue where the http.route attribute on HTTP server spans reported
+  only a plain route template for some routing options. Therefore, the HTTP server span
+  display name has also been changed to a more meaningful value.
   ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
 
 ## 1.13.0
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
@@ -10,7 +10,6 @@
   has been also changed to a more meaningful value.
   ([#3338](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338))
   
-
 ## 1.13.0
 
 Released 2025-Oct-22
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
@@ -1,0 +1,80 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#if !NETSTANDARD
+
+using System.Diagnostics;
+using System.Text;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Patterns;
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation;
+
+internal static class RouteAttributeHelper
+{
+    public static void SetRouteAttributeTag(this Activity activity, RouteEndpoint endpoint, HttpRequest request)
+    {
+        var routePattern = GetRoutePattern(endpoint.RoutePattern, request.RouteValues);
+
+        if (!string.IsNullOrEmpty(routePattern))
+        {
+            TelemetryHelper.RequestDataHelper.SetActivityDisplayName(activity, request.Method, routePattern);
+            activity.SetTag(SemanticConventions.AttributeHttpRoute, routePattern);
+        }
+    }
+
+    private static string GetRoutePattern(RoutePattern routePattern, RouteValueDictionary routeValues)
+    {
+        var sb = new StringBuilder();
+
+        for (var i = 0; i < routePattern.PathSegments.Count; i++)
+        {
+            var segment = routePattern.PathSegments[i];
+
+            foreach (var part in segment.Parts)
+            {
+                if (part is RoutePatternLiteralPart literalPart)
+                {
+                    sb.Append(literalPart.Content);
+                }
+                else if (part is RoutePatternParameterPart parameterPart)
+                {
+                    switch (parameterPart.Name)
+                    {
+                        case "area":
+                        case "controller":
+                        case "action":
+                            routePattern.RequiredValues.TryGetValue(parameterPart.Name, out var parameterValue);
+                            if (parameterValue != null)
+                            {
+                                sb.Append(parameterValue);
+                                break;
+                            }
+
+                            goto default;
+                        default:
+                            if (routeValues.ContainsKey(parameterPart.Name))
+                            {
+                                sb.Append('{');
+                                sb.Append(parameterPart.Name);
+                                sb.Append('}');
+                            }
+
+                            break;
+                    }
+                }
+            }
+
+            if (i < routePattern.PathSegments.Count - 1)
+            {
+                sb.Append('/');
+            }
+        }
+
+        return sb.ToString();
+    }
+}
+
+#endif
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
@@ -40,9 +40,9 @@ internal static class RouteAttributeHelper
         if (routePattern.PathSegments.Count == 0)
         {
             // RazorPage default route
-            if (routePattern.Defaults.ContainsKey("page"))
+            if (routePattern.Defaults.TryGetValue("page", out var pageValue))
             {
-                return routePattern.Defaults["page"]?.ToString()?.Trim('/')
+                return pageValue?.ToString()?.Trim('/')
                     ?? string.Empty;
             }
 
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
@@ -14,6 +14,16 @@ namespace OpenTelemetry.Instrumentation.AspNetCore.Implementation;
 
 internal static class RouteAttributeHelper
 {
+    public static void AddRouteAttribute(this TagList tags, RouteEndpoint endpoint, HttpRequest request)
+    {
+        var routePattern = GetRoutePattern(endpoint.RoutePattern, request.RouteValues);
+
+        if (!string.IsNullOrEmpty(routePattern))
+        {
+            tags.Add(new KeyValuePair<string, object?>(SemanticConventions.AttributeHttpRoute, routePattern));
+        }
+    }
+
     public static void SetRouteAttributeTag(this Activity activity, RouteEndpoint endpoint, HttpRequest request)
     {
         var routePattern = GetRoutePattern(endpoint.RoutePattern, request.RouteValues);
@@ -29,15 +39,14 @@ internal static class RouteAttributeHelper
     {
         var sb = new StringBuilder();
 
-        for (var i = 0; i < routePattern.PathSegments.Count; i++)
+        foreach (var segment in routePattern.PathSegments)
         {
-            var segment = routePattern.PathSegments[i];
-
             foreach (var part in segment.Parts)
             {
                 if (part is RoutePatternLiteralPart literalPart)
                 {
                     sb.Append(literalPart.Content);
+                    sb.Append('/');
                 }
                 else if (part is RoutePatternParameterPart parameterPart)
                 {
@@ -50,30 +59,29 @@ internal static class RouteAttributeHelper
                             if (parameterValue != null)
                             {
                                 sb.Append(parameterValue);
+                                sb.Append('/');
                                 break;
                             }
 
                             goto default;
                         default:
-                            if (routeValues.ContainsKey(parameterPart.Name))
+                            if (!parameterPart.IsOptional ||
+                                (parameterPart.IsOptional && routeValues.ContainsKey(parameterPart.Name)))
                             {
                                 sb.Append('{');
                                 sb.Append(parameterPart.Name);
                                 sb.Append('}');
+                                sb.Append('/');
                             }
 
                             break;
                     }
                 }
             }
-
-            if (i < routePattern.PathSegments.Count - 1)
-            {
-                sb.Append('/');
-            }
         }
 
-        return sb.ToString();
+        // Remove the trailing '/'
+        return sb.ToString(0, sb.Length - 1);
     }
 }
 
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
@@ -37,6 +37,18 @@ internal static class RouteAttributeHelper
 
     private static string GetRoutePattern(RoutePattern routePattern, RouteValueDictionary routeValues)
     {
+        if (routePattern.PathSegments.Count == 0)
+        {
+            // RazorPage default route
+            if (routePattern.Defaults.ContainsKey("page"))
+            {
+                return routePattern.Defaults["page"]?.ToString()?.Trim('/')
+                    ?? string.Empty;
+            }
+
+            return string.Empty;
+        }
+
         var sb = new StringBuilder();
 
         foreach (var segment in routePattern.PathSegments)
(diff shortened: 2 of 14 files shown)


src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
```diff
@@ -37,6 +37,18 @@ public static void SetRouteAttributeTag(this Activity activity, RouteEndpoint en
 
     private static string GetRoutePattern(RoutePattern routePattern, RouteValueDictionary routeValues)
     {
+        if (routePattern.PathSegments.Count == 0)
+        {
+            // RazorPage default route
+            if (routePattern.Defaults.ContainsKey("page"))
+            {
+                return routePattern.Defaults["page"]?.ToString()?.Trim('/')
```
martincostello: You can use `TryGetValue()` to avoid two operations for `"page"`.

RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338/commits/b46f36e91b0b0353f2532c5a57ce357f339398ed

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestResult.cs
```diff
@@ -22,6 +22,8 @@ protected RoutingTestResult(TestCase testCase, string anchorPrefix)
 
     public string? IdealHttpRoute { get; set; }
 
+    public bool IdeaHttpRouteMatch { get; set; }
```
martincostello:
```suggestion
    public bool IdealHttpRouteMatch { get; set; }
```

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+                else if (part is RoutePatternParameterPart parameterPart)
+                {
+                    switch (parameterPart.Name)
+                    {
+                        case "area":
+                        case "controller":
+                        case "action":
+                            routePattern.RequiredValues.TryGetValue(parameterPart.Name, out var parameterValue);
+                            if (parameterValue != null)
```
martincostello:
```suggestion
                            if (routePattern.RequiredValues.TryGetValue(parameterPart.Name, out var parameterValue) && parameterValue != null)
```

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestResult.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338/commits/ed8680ac5ca196eb820163d471ce2ac1aea81fa3

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
RassK: `parameterValue != null` does not help much, since `.ToString()` can return null. So all of that can be shortened by one `?`.

martincostello: What I was getting at more is that you're not checking the result of `TryGetValue()`.

Either you should (which I think is best), or you should explicitly discard it to `_` - otherwise it looks like a bug.

RassK: ah sorry, wrong line, my bad.

RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338/commits/5361b9fa7ebc997fbb89623dd19c5f6330ebaf43

RassK:
I'll recheck http route tests readme and update the changelog. So after that, this change should be ready to go.

martincostello (approved):

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net10.0.md
```diff
@@ -38,7 +38,7 @@
 | :broken_heart: | ConventionalRouting | [Non-default action with route parameter and query string](#metrics__conventionalrouting-non-default-action-with-route-parameter-and-query-string) |
 | :broken_heart: | ConventionalRouting | [Non-default action with query string](#metrics__conventionalrouting-non-default-action-with-query-string) |
 | :green_heart: | ConventionalRouting | [Not Found (404)](#metrics__conventionalrouting-not-found-404) |
-| :green_heart: | ConventionalRouting | [Route template with parameter constraint](#metrics__conventionalrouting-route-template-with-parameter-constraint) |
+| :broken_heart: | ConventionalRouting | [Route template with parameter constraint](#metrics__conventionalrouting-route-template-with-parameter-constraint) |
```
Kielek, line 41: There is some green_heart to broken_heart changes across all net targets. Is it intentional?

RassK: yes, you can recheck some ideal route changes. It's caused because of the different ideal route (cleaned constraints, excess symbols like '/', removing optionals if not present, etc)

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 * Add support for .NET 10.0.
   ([#2822](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/2822))
 
+* Fixes an issue where `http.route` attribute reported only a plain route
```
Kielek: `http.route` for spans? I think you are not touching metrics for most cases.

RassK: true, needs some rewording, so that only spans were affected.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/README.net10.0.md
alanwest: We shouldn't introduce more broken_hearts. If you're making a behavior change that you think is correct then the test harness should be adjusted to produce a green_heart. The main thing these markdown files were meant to highlight is remaining gaps we should aim to get fixed.

For code that we do not control, my hope was that we could push on the ASP.NET team to consider making changes.

Also behavior must remain identical (even if the behavior is not ideal) between the attributes added to metrics and those added to spans. Is this still true with this PR?

RassK: There is no way that hearts will remain the same, when the expectation was also wrong.
New expectations are in line with AspNet package, spec and consistent output.

There is also no way that spans and metrics can have identical behaviour. Since one part is what we control and another is uncontrollable. I don't know what is more ideal, either ASP.NET Core has to adapt faster or SDK apis are extended for packages around API to be able to override the src behaviour.

src/OpenTelemetry.Instrumentation.AspNetCore/CHANGELOG.md
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338/commits/49b93a12cb79761a0c6814dd5897fe630a1a6d2e

Kielek:
I have some conversation with @alanwest last week. Some notes:
1. He would like to avoid differences in the `http.route` related parameters between metrics and the traces. The main reason behind this, is the fact that some backends are using it to correlate these signals based on this attribute.
2. @JamesNK, is there any chance that you can advice how to fix `http.route` spans for metrics? Do you see any options to modify these attributes on OTel instrumentation side? If not, can you apply changes directly in the AspNetCore?
3. Not directly related, but AspNetCore is creating activities, but without needed attributes, so we here recreating such activities in this package. Again, @JamesNK, do you see any option to fully manage these activities+attributes by the AspNetCore directly?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
```diff
@@ -0,0 +1,99 @@
...
+        var routePattern = GetRoutePattern(endpoint.RoutePattern, request.RouteValues);
+
+        if (!string.IsNullOrEmpty(routePattern))
+        {
+            TelemetryHelper.RequestDataHelper.SetActivityDisplayName(activity, request.Method, routePattern);
+            activity.SetTag(SemanticConventions.AttributeHttpRoute, routePattern);
+        }
+    }
+
+    private static string GetRoutePattern(RoutePattern routePattern, RouteValueDictionary routeValues)
```
JamesNK, line 38: You need a very large suite of unit tests to ensure every possible route is correctly converted into a string.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
```diff
@@ -0,0 +1,99 @@
...
+                {
+                    switch (parameterPart.Name)
+                    {
+                        case "area":
+                        case "controller":
+                        case "action":
+                            if (routePattern.RequiredValues.TryGetValue(parameterPart.Name, out var parameterValue) && parameterValue != null)
+                            {
+                                sb.Append(parameterValue);
+                                sb.Append('/');
```
JamesNK, line 73: Why not look for separator part for adding separator?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
```diff
@@ -0,0 +1,99 @@
...
+                                break;
+                            }
+
+                            goto default;
+                        default:
+                            if (!parameterPart.IsOptional ||
+                                (parameterPart.IsOptional && routeValues.ContainsKey(parameterPart.Name)))
+                            {
+                                sb.Append('{');
+                                sb.Append(parameterPart.Name);
```
JamesNK, line 83: What happens if there are default values? Or it is catch all? Or has constraints? Or has optional path extension.

There are also complex segments. Are those being handled correctly? See https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-9.0#complex-segments

Maybe these scenarios are already handled correctly, but unit tests to verify are important.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
```diff
@@ -59,26 +65,29 @@
...
     "name": "Area using `area:exists`, non-default action",
     "testApplicationScenario": "ConventionalRouting",
     "httpMethod": "GET",
     "path": "/MyArea/ControllerForMyArea/NonDefault",
     "expectedStatusCode": 200,
-    "currentHttpRoute": "{area:exists}/{controller=ControllerForMyArea}/{action=Default}/{id?}",
-    "expectedHttpRoute": "{area:exists}/ControllerForMyArea/NonDefault/{id?}"
+    "currentHttpRoute": null,
+    "currentMetricRoute": "{area:exists}/{controller=ControllerForMyArea}/{action=Default}/{id?}",
+    "expectedHttpRoute": "MyArea/ControllerForMyArea/NonDefault"
```
JamesNK, line 80: Aren't you just modifying `area`, `controller` and `page`? Isn't the expected route `"MyArea/ControllerForMyArea/NonDefault/{id?}"`. Or is it removed because it's optional?

I don't think removing it has any value. It's still a dynamic placeholder in the route and should be present according to the spec.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
JamesNK, line 38: Unrelated, but I think an empty string route should resolve to `/`. An empty string by itself is confusing. We're already doing that with `http.route` tag in metrics based on customer feedback.

JamesNK:
Metric tags are accessible via https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.features.ihttpmetricstagsfeature?view=aspnetcore-9.0 on HttpContext. You could get and modify `http.route` there.

We'll do work to HTTP semantic convention activity attributes in .NET 11.

If what you're doing here is the correct way to represent `http.route` then we can copy this into ASP.NET Core.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
```diff
@@ -199,7 +219,8 @@
     "path": "/MinimalApiUsingMapGroup",
     "expectedStatusCode": 200,
     "currentHttpRoute": null,
-    "expectedHttpRoute": "/MinimalApiUsingMapGroup/"
+    "currentMetricRoute": "/MinimalApiUsingMapGroup/",
+    "expectedHttpRoute": "MinimalApiUsingMapGroup"
```
JamesNK, line 223: Why remove preceeding `/`?

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
JamesNK, line 38: This always runs, right? Should it only run if the route needs to change? i.e. there are area/controller/action/page attributes and a value is present in the requests route values collection.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
```diff
@@ -42,7 +46,8 @@
     "path": "/SomePath/SomeString/2",
     "expectedStatusCode": 200,
     "currentHttpRoute": null,
-    "expectedHttpRoute": "SomePath/{id}/{num:int}"
+    "currentMetricRoute": "SomePath/{id}/{num:int}",
+    "expectedHttpRoute": "SomePath/{id}/{num}"
```
JamesNK, line 50: The route constraint is being lost

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
```diff
@@ -189,7 +208,8 @@
     "path": "/MinimalApi/123",
     "expectedStatusCode": 200,
     "currentHttpRoute": null,
-    "expectedHttpRoute": "/MinimalApi/{id}"
+    "currentMetricRoute": "/MinimalApi/{id}",
+    "expectedHttpRoute": "MinimalApi/{id}"
```
JamesNK, line 212: What happens if a minimal api route has a parameter called `action`? Will it be rewritten? There is nothing stopping a minimal API having a parameter with that name and it isn't a static part of the route.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
RassK: This is intentional change:

pros:
- produces similar routes as asp.net mvc5
- reduces cardinality
- simplifies the template

negs:
- no visibility into same routes but different constraints (possibly very rare case)

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
RassK: There are 2 ways, either to include `/` always or not to include. The same route must not produce 2 different versions.
So if there is no extra meaning (and a need), then there is no point to include it. Same in ASP.NET MVC5.

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
RassK: currently yes, unless there is a framework marker so we can check if it was included by the framework.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
RassK: Correct. It would be a nice optimization. Also connected to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338#discussion_r2529243454 (so it runs only if the framework provides the values).

RassK: I'm wondering if it should be removed completely if it's empty (empty means no route found), since:

> [5] http.route: MUST NOT be populated when this is not supported by the HTTP server framework as the route attribute should have low-cardinality and the URI path can NOT substitute it. SHOULD include the [application root](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-definitions) if there is one.

src: https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/

test/OpenTelemetry.Instrumentation.AspNetCore.Tests/RouteTests/RoutingTestCases.json
RassK: `?` is a constraint, so connected to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3338#discussion_r2559674266

for `?` it produces 2 perfectly valid routes if the route is using optional path arg and supports both cases (with and without arg).

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
RassK: I agree, that there are probably loads of assumptions (if a general case is working).

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
RassK: Unfortunately the test infra is doing a loads of assumptions. In that case it needs to be rewritten so `HttpContext` is correctly populated by the framework and not by the assumptions (doing manual construction).

RassK:
My opinion about these topics:

> * He would like to avoid differences in the `http.route` related parameters between metrics and the traces. The main reason behind this, is the fact that some backends are using it to correlate these signals based on this attribute.

That is already so broken and off the spec that I'm not sure which parts could be trusted for machine processing.
There is practically no cardinality at all for some apps.

> * @JamesNK, is there any chance that you can advice how to fix `http.route` spans for metrics? Do you see any options to modify these attributes on OTel instrumentation side? If not, can you apply changes directly in the AspNetCore?

Since SDK is exporting them, then additional processor should be available for API users as well as SDK users already have that option.

> * Not directly related, but AspNetCore is creating activities, but without needed attributes, so we here recreating such activities in this package. Again, @JamesNK, do you see any option to fully manage these activities+attributes by the AspNetCore directly?

Same as above.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/RouteAttributeHelper.cs
RassK: @JamesNK there are no separator parts although routePattern is `"{controller=ConventionalRoute}/{action=Default}/{id?}"`. Only 3 times `RoutePatternParameterPart`.

According to the doc `RoutePatternSeparatorPart` is for something else.

JamesNK:
I’m away. I’ll reply next week

JamesNK:
I updated OTEL `http.route` usage in ASP.NET Core to replace static parameters. It's currently added to metrics, and will be used with Activity tags [soon](https://github.com/dotnet/aspnetcore/pull/64851).

The PR to created improved routes is here: https://github.com/dotnet/aspnetcore/pull/64854

I think the approach I used (using RequiredValues collection) is better. The rest of the route unchanged except the replaced parameter. And it doesn't rely on hardcoded parameter names, e.g. it won't try to replace a "controller" parameter in a minimal API route.

RassK:
I know the spec is quite weak about framework specifics in the `http.route`. The goal is only to get a "statistically interesting unique collection" of routes. 

I would personally hate to see "unnecessary paddings" like `:int`, `?`, `{id?}` when id is missing. I'm not sure that they give much value contextually as statistically interesting tokens. ASP.NET Fx (like mvc4) does not do those, so not adding them would unify values further. 

src: https://github.com/open-telemetry/opentelemetry-specification/blob/v1.50.0/specification/trace/api.md#span
> The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable.

Secondly those tokens should be present then also in the Activity DisplayName, which makes it even more, not making sense.
Also because `The span name SHOULD be the most general string` would not be true.

Thirdly, those tokens can add cardinality is some cases.

src: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
`HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) target available. If there is no (low-cardinality) {target} available, HTTP span names SHOULD be {method}.`

CC: @JamesNK @alanwest @martincostello @Kielek

JamesNK:
> I would personally hate to see "unnecessary paddings" like `:int`, `?`, `{id?}` when id is missing. I'm not sure that they give much value contextually as statistically interesting tokens. ASP.NET Fx (like mvc4) does not do those, so not adding them would unify values further.

People will expect `http.route` to be the route they've definied in their app. I'm ok with modifying [conventional routes](https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-10.0#crd6) with constant values because in that case you're defining a route convention that applies to many actions. But rewriting the route that is in an attribute or in MapGet violates the principal of least surprise. People expect `http.route` == `[HttpGet("my-route")`] or `MapGet("my-route", () => {})`.

And things like constraints are important information. This is possible:

```cs
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet("/product/{id:alpha}", (string id) => "alpha");
app.MapGet("/product/{id:int}", (string id) => "int");

app.Run();
```

Stripping the constraint would make the matched route ambigious.

> > The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable.

Constraints are important information. And optional parameters are part of the route, even if they're not matched.

Even if we wanted to optionally add optional parameters to the route string, it would mean a performance cost because it couldn't be statically determined. [The change I made to ASP.NET Core](https://github.com/dotnet/aspnetcore/pull/64854) means that the final route template is generated [when the endpoint is created](https://github.com/dotnet/aspnetcore/blob/1e1e2f767e823a397fb66be4d1502c2e8e3f83e3/src/Http/Routing/src/RouteEndpointBuilder.cs#L111). There is no per-request overhead.

RassK:
This basically means that spec must separate HTTP span name from including `http.route`. I do agree it's important information (as of `http.route`) but it also clutters HTTP span name. This way there is no reason to replace "static partials" in the `http.route`, but HTTP span name for sure requires it.

martincostello:
@RassK Should we close this PR? It's been open for over 6 months with zero forward progress.

RassK:
> @RassK Should we close this PR? It's been open for over 6 months with zero forward progress.

Yes, I can update it if there is any progress. So far seems it's stuck in spec and even if there is going to be any progress, then most probably it needs native support to avoid doing performance hitting work in hot paths.

PR #3347, opened 2025-10-28
[Resources.Process] implement latest semantic convention spec

### LINKED ISSUE #2892: [feature request] update processruntime resource detectors

thompson-tomo:
Fixes #2892 
Design discussion issue #

## Changes

This adds in additional recommend attributes to describe process. Implementation has been based on latest spec https://github.com/open-telemetry/semantic-conventions/blob/main/docs/resource/process.md


src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +14,26 @@ internal sealed class ProcessDetector : IResourceDetector
...
+        using var currentProcess = System.Diagnostics.Process.GetCurrentProcess();
+        var attributes = new List<KeyValuePair<string, object>>(9)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
+            new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
+            new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString("O") ?? string.Empty),
+            new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
+            new(ProcessSemanticConventions.AttributeProcessWorkingDir, Environment.CurrentDirectory),
+
+            new(ProcessSemanticConventions.AttributeProcessExecName, currentProcess.ProcessName),
```
matt-hensley: Some process related properties may throw exceptions. For example, [StartTime throws 3 exceptions](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.starttime), and there is likely no way to detect errors like `Win32Exception` ahead of time.

thompson-tomo: Looking at the exceptions the only one we are interested in is win32 so have added a try catch and also exited earlier if process has exited.

The others are for scenarios we don't support ie windows 98.

The current directory does throw a number of exceptions but they are for the set method which is not used.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,40 @@ internal sealed class ProcessDetector : IResourceDetector
...
         {
-            using var process = System.Diagnostics.Process.GetCurrentProcess();
-            return process.Id;
+            attributes.Add(new(ProcessSemanticConventions.AttributeProcessStartTime, currentProcess.StartTime.ToString("O") ?? DateTime.Now.ToString("O")));
         }
-#endif
+        catch (Win32Exception)
+        {
+            attributes.Add(new(ProcessSemanticConventions.AttributeProcessStartTime, DateTime.Now.ToString("O")));
+        }
```
Kielek: Fallbacks here are not true IMO. If you are not detect attributes, just omit it.

thompson-tomo: I thought about that but in the case here creation time is an identifying attribute and all identifying should actually be required hence https://github.com/open-telemetry/weaver/issues/986 to address it.

Also Using The current time enables the functionality of distinguishing between telemetry from process starts, leaving it out you lose that.

Kielek: Lack of the data is better than the wrong data.

thompson-tomo: Are you saying that all the process entity attributes should be omitted if the creation time can not be set given it is an identifying attribute?

For me the exact value of the atrribute is not as meaningful as having a value there which can then be used to achieve the use case.

Kielek: I am not against keeping part of the attributes. Part of data seems to be fine. I am against putting obviously wrong data in the attributes/any other signal.

In this case, add this attribute conditionally if it is available. You can document it properly in the README. Small deviations from the semantics should be allowed if we have technical blockers for implementation.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,39 @@ internal sealed class ProcessDetector : IResourceDetector
...
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
+            new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
+            new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
+            new(ProcessSemanticConventions.AttributeProcessWorkingDir, Environment.CurrentDirectory),
+
+            new(ProcessSemanticConventions.AttributeProcessExecName, currentProcess.ProcessName),
+            new(ProcessSemanticConventions.AttributeProcessInteractive, Environment.UserInteractive),
 #if NET
             new(ProcessSemanticConventions.AttributeProcessPid, Environment.ProcessId),
```
Kielek: You already have `currentProcess.Id` here. Is there any reason to go through Environment and potentially fetching process data from the scratch?

What is more, there will be common scenario both for .NET/non .NET

src/OpenTelemetry.Resources.Process/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Extended `ProcessDetector` to include `process.executable.path`,
+`process.working_directory`, `process.args_count`,`process.creation.time`,
+`process.executable.name`, `process.interactive` as well as `process.title`
+as per the latest OpenTelemetry Semantic Convention resource/entity definition.
+ ([#2971](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3347))
```
Kielek:
```suggestion
* Extended `ProcessDetector` to include `process.executable.path`,
  `process.working_directory`, `process.args_count`,`process.creation.time`,
  `process.executable.name`, `process.interactive` as well as `process.title`
  as per the latest OpenTelemetry Semantic Convention resource definition.
  ([#3347](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3347))
```

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,39 @@ internal sealed class ProcessDetector : IResourceDetector
...
+        if (currentProcess.HasExited)
+        {
+            return Resource.Empty;
+        }
+
+        var attributes = new List<KeyValuePair<string, object>>(9)
         {
             new(ProcessSemanticConventions.AttributeProcessOwner, Environment.UserName),
+            new(ProcessSemanticConventions.AttributeProcessArgsCount, Environment.GetCommandLineArgs().Length),
+            new(ProcessSemanticConventions.AttributeProcessTitle, currentProcess.MainWindowTitle),
```
Kielek: I am not sure if it should be used here. Main windows title is not the process title. Check https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.mainwindowtitle?view=net-10.0

It is recommended attribute, so it can be omitted in this PR.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,39 @@ internal sealed class ProcessDetector : IResourceDetector
...
+            new(ProcessSemanticConventions.AttributeProcessWorkingDir, Environment.CurrentDirectory),
```
Kielek, line 35: Based what on I see is not reliable and might change in time.
alternative approach `currentProcess.StartInfo.WorkingDirectory` is not working while I was testing.

As it is conditionally required and we have AttributeProcessArgsCount, I would drop this attribute.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,39 @@ internal sealed class ProcessDetector : IResourceDetector
...
+
+            new(ProcessSemanticConventions.AttributeProcessExecName, currentProcess.ProcessName),
```
Kielek: Are you sure that it met SHOULD criteria from OTel sem. conv?
The name of the process executable. On Linux based systems, this SHOULD be set to the base name of the target of /proc/[pid]/exe. On Windows, this SHOULD be set to the base name of GetProcessImageFileNameW.

If not, it is conditionally required, and we have non-controversial in place AttributeProcessArgsCount. Potentially, it can be removed from this PR.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -14,20 +16,39 @@ internal sealed class ProcessDetector : IResourceDetector
...
+            new(ProcessSemanticConventions.AttributeProcessInteractive, Environment.UserInteractive),
```
Kielek: Are you sure that
https://learn.microsoft.com/en-us/dotnet/api/system.environment.userinteractive?view=net-10.0
is compliant with sem conf description: `Whether the process is connected to an interactive shell.`?

It can be omitted, as it is recommended attribute.

src/OpenTelemetry.Resources.Process/README.md
```diff
@@ -53,7 +53,9 @@ using var loggerFactory = LoggerFactory.Create(builder =>
 The resource detectors will record the following metadata based on where
 your application is running:
 
-- **ProcessDetector**: `process.owner`, `process.pid`.
+- **ProcessDetector**: `process.owner`, `process.pid`, `process.executable.path`,
+`process.working_directory`, `process.args_count`,`process.creation.time`,
+`process.executable.name`, `process.interactive`, `process.title`.
```
Kielek, line 58: lack of info that process.executable.name is avaialbe only on .NET

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
thompson-tomo: Have updated it to use the main module name.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
thompson-tomo: It works for me in my testing. Resource attributes are allowed to change, they just need to be a descriptive attribute which this is. The description of the property mentions that this is the working directory.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
thompson-tomo: Good spot have corrected how this is sourced.

src/OpenTelemetry.Resources.Process/ProcessDetector.cs
thompson-tomo: Yes but to further increase the checks, I now check if stdio is connected.

PR #3367, opened 2025-10-30
Add WantedResourceAttributes option

mattsains:
## Changes

Allows users of the Geneva exporter to specify which resource attributes they want sent to Geneva. This prevents excessive trace data from being exported if there are many resource attributes which do not need to be stored.


rajkumar-rangaraj:
I strongly feel this capability does not belong at the exporter level.
Filtering / selection of resource attributes should happen when the Resource is built or in the Collector pipeline, not inside the Geneva exporter.

In the current SDK, customers already have a clean way to control which resource attributes get emitted by using:

```
ResourceBuilder.CreateEmpty()
    .AddAttributes(...)
```

This supports the model where the application defines the resource contract and only includes what it wants to propagate.
Adding an allow-list at the exporter layer introduces:
* Two places for resource shaping logic (builder + exporter)
* Divergence from upstream OTel behavior
* A risk of exporter-specific resource semantics

PR #3368, opened 2025-10-31
Fix identifier parsing and improve FROM clause parsing

### LINKED ISSUE #3351: [bug] EntityFrameworkCore Instrumentation hangs when tracing certain queries with umlaut in the variable names
### Component

OpenTelemetry.Instrumentation.EntityFrameworkCore

### Package Version

| OpenTelemetry                         | 1.13.1   |
| OpenTelemetry.Exporters.Console       | 1.13.1   |
| OpenTelemetry.Instrumentation.EntityFrameworkCore | 1.13.0-beta.1   |

### Runtime Version

net8.0

### Description

Certain queries with German umlaut like ä, ö, ü in the variable name(s) used in the query cause the tracing to hang (before exporting trace data). 

The problem only exists with the above mentioned version of `OpenTelemetry.Instrumentation.EntityFrameworkCore  1.13.0-beta.1`.
The previous version `1.12.0-beta.2` works fine.

Example of a query where tracing hangs:

```csharp
var geräteIds = new List<long>() { 1, 2, 3 };
var x = db.Tests.AsNoTracking().Where(g => geräteIds.Contains(g.Id));
```

Example of the query string of that query:

```sql
DECLARE @__geräteIds_0 nvarchar(4000) = N'[1,2,3]';

SELECT [t].[Id]
FROM [Tests] AS [t]
WHERE [t].[Id] IN (
    SELECT [g].[value]
    FROM OPENJSON(@__geräteIds_0) WITH ([value] bigint '$') AS [g]
)
```

### Steps to Reproduce

- Have a Microsoft SQL Server 2022 instance with a db `tests` and its connection string ready
- have a table `Tests` with a primary key column ID (bigint)
- the collation of the db tested was `Latin1_General_CI_AS` but it does not seem to make a difference
- create a console application
- install the above mentioned nuget packages
- fill in your actual connection string
- run the application

```csharp
using Microsoft.EntityFrameworkCore;
using OpenTelemetry;
using OpenTelemetry.Trace;

   .AddEntityFrameworkCoreInstrumentation()
   .AddConsoleExporter()
   .Build();

using var db = new UmlautDbContext();

var geräteIds = new List<long>() { 1, 2, 3 };
var x = db.Tests.AsNoTracking().Where(g => geräteIds.Contains(g.Id));

Console.WriteLine(x.ToQueryString()); // does not influence the result whether it hangs or not

foreach (var item in x) // hangs here
{
    Console.WriteLine(item.Id);
}

public class TestEntity
{
    public long Id { get; set; }
}

public class UmlautDbContext : DbContext
{
    public DbSet<TestEntity> Tests { get; set; }
    protected override void OnConfiguring(DbContextOptionsBuilder options) => options.UseSqlServer("Server=...;");
}
```

Docker compose for the required SQL Server

```yaml
services:
  sqlserver:
    image: mcr.microsoft.com/mssql/server:2022-latest
    ports:
      - 1433:1433
    environment:
      - ACCEPT_EULA=Y
      - MSSQL_SA_PASSWORD=ProvideASecurePasswordHere!
      - MSSQL_COLLATION=Latin1_General_CI_AS
```

### Expected Result

It should print the existing item in the Tests table (if any) and not hang.

### Actual Result

It hangs as soon as data is accessed.

### Additional Context

_No response_

### Tip

stevejgordon:
Fixes #3351

## Changes

Primarily this fixes #3351 by ensuring we check all valid letters when parsing identifiers.

Also adds support for delimited table identifiers.

This also makes the handling of comma separated tables more robust by tracking the exit of a FROM clause.

### DIFF
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -468,17 +468,16 @@ internal static class SqlProcessor
             }
 
             var reservedKeywordSpan = reservedKeyword.AsSpan();
-            var isMatch = true;
+
             for (var charPos = 0; charPos < tokenSpan.Length; charPos++)
             {
                 if ((tokenSpan[charPos] | 0x20) != (reservedKeywordSpan[charPos] | 0x20))
                 {
-                    isMatch = false;
-                    break;
+                    return false;
                 }
             }
 
-            return isMatch;
+            return true;
         }
     }
 
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -762,7 +762,7 @@ internal static class SqlProcessor
 
         /// <summary>
         /// Will be set if a keyword has been matched by the parser.
-        /// Not all keywords are neccessarily matched.
+        /// Not all keywords are necessarily matched.
         /// </summary>
         public SqlKeywordInfo? PreviousParsedKeyword; // 8 bytes (reference type)
 
--- a/src/Shared/SqlProcessor.cs
+++ b/src/Shared/SqlProcessor.cs
@@ -15,6 +15,7 @@ internal static class SqlProcessor
     private const char SanitizationPlaceholder = '?';
     private const char SpaceChar = ' ';
     private const char CommaChar = ',';
+    private const char OpenSquareBracketChar = '[';
     private const char OpenParenChar = '(';
     private const char CloseParenChar = ')';
     private const char DashChar = '-';
@@ -34,6 +35,12 @@ internal static class SqlProcessor
 #if NET
     private static readonly SearchValues<char> WhitespaceSearchValues = SearchValues.Create(WhitespaceChars);
 #endif
+
+    // This is not an exhaustive list but covers the majority of common reserved SQL keywords that may follow a FROM clause.
+    // This is used when determining if the previous token is a keyword in order to identify the end of a comma separated FROM clause.
+    private static readonly string[] FromClauseReservedKeywords = [
+        "AS", "JOIN", "CROSS", "WHERE", "BY", "HAVING", "WINDOW", "LIMIT", "OFFSET", "TABLESAMPLE", "WITH", "PIVOT", "UNPIVOT"
+    ];
 
     // We can extend this in the future to include more keywords if needed.
     // The keywords should be ordered by frequency of use to optimize performance.
@@ -148,9 +155,9 @@ internal static class SqlProcessor
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static bool IsAsciiIdentifierChar(char c) =>
 #if NET
-        char.IsAsciiLetter(c) || char.IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
+        char.IsLetter(c) || char.IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
 #else
-        IsAsciiLetter(c) || IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
+        char.IsLetter(c) || IsAsciiDigit(c) || c == UnderscoreChar || c == DotChar;
 #endif
 
     private static SqlStatementInfo SanitizeSql(ReadOnlySpan<char> sql)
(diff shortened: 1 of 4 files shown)


src/Shared/SqlProcessor.cs
```diff
@@ -35,6 +36,12 @@ internal static class SqlProcessor
     private static readonly SearchValues<char> WhitespaceSearchValues = SearchValues.Create(WhitespaceChars);
 #endif
 
+    // This is not an exhaustive list but covers the majority of common reserved SQL keywords that may follow a FROM clause.
+    // This is used when determining if the previous token is a keyword in order to identify the end of a comma separated FROM clause.
+    private static readonly string[] FromClauseReservedKeywords = [
+        "AS", "JOIN", "CROSS", "WHERE", "BY", "HAVING", "WINDOW", "LIMIT", "OFFSET", "TABLESAMPLE", "WITH", "PIVOT", "UNPIVOT"
```
martincostello: Nit: can we sort these? Makes it easier to see if something needs adding/is missing in the future.

src/Shared/SqlProcessor.cs
```diff
@@ -401,14 +409,42 @@ private static void ParseNextToken(
...
+                                    break;
+                                }
+                            }
+
+                            if (isMatch)
+                            {
+                                state.InFromClause = false;
+                            }
+                        }
+                    }
```
martincostello, line 427: For .NET 10, we could do this with `SearchValues<T>`, right?

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -857,5 +857,31 @@
       ],
       "db.query.summary": "ALTER TABLE MyTable"
     }
+  },
+  {
+    "name": "query_with_german_characters",
```
martincostello:
```suggestion
    "name": "query_with_diacritic_characters",
```

src/Shared/SqlProcessor.cs
stevejgordon: I've sorted in order of most likely to appear next after `FROM` to subtly optimise the lookup.

src/Shared/SqlProcessor.cs
stevejgordon: Yeah we probably can. It would be a micro optimisation for this case. I can either do that in a follow up PR (so as to get the main bug fix out sooner) or some time next week I can update it in this PR.

src/Shared/SqlProcessor.cs
martincostello: OK, in that case can you just add a comment to that effect?

src/Shared/SqlProcessor.cs
martincostello: Later is fine, as #2822 hasn't been merged yet anyway.

I thought I'd just mention it in case you'd rework this in any way to make that easier later, and I think the code would be easier to read (e.g. not needing to understand `0x20` is being used to make it case insensitive).

src/Shared/SqlProcessor.cs
stevejgordon: I've added a comment.

src/Shared/SqlProcessor.cs
stevejgordon: I've refactored slightly to make this code a little more self descriptive.

src/Shared/SqlProcessor.cs
```diff
@@ -469,6 +459,27 @@ private static void ParseNextToken(
...
+            for (var charPos = 0; charPos < tokenSpan.Length; charPos++)
+            {
+                if ((tokenSpan[charPos] | 0x20) != (reservedKeywordSpan[charPos] | 0x20))
+                {
+                    isMatch = false;
+                    break;
+                }
+            }
+
+            return isMatch;
```
martincostello: Would it make any difference to just `return false` instead of assign to `isMatch`, and just `return true` at the end?

martincostello (approved):

src/Shared/SqlProcessor.cs
stevejgordon: Yes it would, I've updated.

martincostello:
@stevejgordon Would you mind adding a CHANGELOG entry for both EFCore and SqlClient for the bug fix please?

stevejgordon:
> Would you mind adding a CHANGELOG entry for both EFCore and SqlClient for the bug fix please?

Done!

PR #3372, opened 2025-11-01
[StackExchangeRedis] Support Sdk.SuppressInstrumentation. 

### LINKED ISSUE #3371: [feature request] [Instrumentation.StackExchangeRedis] should support Sdk.SuppressInstrumentation.
### Component

OpenTelemetry.Instrumentation.StackExchangeRedis

using (SuppressInstrumentationScope.Begin())
{
    database.StringGet("1"); // Can't suppress activity.
}

Activity can be suppressed in SuppressInstrumentationScope

``` diff
-if (parent == null || parent.IdFormat != ActivityIdFormat.W3C)
+if (parent == null || parent.IdFormat != ActivityIdFormat.W3C || Sdk.SuppressInstrumentation)
{
    return this.defaultSession;
}
```

### Additional context

_No response_

### Tip

pengweiqhca:
Fixes #3371
Design discussion issue #


src/OpenTelemetry.Instrumentation.StackExchangeRedis/OpenTelemetry.Instrumentation.StackExchangeRedis.csproj
```diff
@@ -28,9 +28,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Configuration" />
-    <PackageReference Include="Microsoft.Extensions.Options" />
-    <PackageReference Include="OpenTelemetry.Api.ProviderBuilderExtensions" />
+    <PackageReference Include="OpenTelemetry" />
```
Kielek, line 30: I am concerned about this change. During stabilization effort, this package was removed (at least from HttpClient). We still do not have good solution to make similar things referencing only OpenTelemetry/Api package.

@open-telemetry/dotnet-maintainers, do you see any better option here (even long term, by adding this functionality on API level)?

matt-hensley: Brought this up at the .NET SIG on Nov 4 2025. Maintainers are looking at moving this functionality out of the SDK package.

Suppression is not currently part of the spec but is useful enough to find a better solution, even though it would technically make the API out of compliance.

pengweiqhca: @Kielek Use `Func<bool>` to replace `Sdk.SuppressInstrumentatio`?

Kielek:
@matt-hensley, @pengweiqhca,

I have discussed this problem internally with @pellared. There is another option you can consider and it should met OpenTelemetry design.

The correct way to determine if you need to create span is Sampler. You could try to write your own and inject it into the OpenTelemetry tracer. What is more some attributes should be available for Sampler to allow making this decision:
https://github.com/open-telemetry/semantic-conventions/blob/d48be4318100be7a8899b713c07db0471b5b1b3d/docs/database/redis.md

What do you think about this solution?

Historically, `Sdk.SuppressInstrumentation` was designed mostly for Exporters (OTLP) to avoid instrumentation for calls during this process.

pengweiqhca:
@Kielek I know Sampler and am already using it, but it’s not suitable for my scenario.

```
for (var index = 0; index < xxx; index++)
{
    using var activity = ActivitySource.StartActivity();

    // I only want to get the full trace the first time and ignore it afterward, but Redis can't be suppressed.
    using var scope = index == 0 ? null : SuppressInstrumentationScope.Begin();

    // database, redis, http, etc.
}
```

PR #3402, opened 2025-11-05
[Infra] Add net10.0 target

martincostello:
Noticed while doing final validation for #3399 and #3401 before release.

## Changes

- Add `net10.0` target to relevant packages that was missed from #2822. Changes are based on [what we do in open-telemetry/opentelemetry-dotnet](https://github.com/open-telemetry/opentelemetry-dotnet/blob/d484e3346bf3611c04264343ee5a664ff815714a/build/Common.props#L33C5-L33C141).
- Fix obsolete warning for .NET 10.
- Fix `CHANGELOG`s to remove .NET 10 change where the TFMs were not changed.
- Update dependency versions for Microsoft.Extensions.* for .NET Standard and .NET Framework to 10.0.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,8 +45,8 @@
   -->
   <ItemGroup Label="Production .NET packages">
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="4.14.0" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="[9.0.0,)" />
-    <PackageVersion Include="Microsoft.Extensions.Options" Version="[9.0.0,)" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="[10.0.0-rc.2.25502.107,)" />
+    <PackageVersion Include="Microsoft.Extensions.Options" Version="[10.0.0-rc.2.25502.107,)" />
     <PackageVersion Include="Microsoft.Owin" Version="[4.2.2,)" />
     <PackageVersion Include="Microsoft.Win32.Registry" Version="[5.0.0,)" />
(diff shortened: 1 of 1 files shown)


martincostello:
Open to suggestions on different ways to refactor the properties that decide the TFMs. The intention here is that `net10.0` is added to any library that already targeted `net8.0` and/or `net9.0`, but not added to any that only target `net4xx` and/or `netstandard2.x`.

src/Shared/ServerCertificateValidationProvider.cs
```diff
@@ -45,7 +45,12 @@ private static bool LoadCertificateToTrustedCollection(X509Certificate2Collectio
     {
         try
         {
+#if NET10_0_OR_GREATER
+            collection.Add(X509CertificateLoader.LoadCertificateFromFile(certFileName));
```
martincostello, line 49: Resolves [SYSLIB0057: X509Certificate2 and X509Certificate constructors for binary and file content are obsolete](https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib0057)

martincostello:
Dependencies checked for AspNetCore and Http which now have a `net10.0` target:

image

image

martincostello:
Also noticed these haven't been updated for the .NET Framework and .NET Standard targets to match https://github.com/open-telemetry/opentelemetry-dotnet/issues/6489:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/a5f3fc04bd3c6ba6e8e129d204ee93e9a449b34f/Directory.Packages.props#L48-L49

Kielek (approved):
In the follow up we could do something with name for `TargetFrameworksForLibraries`.

I am not sure what it should be.

PR #3530, opened 2025-11-14
[Infra] Try to fix OSSF Scorecard false positive

martincostello:
## Changes

- Try and fix `Warn: nugetCommand not pinned by hash` false positive which marks down the `Pinned-Dependencies` rule by removing explicit MSBuild property value passing from workflow.
- Avoid interpolation in arguments.


martincostello:
Looks like that didn't work 😞

Kielek:
You can push changes to your fork-main and execute checks there. It will probably detect other things, but it should be not related to the Pinned-Dependencies

PR #736, opened 2022-10-27
[Exporter.Geneva] Update scopes

utpilla:
### Exporting scopes in GenevaExporter:

1. Export scopes which have a non-null key as individual columns (each key-value pair from the scopes is exported as its own column; these columns would also be taken into consideration when the CustomFields option is applied)
2. GenevaExporter would not check for duplicate keys (if the scopes and log message together have any duplicate keys, they would all be exported)

### Examples

**1. Key is null**
```csharp
using (logger.BeginScope("MyScope"))
{
    logger.LogInformation("Hello from {food} {price}.", "artichoke", 3.99);
}
```
No scopes are exported in this case as there is no key provided for the scope.

**2. Using formatted string in scopes**
```csharp
using (logger.BeginScope("MyInnerInnerScope with {name} and {age} of custom", "John Doe", 35))
{
    logger.LogInformation("Hello from {food} {price}.", "artichoke", 3.99);
}
```
Scopes exported:

|	      name|	age|
|----------:|---:|
|	"John Doe"|	 35|

**3. Using `IEnumerable<KeyValuePair<string, object>>` in scopes**
```csharp
using (logger.BeginScope(new List<KeyValuePair<string, object>>
{
    new KeyValuePair<string, object>("MyKey1", "MyValue1"),
    new KeyValuePair<string, object>("MyKey2", "MyValue2"),
}))
{
    logger.LogInformation("Hello from {food} {price}.", "artichoke", 3.99);
}
```
Scopes exported:

|MyKey1	|MyKey2|
|-----:|------:|
|MyValue1	|MyValue2|

**4. Duplicate Keys in scopes**

```csharp
using (logger.BeginScope(new List<KeyValuePair<string, object>>
{
    new KeyValuePair<string, object>("MyKey1", "MyValue1"),
    new KeyValuePair<string, object>("MyKey1", "MyValue1"),
}))
{
    logger.LogInformation("Hello from {food} {price}.", "artichoke", 3.99);
}
```
Scopes exported:

|MyKey1	|MyKey1|
|-----:|------:|
|MyValue1	|MyValue1|

Please provide a brief description of the changes here.

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #3531, opened 2025-11-14
[Exporter.Geneva] implement resource attributes for logs

mattsains:
## Changes

This follows up on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3214 by implementing similar behaviour for including resource attributes in Geneva Logs.

### DIFF
--- a/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
+++ b/test/OpenTelemetry.Exporter.Geneva.Tests/GenevaLogExporterTests.cs
@@ -664,14 +664,16 @@ public class GenevaLogExporterTests
             {
                 PrepopulatedFields = new Dictionary<string, object>
                 {
-                    ["cloud.role"] = "BusyWorker",
-                    ["cloud.roleInstance"] = "CY1SCH030021417",
                     ["cloud.roleVer"] = "9.0.15289.2",
                     ["prepopulated"] = "prepopulated field",
                 },
             };
 
-            var resource = ResourceBuilder.CreateEmpty().AddAttributes([new KeyValuePair<string, object>("resourceAttr", "from resource")]).Build();
+            var resource = ResourceBuilder.CreateEmpty().AddAttributes([
+                new KeyValuePair<string, object>("resourceAttr", "from resource"),
+                new KeyValuePair<string, object>("service.name", "BusyWorker"),
+                new KeyValuePair<string, object>("service.instanceId", "CY1SCH030021417")])
+                .Build();
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
@@ -858,7 +860,11 @@ public class GenevaLogExporterTests
             var TimeStampAndMappings = ((fluentdData as object[])[1] as object[])[0];
             var timeStamp = (DateTime)(TimeStampAndMappings as object[])[0];
             var mapping = (TimeStampAndMappings as object[])[1] as Dictionary<object, object>;
-            var env_properties = mapping.GetValueOrDefault("env_properties") as Dictionary<object, object> ?? [];
+            var env_properties = new Dictionary<object, object>();
+            if (mapping.ContainsKey("env_properties"))
+            {
+                env_properties = mapping["env_properties"] as Dictionary<object, object> ?? [];
+            }
 
             void AssertField(bool isDedicated, string fieldValue)
             {
@@ -1712,7 +1718,12 @@ public class GenevaLogExporterTests
         foreach (var item in exporterOptions.PrepopulatedFields)
         {
             var partAValue = item.Value as string;
-            var partAKey = MsgPackExporter.V40_PART_A_MAPPING.GetValueOrDefault(item.Key, item.Key);
+            var partAKey = item.Key;
+            if (MsgPackExporter.V40_PART_A_MAPPING.ContainsKey(item.Key))
+            {
+                partAKey = MsgPackExporter.V40_PART_A_MAPPING[item.Key];
+            }
+
             Assert.Equal(partAValue, mapping[partAKey]);
         }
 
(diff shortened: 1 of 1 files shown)


src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -88,9 +88,14 @@ public GenevaLogExporter(GenevaExporterOptions options)
                 throw new NotSupportedException($"Protocol '{connectionStringBuilder.Protocol}' is not supported");
         }
 
+        Resources.Resource ResourceProvider()
+        {
+            return connectionStringBuilder.HonorResourceAttributes ? this.ParentProvider.GetResource() : Resources.Resource.Empty;
```
cijothomas, line 93: is this a simple boolean setting, which, when enabled, will add all resource attributes to the log?

mattsains: yes, it's a connection string-based opt-in feature switch

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -366,27 +413,44 @@ internal ArraySegment<byte> SerializeLogRecord(LogRecord logRecord)
...
+                        cursor = MessagePackSerializer.Serialize(buffer, cursor, entry.Value);
+                        envPropertiesCount += 1;
+                    }
                 }
-                else
+            }
+
+            foreach (var entry in this.propertiesEntries)
+            {
+                // A prepopulated env_properties entry should not be added if the same key exists in the log,
```
cijothomas, line 443: GenevaExporter previously did not attempt de-duplication for performance reasons (neither did upstream OTel sdk). I don't think we need to do it now either as this will affect perf.

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
cijothomas: Got it. I spoke to @rajkumar-rangaraj also, and IMHO, it should be a per-attribute opt-in instead of adding *all* resource attributes. I understand users can control Resource by limiting things added to Resource, but Resource is *not* per exporter, so it's possible they need full resource in another exporter but limited one in Geneva, which is impossible to achieve now.

Few reasons why I think this should be enabled on per-resource-attribute basis.
1. Consistency with other Geneva - elsewhere, this is opt-in on per-attribute basis. eg: [OTel Rust](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/src/logs/processor.rs#L142-L171)
2. Given GenevaExporter can *only* operate with a local agent, it is much better for the agent to add most of the resource information itself. (This is quite different than OTLP approach where agent is likely remote and cannot deduce resource information itself)
3. Payload size limitations - both etw and linux user_events transport have strict hard limit of 64KB, so adding all the resource attributes is not a good idea.
4. Unlike OTLP/Zipkin exporters, GenevaExporter need to send Resource attribute with **each** event - this is already bad (but due to good reasons to make each event self-contained), and is better to selectively send a subset of attributes.

mattsains: does this comment apply to the other PR relating to traces? https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3214

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
mattsains: it was previously not possible to have a duplicate in env_properties. Do you think I should remove this check for duplicates between resource attributes and log fields? How do I weigh sending bad data to the agents versus performance impact?

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
mattsains: I actually attempted to make the change you are suggesting for traces, but it received negative feedback so I decided not to pursue it: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3367#event-20894689667

cijothomas: > does this comment apply to the other PR relating to traces? #3214

Yes.

cijothomas: > I actually attempted to make the change you are suggesting for traces, but it received negative feedback so I decided not to pursue it: [#3367 (comment)](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3367#event-20894689667)

Yes I am aware. I am asking Raj to reconsider, so as to be consistent with the prior work (OTel Rust, which has this as a stable feature already). OTel Rust did it the way it did for the reasons I shared in earlier comment.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
cijothomas: For Logs, duplicates were possible before. (eg: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/736).

The current behavior is to let Agent deal with the duplicate data. It may/may-not be ideal, but that is the current state. (IIRC, metrics also don't do de-duplication of attributes. Agent deals with de-duplication)

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -33,26 +34,33 @@ internal sealed class MsgPackLogExporter : MsgPackExporter, IDisposable
...
-    private readonly List<string>? prepopulatedFieldKeys;
     private readonly byte[] bufferEpilogue;
     private readonly IDataTransport dataTransport;
+    private readonly Func<Resource> resourceProvider;
+
+    // These are values that are always added to the body as dedicated fields
+    private readonly Dictionary<string, object> prepopulatedFields;
+
+    // These are values that are always added to env_properties
+    private readonly Dictionary<string, object> propertiesEntries;
```
rajkumar-rangaraj, line 50: FrozenDictionary is optimized for read-heavy scenarios with zero-allocation lookups and perfect hashing. Switching to regular Dictionary adds a lookup overhead

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -227,15 +271,20 @@ internal ArraySegment<byte> SerializeLogRecord(LogRecord logRecord)
         ushort cntFields = 0;
         var idxMapSizePatch = cursor - 2;
 
-        if (this.prepopulatedFieldKeys != null)
+        this.AddResourceAttributesToPrepopulated();
```
rajkumar-rangaraj, line 274: This method runs on every log record, which is concerning for the hot path. While the idempotency guards prevent duplicate additions, we're still invoking the resourceProvider delegate, iterating all resource attributes, and performing dictionary lookups per-record. Additionally, the .Any() calls allocate an enumerator and closure on each invocation.

Issue #1807, opened 2024-05-16
Opt-out Optional Fields


SuyliuMS:
### Component

OpenTelemetry.Exporter.Geneva

### What is the expected behavior?

According to the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#severity-fields, SeverityText and SeverityNumber are both optional fields. But OpenTelemetry.Exporter.Geneva now don't support to not log SeverityText and SeverityNumber.
Actually there are more optional fields that can't be opt-out now, maybe you can fix in one time.

### Which alternative solutions or features have you considered?

Maybe we can set a switch in GenevaExporterOptions to control if we want to log SeverityText and SeverityNumber.

reyang:
> Maybe we can set a switch in GenevaExporterOptions to control if we want to log SeverityText and SeverityNumber.

Thanks for the suggestion @SuyliuMS! It makes sense to me.

PR #3550, opened 2025-11-20
Add configuration of optional fields in GenevaExporter

hannahhaering:
Added a question in comments, I will add a changelog entry once the question has been resolved.

Implements #1807

## Changes

Added the possibility to disable logging of SeverityText and SeverityNumber via GenevaExporterOptions.


hannahhaering:
The issue mentioned that other fields should be optional as well. I think that the other optional fields are optional in the sense of that they may be skipped if they don't have a value. Would it make sense to add a configuration option to disable them like I did in this PR for the severity fields?

rajkumar-rangaraj:
@hannahhaering Thanks for your contributions. At this time, we do not plan to expand the public API surface.

PR #3552, opened 2025-11-20
[Exporter.Geneva] Opt in resources

mattsains:
This is a re-submission of https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3367 , after @cijothomas recommend that it should be re-introduced.

I did change the name of the config option to WithResourceAttributes to match the Rust option in the [user events log exporter](https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/src/logs/processor.rs#L142-L171)

I also changed the logic when this field is not provided. The default is now to include no resource attributes except for mapped fields (eg., service.name -> Part A cloud extension)

## Changes

Allows users of the Geneva exporter to specify which resource attributes they want sent to Geneva. This prevents excessive trace data from being exported if there are many resource attributes which do not need to be stored.

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
@@ -14,7 +14,7 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.get -> System.Collections.Generic.IEnumerable<string!>?
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.set -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.init -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.get -> System.Collections.Generic.IEnumerable<string!>?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.EventNameExportMode.get -> OpenTelemetry.Exporter.Geneva.EventNameExportMode
--- a/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
@@ -13,8 +13,8 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterHelperExtensions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WantedResourceAttributes.get -> System.Collections.Generic.IEnumerable<string!>?
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WantedResourceAttributes.set -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WithResourceAttributes.get -> System.Collections.Generic.IEnumerable<string!>?
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WithResourceAttributes.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.get -> System.Collections.Generic.IEnumerable<string!>?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.EventNameExportMode.get -> OpenTelemetry.Exporter.Geneva.EventNameExportMode
--- a/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
@@ -13,6 +13,8 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterHelperExtensions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WantedResourceAttributes.get -> System.Collections.Generic.IEnumerable<string!>?
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WantedResourceAttributes.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.get -> System.Collections.Generic.IEnumerable<string!>?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.EventNameExportMode.get -> OpenTelemetry.Exporter.Geneva.EventNameExportMode
--- a/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
@@ -13,8 +13,8 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterHelperExtensions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WithResourceAttributes.get -> System.Collections.Generic.IEnumerable<string!>?
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.WithResourceAttributes.set -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.get -> System.Collections.Generic.IEnumerable<string!>?
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.get -> System.Collections.Generic.IEnumerable<string!>?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.EventNameExportMode.get -> OpenTelemetry.Exporter.Geneva.EventNameExportMode
--- a/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
+++ b/src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
@@ -14,7 +14,7 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.get -> System.Collections.Generic.IEnumerable<string!>?
-OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.init -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.get -> System.Collections.Generic.IEnumerable<string!>?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.CustomFields.set -> void
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.EventNameExportMode.get -> OpenTelemetry.Exporter.Geneva.EventNameExportMode
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
@@ -155,6 +155,12 @@ public class GenevaExporterOptions
                         throw new ArgumentException($"Type `{entry.Value.GetType()}` (key = `{entry.Key}`) is not allowed. Only bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double, and string are supported.");
                 }
 
+                if (entry.Key == Schema.V40.PartA.Ver)
+                {
+                    // This key is now set by the exporters, and is only accepted for legacy reasons.
+                    continue;
+                }
+
                 copy[entry.Key] = val; // shallow copy
             }
 
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
@@ -11,10 +11,7 @@ namespace OpenTelemetry.Exporter.Geneva;
 /// </summary>
 public class GenevaExporterOptions
 {
-    private IReadOnlyDictionary<string, object> fields = new Dictionary<string, object>(1)
-    {
-        [Schema.V40.PartA.Ver] = "4.0",
-    };
+    private IReadOnlyDictionary<string, object> fields = new Dictionary<string, object>();
 
     private IReadOnlyDictionary<string, string>? tableNameMappings;
 
@@ -118,16 +115,9 @@ public class GenevaExporterOptions
         {
             Guard.ThrowIfNull(value);
 
-            var schemaVersion = "4.0";
-
-            if (value.ContainsKey(Schema.V40.PartA.Ver))
+            if (value.ContainsKey(Schema.V40.PartA.Ver) && value[Schema.V40.PartA.Ver] as string is not "4.0")
             {
-                schemaVersion = value[Schema.V40.PartA.Ver] as string;
-            }
-
-            if (schemaVersion is not "2.1" and not "4.0")
-            {
-                throw new ArgumentException("Unsupported schema version, only 2.1 and 4.0 are supported.");
+                throw new ArgumentException("Unsupported schema version, only 4.0 is supported.");
             }
 
             if (value.ContainsKey(Schema.V40.PartA.Name))
@@ -140,7 +130,7 @@ public class GenevaExporterOptions
                 throw new ArgumentException("Event timestamp cannot be pre-populated.");
             }
 
-            var copy = new Dictionary<string, object>(value.Count + 1) { [Schema.V40.PartA.Ver] = schemaVersion };
+            var copy = new Dictionary<string, object>(value.Count + 1);
             foreach (var entry in value)
             {
                 var val = entry.Value;
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
@@ -31,14 +31,14 @@ public class GenevaExporterOptions
     public IEnumerable<string>? CustomFields { get; set; }
 
     /// <summary>
-    /// Gets or sets ResourceFieldNames.
+    /// Gets ResourceFieldNames.
     ///
     /// ResourceFieldNames specifies which resource attribute fields should be sent to Geneva.
     ///
     /// Any resource attributes not in ResourceFieldNames are ignored.
     /// If ResourceFieldNames is not provided, no resource attributes will be sent to Geneva.
     /// </summary>
-    public IEnumerable<string>? ResourceFieldNames { get; set; }
+    public IEnumerable<string>? ResourceFieldNames { get; init; }
 
     /// <summary>
     /// Gets or sets the exception stack trace export mode.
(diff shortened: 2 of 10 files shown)


src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -33,6 +33,16 @@ public class GenevaExporterOptions
     /// </summary>
     public IEnumerable<string>? CustomFields { get; set; }
 
+    /// <summary>
+    /// Gets or sets WantedResourceAttributes.
+    ///
+    /// WantedResourceAttributes specifies which resource attribute fields should be sent to Geneva.
+    ///
+    /// Any resource attributes not in WantedResourceAttributes are ignored.
+    /// If WantedResourceAttributes is not provided, all resource attributes will be sent to Geneva.
```
cijothomas: all will be sent or none will be sent?

cijothomas: In my opinion, it should be none, so the behavior is back-compatible. To get resource attributes, users must explicitly opt-in.

mattsains: if a customer really wants all resource attributes, how might they accomplish that? By the way, there is a feature flag which is default off that controls this feature

mattsains: I've made the change you requested

mattsains: do you think the special resource attributes, service.name and service.instanceId, which are mapped to part A cloud extension, should also be affected by WithResourceAttributes being set to null?

cijothomas: > By the way, there is a feature flag which is default off that controls this feature

I am not sure if we need to feature flag this at all. The `WithResourceAttribute` is sufficient on its own right?


> if a customer really wants all resource attributes, how might they accomplish that?

By listing each attribute. Not ideal, but I also don't think this will be the common case with GenevaExporter. If there is strong need for such a capability, we can introduce a boolean 'with_all_resource`/similar that adds all attributes effectively ignore WithResourceAttributes. Alternatively, "*" can be used to mean - get all attributes - we do such thing elsewhere in GenevaExporter.

mattsains: I can remove the feature flag in a future PR

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -342,6 +351,11 @@ internal int AppendResourceAttributesToBuffer(byte[] buffer, int cursor)
                 }
             }
 
+            if (!isDedicatedField && (this.WithResourceAttributes == null || !this.WithResourceAttributes.Contains(entry.Key)))
```
cijothomas: shouldn't we short-circuit this entire method if `this.WithResourceAttributes == null` ?

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -33,6 +33,16 @@ public class GenevaExporterOptions
...
 
+    /// <summary>
+    /// Gets or sets WithResourceAttributes.
+    ///
+    /// WithResourceAttributes specifies which resource attribute fields should be sent to Geneva.
+    ///
+    /// Any resource attributes not in WithResourceAttributes are ignored.
+    /// If WithResourceAttributes is not provided, no resource attributes will be sent to Geneva.
+    /// </summary>
+    public IEnumerable<string>? WithResourceAttributes { get; set; }
```
cijothomas: do clarify that this is replacement for prepopulatedfields, and prepopulatedfields should be removed before enabling this to avoid sending duplicate fields.

cijothomas: nit: is this still controlled by a feature flag in connection string?

mattsains: it's not a replacement for prepopulated fields. Prepopulated fields still work and take priority over resource attributes

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: I decided that service.name and service.instanceId fields should still be used even if they're not explicitly mentioned in WithResourceAttributes, because they have a special mapping into part A cloud extension. So I don't believe the code should skip this section if it is one of those two resource attributes

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
mattsains: please refer to this previous conversation: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3552#discussion_r2547670562

rajkumar-rangaraj: @cijothomas Is that how it is handled in Rust?

rajkumar-rangaraj: Now I recall there is no concept of prepopulated in Rust

rajkumar-rangaraj: Should this be -->  public  IEnumerable<string>? WithResourceAttributes { get; **init**; }

Also, `AllowedResourceAttributes` or `ResourceAttributeFilter` naming might fit better here. No strong preference will leave it to you.

rajkumar-rangaraj: However, I think it’s a good idea to support having just one of them enabled since they serve the same purpose. It also makes sense to ignore prepopulated fields when this is set.

cijothomas: Given prepopulated_fields was a temporary solution until Resource support comes, I think it makes sense to ignore prepopulated fields when this is set. By default, since this is not set, we'll be fully back-compatible. If user opts-in to this, they are acknowledging that pre-populated_fields no longer works.

mattsains: sure - would you like me to do the same to PrepopulatedFields?

mattsains: prepopulated fields is not ignored if resourcefieldnames is specified

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
mattsains: feature flag has been removed, and the behaviour of the attribute is that no resource attributes are sent if it is null

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
```diff
@@ -103,21 +110,14 @@ public IReadOnlyDictionary<string, string>? TableNameMappings
...
-        get => this.fields;
+        get => this.prepopulatedFields;
         set
         {
             Guard.ThrowIfNull(value);
 
-            var schemaVersion = "4.0";
-
-            if (value.ContainsKey(Schema.V40.PartA.Ver))
+            if (value.ContainsKey(Schema.V40.PartA.Ver) && value[Schema.V40.PartA.Ver] as string is not "4.0")
```
rajkumar-rangaraj, line 118: @cijothomas I asked @mattsains to keep only 4.0 and remove 2.1. Would removing 2.1 here be a breaking change? In your experience, have you seen any customer using 2.1?

src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
```diff
@@ -13,6 +13,8 @@ OpenTelemetry.Exporter.Geneva.GenevaExporterHelperExtensions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.get -> string?
 OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ConnectionString.set -> void
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.get -> System.Collections.Generic.IEnumerable<string!>?
+OpenTelemetry.Exporter.Geneva.GenevaExporterOptions.ResourceFieldNames.init -> void
```
rajkumar-rangaraj: I recall, asking you to use `init`, but let’s switch to `set` instead for API consistency.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -320,71 +295,141 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
...
+                this.prepopulatedFields.Add(key, value);
             }
         }
 
-        return cursor;
+        foreach (var entry in this.prepopulatedFields)
+        {
+            cursor = AddPartAField(buffer, cursor, entry.Key, entry.Value);
+        }
+
```
rajkumar-rangaraj, line 424: After `CreateFraming()` writes the prepopulated fields to the buffer, the dictionary is only used for its `Count`. Consider storing the count as an `int` or `short` field and setting the dictionary to `null` to release the memory.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -125,6 +131,17 @@ public MsgPackTraceExporter(GenevaExporterOptions options, Func<Resource> resour
                 throw new NotSupportedException($"Protocol '{connectionStringBuilder.Protocol}' is not supported");
         }
 
+        if (options.ResourceFieldNames != null)
+        {
+            foreach (var wantedResourceAttribute in options.ResourceFieldNames)
+            {
+                if (PART_A_MAPPING_DICTIONARY.Values.Contains(wantedResourceAttribute))
+                {
+                    throw new ArgumentException($"'{wantedResourceAttribute}' cannot be specified through a resource attribute. Remove it from ResourceFieldNames");
```
rajkumar-rangaraj, line 142:
```suggestion
    
throw new ArgumentException(
    $"'{wantedResourceAttribute}' is a reserved Part A field. The resource attribute is automatically mapped to this field when present. Remove it from ResourceFieldNames");
```

rajkumar-rangaraj:
Consider expanding the test coverage for these scenarios

* Reserved field name in `ResourceFieldNames` throws
* Invalid type in resource attribute
* `service.name` auto-mapping with `ResourceFieldNames`
* Empty `ResourceFieldNames = []` vs null

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: I think the original message is better for two reasons. First, I think that internal customers using this SDK won't know what a part A field is. I certainly didn't know before joining this team. Secondly, It's not true that all part A fields are automatically mapped from resource attribute. Only two - service.name and service.instanceId - are.

src/OpenTelemetry.Exporter.Geneva/.publicApi/PublicAPI.Shipped.txt
mattsains: I'll change it back

src/OpenTelemetry.Exporter.Geneva/GenevaExporterOptions.cs
cijothomas: I don't think GenevaExporter really supported 2.1 ever... And to the best of my knowledge, nobody is using it either (since its not supported really!)

mattsains:
> Consider expanding the test coverage for these scenarios
> 
>     * Reserved field name in `ResourceFieldNames` throws
> 
>     * Invalid type in resource attribute
> 
>     * `service.name` auto-mapping with `ResourceFieldNames`
> 
>     * Empty `ResourceFieldNames = []` vs null

Except for service.name mapping, which was already tested, I've added all of these tests in the new revision

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: made this change.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -58,28 +58,36 @@ internal sealed class MsgPackTraceExporter : MsgPackExporter, IDisposable
...
+    // so constructing a whole new data structure for it is overkill.
+    private readonly IEnumerable<string>? resourceFieldNames;
     private readonly bool shouldIncludeTraceState;
+    private readonly string partAName;
     private readonly Func<Resource> resourceProvider;
-    private readonly List<string> prepopulatedFields;
-    private readonly Dictionary<string, object> propertiesEntries;
 
-    private byte[] bufferPrologue;
+    private byte[]? bufferPrologue;
```
xiang17, line 81: Why would this ever be null?

xiang17: I see you made a change later where you call it in `CreateFraming` which is called in `SerializeActivity`. Do you need any fields which are not available in constructor?

There may be issues to move out of constructor.
1. The design principle is to fail as early as possible if configuration is incorrect. And we shouldn't crash customers' apps while running (after construction is finished).
2. `SerializeActivity` is in hot path, needs to be thread-safe and should not throw exceptions.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -320,71 +299,154 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
...
+        System.Buffer.BlockCopy(buffer, 0, this.bufferPrologue, 0, cursor - 0);
+
+        // Now generate the epilogue
+        cursor = MessagePackSerializer.Serialize(buffer, 0, new Dictionary<string, object> { { "TimeFormat", "DateTime" } });
+
+        this.bufferEpilogue = new byte[cursor - 0];
+        System.Buffer.BlockCopy(buffer, 0, this.bufferEpilogue, 0, cursor - 0);
+
+        // We can release the prepopulatedFields dictionary because we finished using it in the prologue
+        this.prepopulatedFields = null;
```
xiang17: I have a hunch that this might cause race condition.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
xiang17: > After `CreateFraming()` writes the prepopulated fields to the buffer, the dictionary is only used for its `Count`.

I doubt this. `SerializeActivity` could be called in multi-threading environment. A second thread would need to set up `bufferPrologue` again with `CreateFraming()`.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -320,71 +299,154 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
...
+                            case int:
+                            case uint:
+                            case long:
+                            case ulong:
+                            case float:
+                            case double:
+                            case string:
+                                break;
+                            case null:
+                                throw new ArgumentNullException(key, "Resource attribute must not have a null value.");
```
xiang17: This could crash customer's app.

cijothomas: We need to handle this, and follow the principle of never throw exception. (except in the ctor() )

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: yes, the data I need is the resource attributes, which isn't available in the constructor

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: What do you suggest, @cijothomas ?

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: I've removed this part, it was based on a recommendation that ended up being incorrect

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
rajkumar-rangaraj: Log to event source - https://github.com/mattsains/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Internal/ExporterEventSource.cs

cijothomas: Maybe put `"<NULL>"` as the value, so at least users know something is broken... 

I need to recheck to see if this is actually going to crash user app or just fail the export? Even failing entire export is bad just because of one attribute, so above is probably better..

cijothomas: > Log to event source - https://github.com/mattsains/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Internal/ExporterEventSource.cs

I wouldn't suggest that..The exporter is designed to be re-entrant safe... so logging here could cause stackoverflow, if someone enables a EventSource->ILogger bridge.. Low possibility, but as a general principle, we should not log here. If we really need to do logging, then it must be under SuppressionGuard.


Rust has a comment on this: https://github.com/open-telemetry/opentelemetry-rust-contrib/blob/main/opentelemetry-user-events-logs/src/logs/processor.rs#L50-L55

mattsains: for null I could also just drop the attribute, as an alternative.

How about the next exception, the ArgumentException, on line 379? There is actually already validation in the Resource class to avoid anything except the types listed, and also arrays of primitive types, like int[], etc. As far as I can tell, only the array types can cause this exception to be thrown. Maybe we could just handle them idk

rajkumar-rangaraj: > for null I could also just drop the attribute, as an alternative.

Silently dropping an attribute may create supportability issues later. Especially when a customer reports that specific attributes are “missing.” We’ve seen the `"<NULL>"` pattern before, and it provides clearer diagnostics. That seems like a better option here.

Regarding line 378, we could use value as `value.GetType().ToString()` to capture the unexpected type.
Thoughts?

mattsains: > Regarding line 378, we could use value as value.GetType().ToString() to capture the unexpected type.

And then?

`"<int[] is not a supported type>"` ?

rajkumar-rangaraj: This is how it's handled in most exporter, it's an abstract - https://github.com/open-telemetry/opentelemetry-dotnet/blob/49bfa23e825cf7320cf1c70f671313600a58d903/src/Shared/TagWriter/TagWriter.cs#L42

rajkumar-rangaraj: This could be a follow up based on the customer ask, not need to cover now.

mattsains: Raj has discovered that it's not possible to have null in a resource value: https://github.com/open-telemetry/opentelemetry-dotnet/blob/49bfa23e825cf7320cf1c70f671313600a58d903/test/OpenTelemetry.Tests/Resources/ResourceTests.cs#L34

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
```diff
@@ -456,6 +463,341 @@ public void GenevaTraceExporter_Prepopulated_Overwrites_Resource()
         }
     }
 
+    [Fact]
+    public void GenevaTraceExporter_InvalidResourceAttrType_Throws()
```
rajkumar-rangaraj: Is the test name correct?

mattsains: no, I need to update it

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -320,71 +297,164 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
...
+                                // Try to construct a value that communicates that the type is not supported.
+                                try
+                                {
+                                    var stringValue = Convert.ToString(value, CultureInfo.InvariantCulture);
+                                    if (stringValue == null)
+                                    {
+                                        value = "<Unsupported type>";
+                                    }
+
+                                    value = $"<Unsupported type: {stringValue}>";
```
rajkumar-rangaraj: The value here is always set using this line, so anything set beforehand will be overridden.

value = $"<Unsupported type: {stringValue}>";

rajkumar-rangaraj (approved):
LGTM, this needs a fix - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3552/changes#r2620584269

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
mattsains: oh you're right, I'm missing a break in the if statement

mattsains: I'm not sure how to trigger this branch otherwise there would be a test for it

test/OpenTelemetry.Exporter.Geneva.Tests/GenevaTraceExporterTests.cs
mattsains: updated.

xiang17 (approved):

Issue #3554, opened 2025-11-24
[feature request] Make db.statement sanitization configurable for SqlClient/EFCore  (preserve SQL comments / EF TagWith tags)


alexander-kucherov:
### Component

OpenTelemetry.Instrumentation.SqlClient

### Is your feature request related to a problem?

Make db.statement sanitization configurable in .NET (preserve SQL comments / EF TagWith tags)

### Summary

The .NET SQL instrumentation currently always sanitizes SQL statements, including stripping comments, before setting `db.statement` / `db.query.text` on spans.
This makes it impossible to rely on SQL comments for tagging (e.g., EF Core `TagWith`) in telemetry, even when users are willing to accept the risk of collecting raw SQL.

This issue proposes making db.statement sanitization configurable in .NET (similar to the existing `db-statement-sanitizer` feature in the Java agent), while keeping sanitization enabled by default.

### Current behavior

`SqlProcessor` sanitizes SQL by:

- Replacing all literal values (strings, numbers, hex) with `?`.
- Special-casing `IN (...)` lists to a single `?`.
- Removing both single-line (`-- ...`) and multi-line (`/* ... */`) comments entirely from the statement.

`ApplyConventionsForQueryText(Activity activity, string? commandText, bool emitOldAttributes, bool emitNewAttributes, bool sanitizeQuery = true)` already exposes a `sanitizeQuery` boolean, but this parameter is not wired to any public configuration and effectively always uses the default `true`.

As a result, SQL comments never appear in `db.statement` / `db.query.text` on spans, regardless of user intent.

### Problem

Many applications use SQL comments as a lightweight tagging mechanism, for example via EF Core `TagWith`, to associate queries with features, use-cases, or higher-level application concepts. When the sanitizer strips comments, these tags are lost from telemetry, even though:

- The actual database sees the tagged SQL and executes it just fine.
- Some users are willing (or explicitly configured) to allow raw SQL in their observability pipeline.
- Some deployments already implement downstream redaction at the collector or log processor level.

In short, the current behavior is "always sanitize, always drop comments", with no way to opt out for environments that consciously accept the trade-off.

### Proposed solution

Introduce a configurable "db-statement sanitizer" option for .NET database instrumentations and connect it to the existing `sanitizeQuery` parameter:

1. Add a public configuration option, for example on the relevant instrumentation options type(s), such as:

[...]

PR #3555, opened 2025-11-24
Feature/db statement sanitizer #3554

alexander-kucherov:
My proposal for #3554 3554

## Changes

Introduce a configurable "db-statement sanitizer" option for .NET database instrumentations and connect it to the existing `sanitizeQuery` parameter:

1. Add a public configuration option, for example on the relevant instrumentation options type(s), such as:

2. Wire this option into the code path that calls `ApplyConventionsForQueryText`, passing `sanitizeQuery: options.DbStatementSanitizerEnabled`.

3. Keep `DbStatementSanitizerEnabled` defaulting to `true` to preserve the current safe-by-default behavior and stay consistent with the OpenTelemetry guidance that `db.statement` should generally be sanitized by default.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
@@ -175,7 +175,7 @@ internal sealed class EntityFrameworkDiagnosticListener : ListenerHandler
                                 case CommandType.Text:
                                     // Only SQL-like providers support sanitization as we are not
                                     // able to sanitize arbitrary commands for other query dialects.
-                                    bool sanitizeQuery = IsSqlLikeProvider(providerName) && this.options.DbStatementSanitizerEnabled;
+                                    bool sanitizeQuery = this.options.DbStatementSanitizerEnabled && IsSqlLikeProvider(providerName);
 
                                     DatabaseSemanticConventionHelper.ApplyConventionsForQueryText(
                                         activity,


src/OpenTelemetry.Instrumentation.SqlClient/SqlClientTraceInstrumentationOptions.cs
```diff
@@ -22,6 +22,7 @@ public class SqlClientTraceInstrumentationOptions
 {
     internal const string ContextPropagationLevelEnvVar = "OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_CONTEXT_PROPAGATION";
     internal const string SetDbQueryParametersEnvVar = "OTEL_DOTNET_EXPERIMENTAL_SQLCLIENT_ENABLE_TRACE_DB_QUERY_PARAMETERS";
+    internal const string DbStatementSanitizerEnabledEnvVar = "OTEL_INSTRUMENTATION_COMMON_DB_STATEMENT_SANITIZER_ENABLED";
```
martincostello: If this isn't in the OTel spec, I don't think we'd add this (regardless of whether Java already uses it). It would just be an in-code option.

alexander-kucherov: @martincostello , it was used here https://opentelemetry.io/docs/zero-code/java/agent/instrumentation/#db-statement-sanitization
But no problem, let's leave only in-code option. Just removed this line

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -175,7 +175,7 @@ public override void OnEventWritten(string name, object? payload)
                                 case CommandType.Text:
                                     // Only SQL-like providers support sanitization as we are not
                                     // able to sanitize arbitrary commands for other query dialects.
-                                    bool sanitizeQuery = IsSqlLikeProvider(providerName);
+                                    bool sanitizeQuery = IsSqlLikeProvider(providerName) && this.options.DbStatementSanitizerEnabled;
```
martincostello: Swapping the conditions would be _slightly_ more performant (avoid the string check if disabled).

alexander-kucherov: Thanks! Fixed.

alexander-kucherov:
@martincostello just a gentle ping to see if you could review this PR when it’s convenient.​

martincostello:
> Will review properly if we decide we do want to implement this functionality.

PR #3593, opened 2025-12-09
Enhance `OpAmpClient` with subscribe and unsubscribe

stevejgordon:
Fixes N/A

## Changes

Updated `PublicAPI.Unshipped.txt` to include new public API entries for `IOpAmpListener<TMessage>` and its `HandleMessage` method, as well as `Subscribe<T>` and `Unsubscribe<T>` methods in `OpAmpClient`.

Removed `IOpAmpListener` empty marker interface. 

Made `IOpAmpListener<TMessage>` interface public, adding XML documentation.

Removed `IOpAmpMessage` empty marker interface, replacing it with a public empty abstract class `OpAmpMessage`.

Added `Subscribe<T>` and `Unsubscribe<T>` methods to `OpAmpClient` in `OpAmpClient.cs`, allowing users to register listeners for messages.

Renamed `PlainHttpTransportTest.cs` to `PlainHttpTransportTests.cs` for consistency.

Added `OpAmpClientTests.cs` to test subscription and unsubscription functionality in `OpAmpClient`.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
+++ b/src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
@@ -4,7 +4,6 @@
 using OpenTelemetry.Internal;
 using OpenTelemetry.OpAmp.Client.Internal;
 using OpenTelemetry.OpAmp.Client.Internal.Listeners;
-using OpenTelemetry.OpAmp.Client.Internal.Listeners.Messages;
 using OpenTelemetry.OpAmp.Client.Internal.Services;
 using OpenTelemetry.OpAmp.Client.Internal.Services.Heartbeat;
 using OpenTelemetry.OpAmp.Client.Internal.Transport;


stevejgordon:
This required making the `IOpAmpListener`, `IOpAmpListener<T>` and `IOpAmpMessage` interfaces public. Some of these are empty marker interfaces required for generic support of subscription which upsets the compiler (CA1040). Can we ignore these?

I also haven't moved these from internal directory (yet). Should we do so now that they are public? I notice that the message implementations are still in the internal directory.

src/OpenTelemetry.OpAmp.Client/Internal/Listeners/IOpAmpListener.cs
```diff
@@ -3,12 +3,23 @@
...
+/// <summary>
+/// Marker interface for OpAmp listeners.
+/// </summary>
+public interface IOpAmpListener
 {
 }
 
-internal interface IOpAmpListener<TMessage> : IOpAmpListener
+/// <summary>
+/// A listener capable of handling OpAmp messages of a specific type.
```
martincostello: For non-code identifiers should we use `OpAMP` rather than `OpAmp`? Asking as it's written differently in the CHANGELOG.

src/OpenTelemetry.OpAmp.Client/Internal/Listeners/IOpAmpMessage.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace OpenTelemetry.OpAmp.Client.Internal.Listeners;
 
-internal interface IOpAmpMessage
+/// <summary>
+/// Marker interface for OpAmp messages.
+/// </summary>
+public interface IOpAmpMessage
```
martincostello: If we keep the marker interfaces, could simply the declarations like below:

```diff
- public interface Foo
+ public interface Foo;
- {
- }
```

martincostello: Thinking aloud, maybe for these an empty base class could be used for this instead? Solely as an exercise to avoid an empty interface though.

```csharp
public abstract class OpAmpMessage;

public interface IOpAmpListener<TMessage> : IOpAmpListener
    where TMessage : OpAmpMessage
{
    void HandleMessage(TMessage message);
}
```

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        });
+        client.Subscribe(mockListener);
+
+        await client.StartAsync();
+
+        // We don't currently have a direct way to send a message from the client to the server to trigger a response, so
+        // this depends on the heartbeat messages from the server to the client.
+
+        mockListener.WaitForMessages(TimeSpan.FromSeconds(2)); // Wait for the initial identification message response
+        mockListener.WaitForMessages(TimeSpan.FromSeconds(2)); // Wait for the heartbeat response
```
martincostello: What about?
```suggestion
        // Wait for the initial identification plus the heartbeat responses
        var wait = TimeSpan.FromSeconds(2 + 2);

        mockListener.WaitForMessages(wait);
```

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        client.Unsubscribe(mockListener);
+
+        mockListener.WaitForMessages(TimeSpan.FromSeconds(3)); // Wait to verify no more messages arrive after unsubscribing.
```
martincostello: Actually, is there a way we can fake the time in this test so it doesn't need to wait actual wall-clock seconds for things to (not) happen?

martincostello:
> Some of these are empty marker interfaces required for generic support of subscription which upsets the compiler (CA1040). Can we ignore these?

If there's a way we could design it to not need them then I think that would be preferable personally, but I can see how at least the first one makes the code easy to deal with the fact that real meat is a generic interface where we won't easily know the type of the message in advance.

If there's places doing `if (foo is IOpAmpListener)` could they just be replaced with a similar check that just does the equivalent of _"implements an interface of `typeof(IOpAmpListener<>)`"_?

stevejgordon:
@martincostello At least for `IOpAmpListener` I think we can remove it and switch to using `object` for the list of listeners that are subscribed. I wasn't sure about messing with the existing design, but since we're making these public, I agree, it would be nice to avoid them. `IOpAmpMessage` is harder to fully avoid but I have switched it to an abstract class instead.  WDYT?

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
stevejgordon: I didn't see a super clean way to do it. The heartbeat is triggered from the `HeartbeatService`. I looked at using `TimeProvider` with an internal hook for tests to provide a fake one, but it's not supported on all targets and feels a bit clunky. Longer term, we'll expose methods to send client to server messages which might be a nicer way to test this, waiting for those responses. I think that's a separate PR though. One other option for now is updating the mock server to provide a method to force a custom message out so we're can trigger the messages immediately. This was a bit of a hack to avoid that but I'll see what's involved.

stevejgordon: I've refactored the tests a bit. We still need to wait 1 second for the verification of no response, but we can now trigger heartbeats on demand from the client itself. This still feels a little ugly but we can remove this once we implement an API to support consumers sending messages, at which point we can switch to using the public API and a custom message. Do you feel this is reasonable for now?

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
```diff
@@ -13,10 +13,10 @@ namespace OpenTelemetry.OpAmp.Client.Internal;
 
 internal sealed class FrameProcessor
 {
-    private readonly ConcurrentDictionary<Type, IReadOnlyList<IOpAmpListener>> listeners = [];
+    private readonly ConcurrentDictionary<Type, IReadOnlyList<object>> listeners = [];
```
martincostello, line 16: I think this is Advent-of-Code recency making me notice this, but given that we only iterate over the list or change the size +/- 1 with object identity, I wonder if `I{ReadOny}Set<T>` backed with a `HashSet<T>` would be a bit more efficient?

Then could be something like the below?

```csharp
(_, set) => [.. set, listener];
(_, set) => [.. set.Except(listener)];
```

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        // Wait for the initial identification message response
+        mockListener.WaitForMessages(TimeSpan.FromSeconds(1));
+
+        Assert.Single(mockListener.Messages);
+
+        await client.SendHeartbeatAsync(new HealthReport
+        {
+            StartTime = GetCurrentTimeInNanoseconds(),
+            StatusTime = GetCurrentTimeInNanoseconds(),
```
martincostello, line 41: I assume not, but would it make any difference if `StartTime` and `StatusTime` weren't the exact same value from the time to execute twice?

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
stevejgordon: I'm not adverse to looking at that, although perhaps we could make that a follow up PR after reviewing benchmarks etc. I was also thinking that perhaps a linked list approach might also work, similar to the internals of ActivityListeners I believe.

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
stevejgordon: No, the response from the fake server should be pretty instant. I put 1 second as a precaution since we're trying to test for the lack of a response.

src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -1,14 +1,14 @@
-OpenTelemetry.OpAmp.Client.Internal.Listeners.IOpAmpListener
 OpenTelemetry.OpAmp.Client.Internal.Listeners.IOpAmpListener<TMessage>
-OpenTelemetry.OpAmp.Client.Internal.Listeners.IOpAmpListener<TMessage>.HandleMessage(TMessage message) -> void
-OpenTelemetry.OpAmp.Client.Internal.Listeners.IOpAmpMessage
+OpenTelemetry.OpAmp.Client.Internal.Listeners.IOpAmpListener<TMessage>.HandleMessage(TMessage! message) -> void
```
Kielek, line 2: It is probably leftover from previous PRs.

Could you please consider separate PR cleaning it up with some short note that Internal infix was removed from the namespaces?

I do not think that we should expose any Internal API

stevejgordon: Just to make sure I understand correctly, in a new PR, we should both move out of the "Internal" folder and remove from the "Internal" namespace any types we made public in the PR and previously?

Kielek: I think that it is the best what we can do with this kind of changes..

stevejgordon:
@martincostello and @Kielek - Do you think this current set is okay to merge? I'll start on the PR to tidy up the internal namespace and directory as well.

martincostello:
I'm ok with moving the namespaces in a follow-up to make the diff less crazy.

src/OpenTelemetry.OpAmp.Client/Internal/Listeners/IOpAmpListener.cs
stevejgordon: I believe all of these are fixed now.

src/OpenTelemetry.OpAmp.Client/Internal/Listeners/IOpAmpMessage.cs
stevejgordon: Applied for the empty abstract class.

stevejgordon: We've avoided this by storing `object` as the subscriber so the marker interface is no longer required.

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
stevejgordon: I think we've reduced the delay as much as possible, for now.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
stevejgordon: Raised https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3610

src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
stevejgordon: Raised #3609

stevejgordon:
@Kielek Do you think it's okay to merge this one as-is? I'll then work on the namespaces. After that, I'm hoping we could release a new alpha so I can try the bits.

Kielek:
LGTMRS

PR #3614, opened 2025-12-12
[OpAMP] Expose public `RemoteConfigMessage`

### LINKED ISSUE #3613: [OpAMP] Publicly expose types for server to agent messages
It is currently not possible for a consumer to create an `IOpAmpListener` for any of the concrete nessage types. These are currently internal. 

We should expose these types on the public API. A consideration is that doing so in the current design would also require the generated types from the .proto files to be public as those are directly exposed on properties of the messages. This is likely something we want to avoid.

Instead, we could consider making the message types wrappers for the underying generated types, exposing the properties as neccesary. Any properties should not directly expose generated types from the proto. We may therefore need to create some public types that expose the data on message properties in a more friendly form, for example, making `AgentConfigMap` a dictionary.

In a rough prototype I notice an issue with the namespace used for the generate types, currently, `OpAmp.Proto.V1` which makes it difficult for the compiler to resolve. We may need to consider changing the generated namespace, to something like `OpenTelemetry.OpAmp.Client.Internal.Proto,V1`, including `Internal` in the name, for consistency.

@martincostello Do you want to assign to me. What I propose is that I create a draft PR for one message as an example so we can discuss the API design. I'm specifically blocked by at least the server-to-agent remote config message so I plan to start there. Does that sound reasonable?

stevejgordon:
Fixes #3613

## Changes

- Expose RemoteConfigMessage on the public API
- Add supporting types for remote config
- Update unshipped public API


stevejgordon:
A few things for discussion:

This creates new public types for the properties of the message. Therefore this is some overhead in copying from the internal proto types. I think this is to be expected and fairly unavoidable.

This currently ignore CA1819 and exposes the raw bytes for the Body of the config. That feels like it makes sense as the consumer needs the bytes in order to handle the config in whatever format they expect. We could store the bytes in a field and consider exposing `ReadonlySpan<byte>` though.

```c#
/// <summary>
/// Represents an agent configuration file.
/// </summary>
public class AgentConfigFile
{
    private readonly byte[] bodyBytes;

    internal AgentConfigFile(global::OpAmp.Proto.V1.AgentConfigFile agentConfigFile)
    {
        this.bodyBytes = agentConfigFile.Body?.ToByteArray() ?? [];
        this.ContentType = agentConfigFile.ContentType;
    }

    /// <summary>
    /// Gets the raw bytes of the configuration file.
    /// </summary>
    public ReadOnlySpan<byte> Body => this.bodyBytes;

    /// <summary>
    /// Gets the MIME Content-Type that describes the data contained in the <see cref="Body"/>".
    /// </summary>
    public string? ContentType { get; }
}
```

At least for `AgentConfigFile` I left it unsealed as a consumer could create their own derived type that handles any deserialising and exposes their specific properties etc. We might want to seal other types, which I tend to do until there's an obvious case to unseal, but I'm not sure what we prefer in this repo overall.

At the moment, I've focused on making these ready only to support listeners for sever to agent messages. We may later want to make some properties settable for agent to server use cases.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigDictionary.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Messages;
+
+/// <summary>
+/// Represents a collection of agent configuration files, indexed by agent name.
+/// </summary>
+public class AgentConfigDictionary : Dictionary<string, AgentConfigFile>
```
martincostello: Is the default key comparer to use here appropriate? If so I suggest making it explicit.

Asking the question as if there was a use-case to use a different comparer, users wouldn't be able to change it as there's no public constructors to do so.

In that scenario it might be better to either not have the custom class, or to instead declare the property as `IDictionary<string, AgentConfigFile>` so other implementations could be used.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
```diff
@@ -0,0 +1,28 @@
...
+        this.Body = agentConfigFile.Body?.ToByteArray() ?? [];
+        this.ContentType = agentConfigFile.ContentType;
+    }
+
+    /// <summary>
+    /// Gets the raw bytes of the configuration file.
+    /// </summary>
+#pragma warning disable CA1819 // Properties should not return arrays
+    public byte[] Body { get; }
+#pragma warning restore CA1819 // Properties should not return arrays
```
martincostello: Are there any benefits to using an interface type if this is essentially a DTO, rather than requiring a concrete type?

Then we wouldn't need to add the suppression and if we opened up the setter people could use other types and we could handle converting it to the type needed for protobuf under the hood.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigDictionary.cs
stevejgordon: This is a basic stub for now and switching to a property is probably better. This POC was more about the concept than this specific implement which we can flesh out. Sorry, I probably wasn't clear on that.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
stevejgordon: Again, I didn't give this detail too much thought but I'm not sure if that overcomplicates things when the data is really just some bytes. It's kind of a low-level API due to the design of the remote config feature. In our Elastic use case for example, we'll have some bytes representing JSON the specifies our central config options. We'd therefore check the content type is `application/json` and pass the bytes into STJ to deserialise them into our type representing the configuration payload. I think most other use cases would be the same.

stevejgordon: We could use `ICollection<byte>` here, or as per https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3614#issuecomment-3646216091, we could just expose the ROS.

stevejgordon:
@martincostello I think my main question for now, is whether this approach to exposing the public message types seems generally reasonable in this general form. We can then shape the API for each type we need for message properties. 

We can then either, do a big bang to make all server-to-agent messages public, or do each one in a separate PR. I'd be quite keen to get remote config in sooner rather than later so we can experiement with using OpAMP for central config. That might identify other refinements we should do, before we focus on all message types.

martincostello:
It makes sense to me to separate the public API from the protobuf stuff, as that's essentially the same idea as not making EFCore data types public either. Otherwise we're effectively making the generated code public API, which is probably a recipe for long-term sadness.

stevejgordon:
@martincostello I've pushed a refined public API which I think is a nicer public API. `AgentConfigFile` now exposes a method to get the body bytes as well as a higher-performance API by exposing `TryGetBody`. There's no longer a specific dictionary type and instead `RemoteConfigMessage` just exposes `IReadOnlyDictionary<string, AgentConfigFile>`. I've made the comparer `Ordinal` which I think is reasonable. The spec doesn't specify anything around file name case sensitivity.

One consideration is testability for users when subscribing the messages. Right now, the ctor is internal and builds the message based on the protobuf types. We could consider providing an interface for each message to ease testability of subscribers? A risk is if the API expands, we end up modifying the interface and breaking those tests. Another option is designing this so the exposed properties can be set, but that's not straightforward for `AgentConfigFile` as we store the `ByteString` to back the methods and some properties. I don't really like exposing these publically as settable just for testing. Open to thoughts on this scenario.

If this looks good in principle, I can add some tests.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <summary>
+    /// Gets the name of this configuration file.
+    /// </summary>
+    public string Name { get; }
+
+    /// <summary>
+    /// Returns the configuration file body as a byte array.
+    /// </summary>
+    /// <returns>A byte array containing the contents of the message body. The array is empty if the body has no content.</returns>
+    public byte[] GetBodyBytes() => this.body.ToByteArray() ?? [];
```
martincostello: I wonder if `byte[]?` would be better to be able to distinguish between "not set" and "genuinely empty" (assuming such a distinction is meaningful here).

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
```diff
@@ -0,0 +1,72 @@
...
+            return true;
+        }
+
+        try
+        {
+            this.body.Span.CopyTo(destination);
+            bytesWritten = this.body.Length;
+            return true;
+        }
+        catch (ArgumentException)
```
martincostello: Maybe better to check the length to avoid the exception like [`Encoding.TryGetBytes()`](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs#L741-L752) does?

martincostello (commented):
LGTM so far - just a few inline thoughts.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
stevejgordon: It's unclear to me from the spec if this could ever be null. In theory, not as it should only send the config if there's something to send. My gut is that the expected scenario is there is always some data so avoiding making this nullable and requiring the user to null check is slightly more pleasant. The spec specifically calls out that the key may be empty, but doesn't specify scenarios were the config could be null.

stevejgordon:
@martincostello A few changes as per your suggestion. I've also exposed the config hash in the same way as for the body bytes. I've added unit test coverage also. I've marked ready for review and scoped this PR to specifically this message type. I've also raised #3617 to expose a way to confiure the client to identify that the client accepts this message which I'll tackle in a subsequent PR.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
stevejgordon: Resolved in the latest changes

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/AgentConfigFile.cs
Kielek: One topic to the public contract, I have missed in previous PR.

Shouldn't `ReadOnlySpan<byte>` be a better option here? the body, exposes Span property. We could avoid allocations.

It is related to other places in the public API

stevejgordon: @Kielek We did have that exposed at one point but I think at the time I'd missed that `ByteString` has a span directly accessible. I'll update as we can get rid of the methods and TryGet pattern, just relying on exposing the Span for consumers to do whatever they need with.

martincostello (approved):

stevejgordon:
@martincostello and @Kielek - The latest commit massively simplifies the public API and just exposes the ROS<byte>, letting the user use that as neccesary when their subscriber(s) handle the message.

martincostello (approved):

rajkumar-rangaraj (approved):

Kielek (approved):
LGTMRS

Issue #3574, opened 2025-12-01
[bug] Fix some edge cases in query sanitization for EFCore and SqlClient


martincostello:
### Component

OpenTelemetry.Instrumentation.SqlClient

### Package Version

| Package Name                          | Version |
| ------------------------------------- | ------- |
| OpenTelemetry.Instrumentation.EntityFrameworkCore | 1.14.0-beta.2 |
| OpenTelemetry.Instrumentation.SqlClient | 1.14.0-beta.1 |

### Runtime Version

Any

### Description

From discussion here https://github.com/dotnet/efcore/issues/29281#issuecomment-3597064740:

> I can see various issues in a casual look: for example, table and column names may include a single-quote, and that look like it would be identified as a literal to be sanitized in the current code. To compound the problem, different databases differ here: SQL Server allows quoting table/column names via brackets (so `[Table'WithSingleQuote]`), but MySQL uses backticks, etc. And that's just one problem.

### Steps to Reproduce

Use SQL syntax as described above.

### Expected Result

Queries are correctly parsed.

### Actual Result

Queries may be incorrectly parsed.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

alanwest:
Yes there are definitely going to be new cases we'll need to support for other database systems.

Are we at least good with SQL server? My original implementation was ok with SQL server's bracketed identifiers. We should validate that it is still ok.

For MySQL the backticked identifiers is definitely a thing. Postgres also has some different syntax for guids which will need to be handled. Oracle likely has unique needs as well.

martincostello:
This is probably the bit that needs checking we have coverage for and that it's working:

> table and column names may include a single-quote, and that look like it would be identified as a literal to be sanitized in the current code

stevejgordon:
@martincostello I can add some test cases for that specific example and see whether our code handles them. I'm unlikely to have cycles until W/C 15th though. Feel free to assign me and I'll try to take a proper look once some other priorities are completed.

stevejgordon:
[...]

PR #3616, opened 2025-12-12
Add tests for MSSQL bracketed identifers

alanwest:
Related to #3574

Adds a couple test cases for MSSQL's bracketed identifiers. This PR just illustrates status quo. When encountering a single quote in a bracketed identifier `db.query.text` and `db.query.summary` are prematurely truncated.

This issue is related to a [discussion](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3116#discussion_r2349205949) about making our parser overly aggressive and taking a stance to prioritize security over correctness.

I support making improvements to handle these type of scenarios more elegantly so long as we strive not to compromise security. I'd also support moving forward with releasing SqlClient instrumentation without addressing all edge cases. I suspect we'll continue to find new edge cases over time. These could be addressed in future improvements.

### DIFF
--- a/test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
+++ b/test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
@@ -883,5 +883,31 @@
       ],
       "db.query.summary": "SELECT TableA TableB"
     }
+  },
+  {
+    "name": "bracketed_identifier",
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "SELECT [a], [b] FROM [TableA], [TableB]"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT [a], [b] FROM [TableA], [TableB]"
+      ],
+      "db.query.summary": "SELECT TableA TableB"
+    }
+  },
+  {
+    "name": "bracketed_identifier_with_single_quote",
+    "input": {
+      "db.system.name": "other_sql",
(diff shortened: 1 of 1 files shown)


test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -883,5 +883,31 @@
...
+    "input": {
+      "db.system.name": "other_sql",
+      "query": "SELECT [a], [b] FROM [Table'A], [TableB]"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT [a], [b] FROM [Table"
+      ],
+      "db.query.summary": "SELECT Table"
+    }
```
martincostello, line 911: Rather than add a test case that passes when it shouldn't, it might be better to add metadata we can use in the tests to skip it because we know it's broken.

For example, a `"skip": "https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3574"` property?

Then the test can skip itself if the property is present with a non-empty value.

alanwest: I'm going to close this PR, but I think a skip property will be useful as we identify other cases we do not yet support.

alanwest:
Closing for now. A solution is in the works. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3574#issuecomment-3654327722

PR #3618, opened 2025-12-16
[OpAMP] Add remote config settings and ensure correct capabilities

### LINKED ISSUE #3617: [OpAMP] Provide settings to enable remote configuration messages
Add a settings type to represent remote configuration settings to support enabling this feature.

stevejgordon:
Fixes #3617

## Changes

- Introduce `RemoteConfigSettings` to allow configuring remote configuration support in the OpAMP client.
- Update capability advertisement logic to include `AcceptsRemoteConfig` when enabled.
- Add and update tests to verify capability flags and heartbeat behavior.
- Update unshipped public API.

## Design notes

For this initial PR, this exposes a single property on `RemoteConfigSettings` to identify that the client can handle remote configuration. The intent is that this can be later expanded once we expose APIs for sending the effictive config and remote config status.

I've also updated the advertised capabilities to not set the health or heartbeat flags when heartbeat is disabled based on an assumption of how that should work. I believe we should only advertise features we know the client implements (and are enabled) or consumers implement by confguring the client and registering subscribers.

I also wonder if we should avoid dispatching the `RemoteConfigMessage` to listeners if this feature is disabled. In theory, we should never receive it from the server unless we advertise the capability, but we could take a precaution to drop the message should it somehow arrive. This only really matters if for some reason a listener is registered for `RemoteConfigMessage` but the feature is disabled in the settings.

### Alternative design to consider

We may want to consider avoiding this extra `RemoteConfigSettings` type entirely and instead, we could potentially infer the capability based on registered subscribers. I've not prototyped that to know if it's definititely possible and it would also only work if all subscribed `IOpAmpListerner`s are added registered before `StartAsync` is called. In one respect, I like the idea that all a consumer would need to do is register the listener and our implementation works out the rest, but it might be a little too much magic. We may want to support binding configuration from other sources at some point so having a settings class we can bind to is useful there. If we don't include the settings type, it would also prevent a consumer from easily disabling the feature through code (temporily) but leaving their listener registered. 

### DIFF
--- a/test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTests.cs
+++ b/test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTests.cs
@@ -43,7 +43,7 @@ public class PlainHttpTransportTests
         var receivedTextData = clientReceivedFrames.First().CustomMessage.Data.ToStringUtf8();
 
         Assert.Single(serverReceivedFrames);
-        Assert.Equal(mockFrame.Uid, serverReceivedFrames.First().InstanceUid);
+        Assert.Equal(mockFrame.Uid, serverReceivedFrames[0].InstanceUid);
 
         Assert.Single(clientReceivedFrames);
         Assert.StartsWith("This is a mock server frame for testing purposes.", receivedTextData);
--- a/test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTests.cs
+++ b/test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTests.cs
@@ -42,8 +42,8 @@ public class PlainHttpTransportTests
         var clientReceivedFrames = mockListener.Messages;
         var receivedTextData = clientReceivedFrames.First().CustomMessage.Data.ToStringUtf8();
 
-        Assert.Single(serverReceivedFrames);
-        Assert.Equal(mockFrame.Uid, serverReceivedFrames[0].InstanceUid);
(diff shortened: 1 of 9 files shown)


test/OpenTelemetry.OpAmp.Client.Tests/PlainHttpTransportTests.cs
```diff
@@ -43,7 +43,7 @@ public async Task PlainHttpTransport_SendReceiveCommunication(bool useSmallPacke
         var receivedTextData = clientReceivedFrames.First().CustomMessage.Data.ToStringUtf8();
 
         Assert.Single(serverReceivedFrames);
-        Assert.Equal(mockFrame.Uid, serverReceivedFrames.First().InstanceUid);
+        Assert.Equal(mockFrame.Uid, serverReceivedFrames[0].InstanceUid);
```
martincostello: This could be this I think:

```csharp
var frame = Assert.Single(serverReceivedFrames);
Assert.Equal(mockFrame.Uid, frame.InstanceUid);
```

martincostello (approved):

stevejgordon:
@Kielek If this looks okay, I'm wondering if we can get a new alpha release with the changes so far? I'm on leave for two weeks starting Monday so I'll pick up the other message types etc. when I'm back to work.

Kielek (approved):

Kielek:
This PR should be merged shortly. When it will be done, please create issues based on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/new?template=release_request.yml

PR #3626, opened 2025-12-17
[EFCore] Fix integration test

martincostello:
## Changes

Remove integration test for SQL Server using the old conventions after changes in #3592.

The test only runs on Linux, so is easy to overlook if not changing the EFCore files and working on Windows locally.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [ ] ~~Appropriate `CHANGELOG.md` files updated for non-trivial changes~~
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #3627, opened 2025-12-17
Support escaped identifiers in `SqlProcessor`

### LINKED ISSUE #3574: [bug] Fix some edge cases in query sanitization for EFCore and SqlClient

stevejgordon:
Fixes #3574

## Changes

- Update parsing logic to handle escaped identifiers
- Add test cases to validate parsing of escaped identifiers
- Add benchmark instructions
- Run benchmarks before and after changes
- Reduce string allocations for sanitized SQL when it matches the original input
- Minor performance optimisations to reduce impact of changes

NOTE: This changes adds a small amount of execution time overhead. I've tried to counter that with some other minor optimisations. It's still a few 10s of nanonseconds slower but I think this is acceptable and unavoidable.

I did however have a bit of an apithany and realised we can reduce string allocations quite heavily for statements where there are no literals to sanitise. We simply return the original string instance when we have not sanitized anything.

### Before
```
| Method          | Sql                  | Mean     | Error   | StdDev  | Allocated |
|---------------- |--------------------- |---------:|--------:|--------:|----------:|
| GetSanitizedSql | CREAT(...)s(Id) [56] | 229.2 ns | 1.90 ns | 1.78 ns |     208 B |
| GetSanitizedSql | DELET(...) = 42 [32] | 156.0 ns | 0.84 ns | 0.71 ns |     128 B |
| GetSanitizedSql | INSER(...)3e-5) [76] | 269.6 ns | 1.94 ns | 1.72 ns |     192 B |
| GetSanitizedSql | SELEC(...)ls od [39] | 172.7 ns | 3.40 ns | 4.54 ns |     184 B |
| GetSanitizedSql | SELE(...)tory [111]  | 261.3 ns | 4.58 ns | 5.63 ns |     424 B |
| GetSanitizedSql | SELEC(...)table [69] | 121.7 ns | 2.43 ns | 3.32 ns |     128 B |
| GetSanitizedSql | SELEC(...) c.Id [74] | 247.3 ns | 3.98 ns | 3.73 ns |     248 B |
| GetSanitizedSql | SELE(...)_id) [101]  | 292.2 ns | 3.28 ns | 2.90 ns |     312 B |
| GetSanitizedSql | UPDAT(...) = 42 [44] | 178.9 ns | 2.44 ns | 2.28 ns |     144 B |
```

### After
```
| Method          | Sql                  | Mean     | Error   | StdDev  | Allocated |
|---------------- |--------------------- |---------:|--------:|--------:|----------:|
| GetSanitizedSql | CREAT(...)s(Id) [56] | 250.9 ns | 1.01 ns | 0.84 ns |      72 B |
| GetSanitizedSql | DELET(...) = 42 [32] | 169.7 ns | 1.34 ns | 1.19 ns |     128 B |
| GetSanitizedSql | INSER(...)3e-5) [76] | 312.7 ns | 2.03 ns | 1.90 ns |     192 B |
| GetSanitizedSql | SELEC(...)ls od [39] | 202.3 ns | 1.17 ns | 1.04 ns |      80 B |
| GetSanitizedSql | SELE(...)tory [111]  | 354.8 ns | 2.45 ns | 2.29 ns |     176 B |
| GetSanitizedSql | SELEC(...)table [69] | 150.3 ns | 1.37 ns | 1.22 ns |     128 B |
| GetSanitizedSql | SELEC(...) c.Id [74] | 316.1 ns | 1.30 ns | 1.22 ns |      72 B |
| GetSanitizedSql | SELE(...)_id) [101]  | 375.6 ns | 1.84 ns | 1.72 ns |      88 B |
| GetSanitizedSql | UPDAT(...) = 42 [44] | 206.1 ns | 1.26 ns | 1.11 ns |     144 B |
```

@martincostello / @alanwest - What (if any) changelog should we update?


stevejgordon:
@martincostello An EF Core test is failing. I may not get chance to look at that today so if you have any ideas, please let me know!

stevejgordon:
@martincostello I think this failure is unrelated to the changes here. As far as I can see the `SqlClientDiagnosticListener` hard codes the use of new attributes [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs#L158-L171). Which in turn means the span name used the summary "select" rather than the database name "master".

martincostello:
> @martincostello An EF Core test is failing. I may not get chance to look at that today so if you have any ideas, please let me know!

#3626

src/Shared/SqlProcessor.cs
```diff
@@ -154,20 +155,56 @@ private static bool IsAsciiDigit(char c) =>
...
+            // Double closing brackets (]]) are treated as an escaped bracket within the identifier.
+            // A single closing bracket ends the identifier.
+            if (currentChar == '\0')
+            {
+                return false;
+            }
+
+            if (currentChar == CloseSquareBracketChar)
+            {
+                var nextChar = currentPosition + 1 < sql.Length ? sql[currentPosition + 1] : '\0';
```
martincostello: Rather than use the null character as a surrogate for the equality to make it fail, just return false if it would be out of bounds?

src/Shared/SqlProcessor.cs
```diff
@@ -233,8 +270,13 @@ private static SqlStatementInfo SanitizeSql(ReadOnlySpan<char> sql)
             }
         }
 
+        var sanitizedSqlSpan = buffer.Slice(0, state.SanitizedPosition);
+
+        // If the sanitized SQL is identical to the input SQL, we can reuse the original string instance.
+        var sanitizedSql = sanitizedSqlSpan.SequenceEqual(sqlSpan) ? sql : sanitizedSqlSpan.ToString();
```
martincostello, line 279: Nice optimisation.

src/Shared/SqlProcessor.cs
```diff
@@ -410,31 +455,39 @@ private static void ParseNextToken(
...
-
                 // Special handling: if we are in a FROM clause, check if this identifier is a reserved keyword
                 // that indicates the end of the FROM clause.
                 if (state.InFromClause)
                 {
-                    foreach (var reservedKeyword in FromClauseReservedKeywords)
+                    var isReservedKeyword = false;
+
+                    // Range of shortest ("BY") to longest ("TABLESAMPLE") keywords
+                    if (length >= 2 && length <= 11)
```
martincostello: Would the compiler optimise this down to int constants so it could be slightly more self-explanatory?

```csharp
if (length >= "BY".Length && length <= "TABLESAMPLE".Length)
```

src/Shared/SqlProcessor.cs
```diff
@@ -443,35 +496,44 @@ private static void ParseNextToken(
...
+            if (state.InEscapedIdentifier && currentChar == CloseSquareBracketChar)
+            {
+                state.InEscapedIdentifier = false;
+            }
+
             // If we are in a FROM clause, we want to capture the next identifier following a comma or open square bracket.
             // Commas may occur when listing multiple tables in a FROM clause.
             // Brackets may occur when using schema-qualified or delimited identifiers.
-            state.CaptureNextTokenInSummary = state.InFromClause && (currentChar == CommaChar || currentChar == OpenSquareBracketChar);
+            state.CaptureNextNonKeywordTokenAsIdentifier = state.InFromClause && (currentChar == CommaChar || currentChar == OpenSquareBracketChar);
```
martincostello: I _think_ this is equivalent:

```suggestion
            state.CaptureNextNonKeywordTokenAsIdentifier = state.InFromClause && (currentChar is CommaChar or OpenSquareBracketChar);
```

test/OpenTelemetry.Contrib.Shared.Benchmarks/OpenTelemetry.Contrib.Shared.Benchmarks.sln
martincostello, line 1: Convert to `.slnx`.

stevejgordon: I'm not sure why we had a solution file so I've removed it for now. We don't in other benchmark folders.

src/Shared/SqlProcessor.cs
stevejgordon: Updated as suggested in my latest commit.

src/Shared/SqlProcessor.cs
stevejgordon: I've taken a slightly different approach in my latest commit. We now compute this statically once based on the array values. That way, if we ever add anything, we don't need to remember to update the values.

src/Shared/SqlProcessor.cs
stevejgordon: I've refactored the logic in my latest commit.

martincostello (approved):

martincostello:
> What (if any) changelog should we update?

Missed this the first time. I guess we should update the EFCore and SqlClient CHANGELOGs to note there's been miscellaneous fixes for the SQL sanitization.

martincostello:
IDE suggestion I saw while I had this locally for #3628, the last `return` can be simplified to:

```csharp
return currentKeyword.SqlKeyword is SqlKeyword.Select
    && state.FirstSummaryKeyword is not SqlKeyword.Create
    && state.PreviousParsedKeyword?.SqlKeyword is not SqlKeyword.Union;
```

stevejgordon:
@martincostello I've updated with your suggestion + changelogs

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Improve SQL parsing for sanitization and summary generation. Support additional cases
+  for escaped identifiers. Optimize performance of parsing logic.
```
martincostello:
```suggestion
* Improve SQL parsing for sanitization and summary generation. Support additional
  cases for escaped identifiers. Optimize performance of parsing logic.
```

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   the stable conventions. Now, only the stable conventions are supported.
   ([#3592](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3592))
 
+* Improve SQL parsing for sanitization and summary generation. Support additional cases
+  for escaped identifiers. Optimize performance of parsing logic.
```
martincostello:
```suggestion
* Improve SQL parsing for sanitization and summary generation. Support additional
  cases for escaped identifiers. Optimize performance of parsing logic.
```

test/OpenTelemetry.Contrib.Shared.Tests/SqlProcessorAdditionalTestCases.json
```diff
@@ -883,5 +883,135 @@
...
+    "name": "select_from_table_with_space_in_name",
+    "input": {
+      "db.system.name": "microsoft.sql_server",
+      "query": "SELECT * FROM [Order Details]"
+    },
+    "expected": {
+      "db.query.text": [
+        "SELECT * FROM [Order Details]"
+      ],
+      "db.query.summary": "SELECT Order Details"
```
alanwest: This is a relatively minor thing in my opinion, but when a query includes bracketed identifiers would it be difficult to keep them in the query summary?

The value of keeping the brackets is particularly evident in this scenario where the identifier has spaces in it.

stevejgordon: That's a great point! I've updated to logic so that we include the brackets around all escaped identifiers.

PR #3410, opened 2025-11-07
Rework internal logic for determining the trace id created by AWS Lambda

normj:
## Changes

In `Amazon.Lambda.Core` we have reworked the logic that for determining the trace id. The `OpenTelemetry.Instrumentation.AWSLambda` package has been updated to take advantage of the new API added in `Amazon.Lambda.Core` for getting the current trace id.

I didn't add any new tests as the existing tests covered the the internal change and users will not see any difference.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #3629, opened 2025-12-19
Fix issue with ADOT auto-instrumentation and AWS Lambda Trace Id

normj:
## Fixes
Fix issue when using ADOT auto-instrumentation which includes a version of OpenTelemetry.Instrumentation.AWSLambda in the Lambda layer. In this scenario nothing is forcing the NuGet resolution to include the updated version of Amazon.Lambda.Core that has the [reworked logic](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3410) for determining the trace id.

This is causing us some breaking issues with ADOT and are looking to get this fix released as soon as possible.

## Changes
Add defensive code around accessing the `LambdaTraceProvider.CurrentTraceId` and fallback to the environment variable if we get a TypeLoadException. We need to try and use `LambdaTraceProvider.CurrentTraceId` to support Lambda's recently released Managed Instances where there is concurrency and an environment variable will not work. For users using Managed Instances feature they are required to use the updated version of Amazon.Lambda.Core in their Lambda function.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
@@ -32,6 +32,11 @@ internal class AWSLambdaUtils
         return headers.TryGetValue(name, out var value) ? [value] : [];
     };
 
+    // Added volatile for abundance of caution because in theory, particular in using Lambda's Mananged Instances,
+    // during initial startup multiple invocations invoking GetXRayParentContext() at the same time before the first
+    // TraceProviderIsolated.CurrentTraceId is called.
+    private static volatile bool traceProviderIsolatedFailed;
+
     private readonly AWSSemanticConventions semanticConventionBuilder;
 
     public AWSLambdaUtils(AWSSemanticConventions semanticConventionBuilder)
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
```diff
@@ -40,7 +41,19 @@ public AWSLambdaUtils(AWSSemanticConventions semanticConventionBuilder)
...
-        var tracerHeaderValue = LambdaTraceProvider.CurrentTraceId;
+        string? tracerHeaderValue;
+        try
+        {
+            // Prefer the TraceProviderIsolated.CurrentTraceId over the environment variable to support
+            // AWS Lambda's Managed Instances feature. With Managed Instances there are multiple invocations
+            // per execution and environment variable is not set since there isn't a single value to set to.
+            tracerHeaderValue = TraceProviderIsolated.CurrentTraceId;
+        }
+        catch (TypeLoadException)
```
martincostello, line 65: Is there a way to cache which of the two approaches works for a given process? Otherwise I think we'd take a perf hit for each call from the try-catch?

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/TraceProviderIsolated.cs
```diff
@@ -0,0 +1,32 @@
...
+/// will be thrown when parent calling method is called. This gives the parent calling method a chance to catch the
+/// <see cref="TypeLoadException"/> and handle it appropriately.
+/// </summary>
+/// <remarks>
+/// This situation where an older version of Amazon.Lambda.Core is being used can happen when the Lambda function
+/// is being auto-instrumented and the version of OpenTelemetry.Instrumentation.AWSLambda
+/// is not added explicitly to the Lambda function. In this scenario normal NuGet dependency resolution is not done
+/// forcing a version of Amazon.Lambda.Core that will have the LambdaTraceProvider.CurrentTraceId property.
+/// </remarks>
+internal class TraceProviderIsolated
```
martincostello:
```suggestion
internal static class TraceProviderIsolated
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs
normj: Added a tracking boolean if calling `TraceProviderIsolated.CurrentTraceId` failed in the past.

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/TraceProviderIsolated.cs
normj: Done

normj:
@martincostello Addressed your comments

PR #3637, opened 2025-12-19
[Exporter.Geneva] Adding flag to disable enriching spans with httpUrl if the flag is set

mariamgerges:
## Changes

Added an environment flag that enables or disables adding `httpUrl` to the produced Http server spans from `OpenTelemetry.Exporter.Geneva`

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -13,6 +13,9 @@ Released 2026-Jan-21
 * Updated OpenTelemetry core component version(s) to `1.15.0`.
   ([#3721](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3721))
 
+* Added an environment variable `HTTP_URL_IS_SET` check to `OpenTelemetry.Exporter.Geneva` 
+  to enable/disable adding `httpUrl`for HTTP server spans.
+
 ## 1.14.0
 
 Released 2025-Nov-13


cijothomas:
Not sure if this should be an Exporter setting really... Can you elaborate the issue this is attempting to solve?

mariamgerges:
> Not sure if this should be an Exporter setting really... Can you elaborate the issue this is attempting to solve?

sorry for the late reply since I was OOF, we need this flag to be able to set the `httpUrl` from our SDK, since we apply redaction to the url, so we need to be able to set it. this flag allows us to set it, instead of it being set by the exporter.

evgenyfedorov2:
@rajkumar-rangaraj could you please take a look?

evgenyfedorov2:
> Not sure if this should be an Exporter setting really... Can you elaborate the issue this is attempting to solve?

Hi @cijothomas! Exporter forcibly adds `httpUrl` already, we are not changing that (we are also not challenging the fact that it is added at the exporter layer, maybe instrumentation layer was more appropriate, but it is not what we are discussing here). The thing we are trying to achieve here is to opt out from getting `httpUrl` added! That's the only thing - just let us opt out.
Otherwise, especially internally at Microsoft, it is not possible to use the whole Geneva exporter because `httpUrl` may contain sensitive data (unredacted). There is no way to redact it either because it is added at the exporter layer which is too late in the pipeline!

cijothomas:
> > Not sure if this should be an Exporter setting really... Can you elaborate the issue this is attempting to solve?
> 
> Hi @cijothomas! Exporter forcibly adds `httpUrl` already, we are not changing that (we are also not challenging the fact that it is added at the exporter layer, maybe instrumentation layer was more appropriate, but it is not what we are discussing here). The thing we are trying to achieve here is to opt out from getting `httpUrl` added! That's the only thing - just let us opt out. Otherwise, especially internally at Microsoft, it is not possible to use the whole Geneva exporter because `httpUrl` may contain sensitive data (unredacted). There is no way to redact it either because it is added at the exporter layer which is too late in the pipeline!

Got it. Looks like you need an optOut for `httpUrl`. Might be better to expand ExporterOptions to support such capability to opt-out of certain fields. (I have seen similar asks for Logs too - users want some fields to be never emitted.)
Introducing ad-hoc ENV variable might not be a good strategy.

mariamgerges:
@rajkumar-rangaraj can you please check?

rajkumar-rangaraj:
>Otherwise, especially internally at Microsoft, it is not possible to use the whole Geneva exporter because httpUrl may contain sensitive data (unredacted). There is no way to redact it either because it is added at the exporter layer which is too late in the pipeline!

This redaction won't be enabled by default - customers would need to explicitly opt-in either at the instrumentation level or exporter level (if we introduce it). It's better to have this control in one place. If the plan is to change the default behavior, then it makes sense to add it in the exporter.

@mariamgerges Instead of adding a flag to suppress `httpUrl`, we could replace the component attribute values (`url.scheme`, `server.address`, `url.path`) with redacted strings at the instrumentation layer using Filter/Enrich. The exporter will then construct `httpUrl` from the already-redacted values.

This approach keeps the data visible (redacted rather than missing) which is better for supportability, and aligns with SDK design guidance.

PR #3646, opened 2025-12-30
Geneva exporter: Support resource attributes in log exporter

mattsains:
Related to https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3214 

## Changes
* Allow specifying resource attributes to export in Geneva Log exporter.
* Changed behaviour of trace exporter to make resource attributes and prepopulated fields mutually exclusive

* [] Changes in public API reviewed (if applicable)

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
@@ -9,6 +9,7 @@ using OpenTelemetry.Exporter.Geneva.MsgPack;
 using OpenTelemetry.Exporter.Geneva.Tld;
 using OpenTelemetry.Internal;
 using OpenTelemetry.Logs;
+using OpenTelemetry.Resources;
 
 namespace OpenTelemetry.Exporter.Geneva;
 
@@ -93,7 +94,7 @@ public class GenevaLogExporter : GenevaBaseExporter<LogRecord>
             var msgPackLogExporter = new MsgPackLogExporter(options, () =>
             {
                 // this is not equivalent to passing a method reference, because the ParentProvider could change after the constructor.
-                return this.ParentProvider.GetResource();
+                return this.ParentProvider?.GetResource() ?? Resource.Empty;
             });
             this.IsUsingUnixDomainSocket = msgPackLogExporter.IsUsingUnixDomainSocket;
             this.exportLogRecord = msgPackLogExporter.Export;
--- a/src/OpenTelemetry.Exporter.Geneva/GenevaTraceExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/GenevaTraceExporter.cs
@@ -6,6 +6,7 @@ using System.Runtime.InteropServices;
 using OpenTelemetry.Exporter.Geneva.MsgPack;
 using OpenTelemetry.Exporter.Geneva.Tld;
 using OpenTelemetry.Internal;
+using OpenTelemetry.Resources;
 
 namespace OpenTelemetry.Exporter.Geneva;
 
@@ -72,7 +73,7 @@ public class GenevaTraceExporter : GenevaBaseExporter<Activity>
             var msgPackTraceExporter = new MsgPackTraceExporter(options, () =>
             {
                 // this is not equivalent to passing a method reference, because the ParentProvider could change after the constructor.
-                return this.ParentProvider.GetResource();
+                return this.ParentProvider?.GetResource() ?? Resource.Empty;
             });
             this.IsUsingUnixDomainSocket = msgPackTraceExporter.IsUsingUnixDomainSocket;
             this.exportActivity = msgPackTraceExporter.Export;


mattsains:
> A while ago we decided that the resource attributes service.name and service.instanceId would be mapped to the message pack data even if they are not present in ResourceFieldNames. I was just thinking and realised that if someone had set cloud.role or cloud.roleInstance in PrepopulatedFields, then after my change, they will not have their logs exported, because the prepopulated field will conflict with the "built-in" resource attribute mapping for service.name and service.instanceId (which gets mapped to cloud.role and cloud.roleInstance even if you didn't specify it in ResourceFieldNames)
> 
> what do you think I should do? I can check to see if it's in resource attributes before I accept it as a prepopulated field, and then I guess just drop the prepopulated field?

I have updated the code and the tests to replace a prepopulated field with the equivalent auto-mapped resource attribute (ie., service.name, service.instanceId). I also added a test for this behaviour

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -227,13 +336,11 @@ internal ArraySegment<byte> SerializeLogRecord(LogRecord logRecord)
         ushort cntFields = 0;
         var idxMapSizePatch = cursor - 2;
 
-        if (this.prepopulatedFieldKeys != null)
+        if (this.prepopulatedFields != null)
         {
-            for (var i = 0; i < this.prepopulatedFieldKeys.Count; i++)
+            foreach (var field in this.prepopulatedFields)
```
rajkumar-rangaraj, line 363: Existing implementation uses `for` loop to avoid the allocation from `foreach`.

mattsains:
We talked more about the prepopulated/resource attribute thing, and decided that the mapping can only happen when prepopulated fields have not been specified. Otherwise, the default resource will silently overwrite prepopulated values.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
mattsains: as discussed, this isn't a concern because this foreach is unlikely to have a heap allocation

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
```diff
@@ -88,21 +95,34 @@ public MsgPackLogExporter(GenevaExporterOptions options)
...
         if (options.PrepopulatedFields != null)
         {
-            this.prepopulatedFieldKeys = [];
-            var tempPrepopulatedFields = new Dictionary<string, object>(options.PrepopulatedFields.Count, StringComparer.Ordinal);
+            this.prepopulatedFields = new Dictionary<string, object>(options.PrepopulatedFields.Count, StringComparer.Ordinal);
             foreach (var kv in options.PrepopulatedFields)
             {
-                tempPrepopulatedFields[kv.Key] = kv.Value;
-                this.prepopulatedFieldKeys.Add(kv.Key);
+                this.prepopulatedFields[kv.Key] = kv.Value;
```
xiang17: Why removing FrozenDictionary? (It was introduced for some performance improvement #1504)

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
xiang17: If I recall correctly, `foreach` inherently cause heap allocation.

It's a major goal to keep heap allocation as low as possible for Geneva exporter. Can you run a benchmark test in [Benchmark folder](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/test/OpenTelemetry.Exporter.Geneva.Benchmarks) to confirm there is no heap allocation introduced by `foreach`? (PR #1504 also included a mini benchmark code)

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
mattsains: The dictionary is no longer read-only, see AddResourceAttributesToPrepopulated which is called during the first Export per thread

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
mattsains: I can't find any benchmarking code in that PR but if you can point me to a specific benchmark test I'd be happy to run it and post the results here

xiang17: There are two places you can reference:

1. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/593fb31da8c9f5168fdda9f22f5092be7dfff49c/test/OpenTelemetry.Exporter.Geneva.Benchmarks/README.md?plain=1#L3
2. Ctrl+F "Log Benchmarks Code" on page https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/1504 and click on it

mattsains: Oops, I don't know how I missed that collapseable section in the PR.

Here's the results of the benchmark on my branch:

| Method                    | IncludeFormattedMessage | Mean       | Error    | StdDev   | Gen0   | Allocated |
|-|-|-|-|-|-|-|
| LoggerWithMessageTemplate | False                   |   944.3 ns |  8.32 ns |  7.78 ns |      - |     104 B | 
| LoggerWithDirectLoggerAPI | False                   | 1,136.2 ns | 10.99 ns |  8.58 ns | 0.0057 |     320 B | 
| LoggerWithSourceGenerator | False                   |   933.9 ns | 11.04 ns | 10.32 ns |      - |      64 B | 
| SerializeLogRecord        | False                   |   316.7 ns |  1.98 ns |  1.76 ns |      - |         - | 
| Export                    | False                   |   638.5 ns |  9.73 ns |  7.59 ns |      - |         - | 
| LoggerWithMessageTemplate | True                    |   961.9 ns |  8.06 ns |  7.54 ns | 0.0010 |     104 B | 
| LoggerWithDirectLoggerAPI | True                    | 1,117.1 ns |  6.50 ns |  5.76 ns | 0.0057 |     320 B |
| LoggerWithSourceGenerator | True                    |   929.5 ns |  8.82 ns |  8.25 ns |      - |      64 B |
| SerializeLogRecord        | True                    |   315.9 ns |  0.87 ns |  0.73 ns |      - |         - | 
| Export                    | True                    |   641.1 ns |  7.51 ns |  5.86 ns |      - |         - |

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
reyang: I guess this can be done by creating a new FrozenDictonary and replacing the existing one?

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Add ResourceFieldNames to filter resource attributes to send to Geneva
   ([#3552](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3552))
+  ([#3646](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3646))
```
reyang:
```suggestion
  ([#3552](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3552),
   [#3646](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3646))
```

Check line 207.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
mattsains: I've updated the code to use a frozen dictionary

xiang17 (approved):

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackLogExporter.cs
xiang17: I also ran the tests on main branch. This confirmed that the PR didn't increase memory allocation. Interestingly, the memory allocation for .NET 10 is higher than .NET 8 (both on main branch):

## .NET 8

| Method                    | IncludeFormattedMessage | Mean       | Error    | StdDev   | Gen0   | Allocated |
|-------------------------- |------------------------ |-----------:|---------:|---------:|-------:|----------:|
| LoggerWithMessageTemplate | False                   | 1,142.1 ns | 22.74 ns | 42.16 ns | 0.0038 |     104 B |
| LoggerWithDirectLoggerAPI | False                   | 1,331.4 ns | 26.41 ns | 36.15 ns | 0.0114 |     320 B |
| LoggerWithSourceGenerator | False                   | 1,122.7 ns | 17.91 ns | 28.41 ns | 0.0019 |      64 B |
| SerializeLogRecord        | False                   |   434.3 ns |  8.42 ns |  8.27 ns |      - |         - |
| Export                    | False                   |   773.8 ns | 15.53 ns | 35.04 ns |      - |         - |
| LoggerWithMessageTemplate | True                    | 1,132.5 ns | 22.65 ns | 44.71 ns | 0.0038 |     104 B |
| LoggerWithDirectLoggerAPI | True                    | 1,285.8 ns | 24.61 ns | 23.02 ns | 0.0114 |     320 B |
| LoggerWithSourceGenerator | True                    | 1,129.4 ns | 22.21 ns | 24.69 ns | 0.0019 |      64 B |
| SerializeLogRecord        | True                    |   434.2 ns |  5.31 ns |  4.71 ns |      - |         - |
| Export                    | True                    |   769.2 ns | 15.25 ns | 31.84 ns |      - |         - |

## .NET 10

| Method                    | IncludeFormattedMessage | Mean       | Error    | StdDev   | Median     | Gen0   | Allocated |
|-------------------------- |------------------------ |-----------:|---------:|---------:|-----------:|-------:|----------:|
| LoggerWithMessageTemplate | False                   | 1,042.3 ns | 20.88 ns | 60.58 ns | 1,045.5 ns | 0.0038 |     112 B |
| LoggerWithDirectLoggerAPI | False                   | 1,242.6 ns | 24.58 ns | 47.36 ns | 1,222.2 ns | 0.0114 |     320 B |
| LoggerWithSourceGenerator | False                   | 1,063.2 ns | 21.24 ns | 44.33 ns | 1,051.1 ns | 0.0019 |      64 B |
| SerializeLogRecord        | False                   |   383.3 ns |  7.46 ns |  9.70 ns |   379.8 ns |      - |         - |
| Export                    | False                   |   745.4 ns | 14.88 ns | 32.35 ns |   741.4 ns |      - |         - |
| LoggerWithMessageTemplate | True                    | 1,102.1 ns | 31.75 ns | 92.60 ns | 1,082.8 ns | 0.0038 |     112 B |
| LoggerWithDirectLoggerAPI | True                    | 1,219.1 ns | 24.17 ns | 33.08 ns | 1,213.6 ns | 0.0114 |     320 B |
| LoggerWithSourceGenerator | True                    | 1,069.4 ns | 23.99 ns | 69.61 ns | 1,065.7 ns | 0.0019 |      64 B |
| SerializeLogRecord        | True                    |   369.3 ns |  4.97 ns |  4.41 ns |   368.6 ns |      - |         - |
| Export                    | True                    |   763.3 ns | 15.09 ns | 41.57 ns |   755.1 ns |      - |         - |

reyang: Interesting, could you look into where the delta is coming from in .NET 10 comparing to 8?

xiang17: Sure.

xiang17: The 8 bytes come from the size increase of `Microsoft.Extensions.Logging.FormattedLogValues`, which is created via `new` in [LoggerExtensions.LogInformation](https://github.com/dotnet/runtime/blob/3e773872a9535e5c5739d379e09d6ef3e7684906/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerExtensions.cs#L510).

I checked the source code: in .NET 10, an additional field `_cachedToString` of type `string?` was introduced in `FormattedLogValues`: [.NET 10 code](https://github.com/dotnet/runtime/blob/release/10.0/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/FormattedLogValues.cs#L28) vs [.NET 8 code](https://github.com/dotnet/runtime/blob/release/8.0/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/FormattedLogValues.cs#L27). This field is a reference, so it's 8 bytes on 64-bit CLR.

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
```diff
@@ -90,17 +90,21 @@ public GenevaLogExporter(GenevaExporterOptions options)
 
         if (useMsgPackExporter)
         {
-            var msgPackLogExporter = new MsgPackLogExporter(options);
+            var msgPackLogExporter = new MsgPackLogExporter(options, () =>
```
rajkumar-rangaraj, line 94: Should this have a null check? If `ParentProvider` hasn't been set when export is called (possible during SDK initialization edge cases), this will throw NRE. Consider:

```
() => this.ParentProvider?.GetResource() ?? Resource.Empty
```

mattsains-msft: I'll change it but I notice that if ParentProvider can really be null, then the existing metric code is also broken: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Metrics/GenevaMetricExporter.cs#L112

rajkumar-rangaraj (approved):

src/OpenTelemetry.Exporter.Geneva/GenevaLogExporter.cs
reyang: What does this mean? Is it a bug or not?

rajkumar-rangaraj: This is not a bug; the wording may have caused the confusion.
In the existing implementation, `this.ParentProvider.GetResource()` itself already returns `Resource.Empty` when no provider is set, so it is safe. We don’t need to add an explicit null check here or manually set `Resource.Empty` again.

reyang: Thanks @rajkumar-rangaraj! Then this sounds wrong and should be reverted back.

> I'll change it

rajkumar-rangaraj: Yes, this needs a fix. I will send a follow-up PR.

PR #3676, opened 2026-01-09
[Exporter.Geneva] Fix exporter histogram negative values

### LINKED ISSUE #912: Histogram with negative values

* OpenTelemetry, 1.4.0-rc.2
* OpenTelemetry.Exporter.Geneva, 1.4.0-rc.1

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can find this information from the `*.csproj` file):
* net7.0

* [ ] Feature Request
* [x] Bug

Allow negative values in histograms

```cs
var m = new Meter("m");
var h = m.CreateHistogram<long>("h");
h.Record(-1);
```

Exception:
```
System.OverflowException: Arithmetic operation resulted in an overflow.
   at System.Convert.ToUInt64(Double value)
   at OpenTelemetry.Exporter.Geneva.GenevaMetricExporter.Export(Batch`1& batch)
```

## Additional Context

Add any other context about the feature request here.

Dytroxx:
Fixes #912.

Geneva metric TLV histogram serialization no longer throws when histogram aggregates include negative values. The exporter switches to a double-based payload for sum/min/max when needed, and uses a safe encoding for bucket bounds when bounds are not uint64-compatible.

Tests updated and a new negative-values histogram test added.


martincostello:
When #3828 is merged, this PR should rebased and extend the fuzz tests to cover the new code.

martincostello:
> When #3828 is merged, this PR should rebased and extend the fuzz tests to cover the new code.

This has been merged now.

Dytroxx:
Update: Currently waiting for triage on this https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/3866
Have to have clear definition how to proceed regarding the precision and handling.

Dytroxx:
The previous implementation addressed negative overflow, but still allowed precision loss through truncation in some histogram paths. This change removes that truncation behavior and preserves aggregate precision.

## Problem
The old behavior could produce incorrect histogram aggregates in two ways:

1. Negative values could not be safely represented in the UInt64 aggregate payload path.
2. Positive non-integer values could be truncated when converted to UInt64.

## What changed
1. Histogram aggregate payload selection is now based on UInt64 compatibility, not only on negativity.
2. Payload 8 (UInt64) is used only when sum/min/max are all:
- finite
- non-negative
- integral
- within UInt64 range

3. Payload 9 (Double) is used for:
- negative values
- non-integer values
- NaN/Infinity

4. Bucket-bound encoding remains independent:
- Payload 12 for UInt64-compatible bounds
- Payload 13 for non-UInt64-compatible bounds

Mixed mode is valid and supported (for example: UInt64 aggregates with double-encoded bounds).

## Downstream compatibility:
Consumers that implicitly assumed histogram aggregates would always arrive as payload 8 may need to correctly handle payload 9 as well.
This is a behavior change at the wire level for affected data, but it is required to avoid incorrect values.

## Tests
Test coverage added/updated

martincostello:
@rajkumar-rangaraj Could you take a look at this one please, as it's a bit stuck without your input as the SME for Geneva.

rajkumar-rangaraj:
Closing this PR as TLV is not the active path.

martincostello:
@rajkumar-rangaraj does that mean we should close #912 as "won't do"?

PR #3716, opened 2026-01-20
[OpAMP] Add effective config reporting

RassK:
## Changes

Adds possibility to report agent's currently effective config / config files.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/Settings/EffectiveConfigSettings.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Settings/EffectiveConfigSettings.cs
@@ -12,8 +12,8 @@ public sealed class EffectiveConfigSettings
     /// Gets or sets a value indicating whether the client can report effective configuration.
     /// </summary>
     /// <value>
-    /// <c>true</c> if effective configuration reporting is enabled; otherwise, <c>false</c>.
-    /// Default is <c>false</c>.
+    /// <see langword="true"/> if effective configuration reporting is enabled; otherwise, <see langword="false"/>.
+    /// Default is <see langword="false"/>.
     /// </value>
     public bool EnableReporting { get; set; }
 }


src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -150,6 +151,33 @@ IFrameBuilder IFrameBuilder.AddCapabilities()
         return this;
     }
 
+    IFrameBuilder IFrameBuilder.AddEffectiveConfig(IEnumerable<EffectiveConfigFile> files)
+    {
+        this.EnsureInitialized();
+
+        var configMap = new AgentConfigMap();
+        var fileMap = new Dictionary<string, global::OpAmp.Proto.V1.AgentConfigFile>();
```
martincostello: I would explicitly add the comparer here to make it clear it's case-(in)sensitive.

Do we need to do any special handling here regarding the fact that Linux has a case-sensitive file system by macOS and Windows don't?

src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
```diff
@@ -71,6 +72,20 @@ static AgentToServer BuildDisconnectMessage(FrameBuilder fb)
         }
     }
 
+    public async Task DispatchEffectiveConfigAsync(IEnumerable<EffectiveConfigFile> files, CancellationToken token)
```
martincostello, line 75: Is there any concern here that the configuration may contain sensitive values/secrets and them being sent to a remote party?

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    /// <param name="filePath">Path of the configuration file.</param>
+    /// <param name="contentType">The content type of the configuration file.</param>
+    /// <param name="filename">The reported filename.</param>
+    /// <returns>Effective config object.</returns>
+    public static EffectiveConfigFile CreateFromFilePath(string filePath, string contentType, string? filename = null)
+    {
+        try
+        {
+            var content = File.ReadAllBytes(filePath);
```
martincostello, line 50: Do we want/need to add an async variant of this method?

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        var tempDir = Path.GetTempPath();
+        var fileName = Path.GetRandomFileName();
+
+        // Manual separator handling (cross-platform)
+        if (!tempDir.EndsWith(Path.DirectorySeparatorChar.ToString(), StringComparison.Ordinal))
+        {
+            tempDir += Path.DirectorySeparatorChar;
+        }
+
```
martincostello: Why not use `Path.Join()`?

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        this.FilePath = path;
+        this.FileName = fileName;
+    }
+
+    public string FilePath { get; }
+
+    public string FileName { get; set; }
+
+    public static TempFile Create(string content)
```
martincostello, line 20: Alternatively you could use [`Path.GetTempFileName()`](https://learn.microsoft.com/dotnet/api/system.io.path.gettempfilename) with `File.WriteAllText()` to simplify this.

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
RassK: interestingly not available in net462.

martincostello: Could use `Path.Combine()` there (or instead of `Join`)?

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
RassK: not available in net462 and netstandard2.0

src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
RassK: Not familiar with all of the keys but quite possible. Even if not now, then in the future.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
martincostello: Could consider adding for just those TFMs that support it in the future.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
RassK: Good question, spec does not cover it. Server and client can also use different platforms, so hard to leave even OS defaults 🤔

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
RassK: `File.WriteAllText()` 👍 

`Path.GetTempFileName(`) needs additional call to `Path.GetFileName()`. Not sure which way is better, I don't have any preference here.

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
RassK: yes, this seems better. Interestingly produces more better paths as well > https://www.code4it.dev/csharptips/path-combine-vs-path-join/#final-comparison

stevejgordon:
At the moment, the initial capabilities the client publishes via `AddCapabilities` doesn't include the `ReportsEffectiveConfig`. We should ensure we send that through. A slightly trickier question is should we always enable that capability since the client supports it, even if we don't know if the consumer of the client will actually report their effective config. Or should we make the capability opt-in via settings, so that the consumer must first enable it. If so should the code check the capabilities that are enabled and guard against including this message when the capability is not set?

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/TempFile.cs
stevejgordon: I slightly prefer `GetTempFileName` + `GetFileName`.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
RassK: After analysing this I'm falling towards case sensitive option. In case the file is on a linux based server, it needs extra effort to find a suitable file if the casing is not matching.

src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
RassK: I think ideally not a problem if https or wss endpoint is used.

src/OpenTelemetry.OpAmp.Client/Settings/EffectiveConfigSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+/// Configuration settings for the effective configuration reporting capability of the client.
+/// </summary>
+public sealed class EffectiveConfigSettings
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the client can report effective configuration.
+    /// </summary>
+    /// <value>
+    /// <c>true</c> if effective configuration reporting is enabled; otherwise, <c>false</c>.
+    /// Default is <c>false</c>.
```
martincostello: Use `<see langword="true|false"/>` instead of `<c>true|false</c>`.

Kielek:
```suggestion
    /// <see langword="true"/> if effective configuration reporting is enabled; otherwise, <see langword="false"/>.
    /// Default is <see langword="false"/>.
```

Issue #1152, opened 2023-04-18
GenevaExporter should log a helpful message when it runs out of buffer capacity to serialize data


utpilla:
GenevaExporters simply catch the `IndexOutOfRangeException` when they run out of buffer capacity to serialize data and log the related exception message

**What's the expectation?**
GenevaExporters should catch the `IndexOutOfRangeException` and log a more helpful error message which clearly tells what went wrong.

Yun-Ting:
I'll be working on this issue.

TimothyMothra:
@CodeBlanch, this issue needs to be moved to the Future milestone.

PR #3828, opened 2026-01-29
[Geneva] Add fuzz tests and add bounds checks

martincostello:
## Changes

- Add bounds checks to resolve TODOs in Geneva exporter.
- Add fuzz tests for Geneva exporter serializers.

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MessagePackSerializer.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MessagePackSerializer.cs
@@ -58,8 +58,6 @@ internal static class MessagePackSerializer
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public static int SerializeNull(byte[] buffer, int cursor)
     {
-        CheckBounds(buffer, cursor, size: 1);
-
         buffer[cursor++] = NIL;
         return cursor;
     }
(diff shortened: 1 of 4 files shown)


martincostello:
I don't fully trust Copilot, but trying to get it to write fuzz tests to hit these two remaining uncovered lines, they always fail because no exception is thrown:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/92ac6366891589acb96abfffb6bcaea0a0e8648d/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MessagePackSerializer.cs#L397

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/92ac6366891589acb96abfffb6bcaea0a0e8648d/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MessagePackSerializer.cs#L414

Non-ASCII strings seem to be written to the buffer anyway and don't write more than the string's reported length.

Either myself and Copilot can't come up with appropriate test input to hit these lines, or the code is dead/unreachable.

martincostello:
From @rajkumar-rangaraj in CNCF Slack:

> Can we simplify this PR to only add bounds check to `SerializeFloat64`? It's the only unsafe method. Everything else uses safe array access where .NET already validates bounds.
>
> More context - Currently, we use a fixed 64K buffer for serialization. When we exceed capacity (e.g., very long log messages), .NET throws IndexOutOfRangeException and we handle it. However, this doesn't give users helpful context about what failed or why.
What we really need is rather than adding bounds checks everywhere, we should catch buffer overflow scenarios and log meaningful diagnostics - like which metric/log exceeded capacity, the data size, truncated content, etc. This helps users troubleshoot instead of just seeing a generic exception. See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1152

PR #3829, opened 2026-02-02
[Instrumentation.Remoting] initial implementation

lewis800:
Hi,

I wanted to revive the amazing work done in #30 with full attribution of work to the original author @mbakalov.

His branch is very out of date, so I decided to fork main and manually move his work over to a new branch on the fork called remoting.

I have kept his work and mainly just done housekeeping renaming and updating to the latest OTEL standards and I have tried to follow the library patterns, but I am sure I am missing some.

Cheers,

Lewis


martincostello:
Thanks for your contribution @lewis800.

Before we get into reviewing the code in this PR, is this an instrumentation library you're willing to contribute **and** support for feature requests/issues going forwards if it were merged and released to NuGet.org?

The maintainers only have a fixed amount of bandwidth and we can't take on the maintenance burden of every possible piece of new instrumentation that might be contributed to the repo, particularly if they're a more niche use case.

lewis800:
Hi @martincostello for the foreseeable future I have the capacity to contribute and support this remoting instrumentation, I will be using it actively and I suspect likely, will be the one coming back with the requests and issues regarding it.

martincostello:
@lewis800 Cool - in that case can you update the PR to make you a Code Owner in the appropriate places. For example:

- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/792ac6b3c0ed76e986b6e8c795bd9932d53bc167/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/README.md?plain=1#L6
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/792ac6b3c0ed76e986b6e8c795bd9932d53bc167/.github/component_owners.yml#L46-L48
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/792ac6b3c0ed76e986b6e8c795bd9932d53bc167/.github/component_owners.yml#L149-L151

The CI setup will also need updating in the relevant places to build, test and publish the new instrumentation as appropriate. You should be able to find all the relevant changes needed by searching the repo for other instrumentations' name and seeing where they appear.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,297 @@
...
+                // if it is available.
+                ActivityContext contextToInject = default;
+
+                // Start new outgoing activity
+                var activity = RemotingActivitySource.StartActivity(ActivityOutName, ActivityKind.Client);
+                if (activity != null)
+                {
+                    if (activity.IsAllDataRequested && reqMsg is IMethodMessage methodMsg)
+                    {
+                        SetStartingActivityAttributes(activity, methodMsg);
```
thompson-tomo: The attributes where possible should be passed into startActivity to enable head sampling.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    // Got here if we don't have a valid Remoting context. This is possible if the client
+                    // didn't have Remoting instrumentation and so didn't initialize the "__CallContext".
+                    // We still have a parent activity from another instrumentation, so we keep it as
+                    // Current (better than creating a new "root" Remoting activity).
+                }
+
+                if (ourActivity != null && ourActivity.IsAllDataRequested && reqMsg is IMethodMessage methodMsg)
+                {
+                    SetStartingActivityAttributes(ourActivity, methodMsg);
```
thompson-tomo: As above we should be passing attributes to the start activity where possible.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    private static void SetStartingActivityAttributes(Activity activity, IMethodMessage msg)
+    {
+        string serviceName = GetServiceName(msg.TypeName);
+        string methodName = msg.MethodName;
+        string fullyQualifiedMethod = $"{serviceName}/{methodName}";  // "SharedLib.IHelloServer/SayHello"
+
+        activity.DisplayName = fullyQualifiedMethod;
+        activity.SetTag(AttributeRpcSystemName, AttributeRpcSystemNameValue);
```
thompson-tomo: Take a look at the latest rpc conventions as there has been a number of breaking changes.

lewis800:
@martincostello I have updated the readme and think I have successfully added the remoting project into the existing workflows

.github/workflows/ci.yml
```diff
@@ -677,6 +689,7 @@ jobs:
       build-test-instrumentation-owin,
       build-test-instrumentation-process,
       build-test-instrumentation-quartz,
+      build-test-instrumentation-Remoting,
```
martincostello:
```suggestion
      build-test-instrumentation-remoting,
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+/// </remarks>
+internal class TelemetryDynamicSink : IDynamicMessageSink
+{
+    internal const string AttributeRpcSystemName = "rpc.system.name";
+    internal const string AttributeRpcSystemNameValue = "netframework_remoting";
+    internal const string AttributeRpcMethod = "rpc.method";
+
+    // Uri like "tcp://localhost:1234/HelloServer.rem"
+    // TODO: semantic conventions don't have an attribute for a full uri of an RPC endpoint, but seems useful?
+    internal const string AttributeRpcRemotingUri = "rpc.netframework_remoting.uri";
```
martincostello: Is there an open issue/pull request somewhere in [open-telemetry/semantic-conventions](https://github.com/open-telemetry/semantic-conventions) suggesting such an attribute be added (or has one been added since this code was originally written)?

We shouldn't invent new attributes and just implement the Semantic Conventions as-is, so this should be removed.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+/// </summary>
+/// <remarks>
+/// See https://docs.microsoft.com/previous-versions/dotnet/netframework-4.0/tdzwhfy3(v=vs.100) for
+/// information about remoting sinks and https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty
+/// for RegisterDynamicProperty behavior.
+/// </remarks>
+internal class TelemetryDynamicSink : IDynamicMessageSink
+{
+    internal const string AttributeRpcSystemName = "rpc.system.name";
+    internal const string AttributeRpcSystemNameValue = "netframework_remoting";
```
martincostello: Unsure on the value this should be.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+/// A dynamic remoting sink that intercepts all calls leaving or entering an AppDomain, see
+/// <see cref="RemotingInstrumentation"/> constructor: RegisterDynamicProperty
+/// code.
+/// </summary>
+/// <remarks>
+/// See https://docs.microsoft.com/previous-versions/dotnet/netframework-4.0/tdzwhfy3(v=vs.100) for
+/// information about remoting sinks and https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty
+/// for RegisterDynamicProperty behavior.
+/// </remarks>
+internal class TelemetryDynamicSink : IDynamicMessageSink
```
martincostello:
```suggestion
internal sealed class TelemetryDynamicSink : IDynamicMessageSink
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+
+    internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.Remoting";
+    private const string ActivityOutName = ActivitySourceName + ".RequestOut";
+    private const string ActivityInName = ActivitySourceName + ".RequestIn";
+
+    private const string SavedAspnetActivityPropertyName = ActivitySourceName + ".SavedAspnetActivity";
+
+    private static readonly ActivitySource RemotingActivitySource = new(ActivitySourceName, typeof(TelemetryDynamicSink).Assembly.GetPackageVersion());
+
+    private static readonly ConcurrentDictionary<string, string> ServiceNameCache = new();
```
martincostello: For testability reasons, this is better to be non-`static` as presumably in a real application using this it would be a singleton, so the cache behaviour would be equivalent.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+
+                    try
+                    {
+                        this.options.Enrich?.Invoke(ourActivity, RemotingInstrumentationEnrichEventNames.OnMessageStart, methodMsg);
+                    }
+                    catch (Exception ex)
+                    {
+                        RemotingInstrumentationEventSource.Log.EnrichmentException(ex);
+                    }
+                }
```
martincostello: This block is effectively identical to lines 93-107. It should be refactored to avoid duplication.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+            bool validClientActivity = bCliSide && activity.OperationName == ActivityOutName;
+            bool validServerActivity = !bCliSide && activity.OperationName == ActivityInName;
+            if (validClientActivity || validServerActivity)
+            {
+                if (replyMsg is IMethodReturnMessage returnMsg)
+                {
+                    if (returnMsg.Exception == null)
+                    {
+                        // Default to "Unset" status as per spec:
+                        // https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#status
```
martincostello: Links like this should be permalinks, not floating to the default branch name.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+        {
+            RemotingInstrumentationEventSource.Log.DynamicSinkException(e);
+        }
+    }
+
+    internal static ActivityTagsCollection BuildSamplingTags(IMethodMessage msg)
+    {
+        string serviceName = GetServiceName(msg.TypeName);
+        string methodName = msg.MethodName;
+        string fullyQualifiedMethod = $"{serviceName}/{methodName}";
```
martincostello: This should be moved to a private helper method so that...

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+            { AttributeRpcMethod, fullyQualifiedMethod },
+            { AttributeRpcRemotingUri, uriString },
+        };
+    }
+
+    internal static void SetPostCreationAttributes(Activity activity, IMethodMessage msg)
+    {
+        string serviceName = GetServiceName(msg.TypeName);
+        string methodName = msg.MethodName;
+        string fullyQualifiedMethod = $"{serviceName}/{methodName}";
```
martincostello: ...it can be re-used.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+        ServiceNameCache.GetOrAdd(typeName, s =>
+            {
+                int pos = s.IndexOf(",", StringComparison.OrdinalIgnoreCase);
+                if (pos >= 0)
+                {
+                    return s.Substring(0, pos);
+                }
+
+                return s;
+            });
```
martincostello: So the intention here is that assembly-qualified type names should use the types' full name as the service name?

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,329 @@
...
+
+    private static void InjectActivityProperties(LogicalCallContext ctx, string key, string value) => ctx.SetData(key, value);
+
+    private static IEnumerable<string> ExtractActivityProperties(LogicalCallContext ctx, string key)
+    {
+        var data = ctx.GetData(key);
+        if (data != null)
+        {
+            return new[] { (string)data };
```
martincostello, line 288: Is the data guaranteed to be a string?

src/OpenTelemetry.Instrumentation.Remoting/Implementation/RemotingInstrumentationEnrichEventNames.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.Instrumentation.Remoting;
+
+/// <summary>
+/// Defines the event names used by the Remoting instrumentation Enrich callback.
+/// </summary>
+public static class RemotingInstrumentationEnrichEventNames
```
martincostello: Do these need to be public?

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSinkProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System.Runtime.Remoting.Contexts;
+
+namespace OpenTelemetry.Instrumentation.Remoting.Implementation;
+
+/// <summary>
+/// A <see cref="IContributeDynamicSink"/> implementation that returns an instance of
+/// <see cref="TelemetryDynamicSink"/> responsible for instrumenting remoting calls.
+/// </summary>
+internal class TelemetryDynamicSinkProvider : IDynamicProperty, IContributeDynamicSink
```
martincostello:
```suggestion
internal sealed class TelemetryDynamicSinkProvider : IDynamicProperty, IContributeDynamicSink
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+| ------ | --- |
+| Stability | [Alpha](../../README.md#beta) |
+| Code Owners | [@lewis800](https://github.com/lewis800) |
+
+[![NuGet version badge](https://img.shields.io/nuget/v/OpenTelemetry.Instrumentation.Remoting)](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Remoting)
+[![NuGet download count badge](https://img.shields.io/nuget/dt/OpenTelemetry.Instrumentation.Remoting)](https://www.nuget.org/packages/OpenTelemetry.Instrumentation.Remoting)
+[![codecov.io](https://codecov.io/gh/open-telemetry/opentelemetry-dotnet-contrib/branch/main/graphs/badge.svg?flag=unittests-Instrumentation.Remoting)](https://app.codecov.io/gh/open-telemetry/opentelemetry-dotnet-contrib?flags[0]=unittests-Instrumentation.Remoting)
+
+This is an instrumentation library, which instruments [.NET Remoting](https://docs.microsoft.com/previous-versions/dotnet/netframework-3.0/72x4h507(v=vs.85))
+and collects telemetry about incoming and outgoing requests on client
```
martincostello:
```suggestion
This is an [Instrumentation
Library](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-library),
which instruments [.NET Remoting](https://docs.microsoft.com/previous-versions/dotnet/netframework-3.0/72x4h507(v=vs.85))
and collects telemetry about incoming and outgoing requests on client
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+and server objects.
+
+.NET Remoting is a [legacy technology](https://docs.microsoft.com/previous-versions/dotnet/netframework-3.0/kwdt6w2k(v=vs.85))
+that shouldn't be used for new .NET applications and [doesn't exist](https://docs.microsoft.com/dotnet/core/porting/net-framework-tech-unavailable#remoting)
+in .NET Core at all. However, if you do have a legacy application you are
```
martincostello:
```suggestion
in .NET 6 and later versions. However, if you do have a legacy application you are
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Configuration
+
+To enable .NET remoting instrumentation, call `AddRemotingInstrumentaion()` on
+the `TracerProviderBuilder` during the application startup in both client and
+server code.
+
+The following example demonstrates adding .NET remoting instrumentation to a
+client console application. This example also sets up the OpenTelemetry Console
+exporter, which requires adding the package [`OpenTelemetry.Exporter.Console`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry.Exporter.Console/README.md)
```
martincostello:
```suggestion
exporter, which requires adding the [`OpenTelemetry.Exporter.Console`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.Console/README.md) package
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+```
+
+When hosting server objects in IIS, adding instrumentation should typically
+be done in `Global.asax.cs` like in the below example.
+
+This example also sets up the OpenTelemetry Jaeger exporter, which requires
+adding the package [`OpenTelemetry.Exporter.Jaeger`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry.Exporter.Jaeger/README.md)
+to the project.
```
martincostello: This exporter is deprecated. The example should use something else.

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+        protected void Application_End(object sender, EventArgs e)
+        {
+            _tracerProvider?.Dispose();
+        }
+    }
+}
+```
+
+Additionally, when using [http channel](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.channels.http.httpchannel?view=netframework-4.8)
```
martincostello:
```suggestion
Additionally, when using [`HttpChannel`](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.channels.http.httpchannel)
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+for remoting, consider registering [`OpenTelemetry.Instrumentation.Http`](https://github.com/open-telemetry/opentelemetry-dotnet/tree/master/src/OpenTelemetry.Instrumentation.Http)
```
martincostello:
```suggestion
for remoting, consider registering [`OpenTelemetry.Instrumentation.Http`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Http)
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+on the client and [`OpenTelemetry.Instrumentation.AspNet`](https://github.com/open-telemetry/opentelemetry-dotnet/tree/master/src/OpenTelemetry.Instrumentation.AspNet)
```
martincostello:
```suggestion
on the client and [`OpenTelemetry.Instrumentation.AspNet`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AspNet)
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+        options.Filter = msg =>
+        {
+            // Only capture calls to and from "RemoteObject"
+            if (msg is IMethodMessage methodMsg)
+            {
+                return methodMsg.TypeName.Contains("RemoteObject");
+            }
+
+            return false;
+        })
```
martincostello:
```suggestion
        options.Filter = message =>
        {
            // Only capture calls to and from "RemoteObject"
            if (message is IMethodMessage methodMessage)
            {
                return methodMessage.TypeName.Contains("RemoteObject");
            }

            return false;
        })
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+    .Build()
+```
+
+The `Filter` takes an [`IMessage`](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.messaging.imessage?view=netframework-4.8)
```
martincostello:
```suggestion
The `Filter` takes an [`IMessage`](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.messaging.imessage)
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+and returns true or false. You can inspect the message to decide if you
```
martincostello:
```suggestion
and returns a boolean. You can inspect the message to decide if you
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+want to instrument it or not.
+
+## Implementation Details
+
+The instrumentation is implemented via custom [`IDynamicMessageSink`](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.idynamicmessagesink?view=netframework-4.8),
```
martincostello:
```suggestion
The instrumentation is implemented via  custom [`IDynamicMessageSink`](https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.idynamicmessagesink) implementation,
```

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+that is registered on the current `AppDomain` when you call
```
martincostello:
```suggestion
that is registered in the current `AppDomain` when you call
```

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentation.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.Instrumentation.Remoting.Implementation;
+using RemotingContext = System.Runtime.Remoting.Contexts.Context;
+
+namespace OpenTelemetry.Instrumentation.Remoting;
+
+internal class RemotingInstrumentation : IDisposable
```
martincostello:
```suggestion
internal sealed class RemotingInstrumentation : IDisposable
```

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentation.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    private static readonly object LockObj = new();
+    private static int regCount;
```
martincostello: Rather than maintain a lock, use `Interlocked`'s `Increment` and `Decrement()`?

```csharp
int count = Interlocked.Increment(ref regCount);

if (count == 1)
{
   RemotingContext.RegisterDynamicProperty(...)
}
```

```csharp
int count = Interlocked.Decrement(ref regCount);

if (count == 0)
{
   RemotingContext.UnregisterDynamicProperty(...)
}
```

src/OpenTelemetry.Instrumentation.Remoting/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,16 @@
+const OpenTelemetry.Instrumentation.Remoting.RemotingInstrumentationEnrichEventNames.OnMessageFinish = "OnMessageFinish" -> string!
```
martincostello: If nullable reference types are being used, they should be enabled here too.

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <item>If the filter returns <see langword="true"/>, the message is instrumented.</item>
+    /// <item>If the filter returns <see langword="false"/> or throws an exception, no instrumentation is performed.</item>
+    /// </list>
+    /// </remarks>
+    public Func<IMessage, bool>? Filter { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether exception will be recorded
+    /// as an <see cref="ActivityEvent"/> or not. Default value: <see
+    /// langword="true"/>.
```
martincostello: This says it's true, but it isn't.

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>For specification details see: <see
+    /// href="https://github.com/open-telemetry/semantic-conventions/blob/main/docs/exceptions/exceptions-spans.md"
+    /// />.</para>
+    /// </remarks>
```
martincostello:
```suggestion
```

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+/// </summary>
+public static class TracerProviderBuilderExtensions
+{
+    /// <summary>
+    /// Enables .NET Remoting instrumentation.
+    /// </summary>
+    /// <param name="builder"><see cref="TracerProviderBuilderExtensions"/> being configured.</param>
+    /// <returns>The instance of <see cref="TracerProviderBuilderExtensions"/> to chain the calls.</returns>
+    public static TracerProviderBuilder AddRemotingInstrumentation(this TracerProviderBuilder builder) =>
+    AddRemotingInstrumentation(builder, configure: null);
```
martincostello:
```suggestion
        AddRemotingInstrumentation(builder, configure: null);
```

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        Action<RemotingInstrumentationOptions>? configure)
+    {
+        Guard.ThrowIfNull(builder);
+
+        builder.AddSource(TelemetryDynamicSink.ActivitySourceName);
+
+        var remotingOptions = new RemotingInstrumentationOptions();
+        configure?.Invoke(remotingOptions);
+
+        builder.AddInstrumentation(activitySource => new RemotingInstrumentation(remotingOptions));
```
martincostello:
```suggestion
        builder.AddInstrumentation(_ => new RemotingInstrumentation(remotingOptions));
```

martincostello: Can this be made lazily so that the configure delegate isn't immediately invoked?

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (success)
+            {
+                obj.DoStuff();
+            }
+            else
+            {
+                Assert.Throws<Exception>(() => obj.DoStuff(exceptionMessage));
+            }
+
+            AppDomain.Unload(ad);
```
martincostello: In a `finally`?

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    [Fact]
+    public void RemotingInstrumentation_RegisterDynamicProperty_OnlyOnce()
+    {
+        // This will register the dynamic property on the current context
+        using var i1 = new RemotingInstrumentation(null);
+
+        // Make sure the second call succeeds as well and won't try to re-register the property
+        using var i2 = new RemotingInstrumentation(null);
+    }
```
martincostello, line 102: Is there a way to assert that proves that it isn't re-registered?

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Here are my suggestions

* netframework_remoting
* dotnet_remoting
* remoting

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Happy to remove it and, instead populate some of the other tags available in the RPC semantic conventions which would help to get this URI

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: This pattern seems to be reused through the other instrumentation libraries, such as for WCF, I can change it if we need to, has it caused problems for testing in the past?

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
lewis800: Used existing patterns to adjust a lazy invoke

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
lewis800: Exposed Regcount as internal to the remoting tests. added in asserts. also passed new options object to remove warnings.

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
lewis800: wrapped work in a try catch, with the ad unload in a finally block

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: I have reworked it to include the attributes on the start activities, retaining a guarded IsAllDataRequested method for more expensive attributes

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: I have reworked it to include the attributes on the start activities, retaining a guarded IsAllDataRequested method for more expensive attributes

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: I've added in server.address and server.port and error.type, I think otherwise I am adhering to version 1.39 of semantic conventions, please lmk if I missed something

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
thompson-tomo: Would suggest raising a semconv pr to add it and get  additional feedback. If the system is only available on net framework then I would use the netframework option otherwise the dotnet option.

lewis800: Can do I'll raise a PR, natively its framework only but there is a port project for .NET Core but its not official

martincostello: [Docs](https://github.com/open-telemetry/semantic-conventions/blob/6d05e92070f0a5c9a5bbcd396d2ebe2103377f8c/docs/rpc/rpc-spans.md?plain=1#L196-L203).

> `rpc.system.name` has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
martincostello: I can't speak for other areas of the codebase without digging through the Git history, but from personal experience this is a testability smell as tests may behave differently based on ordering if values become cached. If I'd have reviewed other code doing the same thing I'd have likely made the exact same comment.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: I've raised [PR](https://github.com/open-telemetry/semantic-conventions/pull/3422) in the semantic conventions repo

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Let me see what I can cook up, by chance is there a good reference instrumentation that has a pattern I could follow?

src/OpenTelemetry.Instrumentation.Remoting/.publicApi/PublicAPI.Unshipped.txt
Kielek: It is enough to put it only in shipped part.

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+dotnet add package --prerelease OpenTelemetry.Instrumentation.Remoting
+```
+
+## Configuration
+
+To enable .NET remoting instrumentation, call `AddRemotingInstrumentaion()` on
+the `TracerProviderBuilder` during the application startup in both client and
+server code.
+
+The following example demonstrates adding .NET remoting instrumentation to a
```
Kielek:
```suggestion
The following example demonstrates adding .NET Framework remoting instrumentation to a
```

I would like to avoid any suggestion that it might work on modern .NET stack.
Probably requires also reformatting this section to avoid long lines

src/OpenTelemetry.Instrumentation.Remoting/README.md
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+```
+
+When hosting server objects in IIS, adding instrumentation should typically
+be done in `Global.asax.cs` like in the below example.
+
+This example also sets up the OpenTelemetry Jaeger exporter, which requires
+adding the package [`OpenTelemetry.Exporter.Jaeger`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/master/src/OpenTelemetry.Exporter.Jaeger/README.md)
```
Kielek: This package is no longer supported, please update to OTLP

Kielek (commented):
Could you please consider splitting this PR to 2 chunks? First - just project scaffolding with empty project, CI changes, etc. No classes should be fine.

It will be easier to review the code itslef.

For sure you are missing CHANGELOG.md near to README.md

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Okay based on the kind input of the semantic conventions repo we have landed on

dotnet.remoting

src/OpenTelemetry.Instrumentation.Remoting/README.md
lewis800: Updated to OpenTelemetry.Exporter.OpenTelemetryProtocol

lewis800: Added a changelog also

src/OpenTelemetry.Instrumentation.Remoting/Implementation/RemotingInstrumentationEnrichEventNames.cs
lewis800: No they do not I'll change it to internal

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Bundled into SetPostCreationAttributes method, which is now no longer static, to de duplicate

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Updated to perma link

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: refactored into private helper GetFullyQualifiedMethod

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: yea fair haha, private helper created

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: Yes it is

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: We only write string values for these keys. 

The cast would only fail if some other code writes the same key in the call context with a non-string value

src/OpenTelemetry.Instrumentation.Remoting/README.md
lewis800: Updated to OTLP exporter

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
lewis800: updated default value

src/OpenTelemetry.Instrumentation.Remoting/README.md
lewis800: Ideally, I would like to not split this out into two PR's if we can avoid it

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
lewis800: I've moved the ActivitySource out of TelemetryDynamicSink and into TelemetryDynamicSinkProvider, which owns and injects it into each sink via the constructor. 

The provider is effectively a singleton, so this keeps it to one ActivitySource while still being non-static

src/OpenTelemetry.Instrumentation.Remoting/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## Unreleased
+
+## 0.0.0-alpha.2469
```
martincostello: We'd determine the version as part of the release process, then our automation will make the relevant updates.

```suggestion
```

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                            return methodMsg.TypeName.Contains("RemoteObject");
+                        }
+
+                        return false;
+                    };
+                })
+            .Build();
+        var domainSetup = AppDomain.CurrentDomain.SetupInformation;
+        var ad = AppDomain.CreateDomain("other-domain", null, domainSetup);
+        try
```
martincostello, line 45: Just to improve readability.

```suggestion
            .Build();

        var domainSetup = AppDomain.CurrentDomain.SetupInformation;
        var ad = AppDomain.CreateDomain("other-domain", null, domainSetup);

        try
```

test/OpenTelemetry.Instrumentation.Remoting.Tests/RemotingInstrumentationTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                        {
+                            return methodMsg.TypeName.Contains("RemoteObject");
+                        }
+
+                        return false;
+                    };
+                })
+            .Build();
+        var domainSetup = AppDomain.CurrentDomain.SetupInformation;
+        var ad = AppDomain.CreateDomain("other-domain", null, domainSetup);
```
martincostello: Can we give this a more meaningful name like `appDomain` or `domain`?

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        builder.AddSource(TelemetryDynamicSink.ActivitySourceName);
+
+        builder.ConfigureServices(services =>
+        {
+            if (configure != null)
+            {
+                services.Configure(configure);
+            }
+        });
```
martincostello: Looks like you can skip this entirely if it's `null`?

```suggestion
        if (configure != null)
        {
            builder.ConfigureServices(configure);
        }
```

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        builder.AddInstrumentation(sp =>
+        {
+            var remotingOptions = sp.GetRequiredService<IOptionsMonitor<RemotingInstrumentationOptions>>().CurrentValue;
+            return new RemotingInstrumentation(remotingOptions);
+        });
```
martincostello, line 48: Would it make sense to refactor this to accept the `IOptionsMonitor<RemotingInstrumentationOptions>` itself so that the instrumentation could just observe configuration changes itself directly?

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentation.cs
```diff
@@ -0,0 +1,50 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.Instrumentation.Remoting.Implementation;
+using RemotingContext = System.Runtime.Remoting.Contexts.Context;
+
+namespace OpenTelemetry.Instrumentation.Remoting;
+
+internal sealed class RemotingInstrumentation : IDisposable
+{
+    internal static int RegCount;
```
martincostello: `RegistrationCount`

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSinkProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    /// <returns>A new instance of <see cref="TelemetryDynamicSink"/>.</returns>
+    public IDynamicMessageSink GetDynamicSink() => new TelemetryDynamicSink(this.options, this.activitySource, this.serviceNameCache);
+
+    /// <inheritdoc />
+    public void Dispose() => this.activitySource.Dispose();
+
+    private static ActivitySource CreateActivitySource()
+    {
+        const string telemetrySchemaUrl = "https://opentelemetry.io/schemas/1.39.0";
```
martincostello: At the point of merging, this should point to whatever the current latest version is (right now it's now [1.40.0](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.40.0)).

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,365 @@
...
+        this.serviceNameCache.GetOrAdd(typeName, s =>
+            {
+                int pos = s.IndexOf(",", StringComparison.OrdinalIgnoreCase);
+                if (pos >= 0)
+                {
+                    return s.Substring(0, pos);
+                }
+
+                return s;
+            });
```
martincostello: Improves the readability by making it more self-documenting:

```suggestion
        // typeName will be a full .NET type name as a string "SharedLib.IHelloServer, SharedLib, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null"
        this.serviceNameCache.GetOrAdd(typeName, assemblyQualifiedTypeName =>
            {
                int index = assemblyQualifiedTypeName.IndexOf(",", StringComparison.OrdinalIgnoreCase);
                if (index >= 0)
                {
                    // Trim to just the type's full name
                    return assemblyQualifiedTypeName.Substring(0, index);
                }

                return assemblyQualifiedTypeName;
            });
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,365 @@
...
+                tags.Add(AttributeServerPort, port.Value);
+            }
+        }
+
+        return tags;
+    }
+
+    private string GetFullyQualifiedMethod(IMethodMessage msg)
+    {
+        string serviceName = this.GetServiceName(msg.TypeName);
```
martincostello: As we're no longer constrained to match the naming from the interface:

```suggestion
    private string GetFullyQualifiedMethod(IMethodMessage message)
    {
        string serviceName = this.GetServiceName(message.TypeName);
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,365 @@
...
+        var data = ctx.GetData(key);
+        if (data != null)
+        {
+            return new[] { (string)data };
+        }
+
+        return Enumerable.Empty<string>();
+    }
+
+    private void SetPostCreationAttributes(Activity activity, IMethodMessage? msg)
```
martincostello, line 339: Same here, `msg` ➡️ `message`

.github/workflows/ci.yml
```diff
@@ -398,6 +399,17 @@ jobs:
...
+  build-test-instrumentation-remoting:
+    needs: detect-changes
+    if: |
+      contains(needs.detect-changes.outputs.changes, 'instrumentation-remoting')
+      || contains(needs.detect-changes.outputs.changes, 'build')
+      || contains(needs.detect-changes.outputs.changes, 'shared')
+    uses: ./.github/workflows/Component.BuildTest.yml
+    with:
+      project-name: Component[OpenTelemetry.Instrumentation.Remoting]
+      code-cov-name: Instrumentation.Remoting
```
martincostello, line 423:
```suggestion
      code-cov-name: Instrumentation.Remoting
      os-list: '[ "windows-latest" ]'
      tfm-list: '[ "net462" ]'
```

.github/component_owners.yml
```diff
@@ -58,6 +58,8 @@ components:
     - Yun-Ting
   src/OpenTelemetry.Instrumentation.Quartz/:
     - maldago
+  src/OpenTelemetry.Instrumentation.Remoting/:
+    - lewis800
```
martincostello:
```suggestion
    - lewis800
```

.github/component_owners.yml
```diff
@@ -159,6 +161,8 @@ components:
     - fred2u
   test/OpenTelemetry.Instrumentation.Quartz.Tests/:
     - maldago
+  test/OpenTelemetry.Instrumentation.Remoting.Tests/:
+    - lewis800
```
martincostello:
```suggestion
    - lewis800
```

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
lewis800: I wouldn't think so unless we are expecting the options to reloadable or we want that as a feature.

martincostello: Well that's kinda what I'm asking.

We're already depending on a type that supports reloading, so is that something worth supporting? Or do the current options not make sense to be ones that can be reloaded?

Kielek:
> Ideally, I would like to not split this out into two PR's if we can avoid it

@lewis800, I see that you are working on infrastructure fixes here. This was my initial call to split it to separate PR.
Once again, it will allow us to focus on correct things separately.

Cost of extraction does not seem so big to avoid this.

src/OpenTelemetry.Instrumentation.Remoting/TracerProviderBuilderExtensions.cs
lewis800: imo, I don't think that these options should be relodable on the fly, specifically the Propagator. filter and record exception I don't really mind. I guess its not that much extra work to make it reloadable ... is that something other instrumentation does? My understanding that with atleast the OTEL SDK u set the settings on startup and thats it unless u restart and re deploy code, unless u are interested in giving user the option to do mid flight ho reloads

martincostello: There's existing code in various places that do support reloading (the whole reasons options monitor exists). Some components need to support reloading in the future as part of us adding support for Declarative Configuration, so if there are settings that make sense to honour it and it's not much work, then we should make it work.

lewis800: Okay I have implemented that and added a test case

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSinkProvider.cs
```diff
@@ -33,10 +36,16 @@ public TelemetryDynamicSinkProvider(RemotingInstrumentationOptions options)
...
 
     /// <inheritdoc />
-    public void Dispose() => this.activitySource.Dispose();
+    public void Dispose()
+    {
+        this.optionsChangeRegistration?.Dispose();
+        this.activitySource.Dispose();
+    }
+
+    internal RemotingInstrumentationOptions GetOptions() => this.options;
```
martincostello: Curious why this is a method rather than a property given that nothing mutates during the method call?

lewis800: Thats an oversight on my part, no good reason why I've changed it to a property

lewis800:
> > Ideally, I would like to not split this out into two PR's if we can avoid it
> 
> @lewis800, I see that you are working on infrastructure fixes here. This was my initial call to split it to separate PR. Once again, it will allow us to focus on correct things separately.
> 
> Cost of extraction does not seem so big to avoid this.

Hi @Kielek I have split it out have a look at #3972 which only contains the scaffold adding in the remoting csproj's to this repo and CICD and templates etc

lewis800:
@martincostello do we have any other pending code items for this PR?

Do u also want me to create another PR remoting-implementation to compliment the new remoting-scaffold?

Kielek:
> Do u also want me to create another PR remoting-implementation to compliment the new remoting-scaffold?

When the scaffolding is merged, feel free to rebase/merge main here.

Kielek:
@lewis800, scaffolding PR is merged. Could you please resolve-conflicts/rebase this PR?

lewis800:
> @lewis800, scaffolding PR is merged. Could you please resolve-conflicts/rebase this PR?

Yep will do

lewis800:
@martincostello  @Kielek should be rebased off main, let me know if we need any further changes or clean up etc

src/Shared/SemanticConventions.cs
```diff
@@ -151,5 +151,9 @@ internal static class SemanticConventions
     public const string AttributeDbQueryText = "db.query.text";
     public const string AttributeDbStoredProcedureName = "db.stored_procedure.name";
 
+    // v1.37.0 RPC spans
+    // https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/rpc/rpc-spans.md
```
Kielek: It is not true, there is no rpc.system.name in 1.37.0

```suggestion
    // v1.40.0 RPC spans
    // https://github.com/open-telemetry/semantic-conventions/blob/v1.40.0/docs/rpc/rpc-spans.md
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,369 @@
...
+/// </summary>
+/// <remarks>
+/// See https://docs.microsoft.com/previous-versions/dotnet/netframework-4.0/tdzwhfy3(v=vs.100) for
+/// information about remoting sinks and https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty
+/// for RegisterDynamicProperty behavior.
+/// </remarks>
+internal sealed class TelemetryDynamicSink : IDynamicMessageSink
+{
+    internal const string AttributeRpcSystemNameValue = "dotnet.remoting";
+    internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.Remoting";
```
Kielek: This is strange. You have decided to accept ActivitySource as a parameter to the constructor, and now, you are hardcoding this value. It should not be the case.

TracerProviderBuilderExtensions also refer here, so you should probably find a better usage for it.

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSink.cs
```diff
@@ -0,0 +1,369 @@
...
+/// code.
+/// </summary>
+/// <remarks>
+/// See https://docs.microsoft.com/previous-versions/dotnet/netframework-4.0/tdzwhfy3(v=vs.100) for
+/// information about remoting sinks and https://docs.microsoft.com/dotnet/api/system.runtime.remoting.contexts.context.registerdynamicproperty
+/// for RegisterDynamicProperty behavior.
+/// </remarks>
+internal sealed class TelemetryDynamicSink : IDynamicMessageSink
+{
+    internal const string AttributeRpcSystemNameValue = "dotnet.remoting";
```
Kielek: I would consider making it private.

src/OpenTelemetry.Instrumentation.Remoting/OpenTelemetry.Instrumentation.Remoting.csproj
```diff
@@ -7,8 +7,6 @@
     <MinVerTagPrefix>Instrumentation.Remoting-</MinVerTagPrefix>
   </PropertyGroup>
 
-  <!-- Do not run Package Baseline Validation as this package has never released a stable version.
-  Remove this property once we have released a stable version and add PackageValidationBaselineVersion property. -->
```
Kielek, line 11: Please keep this comment,

src/OpenTelemetry.Instrumentation.Remoting/OpenTelemetry.Instrumentation.Remoting.csproj
```diff
@@ -17,4 +15,16 @@
     <InternalsVisibleTo Include="OpenTelemetry.Instrumentation.Remoting.Tests" PublicKey="$(StrongNamePublicKey)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Options" />
+    <PackageReference Include="OpenTelemetry" />
```
Kielek: In general, there is a good practice to reference only OpenTelemetry.Api package by the instrumentations. The only downside is probably lack of support for `Sdk.SuppressInstrumentation`. I would not consider this as a blocker as WCF is doing probably the same thing.

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    /// <summary>
+    /// Gets or sets an action to enrich an Activity.
+    /// </summary>
+    /// <remarks>
+    /// <para><see cref="Activity"/>: the activity being enriched.</para>
+    /// <para>string: the name of the event. See <see cref="RemotingInstrumentationEnrichEventNames"/> for available values.</para>
+    /// <para><see cref="IMethodMessage"/>: the Remoting method message from which additional information can be extracted to enrich the activity.</para>
+    /// </remarks>
+    public Action<Activity, string, IMethodMessage>? Enrich { get; set; }
```
Kielek: I think that you should be able to avoid string parameter here.
For most instrumentation libraries we are trying to make it specific:.
Consider something like, I am not a Remoting expert, so probably some better naming is needed here.
```csharp
    public Action<Activity, IMethodMessage>? EnrichWith[Send]Message { get; set; }
    public Action<Activity, IMethodReturnMessage>? EnrichWithReturnMessage { get; set; }
```

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// </list>
+    /// </remarks>
+    public Func<IMessage, bool>? Filter { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether exception will be recorded
+    /// as an <see cref="ActivityEvent"/> or not. Default value: <see
+    /// langword="false"/>.
+    /// </summary>
+    public bool RecordException { get; set; }
```
Kielek: Can we avoid this property? Please check https://opentelemetry.io/blog/2026/deprecating-span-events/

@pellared, could you please advice here?

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+    public Action<Activity, string, IMethodMessage>? Enrich { get; set; }
+
+    /// <summary>
+    /// Gets or sets a <see cref="TextMapPropagator"/> for context propagation. Default value: <see cref="CompositeTextMapPropagator"/> with <see cref="TraceContextPropagator"/> and <see cref="BaggagePropagator"/>.
+    /// </summary>
+    public TextMapPropagator Propagator { get; set; } = new CompositeTextMapPropagator(new TextMapPropagator[]
+    {
+        new TraceContextPropagator(),
+        new BaggagePropagator(),
+    });
```
Kielek:
```suggestion
    public TextMapPropagator Propagator { get; set; } = new CompositeTextMapPropagator([
        new TraceContextPropagator(),
        new BaggagePropagator()
    ]);
```

src/OpenTelemetry.Instrumentation.Remoting/Implementation/RemotingInstrumentationEventSource.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    [Event(2, Message = "DynamicSink message processor threw an exception. Remoting activity will not be recorded. Exception {0}.", Level = EventLevel.Error)]
+    public void DynamicSinkException(string exception)
+    {
+        this.WriteEvent(2, exception);
+    }
+
+    [NonEvent]
+    public void EnrichmentException(Exception ex)
```
Kielek: You are calling it with 2 places. It will be great to distinguish the issues EnrishOnStart/Stop.

Kielek (commented):
I hope this comments make sense, please let me know what do you think.

src/OpenTelemetry.Instrumentation.Remoting/OpenTelemetry.Instrumentation.Remoting.csproj
lewis800: I only added Sdk.SuppressInstrumentation as I saw other instrumentation use it, so I am not set in stone on having / vs not having it. I'll leave it upto you to decide

src/OpenTelemetry.Instrumentation.Remoting/Implementation/RemotingInstrumentationEventSource.cs
lewis800: split them out into two events

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
lewis800: Okay removed

src/OpenTelemetry.Instrumentation.Remoting/RemotingInstrumentationOptions.cs
lewis800: Implemented
* EnrichWithMessage
* EnrichWithReturnMessage

src/OpenTelemetry.Instrumentation.Remoting/OpenTelemetry.Instrumentation.Remoting.csproj
Kielek: @rajkumar-rangaraj, what is your opinion here?

lewis800: @Kielek  I'm happy to just remove it for now and we can add it in back later if its needed, to get this PR merged for now

Kielek: @rajkumar-rangaraj?
IMO option to fully remove is also fully acceptable.

lewis800: Okay pushed and now no longer referencing the SDK

src/OpenTelemetry.Instrumentation.Remoting/Implementation/TelemetryDynamicSinkProvider.cs
```diff
@@ -15,6 +15,7 @@ namespace OpenTelemetry.Instrumentation.Remoting.Implementation;
 /// </summary>
 internal sealed class TelemetryDynamicSinkProvider : IDynamicProperty, IContributeDynamicSink, IDisposable
 {
+    internal const string ActivitySourceName = "OpenTelemetry.Instrumentation.Remoting";
     internal const string DynamicPropertyName = "TelemetryDynamicSinkProvider";
 
     private readonly ActivitySource activitySource = CreateActivitySource();
```
martincostello: Could you update this to use the new `ActivitySourceFactory` that got added this week please (see #4079)?

lewis800: Okay implemented

Kielek (approved):

lewis800:
@Kielek thank you for the approval, I just updated with main are you able to merge it now?

martincostello (approved):

Issue #515, opened 2022-07-19
[Instrumentation]OpenTelemetry.Instrumentation.ConfluentKafka


chad122:
The instrumentation of Confluent.Kafka

cijothomas:
See https://github.com/confluentinc/confluent-kafka-dotnet/issues/1269

csrakowski:
Hi @martincostello,
I have a personal interest in this instrumentation, and some of the open improvement requests.
If this was closed due to lack of maintainer/time, I would be willing to volunteer and give it some attention over the coming month(s).
Let me know if that is indeed the reason, what's needed from my side to take it over.

martincostello:
The preference would be to natively implement the library itself, which is tracked by the issue referenced above.

The maintainers seem to be open to the idea, they just aren't actively working on it.

I would suggest volunteering time to work on that issue in the first instance.

Issue #127, opened 2021-06-04
Add MongoDB instrumentation


glucaci:
There is aleady an implementation from @jbogard.

https://github.com/jbogard/MongoDB.Driver.Core.Extensions.DiagnosticSources

If he would like, we have to see what is needed to integrate it here.
I would like to see this part of the official OpenTelemetry packages.

lucasoares:
We started using MongoDB with .NET recently and this would be much appreciated. Having it as a official opentelemetry package would make it possible for us to start using it. @jbogard any chance on bringing https://github.com/jbogard/MongoDB.Driver.Core.Extensions.DiagnosticSources to openelemetry-dotnet-contrib?

Kielek:
1. MongoDB team promises native support for this: https://jira.mongodb.org/browse/CSHARP-3124. No updates so far.
2. This package will be exactly the same official as the one created by the @jbogard. The only difference will be the namespace and the nuget-owners.
3. It will be great to agree with @jbogard here that the old package will be dropped. Without this we will have 2 packages, and it is the worst case.
4. Db semantic convention going to be stabilized shortly. It is a blocker to release stable version, you can expect only pre-releases.

lucasoares:
> Clarification though - `opentelemetry-dotnet-contrib` does not make it "official" - that repo is just a collection of community-contributed packages. That's the "contrib" part of it. "Official" would be MongoDB support directly. Aspire's official MongoDB package already uses this package so I'm not sure what problems moving repos could possible solve here.
> 
> I would rather not move this over to a separate repo when this functionality should belong in the MongoDB driver itself. Perhaps the better solution here is to wait until the DB semantic convention to become stabilized then open a PR with the MongoDB driver repo itself?

Yeah, I totally agree with you. But company policies are company policies, right? HAHAHA I will wait for it then =)

jbogard:
I'm not suggesting putting your job at risk, but I can't imagine how "this package is used in Microsoft's public packages" wouldn't make this "official". See: https://www.nuget.org/packages/Aspire.MongoDB.Driver

It's official enough for Microsoft and Aspire, but isn't official enough for your company? Hmmmm 🤣

martincostello:
As noted above, existing packages exist, and there's a plan _at some point_ to add native instrumentation to the MongoDB driver.

We won't be adding a competing (and eventually obsolete) implementation here.

[...]

PR #3844, opened 2026-02-08
feat: create surreal instrumentation library

Odonno:
## Changes

Add a new library to do OTEL instrumentation of the .NET SDK for SurrealDB, following the PR started https://github.com/surrealdb/surrealdb.net/pull/219

* Add new library `OpenTelemetry.Instrumentation.SurrealDb`, inspied by the SQL client library
* Add new library `OpenTelemetry.Events.SurrealDb` to share events payload between the .NET client and the OTEL library

### DIFF
--- a/opentelemetry-dotnet-contrib.slnx
+++ b/opentelemetry-dotnet-contrib.slnx
@@ -134,6 +134,7 @@
     <File Path="README.md" />
   </Folder>
   <Folder Name="/src/">
+    <Project Path="src/OpenTelemetry.Events.SurrealDb/OpenTelemetry.Events.SurrealDb.csproj" />
     <Project Path="src/OpenTelemetry.Exporter.Geneva/OpenTelemetry.Exporter.Geneva.csproj" />
     <Project Path="src/OpenTelemetry.Exporter.InfluxDB/OpenTelemetry.Exporter.InfluxDB.csproj" />
     <Project Path="src/OpenTelemetry.Exporter.Instana/OpenTelemetry.Exporter.Instana.csproj" />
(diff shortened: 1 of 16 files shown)


martincostello:
Where possible, we prefer to not add new telemetry packages when the underlying dependency itself has native OpenTelemetry support (e.g the direction we've taken for MongoDB https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/127#issuecomment-3817903172 and are taking for Kafka https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/515#issuecomment-3831077914).

The linked PR appears to do that (it adds dependecies on `OpenTelemetry.Api` and `OpenTelemetry.Api.ProviderBuilderExtensions`), so why do we need an additional package here instead of that being "All-in-One"?

---

If we _do_ end up adding instrumentation here for Surreal, I don't think we want `OpenTelemetry.Events.SurrealDb` - it creates a weird cycle between this repository and Surreal's. Their library would depend on us, but then we would depend on them - fully updating things would take three PRs and releases (update the events and release, PR to Surreal to ingest that change and release, then another PR here to consume that update). There's also no precedent for such a pattern here, and I don't think it's a good idea to create one.

Odonno:
Hello @martincostello 

If not integrating the library, then I would like some guidance about:

* How to access the internal classes that are used the same way by the `SqlClient`?
```xml
<ItemGroup>
    <Compile Include="$(RepoRoot)\src\Shared\Configuration\*.cs" Link="Includes\Configuration\%(Filename).cs" />
    <Compile Include="$(RepoRoot)\src\Shared\DiagnosticSourceListener.cs" Link="Includes\DiagnosticSourceListener.cs" />
    <Compile Include="$(RepoRoot)\src\Shared\DiagnosticSourceSubscriber.cs" Link="Includes\DiagnosticSourceSubscriber.cs" />
    <Compile Include="$(RepoRoot)\src\Shared\EnvironmentVariables\*.cs" Link="Includes\EnvironmentVariables\%(Filename).cs" />
    <Compile Include="$(RepoRoot)\src\Shared\InstrumentationHandleManager.cs" Link="Includes\InstrumentationHandleManager.cs" />
    <Compile Include="$(RepoRoot)\src\Shared\ListenerHandler.cs" Link="Includes\ListenerHandler.cs" />
    <Compile Include="$(RepoRoot)\src\Shared\SemanticConventions.cs" Link="Includes\SemanticConventions.cs" />
  </ItemGroup>
```

* How to avoid bringing un-needed dependencies in the core package? I suppose I can create a new Intrumentation package in the SurrealDB.NET repository, still need to answer the previous question.

I can put the `OpenTelemetry.Events.SurrealDb` into the core package, this not a problem.

martincostello:
> How to access the internal classes that are used the same way by the `SqlClient`.

You wouldn't - the SurrealDB library would have its own code to implement things appropriately.

> How to avoid bringing un-needed dependencies in the core package

Depends on how the maintainers want the work there to be designed. That's a question for them.

PR #3848, opened 2026-02-09
[AWSLambda] Read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `FileInfo.LinkTarget` (.NET 6+) to read the symlink target
- Guarded by `#if NET` preprocessor directive for .NET Framework compatibility
- Silently skips if the symlink doesn't exist

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `AWSLambdaResourceDetector.cs` — Added readlink logic with `AccountIdSymlinkPath` constant
- `AWSLambdaResourceDetectorTests.cs` — Happy path, leading zeros, and missing symlink tests

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: leading zeros preserved as string
- [x] Unit test: missing symlink → attribute absent, no error

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
@@ -52,7 +52,7 @@ internal sealed class AWSLambdaResourceDetector : IResourceDetector
         }
         catch
         {
-            // Symlink doesn't exist or cannot be read — silently skip.
+            // Symlink doesn't exist or cannot be read -- silently skip.
             return null;
         }
     }
--- a/test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
+++ b/test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Runtime.InteropServices;
 using OpenTelemetry.AWS;
 using OpenTelemetry.Instrumentation.AWSLambda.Implementation;
 using Xunit;
@@ -27,6 +28,12 @@ public class AWSLambdaResourceDetectorTests : IDisposable
     [InlineData("000")]
     public void Detect_WithAccountIdSymlink_SetsCloudAccountId(string prefix)
     {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+        {
+            // Symlinks at /tmp/ are only available on Linux (Lambda runtime).
+            return;
+        }
+
         var expectedAccountId = prefix + Random.Shared.NextInt64(100000000, 999999999).ToString();
 
         File.CreateSymbolicLink(SymlinkPath, expectedAccountId);


src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
```diff
@@ -26,6 +28,7 @@ public Resource Detect()
                 .AttributeBuilder
                 .AddAttributeCloudProviderIsAWS()
                 .AddAttributeCloudRegion(AWSLambdaUtils.GetAWSRegion())
+                .AddAttributeCloudAccountID(GetAccountIdFromSymlink())
```
martincostello, line 34: Rather than make the method conditionally return null, instead make all the additions to this file conditional - there's no point having other TFMs add a null attribute value.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        var attributes = resource.Attributes.ToDictionary(x => x.Key, x => x.Value);
+
+        Assert.True(attributes.ContainsKey("cloud.account.id"));
+        Assert.Equal(expectedAccountId, attributes["cloud.account.id"]);
+    }
+
+    [Fact]
+    public void Detect_WithAccountIdSymlink_PreservesLeadingZeros()
+    {
+        const string expectedAccountId = "000123456789";
```
martincostello: I would generate a random value instead.

martincostello (commented):
Please add a CHANGELOG entry.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        File.CreateSymbolicLink(SymlinkPath, expectedAccountId);
+
+        var conventions = new AWSSemanticConventions(SemanticConventionVersion.Latest);
+        var detector = new AWSLambdaResourceDetector(conventions);
+
+        var resource = detector.Detect();
+        var attributes = resource.Attributes.ToDictionary(x => x.Key, x => x.Value);
+
+        Assert.Equal(expectedAccountId, attributes["cloud.account.id"]);
+    }
```
martincostello: These could maybe be collapsed into a single data-driven test with a `string prefix` input:

```csharp
    [Theory]
    [InlineData("")]
    [InlineData("000")]
    public void Detect_WithAccountIdSymlink_SetsCloudAccountId(string prefix)
    {
        var expectedAccountId = prefix + Random.Shared.NextInt64(100000000, 999999999).ToString();

        File.CreateSymbolicLink(SymlinkPath, expectedAccountId);

        var conventions = new AWSSemanticConventions(SemanticConventionVersion.Latest);
        var detector = new AWSLambdaResourceDetector(conventions);

        var resource = detector.Detect();
        var attributes = resource.Attributes.ToDictionary(x => x.Key, x => x.Value);

        Assert.Equal(expectedAccountId, attributes["cloud.account.id"]);
    }
```

src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaResourceDetector.cs
```diff
@@ -34,4 +41,20 @@ public Resource Detect()
...
+    private static string? GetAccountIdFromSymlink()
+    {
+        try
+        {
+            var fileInfo = new FileInfo(AccountIdSymlinkPath);
+            return fileInfo.LinkTarget;
+        }
+        catch
+        {
+            // Symlink doesn't exist or cannot be read -- silently skip.
```
martincostello, line 55: _A_ dash is fine as a replacement for an emdash

```suggestion
            // Symlink doesn't exist or cannot be read - silently skip.
```

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        this.Cleanup();
+    }
+
+#if NET
+    [Theory]
+    [InlineData("")]
+    [InlineData("000")]
+    public void Detect_WithAccountIdSymlink_SetsCloudAccountId(string prefix)
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
martincostello, line 31: I think it would be better to make this a `[SkippableTheory]` conditioned on the OS, rather than have it silently pass and do nothing.

Kielek:
@lukeina2z. @Oberon00, @rypdal, @normj,
Could you please check this PR? What are your recommendations?
Should we merge it/close it/mark package as abandoned?

lukeina2z:
I no longer work for Amazon and have no access to AWS to verify this. Sorry I could not help. Could you please remove me from the component owner list?

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/Implementation/AWSLambdaResourceDetectorTests.cs
RaphaelManke: Is there some general condition, as the whole detector is scoped down to lambdas and could use an early exit based on the non existence of several env vars or platform specific things.

martincostello: That's not necessary. The tests would never run in a true Lambda environment, so test set up has to set things up to simulate it anyway. Just skipping for the OS is fine.

normj:
Instead of relying on the symlink setup by the collector why not make the existing `GetAccountId` in `AWSLambdaHttpUtils` internal and use that. That avoids having to rely on the layer and extracts the account id from parsing the function arn.

RaphaelManke:
That sounds reasonable. The only question is, will I be able to set a resource attribute? Having it as span attribute would be wrong and only be present on the initial span not the nested ones. 

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Instrumentation.AWSLambda/Implementation/AWSLambdaUtils.cs 

Here we already set the account id. But as a span attribute but not as a resource attribute. 

The Problem is that the Lambda environment does not expose the account id at bootstrap time, only for the extension api or at invoke time. 

The impact is also for aws o11y tools. For example the cloudwatch transaction search will not return all spans for an account because the account is not set on all spans of a function. 

The only alternative I see in order  to calculate the account id at bootstrap time is to use the AWS credentials to calculate the account id 

https://hackingthe.cloud/aws/enumeration/get-account-id-from-keys/

normj:
@RaphaelManke you are right during the init phase when we don't have a `ILambdaContext` we don't have access the function arn. To maximize the amount of users that can get the account id what about if we check for the symlink like you have now but if that does not exist and we do have a the `ILambdaContext` we pull it from there. In that case we wouldn't have for the init phase but at least we would have it for each invocation.

RaphaelManke:
I just noticed that the AWS_ACCESS_KEY_ID env var is always available at the beginning. We can use that to calc the account id. 

https://hackingthe.cloud/aws/enumeration/get-account-id-from-keys/

Would you prefer that as it's independent of the extension?

normj:
> I just noticed that the AWS_ACCESS_KEY_ID env var is always available at the beginning. We can use that to calc the account id.
> 
> https://hackingthe.cloud/aws/enumeration/get-account-id-from-keys/
> 
> Would you prefer that as it's independent of the extension?

That is an interesting hack to get the account id from the access key id but that is not a contract that AWS won't possible change sometime in the future as the auth system evolves. I don't want to bake that assumption into this library.

PR #3864, opened 2026-02-13
[Infra] Add tests for all EventSource implementations

martincostello:
Inspired by https://github.com/dotnet/aspnetcore/pull/65408.

## Changes

- Add tests to validate all implementations of `EventSource`.
- Refactor `EventSource` validation to improve assertions.
- Remove redundant comments.
- Rename test classes `EventSourceTests` for consistency.
- Seal all classes that derive from `EventSource` to fix new tests.

Example assertion failure before:

```text
    System.Exception : Method 'AspNetInstrumentationEventSource.RequestIsFilteredOut' is implemented incorrectly.
    ---- System.InvalidOperationException : Sequence contains no matching element
```

Example assertion failure after:

```text
    EventSource 'OpenTelemetry.Instrumentation.AspNet.Implementation.AspNetInstrumentationEventSource' has event ID validation error(s):
    Duplicate EventId 4: methods 'InvalidConfigurationValue' and 'NullActivity' share the same ID.
    Event NullActivity is givien event ID 4 but 5 was passed to WriteEvent.
    Event NullActivity has ID 4 which is already in use.
    Event NullActivity (with ID 4) has the same task/opcode pair as event InvalidConfigurationValue (with ID 4).
```

### DIFF
--- a/test/OpenTelemetry.Resources.OperatingSystem.Tests/EventSourceTests.cs
+++ b/test/OpenTelemetry.Resources.OperatingSystem.Tests/EventSourceTests.cs
@@ -4,9 +4,9 @@
 using OpenTelemetry.Tests;
 using Xunit;
 
-namespace OpenTelemetry.Resources.OperatingSystem.Test;
+namespace OpenTelemetry.Resources.OperatingSystem.Tests;
 
-public class EventSourceTest
+public class EventSourceTests
 {
     [Fact]
     public void EventSourceTests_OperatingSystemResourcesEventSource() =>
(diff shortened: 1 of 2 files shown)


test/OpenTelemetry.Exporter.Geneva.Tests/EventSourceTests.cs
```diff
@@ -0,0 +1,23 @@
...
+using OpenTelemetry.Tests;
+using Xunit;
+
+namespace OpenTelemetry.Exporter.Geneva.Tests;
+
+public class EventSourceTests
+{
+    [Fact]
+    public void EventSourceTests_EtwEventSource() =>
+        EventSourceTestHelper.ValidateEventSourceIds<EtwDataTransport.EtwEventSource>();
```
cijothomas, line 14: nit: unlike the other EventSources, these are not meant for internal logging - they are normal exporting itself, and need not have validations.

matt-hensley (approved):

PR #3881, opened 2026-02-18
Geneva Exporter: Improve multithreaded support

mattsains:
## Changes

Reviewed the trace and log exporter code and guarded against some potential race conditions.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [x] Changes in public API reviewed (if applicable)

PR #3882, opened 2026-02-18
[Exporter.Geneva] Add thread-safety regression tests for MsgPackTraceExporter

rajkumar-rangaraj:
Adds regression tests that verify `MsgPackTraceExporter.SerializeActivity` produces valid MessagePack output under concurrent multi-threaded access.

PR #3214 introduced a race condition where multiple threads calling `SerializeActivity` simultaneously would each invoke `CreateFraming()`, concurrently mutating the shared `prepopulatedFields` dictionary. This caused `Dictionary` internal state corruption and resulted in "Bad forward protocol format" errors on the receiving end.

The fix was delivered in #3881, but no regression tests existed to cover the
multi-threaded scenario. These tests fill that gap.

## Tests added

- **`SerializeActivity_ConcurrentThreads_ProducesValidMsgPack`** — 8 threads
  serialize activities concurrently and validate that every output is a well-formed
  Fluentd Forward Mode MessagePack message.
- **`CreateFraming_CalledMultipleTimes_FieldCountRemainsConsistent`** — Calls
  `CreateFraming()` multiple times and asserts that `prepopulatedFields.Count`
  and `bufferPrologue.Length` remain stable.
- **`SerializeActivity_DifferentThreads_SameFieldCount`** — 8 threads serialize
  activities concurrently and verify the Map16 field count is identical across
  all threads.

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Fixed race condition in `MsgPackTraceExporter` where concurrent threads
+  calling `SerializeActivity` would corrupt the `prepopulatedFields` dictionary,
+  leading to "Bad forward protocol format" errors.
+  ([#3881](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3881))
+
 ## 1.15.0
 
 Released 2026-Jan-21
--- a/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
@@ -4,6 +4,7 @@
 #if NET
 using System.Collections.Frozen;
 #endif
+using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
@@ -86,7 +87,7 @@ internal sealed class MsgPackTraceExporter : MsgPackExporter, IDisposable
 
     // this stores the prepopulated fields until CreateFraming can consume them into the prologue.
     // after CreateFraming is called, this dictionary is set to null, so don't use it after that.
-    private Dictionary<string, object> prepopulatedFields;
+    private ConcurrentDictionary<string, object> prepopulatedFields;
 
     private bool isDisposed;
 
@@ -152,12 +153,12 @@ internal sealed class MsgPackTraceExporter : MsgPackExporter, IDisposable
 
         this.userProvidedPrepopulatedFields = options.PrepopulatedFields != null && options.PrepopulatedFields.Count > 0;
 
-        this.prepopulatedFields = new Dictionary<string, object>(0, StringComparer.Ordinal);
+        this.prepopulatedFields = new ConcurrentDictionary<string, object>(StringComparer.Ordinal);
         if (options.PrepopulatedFields != null)
         {
             foreach (var entry in options.PrepopulatedFields)
             {
-                this.prepopulatedFields.Add(entry.Key, entry.Value);
+                this.prepopulatedFields[entry.Key] = entry.Value;
             }
         }
 
@@ -347,15 +348,6 @@ internal sealed class MsgPackTraceExporter : MsgPackExporter, IDisposable
 
         var resourceAttributes = this.resourceProvider().Attributes;
 
-        if (this.resourceFieldNames != null)
-        {
-            // if ResourceFieldNames is set, we use resource attributes rather than PrepopulatedFields
-            this.prepopulatedFields = new Dictionary<string, object>(0, StringComparer.Ordinal);
-        }
-
-        // this is guaranteed to not be null because it's set in the constructor
-        Guard.ThrowIfNull(this.prepopulatedFields);
-
         foreach (var resourceAttribute in resourceAttributes)
         {
             var key = resourceAttribute.Key;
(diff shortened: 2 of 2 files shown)


test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                        activity.SetTag("thread", threadIndex);
+                        activity.SetTag("iteration", i);
+
+                        // This calls CreateFraming() on the first invocation per thread,
+                        // racing with other threads doing the same.
+                        var serialized = exporter.SerializeActivity(activity);
+
+                        // Validate: the serialized data must be valid MessagePack
+                        // representing a proper Fluentd Forward Mode message.
+                        // Under the race condition, the Map16 header field count
```
mattsains, line 112: I get that this test is a reaction to the bug, but this is the fifth time in this file that the bug is explicitly called out - is that necessary? This test will live on as a general test of multi-threading support, not simply that the exact bug we identified won't come back. In my view, the file inappropriately refers back to the specific issue that caused us to realise the need for this test throughout, and these references should mostly be removed

test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+        var resource = new Resource(resourceAttributes);
+
+        using var exporter = new MsgPackTraceExporter(exporterOptions, () => resource);
+
+        // Call CreateFraming multiple times to simulate the multi-thread scenario.
+        // Each call should produce the same consistent state.
+        // In the buggy code, repeated calls can cause prepopulatedFields to grow
+        // (duplicate entries added from resource attributes), making the Count
+        // diverge from what's in the prologue.
+        exporter.CreateFraming();
```
mattsains, line 237: I don't think it makes sense to call CreateFraming in the tests. Perhaps CreateFraming should be changed to be private.

This test does not do what it says it does - CreateFraming is called twice, but synchronously, so it just does the same thing twice. I think this test should be removed.

test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+        const int threadCount = 8;
+        const int iterationsPerThread = 50;
+        var barrier = new Barrier(threadCount);
+        var exceptions = new System.Collections.Concurrent.ConcurrentBag<Exception>();
+
+        // Create a real activity to serialize
+        using var activitySource = new ActivitySource("ThreadSafetyTest");
+        using var listener = new ActivityListener
+        {
+            ShouldListenTo = _ => true,
```
mattsains, line 79: The other tests following a convention of only listening to the specific activity source name of the test, and comments say this is to avoid interference with other tests. However, I am not sure this is actually possible because the listener will not be attached to any other tests' activity sources. Anyway, not sure what to suggest here but just pointing out it's different to the other tests' layout.

test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                        // representing a proper Fluentd Forward Mode message.
+                        // Under the race condition, the Map16 header field count
+                        // doesn't match actual serialized fields, causing this
+                        // deserialization to throw.
+                        try
+                        {
+                            var data = new byte[serialized.Count];
+                            Array.Copy(serialized.Array!, serialized.Offset, data, 0, serialized.Count);
+
+                            var deserialized = MessagePack.MessagePackSerializer.Deserialize<object>(
```
mattsains, line 120: To improve this test, you can manually construct a MessagePackReader, and make sure that all the bytes have been consumed by the deserialize call. This way we ensure that there are no extra message pack encoded fields that were left off by an incorrect field count. That would also let you remove the test called CreateFraming_CalledMultipleTimes_FieldCountRemainsConsistent

test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                            Assert.Contains("env_dt_traceId", mapping.Keys);
+                            Assert.Contains("env_dt_spanId", mapping.Keys);
+                            Assert.Contains("name", mapping.Keys);
+                            Assert.Contains("kind", mapping.Keys);
+
+                            // Verify resource-derived fields
+                            Assert.Contains("env_cloud_role", mapping.Keys);
+                            Assert.Equal("TestService", mapping["env_cloud_role"]);
+                            Assert.Contains("env_cloud_roleInstance", mapping.Keys);
+                            Assert.Equal("Instance123", mapping["env_cloud_roleInstance"]);
```
mattsains, line 155: I suggest asserting the number of fields in mapping to ensure there are no extra, unintended fields

test/OpenTelemetry.Exporter.Geneva.Tests/MsgPackTraceExporterThreadSafetyTests.cs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    /// <summary>
+    /// Verifies that different threads get consistent serialization results.
+    /// Each thread serializes the same activity and the resulting Map16 field count
+    /// (embedded in the buffer at mapSizePatchIndex) should be identical across threads.
+    /// Under the race condition, different threads can end up with different field counts.
+    /// </summary>
+    [Fact]
+    public void SerializeActivity_DifferentThreads_SameFieldCount()
```
mattsains, line 273: see my comment above discussing a way to roll this test into SerializeActivity_ConcurrentThreads_ProducesValidMsgPack

PR #3887, opened 2026-02-24
[WCF] Modernize and refactor tests

matt-hensley:
## Changes

WCF tests updated to use modern C# features, share constants, and common assertions extracted.

Tests should be functionally identical.

This is in preparation for updating to latest RPC semconv.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfInstrumentationConstants.cs
+++ b/src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfInstrumentationConstants.cs
@@ -5,15 +5,11 @@ namespace OpenTelemetry.Instrumentation.Wcf;
 
 internal static class WcfInstrumentationConstants
 {
-    public const string SoapMessageVersionTag = "soap.message_version";
-    public const string SoapReplyActionTag = "soap.reply_action";
-    public const string SoapViaTag = "soap.via";
-    public const string WcfChannelSchemeTag = "wcf.channel.scheme";
     public const string AttributeSoapMessageVersion = "soap.message_version";
     public const string AttributeSoapReplyAction = "soap.reply_action";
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry.Instrumentation.Wcf/Implementation/WcfInstrumentationConstants.cs
Kielek: worth to introduce same naming convention a in shared SemanticConvention file.

```suggestion
public const string AttributeSoapMessageVersion = "soap.message_version";
public const string AttributeSoapReplyAction = "soap.reply_action";
public const string AttributeSoapVia = "soap.via";
public const string AttributeWcfChannelScheme = "wcf.channel.scheme";
public const string AttributeWcfChannelPathT = "wcf.channel.path";
```

test/OpenTelemetry.Instrumentation.Wcf.Tests/TelemetryDispatchMessageInspectorTests.netfx.cs
```diff
@@ -114,8 +114,6 @@ public async Task IncomingRequestInstrumentationTest(
                                     case WcfEnrichEventNames.BeforeSendReply:
                                         activity.SetTag("server.beforesendreply", WcfEnrichEventNames.BeforeSendReply);
                                         break;
-                                    default:
-                                        break;
                                 }
```
Kielek, line 116: Something change in the C# analyzers. I am pretty sure that I was forced to add empty defaults, could you please double check it?

matt-hensley: I'm surprised the exhaustive case check isn't failing. Probably an edge case or bug.

I put the default case back.

Kielek (approved):

PR #4078, opened 2026-04-08
[EntityFrameworkCore] Add schema URL to traces

martincostello:
Contributes to #4064.

## Changes

Add schema URL to the `ActivitySource`.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] ~~Changes in public API reviewed (if applicable)~~

Issue #4064, opened 2026-04-05
Add schema URL to ActivitySources and Meters


martincostello:
I noticed that various libraries are not adding a schema URL to their `ActivitySource` and/or `Meter` instances.

Where possible, we should add them:

- [x] ~~Instrumentation.AspNetCore~~ - #4066 #4373
- [x] ~~Instrumentation.AWS~~ - #4063
- [x] ~~Instrumentation.AWSLambda~~ - #4063
- [x] ~~Instrumentation.Cassandra~~ - #4639
- [x] ~~Instrumentation.ConfluentKafka~~ - #4636
- [ ] ~~Instrumentation.ElasticsearchClient~~ - ~~#4635~~
- [x] ~~Instrumentation.EntityFrameworkCore~~ - #4078
- [x] ~~Instrumentation.EventCounters~~ - #4640
- [x] ~~Instrumentation.GrpcCore~~ - #4338
- [x] ~~Instrumentation.GrpcNetClient~~ - #4338
- [x] ~~Instrumentation.Hangfire~~ = #4641
- [x] ~~Instrumentation.Http~~ - #4082
- [x] ~~Instrumentation.Owin~~ - #4375
- [x] ~~Instrumentation.Process~~ - #4088
- [x] ~~Instrumentation.Quartz~~ - #4642
- [x] ~~Instrumentation.Runtime~~ - #4087
- [ ] Instrumentation.ServiceFabricRemoting - #5401
- [x] ~~Instrumentation.StackExchangeRedis~~ - #4095
- [x] ~~Instrumentation.Wcf~~ - #4377

Kielek:
Note for anybody who would like to contribute together with @martincostello .
It is not only about providing url. It requires also check if the implementation met the criteria described in the semantic convention.

martincostello:
All the specific violations are now flagged in #4068

martincostello:
Just OpenTelemtry.Instrumentation.ServiceFabricRemoting left.

PR #4079, opened 2026-04-08
[Infra] Add factory for ActivitySource and Meter

martincostello:
Contributes to #4064.

## Changes

Add shared code to minimise duplication when creating instances of `ActivitySource` and `Meter` to help ensure that telemetry schema URLs are set.

While working on #4078 I realised that there was a lot of duplication starting to happen, so this centralises that code and adopts it for use for AspNet and SqlClient instrumentation which are the only two that currently set `TelemetrySchemaUrl`.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNet/AspNetInstrumentation.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNet/AspNetInstrumentation.cs
@@ -4,22 +4,20 @@
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using OpenTelemetry.Instrumentation.AspNet.Implementation;
-using OpenTelemetry.Internal;
+using OpenTelemetry.Trace;
 
 namespace OpenTelemetry.Instrumentation.AspNet;
 
 /// <summary>
-/// Asp.Net Requests instrumentation.
+/// ASP.NET Requests instrumentation.
 /// </summary>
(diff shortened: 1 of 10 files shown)


src/OpenTelemetry.Instrumentation.AspNet/AspNetInstrumentation.cs
```diff
@@ -4,22 +4,20 @@
...
 internal sealed class AspNetInstrumentation : IDisposable
 {
     public static readonly AspNetInstrumentation Instance = new();
 
-    private static readonly (ActivitySource ActivitySource, Meter Meter) Telemetry = CreateTelemetry();
-#pragma warning disable SA1202 // Elements must be ordered by accessibility. Telemetry field should be private and initialized earlier
-    public static readonly ActivitySource ActivitySource = Telemetry.ActivitySource;
-#pragma warning restore SA1202 // Elements must be ordered by accessibility. Telemetry field should be private and initialized earlier
-    public static readonly Meter Meter = Telemetry.Meter;
+    public static readonly Version SemanticConventionsVersion = new(1, 36, 0);
```
Kielek, line 18: I would consider making it as private.

martincostello: I think this is neater than needing to suppress the style warning in ~20 places once all the relevant places are updated to use these new factory methods.

Kielek (approved):

PR #4068, opened 2026-04-07
[Infra] Update banned API analyzers

martincostello:
Relates to #4064.

## Changes

Ban constructors for `Activity` and `Meter` that do not support setting the telemetry schema URL.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [ ] ~~Unit tests added/updated~~
* [ ] ~~Appropriate `CHANGELOG.md` files updated for non-trivial changes~~
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #4087, opened 2026-04-09
[Runtime] Use MeterFactory

martincostello:
## Changes

- Use `MeterFactory` to create `Meter`.
- Allow `null` for the semantic conventions version.
- Fix incorrect `MeterFactory` namespace.
- Add support for optional tags.
- Add support for overriding the meter name.

~~I didn't realise the runtime instrumentation doesn't implement a semantic convention at all (at least I don't _think_ it does), so also special-cases a `0.0.0` version to allow opting-out (to avoid needing a suppression in #4068), but I'm open to alternative approaches - I just want to make it a pit of failure to _not_ specify a version.~~

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
@@ -2,8 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics.Metrics;
+#if NET
 using System.Reflection;
-using OpenTelemetry.Internal;
+#endif
 #if NET
 using JitInfo = System.Runtime.JitInfo;
 #endif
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
```diff
@@ -15,11 +16,7 @@ namespace OpenTelemetry.Instrumentation.Runtime;
 /// </summary>
 internal sealed class RuntimeMetrics
 {
-    internal static readonly Assembly Assembly = typeof(RuntimeMetrics).Assembly;
-    internal static readonly AssemblyName AssemblyName = Assembly.GetName();
-#pragma warning disable IDE0370 // Suppression is unnecessary
-    internal static readonly Meter MeterInstance = new(AssemblyName.Name!, Assembly.GetPackageVersion());
-#pragma warning restore IDE0370 // Suppression is unnecessary
+    internal static readonly Meter MeterInstance = Trace.MeterFactory.Create<RuntimeMetrics>(new(0, 0, 0)); // These metrics are not in the Semantic Conventions
```
Kielek: Alternative option - change the contract of Create to Create<T>(Version? version) and handle there null gracefully. It might be less tricky.

Kielek (approved):

src/OpenTelemetry.Instrumentation.Runtime/RuntimeMetrics.cs
martincostello: Yeah I wasn't entirely sure what do do about this (see PR description). I'll change it to nullable.

PR #4036, opened 2026-03-29
[Resources.Process] Add process.creation.time

martincostello:
Contributes to #1515.

## Changes

Collect new attribute based on [v1.41.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/resource/process.md) of the semantic conventions:

- ~~`process.args_count`~~
- ~~`process.command`~~
- `process.creation.time`
- ~~`process.executable.path`~~
- ~~`process.title`~~
- ~~`process.working_directory`~~

~~Based on [_Selecting process attributes_](https://github.com/open-telemetry/semantic-conventions/blob/v1.40.0/docs/resource/process.md#selecting-process-attributes), I went with `process.executable.path` as it can be used to derive `process.executable.name`, and `process.args_count` so that we do not need to implement redaction for `process.command_line` and/or `process.command_args`.~~

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #4088, opened 2026-04-09
[Instrumentation.Process] Add schema URL

martincostello:
Contributes to #4064.

## Changes

- Add schema URL to metrics.
- Removed the `process.cpu.count` metric.


src/OpenTelemetry.Instrumentation.Process/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Updated OpenTelemetry core component version(s) to `1.15.2`.
   ([#4080](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4080))
 
+* Add instrumentation scope version and schema URL to metrics.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#4088](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4088))
```

Kielek (commented):
It is not fully true.

`process.cpu.count` is not defined in the otel spec.
I think that we should drop it first and then merge it.

Ultimate solution, include metrics:
https://github.com/dotnet/runtime/blob/a534a1c68e656b07d16015e2f46fa55b25bfd0b9/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/RuntimeMetrics.cs#L143-L147

for .NET Fx + .NET8 in Runtime Instrumentation package + link changelog to use other packages during migration.

martincostello:
~~I think it's better to do that as part of https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4036.~~

~~They could go out in the same release anyway.~~

~~I just intentionally split the infrastructure to report the schema (be status quo) out from changing the implementation.~~

Nevermind, I'd forgotten they are different packages that just both have Process in the name.

martincostello:
> I think that we should drop it

Done.

martincostello:
https://github.com/open-telemetry/semantic-conventions/issues/651 is tracking the ask to make this part of the actual Semantic Conventions.

Kielek:
On the other hand we have already https://github.com/open-telemetry/semantic-conventions/blob/main/docs/runtime/dotnet-metrics.md#metric-dotnetprocesscpucount

I would not declare this metric as part of 1.25.0, as it is not defined there.

martincostello:
> I would not declare this metric as part of 1.25.0, as it is not defined there.

It isn't because it's been deleted? If we don't emit it it isn't associated with that schema URL.

Kielek (approved):
I missed that you have already committed fix here.

PR #4128, opened 2026-04-16
[PersistentStorage.*] Support ReadOnlySpan as buffer + target .NET

Kielek:
Needed for https://github.com/open-telemetry/opentelemetry-dotnet/pull/7101/changes#diff-e4cd4e54cc3d2001a98d2c0ee34839696685e49d40de10f73e812f8f8b45fc78R59

## Changes

[PersistentStorage.*] Support ReadOnlySpan as buffer + target .NET
It should avoid creating separate byte array while storing data on the disk.

There is an option to apply breaking changes in the API.  I have found 2 significant usages by nuget.org
https://www.nuget.org/packages/Azure.Monitor.OpenTelemetry.Exporter/
and https://github.com/Zongsoft/framework (minority of downloads).


Kielek:
@rajkumar-rangaraj, looking for you opinion here as the main usage is OTLP + Azure exporter.

src/OpenTelemetry.PersistentStorage.Abstractions/PersistentBlob.cs
```diff
@@ -107,6 +133,9 @@ public bool TryDelete()
 
     protected abstract bool OnTryWrite(byte[] buffer, int leasePeriodMilliseconds = 0);
 
+    protected virtual bool OnTryWriteSpan(ReadOnlySpan<byte> buffer, int leasePeriodMilliseconds)
+        => this.OnTryWrite(buffer.ToArray(), leasePeriodMilliseconds);
```
martincostello, line 137: Just to check I'm following this correctly, so by default the span overload just calls the array one (so won't get the benefits), but our own derived implementations override this so they benefit from the improvement?

martincostello (approved):

src/OpenTelemetry.PersistentStorage.Abstractions/PersistentBlob.cs
Kielek: Yes, IMO it is the only way to avoid breaking change in the API.

xiang17 (approved):

Kielek:
@xiang17, what do you think about follow up PR to fully remove legacy API (based on byte[]). As a follow up, we need to make release 2.0.

xiang17:
> @xiang17, what do you think about follow up PR to fully remove legacy API (based on byte[]). As a follow up, we need to make release 2.0.

Should we wait a while after setting the legacy APIs to `Obsolete`? How long should we wait before removing them fully?

Kielek:
@xiang17, I do not think that it is about time. It is about bumping version to 2.0.0.
The most popular usage is https://www.nuget.org/packages/Azure.Monitor.OpenTelemetry.Exporter/ so I am asking in this context.

PR #4116, opened 2026-04-15
[OpAMP] Apply response size limits for oversized responses

stevejgordon:
## Changes

- Includes improvements to event source logging for diagnostics.
- Includes additional tests to confirm expected size limits are applied.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* ~[ ] Changes in public API reviewed (if applicable)~

PR #4130, opened 2026-04-16
[Infra] Use HttpClientHelpers 

martincostello:
## Changes

- Refactor OpenTelemetry.Sampler.AWS to use the new `HttpClientHelpers` shared code to enforce response size limits.
- Refactor OpenTelemetry.OpAmp.Client to move the code to limit the response size to the new `HttpClientHelpers` shared code.
- Fix stream not being disposed.


src/OpenTelemetry.OpAmp.Client/Internal/OpAmpClientEventSource.cs
```diff
@@ -15,8 +15,7 @@ internal sealed class OpAmpClientEventSource : EventSource
     private const int EventIdInvalidWsFrame = 1;
     private const int EventIdTransportCloseFailure = 2;
     private const int EventIdOversizedResponseContentLength = 3;
-    private const int EventIdOversizedResponseBody = 4;
-    private const int EventIdHttpResponseReceived = 5;
+    private const int EventIdHttpResponseReceived = 4;
```
martincostello, line 18: As we haven't shipped #4116 yet, should be fine to re-number the event.

src/OpenTelemetry.Sampler.AWS/AWSXRaySamplerClient.cs
```diff
@@ -122,28 +124,25 @@ private void Dispose(bool disposing)
         }
     }
 
-    private async Task<string> DoRequestAsync(string endpoint, HttpRequestMessage request)
+    private async Task<string> DoRequestAsync(string endpoint, HttpRequestMessage request, CancellationToken cancellationToken)
     {
         // 1 MB is well above any legitimate X-Ray sampling rules/targets
         // response while still protecting against unbounded reads.
-        const int maxResponseSizeInBytes = 1024 * 1024;
+        const int MaxResponseSizeInBytes = 1024 * 1024;
```
stevejgordon, line 131: Is PascalCasing standard in the repo for local const?

src/Shared/HttpClientHelpers.cs
```diff
@@ -5,30 +5,147 @@
...
+#endif
+            .ConfigureAwait(false);
+
+#if NET
+        await using (stream.ConfigureAwait(false))
+#else
+        using (stream)
+#endif
+        {
+            var length = GetBufferLength(stream, limit);
```
stevejgordon: As some stream types may throw and therefore we get the max limit back, should we also use the content length (when present) to reduce the buffer we allocate here?

src/OpenTelemetry.Sampler.AWS/AWSXRaySamplerClient.cs
martincostello: That's what I've always done - otherwise it's not clear in isolation at usage sites if they're a constant or a variable.

stevejgordon: Fair enough, I've never seen it, but const locals are quite rare. I'm fine with it!

src/OpenTelemetry.Sampler.AWS/AWSXRaySamplerClient.cs
```diff
@@ -24,7 +24,7 @@ public AWSXRaySamplerClient(string host)
         this.httpClient = new HttpClient();
     }
 
-    public async Task<List<SamplingRule>> GetSamplingRules()
+    public async Task<List<SamplingRule>> GetSamplingRules(CancellationToken cancellationToken = default)
```
normj, line 27: I always treat adding new parameters with a default value as still a breaking change since it breaks binary compatibility. Is the logic that since this is probably called at the top of stack where users would be recompiling when they take in the dependency this is okay?

src/Shared/HttpClientHelpers.cs
```diff
@@ -5,30 +5,147 @@
...
+#if NET
+        await using (stream.ConfigureAwait(false))
+#else
+        using (stream)
+#endif
+        {
+            var length = GetBufferLength(stream, limit, httpResponse.Content.Headers.ContentLength);
+
+#if NET
+            var buffer = ArrayPool<byte>.Shared.Rent(length);
```
normj, line 53: This is a pretty large `buffer` to preallocated to just be thrown away at the end of this method. You might get lucky with a 1 to 4 MB buffer being already in the pool.

You could repurpose the `LimitedStream` that was in my original code and wrap the stream in the `HttpResponseMessage`. That way there is no extra allocation for this. You could something like
```
response.Content = new StreamContent(new LimitedStream(response.Content.ReadAsStream(), limit));
```

You might also need to copy the content headers from the original HttpContent.

src/OpenTelemetry.Sampler.AWS/AWSXRaySamplerClient.cs
martincostello: This class is internal, so there's no change to the public API surface. The caller of this doesn't have any public API that allows us to pass through a `CancellationToken` at the moment. I changed this so that if callers have a token, then the compiler will flag a warning to pass it through.

src/Shared/HttpClientHelpers.cs
martincostello: The array given back might be any size as long as it's at least the length we need, and then we'll only use up to the size we asked for - it could be any buffer already created elsewhere by application code and it might be allocation-free at the point of use. It might not be new and will get reused when it goes back into the pool on line 115.

We had 4 different call paths that needed to enforce a limit on the stream size, and the sampler was the odd-one-out in terms of the approach with a wrapped stream, so I moved it to use the same approach as the others already had to minimise how many ways we have to do the same thing.

normj: Understood the ArrayPool will help with the reuse but given the default size of the buffer is 4MB that is going to have be create at least once even though it looks like most responses will just be a few KB. In Lambda's case with constraint CPU/Memory that could affect the cold start.

I can go along with the code as is but it seems an unnecessary large allocation for a safety check that we could avoid.

martincostello: 4MiB is the default, but OpAMP is overriding it via `TransportConstants.MaxMessageSize`, and the AWS sampler is overriding it to `const int MaxResponseSizeInBytes = 1024 * 1024`.

The 4MiB as the default is driven by this change: https://github.com/open-telemetry/opentelemetry-dotnet/pull/7017.

We could always just remove the default and require all callers to specify it.

Kielek (approved):

Kielek:
@normj, @stevejgordon, feel free to rise post-merge comment here if you think that we need to improve something.

stevejgordon:
@Kielek - I raised #4143 to consider one usability issue with debugging, but it's not critical.

PR #4121, opened 2026-04-15
[Resources.Azure] Limit response body size read

martincostello:
## Changes

Limit the length of the HTTP response body that is read from the Azure VM instance metadata service.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #4133, opened 2026-04-16
[OpAMP] Harden WebSocket transport

stevejgordon:
Async receive loop, max-message enforcement, ClientWebSocketFactory, and disposal fixes

## Changes

- Replace background Thread with async Task in WsReceiver so the receive loop is cancellable and disposal is deterministic (blocks until the task drains, prevents use-after-free on the socket)
- Remove the insecure DangerousAcceptAnyServerCertificateValidator default on .NET; users who need custom TLS can now set ClientWebSocketFactory
- Add OpAmpClientSettings.ClientWebSocketFactory (new public API) so callers control socket options (headers, TLS, etc.)
- Fix OpAmpClient.Dispose() to actually stop background services and dispose the transport (previously only dispatcher.Dispose() was called)
- Catch WebSocketException/ObjectDisposedException in the receive path so aborted-socket errors propagate cleanly rather than as unobserved faults
- Catch frame-processing exceptions and emit a FrameProcessingFailure event so a single bad frame cannot kill the long-lived connection
- Fix array-pool return bug: was returning the non-pooled receiveBuffer - Fix multi-buffer sequence construction (workingCount off-by-one) - Add OversizedWebSocketMessage and FrameProcessingFailure EventSource events - Add ThrowIfDisposed guards to all public OpAmpClient and WsTransport methods - Add new tests covering boundary conditions, disposal, factory config, invalid/text
  frames, fragmented oversized messages, and error resilience
- Switch StopAsync from CloseAsync to CloseOutputAsync so the graceful close handshake does not conflict with the concurrent ReceiveAsync in WsReceiver; also fixes a potential hang on .NET Framework where CloseAsync blocks waiting for the server close frame
- Fix single-buffer sequence construction: slice receiveBuffer to totalCount so the frame does not include uninitialised tail bytes

## Behaviour Notes

1. **Security fix (breaking for some users): TLS certificate validation is no longer bypassed**

   The .NET-only `DangerousAcceptAnyServerCertificateValidator` default has been removed. Any user
   connecting to an OpAMP server with a self-signed or non-standard TLS cert will need to configure
   `ClientWebSocketFactory` and set `ClientWebSocket.Options` appropriately. HTTP transport is
   unaffected.

2. **New public API: `OpAmpClientSettings.ClientWebSocketFactory`**

   Callers can now supply a factory that returns a pre-configured `ClientWebSocket`. This is the
   migration path for anyone who was relying on the removed TLS bypass, and it enables setting custom
   request headers, keep-alive settings, etc.

3. **`OpAmpClient.Dispose()` now releases transport and service resources**

   Previously `Dispose()` only called `dispatcher.Dispose()` and leaked the `IOpAmpTransport` and
   background services. `Dispose()` now stops all services and disposes the transport. Callers who
   were relying on `Dispose()` to be a no-op (e.g. GC-only cleanup) will now see that resources are
   actively released.

4. **`Dispose()` is not a graceful shutdown — `StopAsync()` should be called first**

   `Dispose()` aborts the `WebSocket` immediately without sending the agent disconnect message. The
   README and XML docs have been updated to make this explicit. Callers who previously called only
   `Dispose()` and expected a clean server-side unregister should switch to `await StopAsync()`
   followed by `Dispose()` (or `using`).

5. **`StopAsync` switches from `CloseAsync` to `CloseOutputAsync`**

   `CloseAsync` internally calls `ReceiveAsync`, which conflicts with the concurrent `ReceiveAsync`
   running in `WsReceiver` and can cause a deadlock or hang on .NET Framework. `CloseOutputAsync`
   only sends the close frame; the receive loop consumes the server's close response naturally and
   exits.

6. **Frame processing errors no longer crash the receive loop**

   A bad or unrecognisable server frame now logs a `FrameProcessingFailure` warning event and moves
   on to the next frame, rather than crashing the `async void` receive thread silently. This changes
   the failure mode from "connection dies silently" to "frame is dropped, warning is logged,
   connection stays up."

7. **Disposal of `WsReceiver` is now synchronous and deterministic**

   `Dispose()` cancels the receive task and blocks (via `GetAwaiter().GetResult()`) until it
   finishes, preventing the socket from being used concurrently after disposal. The previous
   implementation called `Thread.Join()` but never cancelled the cancellation token inside
   `Dispose()`, so if the caller had not cancelled externally the thread would block indefinitely on
   a live socket. The new path always has a cancellation route: `disposeTokenSource` is cancelled
   before the wait, and `WsTransport.Dispose()` calls `webSocket.Abort()` first, which unblocks
   `ReceiveAsync` almost immediately.

8. **Updated `TestHttpServer` and `TestWebSocketServer`**

   The original `RunningServer` used `new Task(async () => { ... })`, which is a known anti-pattern
   where the outer `Task` completes at the first `await` rather than when the async work finishes —
   meaning `Dispose()` was not actually waiting for the listener loop to complete. The fix replaces
   this with `Task.Run(() => this.ListenAsync(...))`, which correctly tracks the full async
   lifecycle. Initialization signalling is also improved: `AutoResetEvent` is replaced with
   `TaskCompletionSource<bool>` and the signal is set before the accept loop begins, so `Start()`
   returns only when the server is unambiguously ready. Shutdown exception handling is consolidated
   into a shared `IsListenerShutdownException` helper. `TestWebSocketServer.RunServer` also gains a
   new overload that exposes the `HttpListenerContext` alongside the `WebSocket`, enabling tests to
   inspect HTTP upgrade request headers (e.g. for verifying custom headers set via
   `ClientWebSocketFactory`); the existing single-argument overload delegates to the new one and is
   unaffected.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
+++ b/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
@@ -14,6 +14,13 @@
 * Apply response size limits for oversized OpAMP responses.
   ([#4116](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4116))
 
+* Harden WebSocket transport:
+  * Enforce maximum server message size.
+  * Use a cancellable receive loop with deterministic disposal.
+  * Add `OpAmpClientSettings.ClientWebSocketFactory`, and remove the insecure
+  default of accepting any server TLS certificate when connecting on .NET.
+  ([#4133](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4133))
+
 * Ensure heartbeat interval is bounded.
   ([#4136](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4136))
 
--- a/src/OpenTelemetry.OpAmp.Client/Internal/OpAmpClientEventSource.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/OpAmpClientEventSource.cs
@@ -14,8 +14,9 @@ internal sealed class OpAmpClientEventSource : EventSource
     // General events 1-499
     private const int EventIdInvalidWsFrame = 1;
     private const int EventIdTransportCloseFailure = 2;
-    private const int EventIdOversizedResponseContentLength = 3;
-    private const int EventIdHttpResponseReceived = 4;
+    private const int EventIdHttpResponseReceived = 3;
+    private const int EventIdOversizedWebSocketMessage = 4;
+    private const int EventIdFrameProcessingFailure = 5;
 
     // Service events 500-999
     private const int EventIdHeartbeatServiceStart = 500;
@@ -61,10 +62,19 @@ internal sealed class OpAmpClientEventSource : EventSource
         this.WriteEvent(EventIdTransportCloseFailure, exception);
     }
 
-    [Event(EventIdOversizedResponseContentLength, Message = "OpAMP server response discarded: Content-Length ({0} bytes) exceeds the {1}-byte limit. The request was delivered but the server response was not processed.", Level = EventLevel.Warning)]
-    public void OversizedResponseContentLength(long contentLengthBytes, int limitBytes)
+    [NonEvent]
+    public void OversizedWebSocketMessageReceived(int minimumBytes, int limitBytes)
     {
-        this.WriteEvent(EventIdOversizedResponseContentLength, contentLengthBytes, limitBytes);
+        if (this.IsEnabled(EventLevel.Warning, EventKeywords.All))
+        {
+            this.OversizedWebSocketMessage(minimumBytes, limitBytes);
+        }
+    }
+
+    [Event(EventIdOversizedWebSocketMessage, Message = "OpAMP server WebSocket message discarded: message is at least {0} bytes, exceeding the {1}-byte limit. The connection will be closed and the frame will not be processed.", Level = EventLevel.Warning)]
+    public void OversizedWebSocketMessage(int minimumBytes, int limitBytes)
+    {
+        this.WriteEvent(EventIdOversizedWebSocketMessage, minimumBytes, limitBytes);
     }
 
     [NonEvent]
@@ -80,6 +90,21 @@ internal sealed class OpAmpClientEventSource : EventSource
     public void HttpResponseReceived(int bytes)
     {
         this.WriteEvent(EventIdHttpResponseReceived, bytes);
+    }
+
+    [NonEvent]
+    public void FrameProcessingException(Exception ex)
+    {
+        if (this.IsEnabled(EventLevel.Warning, EventKeywords.All))
+        {
+            this.FrameProcessingFailure(ex.ToInvariantString());
+        }
+    }
+
+    [Event(EventIdFrameProcessingFailure, Message = "Failed to process incoming server frame. The frame was dropped: {0}", Level = EventLevel.Warning)]
+    public void FrameProcessingFailure(string exception)
+    {
+        this.WriteEvent(EventIdFrameProcessingFailure, exception);
     }
 
     [Event(EventIdHeartbeatServiceStart, Message = "Heartbeat service started.", Level = EventLevel.Informational)]
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
@@ -8,18 +8,23 @@ using OpenTelemetry.OpAmp.Client.Internal.Utils;
 
 namespace OpenTelemetry.OpAmp.Client.Internal.Transport.WebSocket;
 
+/// <summary>
+/// Reads OpAMP WebSocket messages from the server and dispatches them to the frame processor.
+/// </summary>
 internal sealed class WsReceiver : IDisposable
 {
     private const int RentalBufferSize = 4 * 1024; // 4 KB
     private const int ReceiveBufferSize = 8 * 1024; // 8 KB
 
     private readonly ClientWebSocket ws;
-    private readonly Thread receiveThread;
     private readonly FrameProcessor processor;
+    private readonly CancellationTokenSource disposeTokenSource = new();
 
     private readonly byte[] receiveBuffer = new byte[ReceiveBufferSize];
 
-    private CancellationToken token;
+    private CancellationTokenSource? linkedReceiveTokenSource;
+    private Task? receiveTask;
+    private bool disposed;
 
     public WsReceiver(ClientWebSocket ws, FrameProcessor processor)
     {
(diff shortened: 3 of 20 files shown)


stevejgordon:
The Sampler test for AWS may be related to changes in the test server, but it runs fine locally.

src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Harden WebSocket transport: enforce maximum server message size, use a
+  cancellable receive loop with deterministic disposal, add
+  `OpAmpClientSettings.ClientWebSocketFactory`, and remove the insecure default
+  of accepting any server TLS certificate when connecting on .NET.
+  ([#4133](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4133))
```
martincostello: I'd break the distinct changes into different bullet points.

```
* Harden WebSocket transport
  * Change 1
  * Change 2...
```

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpClientEventSource.cs
```diff
@@ -17,6 +17,8 @@ internal sealed class OpAmpClientEventSource : EventSource
     private const int EventIdOversizedResponseContentLength = 3;
     private const int EventIdOversizedResponseBody = 4;
     private const int EventIdHttpResponseReceived = 5;
+    private const int EventIdOversizedWebSocketMessage = 6;
+    private const int EventIdFrameProcessingFailure = 7;
```
martincostello: Once #4130 is merged the event IDs will change.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
```diff
@@ -8,18 +8,29 @@
...
 namespace OpenTelemetry.OpAmp.Client.Internal.Transport.WebSocket;
 
+/// <summary>
+/// Reads OpAMP WebSocket messages from the server and dispatches them to the frame processor.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Oversized messages: see <see cref="TransportConstants.MaxMessageSize"/> for the size limit
+/// and the note on peak buffered bytes when enforcement happens after each receive operation.
+/// </para>
```
martincostello: Do we need this in the `///` comments?

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -1,87 +1,136 @@
...
+    /// Connects the WebSocket and starts receiving server frames.
+    /// </summary>
+    /// <param name="token">A cancellation token for the connect operation.</param>
+    /// <returns>A task that completes when the connection is established and the receive loop has started.</returns>
+    /// <remarks>
+    /// This method is not idempotent: a second call after a successful start throws
+    /// <see cref="InvalidOperationException"/>. If the connection or receive startup fails, internal state is reset
+    /// so that <see cref="Dispose"/> can still be called cleanly. A new <see cref="WsTransport"/> instance
+    /// (and therefore a new <see cref="ClientWebSocket"/>) is required to reconnect.
+    /// </remarks>
```
martincostello, line 70: Should add the `<exception>` tag to document this too.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -1,87 +1,136 @@
...
 
+    /// <summary>
+    /// Connects the WebSocket and starts receiving server frames.
+    /// </summary>
+    /// <param name="token">A cancellation token for the connect operation.</param>
+    /// <returns>A task that completes when the connection is established and the receive loop has started.</returns>
+    /// <remarks>
+    /// This method is not idempotent: a second call after a successful start throws
+    /// <see cref="InvalidOperationException"/>. If the connection or receive startup fails, internal state is reset
+    /// so that <see cref="Dispose"/> can still be called cleanly. A new <see cref="WsTransport"/> instance
```
martincostello: The middle sentence doesn't feel like something callers should need to care about.

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
```diff
@@ -144,33 +160,85 @@ public Task SendCustomCapabilitiesAsync(IEnumerable<string> capabilities, Cancel
...
+            if (service is IDisposable disposableService)
+            {
+                disposableService.Dispose();
+            }
+        }
+
+        if (this.transport is IDisposable disposableTransport)
+        {
+            disposableTransport.Dispose();
+        }
```
martincostello, line 214: Do we maybe need to consider `IAsyncDisposable` too? Only if any thing we're already using implements one of those.

stevejgordon: Nothing implements IAsyncDisposable (currently)

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
stevejgordon: Good shout, restructured this section.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
stevejgordon: Done + added XML docs to other methods for consistency

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsReceiver.cs
stevejgordon: No, removed

src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
stevejgordon: Done

martincostello:
I _think_ #4121 resolved the Hangfire flakes.

stevejgordon:
> I _think_ #4121 resolved the Hangfire flakes.

Once your PR is merged, I can rebase on top of everything and hopefully we'll see this one go green again!

martincostello (approved):

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -1,87 +1,171 @@
...
+            }
+            catch
+            {
+                // Best effort: Abort may throw if the socket is already in a terminal state.
+            }
 
-        this.receiver.Start(token);
+            Interlocked.Exchange(ref this.startState, 0);
+            throw;
+        }
```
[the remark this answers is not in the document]
stevejgordon: This wasn't the best fix. Added a richer handling of states.

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -1,87 +1,171 @@
...
+        if (this.disposed)
+        {
+            return;
+        }
+
+        this.disposed = true;
+
+        this.webSocket.Abort();
         this.receiver.Dispose();
+        this.webSocket.Dispose();
```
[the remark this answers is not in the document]
stevejgordon: Fixed

src/OpenTelemetry.OpAmp.Client/Internal/Transport/WebSocket/WsTransport.cs
```diff
@@ -1,87 +1,171 @@
...
+    /// </summary>
+    /// <param name="token">A cancellation token for the connect operation.</param>
+    /// <returns>A task that completes when the connection is established and the receive loop has started.</returns>
+    /// <remarks>
+    /// This method is not idempotent: a second call after a successful start throws
+    /// <see cref="InvalidOperationException"/>. If connection establishment or receive-loop startup fails,
+    /// the transport aborts the socket and the instance cannot be reused. A new <see cref="WsTransport"/>
+    /// instance (and therefore a new <see cref="ClientWebSocket"/>) is required to reconnect.
+    /// </remarks>
+    /// <exception cref=" ObjectDisposedException">
```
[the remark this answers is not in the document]
stevejgordon: Fixed

stevejgordon:
@martincostello Hopefully ready now!

PR #4136, opened 2026-04-16
[OpAMP] Ensure heartbeat interval is bounded

stevejgordon:
## Changes

Ensures that a interval larger than `TimeSpan.MaxValue.TotalSeconds` returned by the server is not used when configuring the heartbeat service.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
@@ -11,6 +11,9 @@ internal sealed class HeartbeatService : IBackgroundService, IOpAmpListener<Conn
 {
     public const string Name = "heartbeat-service";
 
+    // Safe upper bound.
+    private static readonly ulong MaxHeartbeatIntervalSeconds = (ulong)TimeSpan.MaxValue.TotalSeconds;
+
     private readonly FrameDispatcher dispatcher;
     private readonly FrameProcessor processor;
     private readonly CancellationTokenSource cts;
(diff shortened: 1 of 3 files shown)


Kielek (approved):

src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
```diff
@@ -11,6 +11,9 @@ internal sealed class HeartbeatService : IBackgroundService, IOpAmpListener<Conn
 {
     public const string Name = "heartbeat-service";
 
+    // Safe upper bound.
+    private static readonly ulong MaxHeartbeatIntervalSeconds = (ulong)TimeSpan.MaxValue.TotalSeconds;
```
martincostello, line 15: Is this upper bound just treated as a sentinel ("wait as long as it takes") to avoid the overflow, or should it be a more reasonable upper bound instead of ~1,200 years?

stevejgordon: This just addresses the overflow. We could choose a sensible upper limit below this, but realisticaly, if the server sends a value, we probably should just honour it, even it's crazy high.

martincostello (approved):

test/OpenTelemetry.OpAmp.Client.Tests/HeartbeatServiceTests.cs
```diff
@@ -11,6 +13,26 @@ namespace OpenTelemetry.OpAmp.Client.Tests;
...
+        using var service = new HeartbeatService(dispatcher, new FrameProcessor());
+        service.Configure(settings);
+
+        var message = new ConnectionSettingsMessage(new ConnectionSettingsOffers
+        {
+            Opamp = new OpAMPConnectionSettings { HeartbeatIntervalSeconds = intervalSeconds },
+        });
+
+        // Must not throw OverflowException.
+        service.HandleMessage(message);
```
martincostello, line 33: Could consuder using this pattern for "don't throw":

```csharp
var exception = Record.Exception(() => service.HandleMessage(message));

Assert.Null(exception);
```

PR #4138, opened 2026-04-16
[OpAMP] Add missing null check for remote config

stevejgordon:
## Changes

Ensures that in cases where the remote config message has no config (which the spec permits) that we handle that scenario.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
+++ b/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Add missing null check for remote config.
+  ([#4138](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4138))
+
 * Ensure heartbeat interval is bounded.
   ([#4136](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4136))
 
--- a/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
+++ b/src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 ## Unreleased
 
-* Add missing null check for remote config.
-  ([#4138](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4138))
-
-* Ensure heartbeat interval is bounded.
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigMessage.cs
```diff
@@ -16,16 +16,25 @@ public class RemoteConfigMessage : OpAmpMessage
 
     internal RemoteConfigMessage(AgentRemoteConfig agentRemoteConfig)
     {
-        this.agentConfigMap = new Dictionary<string, AgentConfigFile>(agentRemoteConfig.Config.ConfigMap.Count, StringComparer.Ordinal);
-
-        foreach (var config in agentRemoteConfig.Config.ConfigMap)
+        // Config may be absent per the OpAmp proto spec — it SHOULD NOT be set when
```
martincostello:
```suggestion
        // Config may be absent per the OpAmp proto spec - it SHOULD NOT be set when
```

test/OpenTelemetry.OpAmp.Client.Tests/Messages/RemoteConfigMessageTests.cs
```diff
@@ -32,6 +32,23 @@ public void Constructor_WithValidConfig_InitializesConfigMap()
         Assert.True(remoteConfigMessage.AgentConfigMap.ContainsKey(Config2Name));
     }
 
+    [Fact]
+    public void Constructor_WithNullConfig_InitializesEmptyDictionary()
+    {
+        // Arrange — the OpAmp proto says Config SHOULD NOT be set when unchanged,
+        // so a valid server response may omit it entirely.
```
martincostello:
```suggestion
        // Arrange
```

martincostello (approved):

src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Add missing null check for remote config.
```
martincostello: I think the various PRs have caused the list to drift from being oldest to newest sorted.

stevejgordon: Yeah, let me take a look

PR #4146, opened 2026-04-17
[Instrumentation.Owin] OwinInstrumentation does not require IConfiguration registration

ysolomchenko:
Fixes # N/A
Design discussion issue # N/A

Found by Codex analysis

## Changes

This change makes startup safer in scenarios where `IConfiguration` is not registered in DI for `OWIN`
The shared file is compiled into five packages:

- OpenTelemetry.Instrumentation.Owin
- OpenTelemetry.Instrumentation.AspNet
- OpenTelemetry.Instrumentation.AspNetCore
- OpenTelemetry.Instrumentation.Http
- OpenTelemetry.Instrumentation.GrpcNetClient

GrpcNetClient does not use this helper, so it is only “binary touched” with no actual runtime behavior changes.

Separate tests for ASP.NET, HttpClient, and ASP.NET Core are not required, as this scenario is already covered by existing tests.


src/Shared/Options/DelegatingOptionsFactoryServiceCollectionExtensions.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.DependencyInjection;
 
 internal static class DelegatingOptionsFactoryServiceCollectionExtensions
 {
+    private static readonly IConfiguration EmptyConfiguration = new ConfigurationBuilder().Build();
```
martincostello: I would create this inline in the two places it's needed, rather than keep this around for all usage.

ysolomchenko: Fixed in [0c8956e](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4146/commits/0c8956eef980281a5d11a89ef62da8338e623e75)

martincostello (approved):

PR #4285, opened 2026-04-22
[OpAMP] Enhance EffectiveConfigFile: stream APIs, size limits, tests

### LINKED ISSUE #4151: [OpAMP] Change effective config `CreateFromFilePath` signature
We've discussed changing `CreateFromFilePath` to accept a `Stream` and max size. This makes the consumer responsible for reading from the file and validating the input (if required) before creating the request.

stevejgordon:
Fixes #4151

## Changes

- Add CreateFromStream and CreateFromStreamAsync to EffectiveConfigFile for safe, size-limited config loading from streams (sync/async) with enforced max byte length.
- Remove CreateFromFilePath.
- Content is now ReadOnlyMemory<byte>.
- Enforce non-null contentType and fileName (default to empty string).
- Improve OpAmpClient API docs with thrown exception details.
- Add comprehensive EffectiveConfigFile tests (validation, limits, event logging).
- Update FrameBuilderTests and OpAmpClientTests for new APIs.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
@@ -11,9 +11,9 @@ OpenTelemetry.OpAmp.Client.Messages.CustomMessageMessage.Capability.get -> strin
 OpenTelemetry.OpAmp.Client.Messages.CustomMessageMessage.Data.get -> System.ReadOnlySpan<byte>
 OpenTelemetry.OpAmp.Client.Messages.CustomMessageMessage.Type.get -> string!
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile
-OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.Content.get -> System.Memory<byte>
+OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.Content.get -> System.ReadOnlyMemory<byte>
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.ContentType.get -> string!
-OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.EffectiveConfigFile(System.Memory<byte> content, string! contentType, string! fileName) -> void
+OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.EffectiveConfigFile(System.ReadOnlyMemory<byte> content, string! contentType = "", string! fileName = "") -> void
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.FileName.get -> string!
 OpenTelemetry.OpAmp.Client.Messages.OpAmpMessage
 OpenTelemetry.OpAmp.Client.Messages.OpAmpMessage.OpAmpMessage() -> void
@@ -88,6 +88,7 @@ OpenTelemetry.OpAmp.Client.Settings.RemoteConfigSettings.AcceptsRemoteConfig.set
 OpenTelemetry.OpAmp.Client.Settings.RemoteConfigSettings.RemoteConfigSettings() -> void
 override OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.Equals(object? obj) -> bool
 override OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.GetHashCode() -> int
-static OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.CreateFromFilePath(string! filePath, string! contentType, string? filename = null) -> OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile!
+static OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.CreateFromStream(System.IO.Stream! stream, string! contentType = "", string! fileName = "", int maxBytes = 524288) -> OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile!
+static OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.CreateFromStreamAsync(System.IO.Stream! stream, string! contentType = "", string! fileName = "", int maxBytes = 524288, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task<OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile!>!
 static OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.operator !=(OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion left, OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion right) -> bool
 static OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.operator ==(OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion left, OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion right) -> bool
--- a/src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
@@ -13,7 +13,7 @@ OpenTelemetry.OpAmp.Client.Messages.CustomMessageMessage.Type.get -> string!
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.Content.get -> System.ReadOnlyMemory<byte>
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.ContentType.get -> string!
-OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.EffectiveConfigFile(System.ReadOnlyMemory<byte> content, string! contentType = "", string! fileName = "") -> void
+OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.EffectiveConfigFile(System.ReadOnlyMemory<byte> content, string! contentType, string! fileName) -> void
 OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.FileName.get -> string!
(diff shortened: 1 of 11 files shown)


src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -88,6 +88,7 @@ OpenTelemetry.OpAmp.Client.Settings.RemoteConfigSettings.AcceptsRemoteConfig.set
 OpenTelemetry.OpAmp.Client.Settings.RemoteConfigSettings.RemoteConfigSettings() -> void
 override OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.Equals(object? obj) -> bool
 override OpenTelemetry.OpAmp.Client.Settings.AnyValueUnion.GetHashCode() -> int
-static OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.CreateFromFilePath(string! filePath, string! contentType, string? filename = null) -> OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile!
+static OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile.CreateFromStream(System.IO.Stream! stream, string! contentType = "", string! fileName = "", int maxBytes = 524288) -> OpenTelemetry.OpAmp.Client.Messages.EffectiveConfigFile!
```
Kielek: Just skimming the API.

The default value attributes are bad practice from the public API perspective and long term of maintenance.

stevejgordon: Fair enough, I can tweak that.

stevejgordon: It was mostly to provide a sensible max bytes default

martincostello: An options class to hold the possible values also has the flexibility to make additions without the complexity of binary breaking changes with default/optional parameters.

stevejgordon: Content type and file name don't need defaults, I did that to mirror the original method null defaults which are not really valid here. On reflection, requiring those explicitly makes sense. For max size, as nothing else is likely to be needed here, I think an options type is overkill. I think either just requiring that explicitly, or having an overload without it (which then applies the default by calling the limited overload) is better. Open to opinions.

martincostello: Maybe instead of `524288`, so something like [`StreamReader` does](https://github.com/dotnet/runtime/blob/e24fb47ed3714cd7e302c5c6f9b57b18ff825e52/src/libraries/System.Private.CoreLib/src/System/IO/StreamReader.cs#L138) and have `-1` as a sentinel for the default size?

src/OpenTelemetry.OpAmp.Client/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Enhance EffectiveConfigFile:
+  * Remove `CreateFromFilePath` factory method.
+  * Add `CreateFromSteam` and `CreateFromStreamAsync` methods which enforce max
+  size limits.
```
martincostello: Nit: easier to read IMHO.

```suggestion
    size limits.
```

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -188,6 +188,7 @@ IFrameBuilder IFrameBuilder.AddEffectiveConfig(IEnumerable<EffectiveConfigFile>
         var fileMap = new Dictionary<string, global::OpAmp.Proto.V1.AgentConfigFile>(StringComparer.Ordinal);
         foreach (var file in files)
         {
+            // Dictionary.Add throws if multiple files share the same FileName (including the empty-string default).
```
martincostello: Should we do something about it, rather than just write a comment for how Add works?

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
```diff
@@ -1,21 +1,46 @@
...
+        if (contentType == null)
+        {
+            throw new ArgumentNullException(nameof(contentType));
+        }
+
+        if (fileName == null)
+        {
+            throw new ArgumentNullException(nameof(fileName));
+        }
+#endif
```
martincostello: If we haven't done it already, maybe we should polyfill this into our existing `Guard` class so we avoid the need to do all the conditional exception handling.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
```diff
@@ -24,44 +49,250 @@ public EffectiveConfigFile(Memory<byte> content, string contentType, string file
...
+            // For seekable streams, CheckSeekableSize already confirmed remaining length is zero.
+            if (!stream.CanSeek && stream.ReadByte() != -1)
+            {
+                ThrowSizeLimitExceeded(maxBytes);
+            }
+
+            return new EffectiveConfigFile(ReadOnlyMemory<byte>.Empty, contentType, fileName);
+        }
+
+        var buffer = ArrayPool<byte>.Shared.Rent(maxBytes);
```
martincostello, line 91: This stuff looks like the code we added recently to encode limits in a shared place - can we add this code there and consume it, rather than make more copies of the pattern we need to verify?

src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
stevejgordon: Yeah, that also isn't my favourite API design for discoverability. I'm now leaning toward just requiring everything as it has a clear purpose and the ctor can be used with a pre-limited block of memory for alternative scenarios.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
stevejgordon: Yes, I meant to flag this for discussion. For now, I've added tests to validate this behaviour as-is and documented the exception on the public API on the OpAmpClient. I'm a bit torn on what we should do here. Deduping is an option but is more silent. At least letting this throw up to the calling code makes the consumer clear on the issue and it's ultimately their problem to fix with unique names.

martincostello: I vaguely remember bringing up something here in a previous PR, because the dictionary is case-sensitive.

If the names come from the client, then I would err towards throwing with a more tailored message.

If the names are related to the server-side config, then maybe just log a warning before the value is overwritten via the indexer?

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
stevejgordon: It's quite similar to the stuff added to HttpClientHelpers, but that code isn't quite the right fit as-is and would need generalising futher (exception messages etc) for this use case. Could investigate that as a low-priority follow-up?

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
stevejgordon: I think just using the Guard classes is enough here. I'll update.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
stevejgordon: Ah, yes [you did](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/3716#discussion_r2708413767) when this went in. This is for an agent to server message (i.e. client to server) and the consumer of the OpAMP client is responsible for providing one or more `EffectiveConfigFile` to report. So we could catch the exception and raise our own with a slightly clearer message.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/EffectiveConfigFile.cs
stevejgordon: Resolved in the latest commit

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
stevejgordon: Resolved in the latest commit

src/OpenTelemetry.OpAmp.Client/.publicApi/PublicAPI.Unshipped.txt
stevejgordon: For now, I've made everything required on both this factory methods and the ctor. In general, I think this is fine for this API, but I'm happy to tweak it further.

test/OpenTelemetry.OpAmp.Client.Tests/Messages/EffectiveConfigFileTests.cs
```diff
@@ -0,0 +1,612 @@
...
+            => throw new NotSupportedException();
+
+        public override void SetLength(long value)
+            => throw new NotSupportedException();
+
+        public override void Write(byte[] buffer, int offset, int count)
+            => throw new NotSupportedException();
+    }
+
+    private sealed class CapturingEventListener : EventListener
```
martincostello: You should be able to use the shared `InMemoryEventListener` instead.

I did a similar de-dupe a few days ago in the main repo: https://github.com/open-telemetry/opentelemetry-dotnet/pull/7119

stevejgordon: Done

RassK (approved):
Seems good to me

martincostello (approved):

PR #4370, opened 2026-05-07
[AspNetCore] Update gRPC Semantic Conventions

martincostello:
## Changes

Update the Semantic Conventions version for gRPC Server spans to v1.41.0.

One thing I'm unsure of is whether or not we need to implement `OTEL_SEMCONV_STABILITY_OPT_IN` or not given that the support is already experimental.

Once that question is answered I'll add/update the tests as appropriate.

* [x] ~~Changes in public API reviewed (if applicable)~~


Kielek:
AutoInstruemntation perspective - never advertised this feature, so no needed.
It is in `rc` mode, there should be 1.42.0 release also. If there is no changes, I would vote for hard switch.

Codex feedback:
Findings
[P1] [HttpInListener.cs:448](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07eb7a8207cee8cfa61f7df6e033c40fb4562c34/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L448) emits rpc.method as only details.RpcMethod. In semantic conventions v1.41.0, gRPC rpc.method is the fully-qualified logical method, e.g. ExampleService/Echo. With OTEL_SEMCONV_STABILITY_OPT_IN=rpc, this drops rpc.service but still emits only Echo, so it is not compliant.

[P1] [HttpInListener.cs:466](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07eb7a8207cee8cfa61f7df6e033c40fb4562c34/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L466) writes rpc.response.status_code as an int. v1.41.0 requires a string gRPC status value such as OK, DEADLINE_EXCEEDED, etc.

[P1] [HttpInListener.cs:436](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07eb7a8207cee8cfa61f7df6e033c40fb4562c34/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L436) sets span status for failed gRPC responses, but the PR does not set error.type to the gRPC status when the status is an error. v1.41.0 makes error.type conditionally required when the operation failed.

[P2] [HttpInListener.cs:285](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07eb7a8207cee8cfa61f7df6e033c40fb4562c34/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L285) only reads grpc.status_code, while [HttpInListener.cs:452](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/07eb7a8207cee8cfa61f7df6e033c40fb4562c34/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L452) removes grpc.status. The v1.41.0 compatibility mapping maps grpc.status to rpc.response.status_code, so dropping it without mapping can lose the required status.

martincostello:
> there should be 1.42.0 release also

I'll do a follow-up for 1.42.0 as there will be some other places for the other gRPC instrumentations that already moved to 1.41 that were merged previously.

martincostello:
Comments should be addressed. There's also some overlap in #4508.

Kielek:
1.42.0 is out, do you want to update in scope of this PR?

martincostello:
> 1.42.0 is out, do you want to update in scope of this PR?

No I'll do that separately with #4508.

src/Shared/RpcSemanticConventionHelper.cs
Kielek, line 1: I think that you can remove this class from this PR.

opentelemetry-dotnet-contrib.slnx
```diff
@@ -205,6 +205,7 @@
     <File Path="src/Shared/RedactionHelper.cs" />
     <File Path="src/Shared/RequestDataHelper.cs" />
     <File Path="src/Shared/ResourceSemanticConventions.cs" />
+    <File Path="src/Shared/RpcSemanticConventionHelper.cs" />
```
Kielek: And revert this file.

Kielek (commented):
Two new comments from Codex and one from my side:
[P2] [HttpInListener.cs:285](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/054ec2eb6a8850ad2ab5dab210fb90fab36f6ae2/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L285) still only reads grpc.status_code, while [HttpInListener.cs:436](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/054ec2eb6a8850ad2ab5dab210fb90fab36f6ae2/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L436) removes grpc.status. The v1.41.0 compatibility mapping says grpc.status maps to rpc.response.status_code; if that tag is present without grpc.status_code, this drops the required RPC status.

[P2] [HttpInListener.cs:428](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/054ec2eb6a8850ad2ab5dab210fb90fab36f6ae2/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L428) only emits rpc.method when the gRPC method parses as service/method. For an available but unrecognized method, v1.41.0 says rpc.method should be _OTHER and rpc.method_original should carry the original value. The current test at [GrpcTests.cs:55](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/054ec2eb6a8850ad2ab5dab210fb90fab36f6ae2/test/OpenTelemetry.Instrumentation.AspNetCore.Tests/GrpcTests.cs#L55) asserts no rpc.method, so this edge remains non-compliant.

martincostello:
Feedback addressed.

Kielek (approved):
LGTM with minor direct feedback

PR #3867, opened 2026-02-14
[Infra] Support .NET 11

martincostello:
_The intention is to keep this PR updated over the course of .NET 11's development as new previews are released._

Fixes #1730.

## Changes

Add support for building and testing with .NET 11.

- Build with .NET 11 SDK
- Add `net11.0` target
- Drop `net8.0` and `net9.0` targets

## TODO

- [ ] Use stable (or RC) SDK, container and NuGet package versions
- [ ] Re-enable disabled tests
- [ ] Remove any hacks or TODOs

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] Changes in public API reviewed (if applicable)

PR #4397, opened 2026-05-15
[SqlClient] Fix trace flags to handle random IDs

martincostello:
## Changes

Update flag formatting when `EnableTraceContextPropagation` is enabled to cater for `ActivityTraceFlags.RandomTraceId` (https://github.com/dotnet/runtime/issues/124509).

Found in #3867 with .NET 11 preview.4.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
@@ -129,7 +129,7 @@ internal sealed class SqlClientDiagnosticListener : ListenerHandler
                         var parameter = setContextCommand.CreateParameter();
                         parameter.ParameterName = ContextInfoParameterName;
 
-                        var tracedflags = ((int)activity.ActivityTraceFlags).ToString("00", CultureInfo.InvariantCulture);
+                        var tracedflags = ((byte)activity.ActivityTraceFlags).ToString("x2", CultureInfo.InvariantCulture);
                         var traceparent = $"00-{activity.TraceId.ToHexString()}-{activity.SpanId.ToHexString()}-{tracedflags}";
 
                         parameter.DbType = DbType.Binary;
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -129,7 +129,7 @@ public override void OnEventWritten(string name, object? payload)
                         var parameter = setContextCommand.CreateParameter();
                         parameter.ParameterName = ContextInfoParameterName;
 
-                        var tracedflags = (activity.ActivityTraceFlags & ActivityTraceFlags.Recorded) != 0 ? "01" : "00";
+                        var tracedflags = ((int)activity.ActivityTraceFlags).ToString("00", CultureInfo.InvariantCulture);
```
martincostello:
```suggestion
                        var tracedflags = ((byte)activity.ActivityTraceFlags).ToString("x2", CultureInfo.InvariantCulture);
```

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
```diff
@@ -129,7 +129,7 @@ public override void OnEventWritten(string name, object? payload)
                         var parameter = setContextCommand.CreateParameter();
                         parameter.ParameterName = ContextInfoParameterName;
 
-                        var tracedflags = (activity.ActivityTraceFlags & ActivityTraceFlags.Recorded) != 0 ? "01" : "00";
+                        var tracedflags = ((byte)activity.ActivityTraceFlags).ToString("x2", CultureInfo.InvariantCulture);
```
Kielek, line 132: I would consider, if some switch case version does will be not more performant (or remove some memory allocations for most cases).

pseudoce:
```
[ForceInline]
publi static string SerializeActivityTags(ActivityTraceFlags flags)
{
swithc flags:
case 0: return "00";
case 1: return "01";
case 2: return "02";
case 3: return "03";
default return flags.ToString("x2", CultureInfo.InvariantCulture);
}
````

It can be follow up for the .NET11 branch, as we will have access to the real flag valeus tehre.

Kielek (approved):

src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlClientDiagnosticListener.cs
martincostello: #4399

PR #4403, opened 2026-05-19
[Extensions.AWS] Enable X-Ray ALB trace context parsing with only Root value

### LINKED ISSUE #867: AWSXRayPropagator does not correctly parse traces without Parent segment
# Issue with OpenTelemetry.Contrib.Extensions.AWSXRay

* OpenTelemetry.Contrib.Extensions.AWSXRay 1.2.0

Runtime version (e.g. `net462`, `net48`, `netcoreapp3.1`, `net6.0` etc. You can find this information from the `*.csproj` file):

* net6.0

* [ ] Feature Request
* [x] Bug

Based on [XRay Docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader) it is possible to have traces with both `Parent` segment and without one. 
Based on [AWS ALB with XRay Docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-services-elb.html) the only segment ALB sends is `Root`.

I expect `AWSXRayPropagator` to be able to correctly parse Xray header if it's passed, with or without `Parent`, with or without `Sampled`.

`AWSXRayPropagator` parses XRay Trace only in full format (e.g. with Root and Parent). If it's missing parent or sampled segment, new context is not created. 

Repro is very simple:
Just pass header like this: 
X-Amzn-Trace-Id: Root=1-63b55b03-7474ebdc42dd561f221078b2

LukaszBancarz:
Fixes #867

## Changes

Fix to the case when AWSXRayPropagator parses XRay Trace only in full format (e.g. with Root and Parent). If it's missing parent or sampled segment, new context is not created. This is a common case in AWS that X-Amzn-Trace-Id value consist of only Root (TraceId) value. In this case a new SpanId is generated and ActivityTraceFlags to None.

* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes - N/A
* [ ] Changes in public API reviewed (if applicable) - N/A

### DIFF
--- a/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
@@ -5,6 +5,9 @@
 * Fix sampling behaviour to be compatible with .NET 11.
   ([#4396](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4396))
 
+* Changed AWSXRayPropagator behavior to parse AWS ALB headers with only Root value in it.
+  ([#4403](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4403))
+
 ## 1.15.1
 
 Released 2026-Apr-21
--- a/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
+++ b/src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Fix sampling behaviour to be compatible with .NET 11.
   ([#4396](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4396))
 
-* Changed AWSXRayPropagator behavior to parse AWS ALB headers with only Root value in it.
+* `AWSXRayPropagator` now parses AWS ALB headers with only a Root value.
   ([#4403](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4403))
 
 ## 1.15.1
--- a/src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
+++ b/src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
@@ -222,7 +222,7 @@ public class AWSXRayPropagator : TextMapPropagator
         var activityTraceId = ActivityTraceId.CreateFromString(traceId);
 
         // Generate a new span ID if Parent is missing (ALB generating only Root param scenario)
-        var activityParentId = parentId.IsEmpty ? ActivitySpanId.CreateRandom() : ActivitySpanId.CreateFromString(parentId);
+        var activityParentId = parentId.IsEmpty ? default : ActivitySpanId.CreateFromString(parentId);
 
         // Default to None if not specified
         var activityTraceOptions = traceOptions == SampledValue ? ActivityTraceFlags.Recorded : ActivityTraceFlags.None;
--- a/src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
+++ b/src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
@@ -214,13 +214,17 @@ public class AWSXRayPropagator : TextMapPropagator
             }
         }
 
-        if (traceId.IsEmpty || parentId.IsEmpty || traceOptions == default)
+        if (traceId.IsEmpty)
         {
             return false;
         }
 
         var activityTraceId = ActivityTraceId.CreateFromString(traceId);
-        var activityParentId = ActivitySpanId.CreateFromString(parentId);
+
+        // Generate a new span ID if Parent is missing (ALB generating only Root param scenario)
+        var activityParentId = parentId.IsEmpty ? ActivitySpanId.CreateRandom() : ActivitySpanId.CreateFromString(parentId);
+
+        // Default to None if not specified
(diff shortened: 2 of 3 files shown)


martincostello (commented):
Please add a CHANGELOG entry for this fix.

src/OpenTelemetry.Extensions.AWS/Trace/AWSXRayPropagator.cs
```diff
@@ -214,13 +214,17 @@ internal static bool TryParseXRayTraceHeader(string rawHeader, out ActivityConte
...
+        if (traceId.IsEmpty)
         {
             return false;
         }
 
         var activityTraceId = ActivityTraceId.CreateFromString(traceId);
-        var activityParentId = ActivitySpanId.CreateFromString(parentId);
+
+        // Generate a new span ID if Parent is missing (ALB generating only Root param scenario)
+        var activityParentId = parentId.IsEmpty ? ActivitySpanId.CreateRandom() : ActivitySpanId.CreateFromString(parentId);
```
Kielek: I would consider this as a wrong pattern.
Is any other language follow this practice?

If I am understand this correctly, if you are not able to find parent span id, you will generate some artificial version.
It will lead to the case where backends will be not able to find any real parent. IMO it can be even worse.
Maybe in such cases it will be better to just treat it as empty parent and prepare rooted activity?

LukaszBancarz: I agree 100%, in this case this should be treated as a root activity meaning no parent span Id. Thank you

src/OpenTelemetry.Extensions.AWS/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Fix sampling behaviour to be compatible with .NET 11.
   ([#4396](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4396))
 
+* Changed AWSXRayPropagator behavior to parse AWS ALB headers with only Root value in it.
```
martincostello:
```suggestion
* `AWSXRayPropagator` now parses AWS ALB headers with only a Root value.
```

test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
```diff
@@ -177,8 +177,14 @@ public void TestExtractTraceHeaderWithoutParentId()
...
             { AWSXRayTraceHeaderKey, "Root=1-5759e988-bd862e3fe1be46a994272793;Sampled=1" },
         };
+        var traceId = ActivityTraceId.CreateFromString(TraceId.AsSpan());
 
-        Assert.Equal(default, this.awsXRayPropagator.Extract(default, carrier, Getter));
+        var result = this.awsXRayPropagator.Extract(default, carrier, Getter);
+
+        Assert.Equal(traceId, result.ActivityContext.TraceId);
+        Assert.Equal(default, result.ActivityContext.SpanId);
+        Assert.True((result.ActivityContext.TraceFlags & ActivityTraceFlags.Recorded) != 0, "because Sampled=1");
```
martincostello: It would be better to include the value of `result.ActivityContext.TraceFlags` in the assertion message.

test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
```diff
@@ -188,8 +194,29 @@ public void TestExtractTraceHeaderWithoutSampleDecision()
...
+        {
+            { AWSXRayTraceHeaderKey, "Root=1-5759e988-bd862e3fe1be46a994272793" },
+        };
+        var traceId = ActivityTraceId.CreateFromString(TraceId.AsSpan());
+
+        var result = this.awsXRayPropagator.Extract(default, carrier, Getter);
+
+        Assert.Equal(traceId, result.ActivityContext.TraceId);
+        Assert.Equal(default, result.ActivityContext.SpanId);
+        Assert.True((result.ActivityContext.TraceFlags & ActivityTraceFlags.None) == 0, "because Sampled is missing");
```
martincostello: Same here.

martincostello (approved):

test/OpenTelemetry.Extensions.AWS.Tests/Trace/AWSXRayPropagatorTests.cs
```diff
@@ -188,8 +194,29 @@ public void TestExtractTraceHeaderWithoutSampleDecision()
...
+        {
+            { AWSXRayTraceHeaderKey, "Root=1-5759e988-bd862e3fe1be46a994272793" },
+        };
+        var traceId = ActivityTraceId.CreateFromString(TraceId.AsSpan());
+
+        var result = this.awsXRayPropagator.Extract(default, carrier, Getter);
+
+        Assert.Equal(traceId, result.ActivityContext.TraceId);
+        Assert.Equal(default, result.ActivityContext.SpanId);
+        Assert.True((result.ActivityContext.TraceFlags & ActivityTraceFlags.None) == 0, $"Expected TraceFlags check to pass because Sampled is missing, but got: {result.ActivityContext.TraceFlags}");
```
Kielek, line 218: This line seems to be odd. `result.ActivityContext.TraceFlags & ActivityTraceFlags.None` is always 0, because `None` is `0`

PR #4419, opened 2026-05-20
[Infra] Add zizmor scan workflow

martincostello:
## Changes

Add a workflow that scans the repository with zizmor for vulnerabilities in GitHub Actions workflows.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -2,9 +2,7 @@ name: zizmor scan
 
 on:
   push:
-    branches: [ 'main' ]
-  schedule:
-    - cron: '0 0 * * *' # Once a day at 00:00 UTC
+    branches: [ 'main*', 'instrumentation*', 'exporter*', 'extensions*' ]
   workflow_dispatch:
 
 permissions: {}
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -24,14 +24,14 @@ jobs:
 
     steps:
 
-    - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      with:
-        filter: 'tree:0'
-        persist-credentials: false
-        show-progress: false
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          filter: 'tree:0'
+          persist-credentials: false
+          show-progress: false
 
-    - name: Scan workflows with zizmor
-      uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
-      with:
-        persona: pedantic
+      - name: Scan workflows with zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+        with:
+          persona: pedantic
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,37 @@
+name: zizmor scan
+
+on:
+  push:
+    branches: [ 'main' ]
+  schedule:
+    - cron: '0 0 * * *' # once in a day at 00:00
+  workflow_dispatch:
+
+permissions: {}
+
+env:
+  FORCE_COLOR: 3
+  TERM: xterm
+
+jobs:
+  lint:
+    runs-on: ubuntu-24.04
+
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    steps:
+
+    - name: Checkout code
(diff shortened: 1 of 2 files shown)


martincostello:
~~I just realised this won't run in the PR as it's from a fork.~~

~~I guess we'd have to merge it and then fix forward from there until everything's resolved, then make it a required status.~~

No, I forgot to add the PR trigger... 🤦

.github/zizmor.yml
```diff
@@ -0,0 +1,3 @@
+rules:
+  secrets-outside-env:
+    disable: true
```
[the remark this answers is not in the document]
martincostello: This is intentional for now as it requires significant changes otherwise.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
+name: zizmor scan
+
+on:
+  push:
+    branches: [ 'main' ]
+  schedule:
+    - cron: '0 0 * * *' # once in a day at 00:00
```
martincostello:
```suggestion
    - cron: '0 0 * * *' # Once a day at 00:00 UTC
```

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
+name: zizmor scan
+
+on:
```
pellared, line 3: How about adding

Why not

```
  pull_request:
    branches: [ 'main*', 'instrumentation*', 'exporter*', 'extensions*' ]
```

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches: [ 'main' ]
+  schedule:
+    - cron: '0 0 * * *' # Once a day at 00:00 UTC
```
pellared: Can you explain why?
Would it not always give the same result as for `main`?

pellared (approved):

.github/workflows/zizmor.yml
martincostello: Good point - I copied this from the CodeQL workflow, but compared to that zizmor is pinned and doesn't have any server-side component so I guess wouldn't ever change.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
+name: zizmor scan
+
+on:
+  push:
+    branches: [ 'main' ]
```
martincostello:
```suggestion
    branches: [ 'main*', 'instrumentation*', 'exporter*', 'extensions*' ]
```

.github/workflows/zizmor.yml
martincostello:
```suggestion
```

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,35 @@
+name: zizmor scan
+
+on:
```
martincostello, line 3:
```suggestion
on:
  pull_request:
```

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,36 @@
...
+  workflow_dispatch:
+
+permissions: {}
+
+env:
+  FORCE_COLOR: 3
+  TERM: xterm
+
+jobs:
+  lint:
```
martincostello:
```suggestion
  zizmor:
    name: zizmor
    timeout-minutes: 20
```

Kielek (approved):
As I understand, when merged and all issues solved, we can include it into PR as mandatory step?

martincostello:
> As I understand, when merged and all issues solved, we can include it into PR as mandatory step?

Yep, that's the idea.

I'll start working on addressing the findings once it's merged as follow-ups. I imagine there will be quite a few trivia-ish ones (like naming all steps) or explicitly configuring actions/checkout inputs.

🤞 there's nothing serious to address.

We can also debate suppressions vs. fixes were neccessary.

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,42 @@
...
+      actions: read # Scan GitHub Actions workflows
+      contents: read # Checkout the repository
+      security-events: write # Store results in the Security tab
+
+    steps:
+
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          filter: 'tree:0'
```
pellared, line 35: why?

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,42 @@
...
+          persist-credentials: false
+          show-progress: false
```
pellared, line 37: why?

pellared (approved):

.github/workflows/zizmor.yml
martincostello: More efficient fetch. I forget the exact details, but it's something we used at my old job and I added to my own personal repos, and it's got copy-pasted from there. You can find it in existing workflows in this repo as well.

.github/workflows/zizmor.yml
martincostello: Reduces noise in CI logs.

Issue #4445, opened 2026-05-25
[feature request]  Add support for unhandled Exceptions logging instrumentation


iblancasa:
### Is your feature request related to a problem?

.NET applications that use OpenTelemetry Logs do not currently have a standard OpenTelemetry-based way to report runtime-level unhandled exception events. Users who want this signal today must add application-specific handlers such as `AppDomain.CurrentDomain.UnhandledException` or host/framework-specific hooks, then manually map those exceptions into log records.

The goal of this issue is to agree whether this repository should provide anything for that use case, and if so, what shape it should take.

Important scope question: this may not be an instrumentation package. The first thing to settle is whether it should be a package, helper, extension API...

## Use Case

Applications often want a final best-effort telemetry signal when an exception reaches a runtime-level unhandled exception path. Error-monitoring products commonly collect this signal today. A vendor-neutral OpenTelemetry path would let applications send the same signal through the existing Logs pipeline instead of adding vendor-specific SDK code or custom logging glue.

This is similar to behavior provided by error-monitoring SDKs such as Sentry, which can install unhandled-exception hooks and report those exceptions through their own pipeline. The request here is for an OpenTelemetry-native path for the same general signal.

This is not intended to guarantee crash reporting. Delivery would always be best effort because process termination may happen before exporters flush.

## Other languages

This same functionality has already been accepted in other OpenTelemetry language contrib repositories:

* Node.js: `instrumentation-runtime-node` now emits logs for uncaught exceptions. The PR was merged in `opentelemetry-js-contrib`: https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3379
* Python: `opentelemetry-instrumentation-exceptions` has maintainer approval in `opentelemetry-python-contrib`: https://github.com/open-telemetry/opentelemetry-python-contrib/pull/4503

## Design proposal

One possible shape is a small Logs-focused package or extension that can register runtime-level exception handlers and emit OpenTelemetry log records using the exception log semantic conventions.

Initial scope, if accepted, could be limited to one hook such as `AppDomain.CurrentDomain.UnhandledException`. Other hooks, such as `TaskScheduler.UnobservedTaskException`, should be discussed separately because their behavior and reliability differ.

## Non-Goals

* Durable crash reporting.
* Guaranteed export before process termination.
* Replacing host/framework-specific exception handling.
* Suppressing exceptions or changing application behavior.
* Adding broad error monitoring features outside OpenTelemetry Logs.

[...]

PR #4446, opened 2026-05-25
Unhandled Exceptions logging instrumentation

iblancasa:
Design discussion issue #4445


Kielek:
I have some doubts about his contribution.
1. @iblancasa, do you have any needs to use this package or you just want to have something similar to other languages/technolgoies?
2. Especially the part related to unhandled exception, it might be tricky to export anything. There is a high chance the process will be terminated before we will be able to export anything/
3. Next topic to be discussed is the package name. We do not have yet any package related to the logs. I am not sure if it fits to the Instrumentation or better to some extensions.
4. If accepted it should be split to infrastructure stuff and the real implementation.

iblancasa:
> 1. @iblancasa, do you have any needs to use this package or you just want to have something similar to other languages/technolgoies?

Actually both. I want a standard OpenTelemetry way to capture last-chance exception telemetry that many error-monitoring systems already collect today. The goal is not only language parity, but giving .NET users a vendor-neutral path to send these failures through the OpenTelemetry. The reason for doing it via logs is because is how semantic conventions suggest to implement this https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-logs/

> 2. Especially the part related to unhandled exception, it might be tricky to export anything. There is a high chance the process will be terminated before we will be able to export anything/

I agree unhandled exception export is best-effort in this case. This package should not promise reliable crash reporting. The `README` already documents that the process may terminate before exporters flush. The intent is to emit through the normal OpenTelemetry  pipeline when the runtime gives us a hook, not to guarantee delivery. Users needing reliable crash persistence still need process supervisors, local durable exporters, crash dumps, or host-specific handling.

> 3. Next topic to be discussed is the package name. We do not have yet any package related to the logs. I am not sure if it fits to the Instrumentation or better to some extensions.

I'm not attached to the name. Totally open to suggestions.

> 4. If accepted it should be split to infrastructure stuff and the real implementation.

It's fine. Can you give me a bit more of info about how this should be done? Or some reference. I'm new contributing to DotNet SDK.

Thank you for your comments @Kielek!

Kielek:
> > 3. Next topic to be discussed is the package name. We do not have yet any package related to the logs. I am not sure if it fits to the Instrumentation or better to some extensions.
> 
> I'm not attached to the name. Totally open to suggestions.

I do not have yet any good name for this, but I see other option, I am not sure if it will be working for you, but we could put it into `Extension` package. @mikegoldsmith, @open-telemetry/dotnet-contrib-approvers,  WDYT? Do you have any better options?

> > 4. If accepted it should be split to infrastructure stuff and the real implementation.
> 
> It's fine. Can you give me a bit more of info about how this should be done? Or some reference. I'm new contributing to DotNet SDK.

You can check https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md#new-projects specially linked PR - https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4041 (there is a chance that you have correctly included some new requirements into this PR, so if it not jus implementation it should be in very first PR).

iblancasa:
Closing this PR to create the proper ones from scratch.

PR #4458, opened 2026-05-29
[Geneva] Batch OTLP-protobuf metric-point UDS sends

### LINKED ISSUE #4457: [Geneva] Batch OTLP-protobuf metric-point UDS sends to reduce per-export syscalls
## Component(s)

`OpenTelemetry.Exporter.Geneva` (metric exporter)

## Is your feature request related to a problem? Please describe.

The Geneva metric exporter performs one Unix-domain-socket `Socket.Send` syscall per individual metric point during each export interval, in [`OtlpProtobufSerializer.SendMetricPoint`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs).

At realistic cardinality (e.g. `CardinalityLimit=4000` across multiple histograms ⇒ up to ~20K points/scrape), this dominates CPU on the exporter thread. Profile evidence from a production high-throughput .NET service shows `Interop+Sys.Send` at **~13% exclusive CPU**, with the Geneva metric exporter as the dominant contributor.

The same pattern was identified in the Rust contrib repo for the Geneva log exporter: open-telemetry/opentelemetry-rust-contrib#306. The .NET Geneva *log* exporter already does buffered batching; only the metric exporter has the per-event syscall pattern.

## Describe the solution you'd like

Batch metric-point events into bounded UDS sends.

Wire-format guarantee: when `prefixBufferWithUInt32LittleEndianLength` is set (the only mode used by `MetricUnixDomainSocketDataTransport`), each serialized event is already framed with a uint32-LE length prefix. The Geneva MDSD receiver already demultiplexes multiple events from a single read by that prefix, so concatenating events into one `Send` requires no wire-format change and no receiver change.

Proposed API additions to `IMetricDataTransport` (internal):

```csharp
bool TryAppendOtlpProtobufEvent(byte[] buffer, int size);
void FlushOtlpProtobufEvents();
```

* `MetricUnixDomainSocketDataTransport` accumulates events into an internal buffer (sized to match the existing per-event allocation, `GenevaMetricExporter.BufferSize` = 65360 bytes) and flushes via the existing UDS path.
* `MetricUnixUserEventsDataTransport` and `MetricWindowsEventTracingDataTransport` preserve per-event semantics — each ETW / user_events write is a logical event for the consumer, so they forward `TryAppend` directly to `SendOtlpProtobufEvent` and treat `Flush` as a no-op.
* `OtlpProtobufSerializer.SendMetricPoint` becomes `TryAppend → Flush+retry → fallback to standalone Send` (for any single event that exceeds the accumulation buffer; preserves current behavior).
* `SerializeAndSendMetrics` calls `FlushOtlpProtobufEvents()` once at the end.
* `TlvMetricExporter` uses a different framing (`BinaryHeader` + `MetricEventType.TLV` per event) and is intentionally out of scope.

This is intended to be transparent — no opt-in feature flag — since the wire format is unchanged.

## Describe alternatives you've considered

* Behind a feature flag — rejected: increases surface area and adds a flag users would never want to leave off.
* Batching at the receiver side — not possible; the bottleneck is in the producer's syscall path.
* Switching to `SOCK_DGRAM` — the existing transport is `SOCK_STREAM`, and dgram has its own per-message kernel limits; not necessary given the existing length-prefix framing.

## Expected impact

* **Syscall reduction**: 100×–1000× depending on metric-point count per scrape.
* **CPU reduction**: in the profile that motivated this fix, `Interop+Sys.Send` was 12.82% exclusive with the metric exporter as the majority contributor. Recovering ~10% exclusive CPU on a high-throughput worker is the realistic target.
* No wire change, no receiver (MDSD) change, no SDK change.

## Additional context

Implementation ready as a PR — will link below. Tests include byte-for-byte equivalence between batched bytes and concatenated per-event sends, fallback for oversize events, and flush-failure semantics.

arman-msft:
Fixes #4457

## Why

The Geneva metric exporter issues one Unix-domain-socket `Socket.Send` syscall per individual metric point in [`OtlpProtobufSerializer.SendMetricPoint`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs). At realistic cardinality (e.g. `CardinalityLimit=4000` across multiple histograms ⇒ up to ~20K points/scrape) this dominates exporter CPU.

Profile evidence from a high-throughput production .NET service shows `Interop+Sys.Send` at **~13% exclusive CPU**, with the Geneva metric exporter as the dominant contributor.

Prior art: the same shape of fix is open for the Rust contrib repo's Geneva log exporter — open-telemetry/opentelemetry-rust-contrib#306.

## What

Add two members to `IMetricDataTransport` (internal):

```csharp
bool TryAppendOtlpProtobufEvent(byte[] buffer, int size);
void FlushOtlpProtobufEvents();
```

* `MetricUnixDomainSocketDataTransport` accumulates serialized events into an internal buffer sized to match the existing per-event allocation (`GenevaMetricExporter.BufferSize` = 65360 bytes) and flushes via the existing UDS path. The buffer offset is reset *before* the underlying `Send` so a throw during flush cannot leak partial state into the next export.
* `MetricUnixUserEventsDataTransport` and `MetricWindowsEventTracingDataTransport` preserve per-event semantics — each ETW / user_events write is a logical event for the consumer, so they forward `TryAppend` directly to `SendOtlpProtobufEvent` and treat `Flush` as a no-op.
* `OtlpProtobufSerializer.SendMetricPoint` becomes `TryAppend → on overflow Flush+retry → fallback to standalone Send` for any single event that exceeds the accumulation buffer (preserves current behavior for oversize events).
* `SerializeAndSendMetrics` calls `FlushOtlpProtobufEvents()` exactly once at the end. Flush failures are surfaced through the same `ExportResult.Failure` path as per-point serialization errors.
* `TlvMetricExporter` uses a different framing (`BinaryHeader` + `MetricEventType.TLV` per event) and is intentionally **out of scope** for this PR.

## Wire-format guarantee

When `prefixBufferWithUInt32LittleEndianLength` is set (the only mode used by `MetricUnixDomainSocketDataTransport`, set in [`OtlpProtobufMetricExporter`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobufMetricExporter.cs)), each serialized event is already framed with a uint32-LE length prefix written by `WriteIndividualMessageTagsAndLength`. The Geneva MDSD receiver already demultiplexes multiple events from a single read by that prefix, so concatenating events into one `Send` is **wire-compatible**: no MDSD change, no SDK change.

This is exercised by an explicit test (`BatchedFlushedBytes_AreByteForByteEqualToConcatenatedPerEventBytes`) that compares the bytes flushed by the batching transport against the concatenation of the bytes the legacy path would have sent for the same in-memory metric snapshot.

## Opt-in / opt-out

Transparent — no feature flag. The wire format is unchanged and the only externally observable difference is `Socket.Send` call count.

## Tests

New file `test/OpenTelemetry.Exporter.Geneva.Tests/OtlpProtobufMetricBatchingTests.cs`:

| Test | Asserts |
|---|---|
| `MultiplePoints_AreFlushedInASingleSend` | Multiple metric points across one export ⇒ exactly one flush; appended-event count matches metric-point count. |
| `BufferTooSmall_ForcesIntermediateFlushes` | When the accumulation buffer cannot hold all events, intermediate flushes occur and all bytes are still delivered. |
| `SinglePointLargerThanAccumulationBuffer_FallsBackToStandaloneSend` | Oversize single event falls back to `SendOtlpProtobufEvent`; no regression. |
| `Flush_IsCalledOnceEvenWhenZeroPointsAreEmitted` | Terminal flush runs at the end of `SerializeAndSendMetrics` regardless of point count. |
| `BatchedFlushedBytes_AreByteForByteEqualToConcatenatedPerEventBytes` | Wire-format equivalence vs. the legacy per-event sends. |
| `FlushFailure_IsReportedAsExportFailure` | A throwing flush surfaces as `ExportResult.Failure`; no buffered state leaks into the next export. |

All 574+ existing Geneva tests continue to pass across `net10.0` / `net9.0` / `net8.0` / `net462` / `net47` / `net471` / `net472` / `net48`.

## CHANGELOG

Added under `## Unreleased` in `src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md`.

## Expected impact

* Syscall reduction: 100×–1000× depending on metric-point count per scrape.
* CPU reduction: in the motivating profile, `Interop+Sys.Send` was 12.82% exclusive with the metric exporter as the majority contributor. Recovering ~10% exclusive CPU on a high-throughput worker is the realistic target.
* No wire change, no MDSD receiver change, no SDK change.

cc @cijothomas @reyang @Kielek (Geneva exporter maintainers — happy to revise based on feedback, especially on the `IMetricDataTransport` API shape).

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Batch OTLP-protobuf metric events into a single Unix Domain Socket `Send`
+  per export. The Geneva metric exporter previously issued one `Socket.Send`
+  syscall per metric point, which on high-cardinality scrapes dominated
+  worker CPU. Events are now accumulated in a transport-side buffer (bounded
+  by the existing 65360-byte per-event buffer size) and flushed once per
+  export. The on-the-wire format is unchanged: each event is already
+  uint32-LE length-prefixed, which the Geneva MDSD receiver uses to
+  demultiplex events from a single read. ETW and Linux `user_events`
+  transports are unaffected (per-event semantics preserved). Single events
+  larger than the accumulation buffer fall back to a standalone send for
+  back-compat.
+
 ## 1.15.2
 
 Released 2026-Apr-21
--- a/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Metrics/OtlpProtobuf/OtlpProtobufSerializer.cs
@@ -208,6 +208,22 @@ internal sealed class OtlpProtobufSerializer
         // Serialize
         this.SerializeResourceMetrics(buffer, resource);
 
+        // Flush any events accumulated by the transport in one final
+        // transport write. Transports that do not batch implement this as a
+        // no-op. The OTLP wire format is unchanged: each event was already
+        // uint32-LE length-prefixed by WriteIndividualMessageTagsAndLength
+        // when prefixBufferWithUInt32LittleEndianLength is set, which the
+        // Geneva MDSD receiver uses to demultiplex events from a single read.
+        try
+        {
+            this.MetricDataTransport.FlushOtlpProtobufEvents();
+        }
+        catch (Exception ex)
+        {
+            this.metricExportResult = ExportResult.Failure;
+            ExporterEventSource.Log.ExporterException("Failed to flush batched OTLP metric events", ex);
+        }
+
         this.ClearScopeMetrics();
 
         return this.metricExportResult;
@@ -746,7 +762,27 @@ internal sealed class OtlpProtobufSerializer
     }
 
     private void SendMetricPoint(byte[] buffer, ref int cursor)
-        => this.MetricDataTransport.SendOtlpProtobufEvent(buffer, cursor);
+    {
+        // Try to append into the transport's accumulation buffer so the
+        // export does one large socket write instead of one per metric point.
+        if (this.MetricDataTransport.TryAppendOtlpProtobufEvent(buffer, cursor))
+        {
+            return;
+        }
+
+        // Accumulation buffer is full: flush what we have and retry once.
+        this.MetricDataTransport.FlushOtlpProtobufEvents();
+
+        if (this.MetricDataTransport.TryAppendOtlpProtobufEvent(buffer, cursor))
+        {
+            return;
+        }
+
+        // Single event larger than the accumulation buffer. Preserve the
+        // pre-batching behavior so back-compat is unchanged for oversize
+        // events and for external IMetricDataTransport consumers.
+        this.MetricDataTransport.SendOtlpProtobufEvent(buffer, cursor);
+    }
 
     private void SerializeResource(byte[] buffer, ref int cursor, Resource resource)
     {
--- a/src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/IMetricDataTransport.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Metrics/Transport/IMetricDataTransport.cs
@@ -24,4 +24,31 @@ internal interface IMetricDataTransport : IDisposable
     void SendOtlpProtobufEvent(
         byte[] body,
         int size);
+
+    /// <summary>
+    /// Attempts to append an OTLP protobuf event to an internal accumulation
+    /// buffer to be flushed later as a single transport write. The bytes
+    /// referenced by <paramref name="body"/> must be copied by the
+    /// implementation because the caller may reuse the buffer immediately.
+    /// </summary>
+    /// <param name="body">The byte array containing the serialized event.</param>
+    /// <param name="size">Length of the serialized event in bytes.</param>
+    /// <returns>
+    /// <c>true</c> if the event was appended (or sent immediately by
+    /// transports that do not batch). <c>false</c> if the accumulation buffer
+    /// would overflow; the caller is expected to call
+    /// <see cref="FlushOtlpProtobufEvents"/> and retry.
+    /// </returns>
+    bool TryAppendOtlpProtobufEvent(
+        byte[] body,
+        int size);
+
+    /// <summary>
+    /// Flushes any events previously buffered by
+    /// <see cref="TryAppendOtlpProtobufEvent"/> as a single transport write.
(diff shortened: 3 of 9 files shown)


cijothomas:
> Prior art: the same shape of fix is open for the Rust contrib repo's Geneva log exporter — https://github.com/open-telemetry/opentelemetry-rust-contrib/issues/306.

Rust's Geneva Exporter and .NET's GenevaExporter just have the name "Geneva" in common. They are fundamentally different things. GenevaExporter in .NET is just codename for writing to kernel tracing facilities. (Linux tracepoints, Windows ETW). The Unixsocket is a temporary thing, that would be removed in the future.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+* Batch OTLP-protobuf metric events into a single Unix Domain Socket `Send`
+  per export. The Geneva metric exporter previously issued one `Socket.Send`
+  syscall per metric point, which on high-cardinality scrapes dominated
+  worker CPU. Events are now accumulated in a transport-side buffer (bounded
```
cijothomas, line 8: what happens when the size exceeds the limit of 65360?

cijothomas (commented):
Not a maintainer for this package anymore, so just sharing my thoughts.
1. UnixSocket is something we should be removing from this exporter, so not worth doing any optimizations to that. The alternative is user_events and is already supported.
2. However, the write-one-metricpoint-per-syscall is bad and requires a fix. The right place for the fix is likely upstream in the SDK itself. See https://github.com/open-telemetry/opentelemetry-specification/pull/4895#issuecomment-3953440119

If we go with this change, make sure to do extensive testing too. (not unit tests, actual integration test.)

arman-msft:
Thanks @cijothomas. Will explore migrating to user_events and follow up with Rajkumar Rangaraj

cijothomas:
> Thanks @cijothomas. Will explore migrating to user_events and follow up with Rajkumar Rangaraj

Just to be clear: Migrating to user_events also won't fully mitigate the issue - we still write each metric point as separate events to user_events. It'll be much faster than socket, so it'll relive immediate pain.

PR #4474, opened 2026-06-08
[Exporter.InfluxDB] Add backpressure support for async metric writes

### LINKED ISSUE #4473: [feature request] InfluxDB backpressure support
### Component

OpenTelemetry.Exporter.InfluxDB

Yes. In production environments with sustained high metric volume, the InfluxDB metrics exporter cannot keep up with the incoming write rate. Pending metric batches accumulate faster than they are exported, which causes unbounded memory growth and can eventually lead to process instability or restarts.

The exporter should support configurable backpressure handling so that memory usage remains bounded when the write rate cannot keep up with the incoming metric rate.
When the pending queue reaches capacity, the exporter should be able to:

* block until space becomes available,
* drop the current batch, or
* evict the oldest queued batch and enqueue the current one instead.

The default behavior should remain unchanged when this feature is not enabled.

None

### Additional context

This behavior was observed in a production deployment. The issue is not that telemetry should be blocked indefinitely, but that the exporter currently has no controlled way to handle overload, which allows memory usage to grow without bound.

### Tip

SKuckluck:
Fixes #4473

### Add backpressure support to the InfluxDB metrics exporter

In production environments with a high volume of OpenTelemetry metrics, the InfluxDB exporter can become a bottleneck when the remote endpoint cannot accept data as fast as the instrumented application produces it. The existing exporter enqueues into an unbounded internal buffer on each export call using the InfluxDB client's internally buffered WriteApi, which provides no mechanism to limit how much data accumulates in memory. One of our production deployments is actively affected by this: metrics are collected at a rate the InfluxDB backend cannot sustain, causing unbounded memory growth until the process is restarted.

This change introduces a configurable backpressure mechanism. When `MaxPendingExports` is set to a value greater than zero, the exporter switches from the synchronous buffered write path to an asynchronous background worker that holds a bounded queue of pending metric batches and writes them one at a time using `IWriteApiAsync`. The caller can choose how the exporter behaves when the queue is full by setting `BackpressureMode` to one of three options: 
* block the exporting thread until space is available (Wait), 
* silently discard the current batch (DropNewest), 
* or evict the oldest queued batch and enqueue the current one instead (DropOldest). 

The default behavior when MaxPendingExports is zero remains unchanged, preserving full backwards compatibility.

`IMetricsWriter` was changed to collect line protocol strings in a caller-provided buffer rather than writing directly to WriteApi. This makes the serialization step reusable for both the existing synchronous path and the new backpressure worker.

The background worker uses a lock to keep queue updates and semaphore signaling consistent. Write failures are captured and returned from ForceFlush and Shutdown. In Wait mode, Flush also waits for blocked enqueue operations to finish before reporting success. Timeout handling uses 64-bit arithmetic to avoid overflow on large values. Disposal is idempotent and waits for the worker to stop cleanly.

### DIFF
--- a/src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
+++ b/src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
@@ -1,5 +1,11 @@
 #nullable enable
+OpenTelemetry.Exporter.InfluxDB.BackpressureMode
+OpenTelemetry.Exporter.InfluxDB.BackpressureMode.DropNewest = 1 -> OpenTelemetry.Exporter.InfluxDB.BackpressureMode
+OpenTelemetry.Exporter.InfluxDB.BackpressureMode.DropOldest = 2 -> OpenTelemetry.Exporter.InfluxDB.BackpressureMode
+OpenTelemetry.Exporter.InfluxDB.BackpressureMode.Wait = 0 -> OpenTelemetry.Exporter.InfluxDB.BackpressureMode
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions
+OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.BackpressureMode.get -> OpenTelemetry.Exporter.InfluxDB.BackpressureMode
+OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.BackpressureMode.set -> void
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Bucket.get -> string?
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Bucket.set -> void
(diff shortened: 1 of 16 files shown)


martincostello:
@SKuckluck Please sign the CLA.

Havret:
Thanks for the detailed write-up and for investigating this in production - the root cause is real. `WriteApi` explicitly documents that backpressure is not implemented in C# ([source comment](https://github.com/influxdata/influxdb-client-csharp/blob/master/Client/WriteApi.cs)) and uses an unbounded Rx buffer, so the OOM scenario you describe is valid.

That said, I wonder if we can solve this with a much simpler change that avoids the custom worker, bounded queue, and the new public API surface entirely.

**The key insight**: the InfluxDB exporter uses `MetricReaderTemporalityPreference.Cumulative`. This means if `Export()` returns `ExportResult.Failure`, no data is lost - the OTel SDK keeps accumulating metric values and the next export cycle will include everything. The `PeriodicExportingMetricReader` is already a natural retry/backpressure mechanism.

The root problem isn't the lack of a bounded queue - it's that `WriteApi.WritePoint()` never tells OTel that writes are failing or falling behind. It silently enqueues into an unbounded buffer, so `Export()` always returns `Success` even when InfluxDB is down.

What do you think? Happy to discuss further.

PR #4498, opened 2026-06-11
[Exporter.Geneva] Reduce per-record allocations

petrroll:
## Changes

Cut hot-path allocations in the Geneva exporter:

* Cache sanitized logger category names with a lock-free copy-on-write (CAS) cache, plus an accelerated fast path that returns the original string when it is already a valid event name. The cache snapshot is a FrozenDictionary on .NET 8+ and a Dictionary elsewhere.
* MetricSerializer.SerializeBase128String now writes UTF-8 directly into the buffer via GetByteCount/GetBytes instead of allocating a temp byte[] and copying.
* MsgPackTraceExporter.SerializeHttpUrl writes the composed HTTP url segments straight into the msgpack buffer (back-patching the STR16 header) instead of building interpolated strings + StringBuilder; the rare >16383-char case still delegates to GetHttpUrl for byte-identical truncation.

Adds byte-for-byte equivalence tests for SerializeHttpUrl and before/after allocation benchmarks for each optimization.

### DIFF
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -5,6 +5,13 @@
 * Updated OpenTelemetry core component version(s) to `1.16.0`.
   ([#4487](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4487))
 
+* Reduced per-log-record and per-span allocations in the Geneva exporter:
+  cache sanitized logger category names (with a fast path that reuses already
+  valid names) and freeze the cache on .NET 8+, serialize metric base-128
+  strings and HTTP url spans directly into the output buffer instead of via
+  temporary arrays/strings.
+  ([#4498](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4498))
+
 ## 1.15.2
 
 Released 2026-Apr-21
--- a/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
@@ -5,14 +5,15 @@
 * Updated OpenTelemetry core component version(s) to `1.16.0`.
   ([#4487](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4487))
 
+* Log error when telemetry data exceeds the serialization buffer capacity.
+  ([#4027](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4027))
+
 * Reduced per-log-record and per-span allocations in the Geneva exporter:
   cache sanitized logger category names (with a fast path that reuses already
   valid names) and freeze the cache on .NET 8+, serialize metric base-128
   strings and HTTP url spans directly into the output buffer instead of via
   temporary arrays/strings.
   ([#4498](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4498))
-* Log error when telemetry data exceeds the serialization buffer capacity.
-  ([#4027](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4027))
 
 ## 1.15.2
 
--- a/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
+++ b/src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
@@ -313,6 +313,97 @@ internal sealed class MsgPackTraceExporter : MsgPackExporter, IDisposable
         return urlStringBuilder.ToString();
     }
 
+    // Serializes the composed HTTP url (scheme://address:port/path?query) for a
+    // server span directly into the msgpack buffer, avoiding the intermediate
+    // interpolated strings, StringBuilder, and ToString() that GetHttpUrl
+    // allocates. The fast path writes each url segment's UTF-8 straight into the
+    // buffer and back-patches the STR16 length header (exactly mirroring
+    // MessagePackSerializer.SerializeUnicodeString's wire format). The rare
+    // over-limit case delegates to GetHttpUrl + SerializeUnicodeString so the
+    // existing char-count truncation semantics are preserved byte-for-byte.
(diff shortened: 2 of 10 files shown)


src/OpenTelemetry.Exporter.Geneva/Internal/Tld/TldLogCommon.cs
```diff
@@ -241,6 +232,138 @@ protected virtual void Dispose(bool disposing)
...
+        var sanitized = SanitizeCategoryName(categoryName);
+
+        // Lock-free copy-on-write update. The cache is a pure memoization of a
+        // deterministic, side-effect-free function, so a lost race is harmless:
+        // the worst case is that the value is recomputed and re-inserted later.
+        // We use compare-and-swap (rather than a plain write) so the update is
+        // guaranteed to make progress and never overwrite a concurrent insert
+        // with a stale snapshot. The loop terminates because the cache only ever
+        // grows up to MaxCachedSanitizedCategoryNames entries.
+        while (true)
```
petrroll: This is ok *based on assumption* (please help me verify) that new unique category names should be relatively bounded. I.e. there can't be thousands of tens of thousands of them.

src/OpenTelemetry.Exporter.Geneva/Internal/Tld/TldLogCommon.cs
```diff
@@ -2,18 +2,25 @@
...
+#else
+using SanitizedCategoryNameCache = System.Collections.Generic.Dictionary<string, string>;
+#endif
 
 namespace OpenTelemetry.Exporter.Geneva.Tld;
 
 internal abstract class TldLogCommon : IDisposable
 {
     protected const int MaxSanitizedEventNameLength = 50;
+    protected const int MaxCachedSanitizedCategoryNames = 1024;
```
petrroll: Does this feel sensible? Could be increased even substantially IMHO.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Reduced per-log-record and per-span allocations in the Geneva exporter:
+  cache sanitized logger category names (with a fast path that reuses already
+  valid names) and freeze the cache on .NET 8+, serialize metric base-128
+  strings and HTTP url spans directly into the output buffer instead of via
+  temporary arrays/strings.
```
martincostello, line 15: Please add a link to this PR like the other entries.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
```diff
@@ -313,6 +313,97 @@ internal static bool CacheIfPartOfHttpUrl(KeyValuePair<string, object?> entry, o
...
 
+    // Serializes the composed HTTP url (scheme://address:port/path?query) for a
+    // server span directly into the msgpack buffer, avoiding the intermediate
+    // interpolated strings, StringBuilder, and ToString() that GetHttpUrl
+    // allocates. The fast path writes each url segment's UTF-8 straight into the
+    // buffer and back-patches the STR16 length header (exactly mirroring
+    // MessagePackSerializer.SerializeUnicodeString's wire format). The rare
+    // over-limit case delegates to GetHttpUrl + SerializeUnicodeString so the
+    // existing char-count truncation semantics are preserved byte-for-byte.
+    internal static int SerializeHttpUrl(byte[] buffer, int cursor, object?[] httpUrlParts, ref ushort cntFields)
```
martincostello: `fieldCount`?

src/OpenTelemetry.Exporter.Geneva/Internal/Tld/TldLogCommon.cs
martincostello: If it can provably terminate then it's OK, but usually I prefer to have some sort of check on these so if we're wrong it won't get stuck in an infinite loop.

For example there could be a `maxAttempts` safety net (or max time using a stopwatch `long` value) and an exception is thrown if it gets exceeded.

`while (true)`s always make me uneasy 😄

martincostello (commented):
Looks reasonable to me.

Are there any changes that could/should be made to extend the fuzz tests?

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
petrroll: Done - added the PR link.

src/OpenTelemetry.Exporter.Geneva/Internal/MsgPack/MsgPackTraceExporter.cs
petrroll: Done

src/OpenTelemetry.Exporter.Geneva/Internal/Tld/TldLogCommon.cs
petrroll: Fixed just to be sure. But it kinda provably always progressed even before.

martincostello (approved):
Looks reasonable to me but I defer to the component owners for final approval.

martincostello:
Needs some style fixes too.

petrroll:
Adjusted

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 * Updated OpenTelemetry core component version(s) to `1.16.0`.
   ([#4487](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4487))
 
+* Reduced per-log-record and per-span allocations in the Geneva exporter:
+  cache sanitized logger category names (with a fast path that reuses already
+  valid names) and freeze the cache on .NET 8+, serialize metric base-128
+  strings and HTTP url spans directly into the output buffer instead of via
+  temporary arrays/strings.
+  ([#4498](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4498))
```
martincostello, line 16: This should go at the end of the Unreleased section.

rajkumar-rangaraj (approved):
LGTM

@xiang17 Could you please review before the merge?

PR #4500, opened 2026-06-11
[Instrumentation.AWS] Reduce allocations

martincostello:
## Changes

Reduce allocations by caching immutable values rather than recomputing them.


normj (approved):
At some point we should think about updating the SDK to add `OperationName` to the `IRequestContent` the generator can make sure the operation name is set as part of the marshaller. That would get rid of all the extra temporary strings stripping `Request` and `Response` from the type names as well as remove the reflection calls.

Kielek (commented):
LGTMRS

Kielek (approved):

PR #4338, opened 2026-05-01
[gRPC] Update Semantic Conventions for client instrumentations

martincostello:
Fixes #4335
Contributes to #4064

## Changes

- Update to [v1.41.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.41.0/docs/rpc/rpc-spans.md) of the gRPC Semantic Conventions.
- ~~Add `net8.0` and `net10.0` TFMs for GrpcCore instrumentation.~~
- Improve gRPC test server fixture.
- Improve GrpcCore tests.
- Remove obsolete code.
- Unskip tests that pass.
- Apply Visual Studio refactor suggestions.
- Add `net462` target to GrpcCore tests.
- Fix broken behaviour in GrpcCore for .NET Framework found through testing.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #4508, opened 2026-06-12
[gRPC] Update gRPC Semantic Conventions to 1.42.0

### LINKED ISSUE #1778: Grpc Instrumentation is broken in NET6.0 or higher, when custom propagator is used
# Bug Report

Grpc Instrumentation creates a sibling activity, and sets it as current, if the propagator is not the default understood by ASP.NET Core.  The expectation was that, Grpc would retrieve the Activity from Activity.Current, and hence it'd be getting the newly created sibling Activity. However, in newer versions, Grpc is *NOT* using Activity.Current to fetch the activity to populate tags. See the change: https://github.com/grpc/grpc-dotnet/pull/1545/files. This means Grpc library is adding tags to the original activity it created, but instrumentation is looking for tags inside the activity it created.

Runtime version: net6.0

Not likely an issue with ASP.NET Core, but this is to be confirmed.

martincostello:
See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4338#issuecomment-4671180315.

## Changes

- Fix 1.41 Semantic Conventions for gRPC:
  - Remove `rpc.service`.
  - Update `rpc.method` to include the service name.
- Updated to 1.42.0 of the Semantic Conventions for gRPC.
- Fix gRPC instrumentation not working when using a custom propagator to resolve #1778.


Kielek:
> Haven't added a CHANGELOG entry is this is just a follow-up for unreleased changes from #4338.

Maybe you could add this after comma `(#4338, #4508)`?

src/OpenTelemetry.Instrumentation.GrpcCore/RpcScope.cs
IliaBrahinets: `TrimStart` can be replaced with `Trim` to be  consistent with Grpc.Net.Client

Matus-p:
i did some additional changes to adhere to 1.41 grpc semantic conventions, see this PR: https://github.com/martincostello/opentelemetry-dotnet-contrib/pull/4
feel free to cherry-pick it to your PR

src/OpenTelemetry.Instrumentation.GrpcCore/RpcScope.cs
Matus-p: similarly this could be updated to use Trim/TrimStart too: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/85e155fce69ec0e48f35e8c02d97a67eeca1501a/src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs#L501

martincostello:
> i did some additional changes to adhere to 1.41 grpc semantic conventions, see this PR: [martincostello#4](https://github.com/martincostello/opentelemetry-dotnet-contrib/pull/4)

Thanks. I'll look to cherry-pick the relevant parts once #4370 is merged.

martincostello:
@Matus-p I incorporated your feedback from my fork. Could you sign the CLA please? Otherwise I'll need to update the latest commit message to remove your co-authorship.

src/OpenTelemetry.Instrumentation.AspNetCore/Implementation/HttpInListener.cs
```diff
@@ -46,10 +45,6 @@ internal class HttpInListener : ListenerHandler
     private static readonly PropertyFetcher<Exception> ExceptionPropertyFetcher = new("Exception");
     private static readonly object CreatedByInstrumentationMarker = new();
 
-    // Caches the display name, rpc.service, and rpc.method derived from the raw gRPC method string.
-    // The set of distinct gRPC method strings is bounded by the number of gRPC endpoints in the app.
-    private static readonly GrpcMethodDetailsCache GrpcMethodCache = new();
```
martincostello, line 51: Now that we don't need to parse the service name, the cache became redundant.

test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.server.cs
Matus-p: AttributeRpcService should not be present and AttributeRpcMethod should contain full name

Matus-p: AttributeNetPeerPort and AttributeNetPeerIp are not updated too .. the test is disabled, probably this is why it it is not updated

martincostello: Thanks - updated, and then I went and fixed the bug and unskipped the test.

Kielek:
Codex findings

[P1] src/OpenTelemetry.Instrumentation.GrpcCore/RpcScope.cs:208 - GrpcCore still does not apply 1.42 gRPC client error rules. StopActivity(int) only writes rpc.response.status_code and stops, so a disposed/cancelled client call still exports CANCELLED with status unset and no error.type. v1.42 says client spans treat every non-OK status as an error. The same path also means exception failures reach StopActivity(...) without setting error.type.
[P2] src/Shared/GrpcTagHelper.cs:37 - SetGrpcMethodAndDisplayNameFromActivity maps every non-null grpc.method directly to rpc.method. The 1.42 compatibility doc says native grpc.method=other must become rpc.method=_OTHER and preserve the original in rpc.method_original; this currently emits rpc.method=other and names the span other.

Kielek:
One more
[P1] src/OpenTelemetry.Instrumentation.GrpcCore/RpcScope.cs:265 - StopActivity(Exception) marks every RpcException as ActivityStatusCode.Error before StopActivity(int) applies the kind-specific gRPC status rules. For server spans, v1.42 only treats UNKNOWN, DEADLINE_EXCEEDED, UNIMPLEMENTED, INTERNAL, UNAVAILABLE, and DATA_LOSS as errors, but the current server test path uses RESOURCE_EXHAUSTED and still expects Error plus error.type. This defeats the new ResolveSpanStatusForGrpcStatusCodeOnServer(...) logic for handled server status codes.

martincostello:
Going to look at adding some weaver tests on top of the latest commit for additional verification.

src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
```diff
@@ -20,12 +20,15 @@
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" VersionOverride="[3.33.2,)" />
     <PackageReference Include="Grpc.Core.Api" />
-    <PackageReference Include="OpenTelemetry.Api" />
+    <PackageReference Include="OpenTelemetry" />
```
Kielek: Why this change is needed?
By default instrumentation packages should reference only API/APIBuilder package.

Kielek (commented):
One more comment from Codex... I am not sure why it cannot handle at once.:
P1 [test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.client.cs:374](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/2b8e2f556dd5d9dce9ab7165979685985a5872e7/test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.client.cs#L374)
This assertion still reads SemanticConventions.AttributeRpcGrpcStatusCode (rpc.grpc.status_code), but the PR now writes rpc.response.status_code. Nearby assertions were updated correctly. This test should assert AttributeRpcResponseStatusCode, otherwise GetTagValue(...) returns null.

src/OpenTelemetry.Instrumentation.GrpcCore/OpenTelemetry.Instrumentation.GrpcCore.csproj
martincostello: The weaver tests need the `Metrics` type - fixed by moving the reference to the test project.

test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.client.cs
```diff
@@ -200,26 +303,22 @@ public void GrpcAndHttpClientInstrumentationWithSuppressInstrumentation()
...
         Assert.Equal($"greet.Greeter/SayHello", grpcSpan4.DisplayName);
-        Assert.Equal(0, grpcSpan4.GetTagValue(SemanticConventions.AttributeRpcResponseStatusCode));
+        Assert.Equal("OK", grpcSpan4.GetTagValue(SemanticConventions.AttributeRpcResponseStatusCode));
     }
 
-#if NET
     [Fact(Skip = "https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1727")]
-#else
-    [Fact]
-#endif
```
martincostello, line 222: This is already in a `#if NET` block.

test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.client.cs
```diff
@@ -267,9 +366,9 @@ public void GrpcPropagatesContextWithSuppressInstrumentationOptionSetToTrue()
             Assert.Equal($"greet.Greeter/SayHello", clientActivity.DisplayName);
             Assert.Equal($"POST /greet.Greeter/SayHello", serverActivity.DisplayName);
             Assert.Equal(clientActivity.TraceId, serverActivity.TraceId);
-            Assert.Equal(clientActivity.SpanId, serverActivity.ParentSpanId);
-            Assert.Equal(0, clientActivity.GetTagValue(SemanticConventions.AttributeRpcGrpcStatusCode));
+            Assert.Equal("OK", clientActivity.GetTagValue(SemanticConventions.AttributeRpcResponseStatusCode));
             Assert.Equal("customValue", serverActivity.GetCustomProperty("customField") as string);
+            Assert.Equal(clientActivity.SpanId, serverActivity.ParentSpanId);
```
martincostello, line 371: Moved last so it's easy to check all the other assertions pass.

Kielek (approved):

PR #4509, opened 2026-06-15
[OpAMP] Add support for reporting remote configuration status

ysolomchenko:
## Why

https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverremote_config_status

## Changes

Add support for reporting remote configuration status


src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -17,6 +17,7 @@ internal sealed class FrameBuilder : IFrameBuilder
     private readonly OpAmpClientSettings settings;
 
     private AgentToServer? currentMessage;
+    private RemoteConfigStatusReport? lastRemoteConfigStatusReport;
```
RassK: This class should not hold state of concrete messages.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -212,6 +218,32 @@ IFrameBuilder IFrameBuilder.AddEffectiveConfig(IEnumerable<EffectiveConfigFile>
         return this;
     }
 
+    IFrameBuilder IFrameBuilder.AddRemoteConfigStatus(RemoteConfigStatusReport status)
+    {
+        this.EnsureInitialized();
+
+        if (RemoteConfigStatusReportEquals(this.lastRemoteConfigStatusReport, status))
```
RassK: Irrelevant, leave this for the client's implementation.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -225,6 +257,25 @@ AgentToServer IFrameBuilder.Build()
     public void Reset()
     {
         this.currentMessage = null;
+        this.lastRemoteConfigStatusReport = null;
+    }
+
+    private static RemoteConfigStatuses MapStatus(RemoteConfigStatusCode status)
```
RassK: this is message specific, wrong place. You can use extra helper to keep this class cleaner.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
```diff
@@ -0,0 +1,48 @@
...
+        ReadOnlyMemory<byte> lastRemoteConfigHash,
+        RemoteConfigStatusCode status,
+        string? errorMessage = null)
+    {
+        if (status is < RemoteConfigStatusCode.Unset or > RemoteConfigStatusCode.Failed)
+        {
+            throw new ArgumentOutOfRangeException(nameof(status), status, "Unsupported remote configuration status.");
+        }
+
+        this.lastRemoteConfigHash = lastRemoteConfigHash.ToArray();
```
RassK, line 43: Any specific reason for the early copy instead of a reference ?

src/OpenTelemetry.OpAmp.Client/Settings/RemoteConfigSettings.cs
```diff
@@ -16,4 +16,13 @@ public sealed class RemoteConfigSettings
...
     public bool AcceptsRemoteConfig { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the client reports remote configuration status.
+    /// </summary>
+    /// <value>
+    /// <c>true</c> if remote configuration status should be reported to the server; otherwise, <c>false</c>.
+    /// Default is <c>false</c>.
+    /// </value>
+    public bool ReportsRemoteConfig { get; set; }
```
RassK: Maybe better to change this to `ReportsRemoteConfigStatus`, I know the proto capability is named differently but imo it does not have a good descriptive name.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
ysolomchenko: fixed in [ebae6f8](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/ebae6f8f6124ae05256d12f15ebc18f9023c6bb9)

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
ysolomchenko: fixed in [ebae6f8](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/ebae6f8f6124ae05256d12f15ebc18f9023c6bb9)

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
ysolomchenko: fixed in [ebae6f8](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/ebae6f8f6124ae05256d12f15ebc18f9023c6bb9)

src/OpenTelemetry.OpAmp.Client/Settings/RemoteConfigSettings.cs
ysolomchenko: fixed in [306f181](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/306f181f8bb99470937750a3a9881a444835a0de)

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
ysolomchenko: Yes, this is a defensive copy. `ReadOnlyMemory` can reference caller-owned mutable memory, so copying here keeps `RemoteConfigStatusReport` immutable and prevents the reported hash from changing after construction.

RassK: Any realistic paths where it could change? It should be pretty much read only field anyway.

ysolomchenko: The normal `ConfigHash` path is immutable, but the constructor accepts arbitrary `ReadOnlyMemory`, which can wrap caller-owned mutable arrays or pooled buffers.

ysolomchenko:
```
var hash = new byte[] { 1, 2, 3 };
var report = new RemoteConfigStatusReport(hash, RemoteConfigStatusCode.Applied);

hash[0] = 9; // report would observe this if it kept a reference
```

RassK: Seems more of a pseudo defence against a possible coding error. I agree that the public layer may cause a threat, but 99% it's the same reference coming from `RemoteConfigMessage`. If you see that the constructor is accepting ReadOnlyMemory<T> (or any reference type), it's already a flag that you need to maintain the state till a response.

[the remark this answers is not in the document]
ysolomchenko: Its not true.

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
```diff
@@ -320,6 +322,87 @@ internal async Task SendsEffectiveConfigFile_FromFile()
...
+
+    [Fact]
+    internal void RemoteConfigStatusReport_CopiesHash()
+    {
+        var hash = new byte[] { 1, 2, 3 };
+
+        var status = new RemoteConfigStatusReport(hash, RemoteConfigStatusCode.Applied);
+        hash[0] = 4;
+
+        Assert.Equal(new byte[] { 1, 2, 3 }, status.LastRemoteConfigHash.ToArray());
```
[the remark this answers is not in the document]
ysolomchenko: Its not true.

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Initializes a new instance of the <see cref="RemoteConfigStatusReport"/> class.
+    /// </summary>
+    /// <param name="lastRemoteConfigHash">Hash bytes from the last received remote configuration.</param>
+    /// <param name="status">The remote configuration status.</param>
+    /// <param name="errorMessage">Optional error message when <paramref name="status"/> is <see cref="RemoteConfigStatusCode.Failed"/>.</param>
+    public RemoteConfigStatusReport(
+        ReadOnlyMemory<byte> lastRemoteConfigHash,
+        RemoteConfigStatusCode status,
+        string? errorMessage = null)
+    {
```
stevejgordon, line 25: Should we also validate that error message is only set, when status is failed. i.e. this guard:

```c#
if (errorMessage != null && status != RemoteConfigStatusCode.Failed)
{
    throw new ArgumentException(
        $"{nameof(errorMessage)} should only be set when {nameof(status)} is {nameof(RemoteConfigStatusCode.Failed)}.",
        nameof(errorMessage));
}
```

stevejgordon, line 25: Should we also validate that the hash is not empty:

```c#
if (lastRemoteConfigHash.IsEmpty)
{
    throw new ArgumentException("Hash must not be empty.", nameof(lastRemoteConfigHash));
}
```

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
```diff
@@ -320,6 +322,85 @@ internal async Task SendsEffectiveConfigFile_FromFile()
         Assert.Equal(fileContentType, actualConfig.ContentType);
     }
 
+    [Fact]
+    internal async Task SendsRemoteConfigStatus_IsDisabledAndThrows()
+    {
```
stevejgordon, line 327: Does this need to spin up the fake server for this assertion? Replace with something like:

```c#
using var client = new OpAmpClient(o => o.Heartbeat.IsEnabled = false);
var statusReport = new RemoteConfigStatusReport(new byte[] { 1, 2, 3 }, RemoteConfigStatusCode.Applied);
Assert.Throws<InvalidOperationException>(() => client.SendRemoteConfigStatusAsync(statusReport));
```

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
Kielek: I think the snapshot is intentional and worth keeping for the public report object: ReadOnlyMemory prevents mutation through this API, but it can still wrap caller-owned mutable or pooled memory. Since the report may be sent later, keeping a reference would make the hash lifetime-dependent.

That said, the current constructor takes ReadOnlyMemory while RemoteConfigMessage exposes ConfigHash as ReadOnlySpan, so normal callers may pay a caller-side copy plus this defensive copy. Since this is new API, I’d consider changing the constructor parameter to ReadOnlySpan<byte> and still copying internally.

Kielek (commented):
Plus some Codex scans results, you do not need to blindly apply them, but leaving it here for consideration:
[P2] Public API makes the primary handoff awkward and double-copies
[RemoteConfigStatusReport accepts ReadOnlyMemory<byte>](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/86907eb15037108b6f26031a3f9379f6410f1148/src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs#L18-L29), but the natural source value, [RemoteConfigMessage.ConfigHash](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/86907eb15037108b6f26031a3f9379f6410f1148/src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigMessage.cs#L43), is ReadOnlySpan<byte>. A caller receiving remote config cannot pass message.ConfigHash directly; they must call ToArray(), then the constructor copies again. Since this is new public API and the constructor already snapshots the hash, I’d change the constructor parameter to ReadOnlySpan<byte>.

[P3] error_message can be sent for successful/non-failed statuses
[RemoteConfigStatusReport stores any errorMessage](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/86907eb15037108b6f26031a3f9379f6410f1148/src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs#L30), and [ToRemoteConfigStatus sends it whenever non-null](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/86907eb15037108b6f26031a3f9379f6410f1148/src/OpenTelemetry.OpAmp.Client/Internal/Utils/RemoteConfigStatusReportExtensions.cs#L16-L19). The OpAMP proto describes error_message as optional when status==FAILED; allowing it with Applied/Applying creates semantically inconsistent reports. I’d either reject errorMessage unless status is Failed, or only serialize it for Failed.

src/OpenTelemetry.OpAmp.Client/Settings/RemoteConfigSettings.cs
```diff
@@ -16,4 +16,13 @@ public sealed class RemoteConfigSettings
     /// Default is <c>false</c>.
     /// </value>
     public bool AcceptsRemoteConfig { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the client reports remote configuration status.
+    /// </summary>
+    /// <value>
+    /// <c>true</c> if remote configuration status should be reported to the server; otherwise, <c>false</c>.
```
stevejgordon, line 24: Would a remark be useful here, along the lines of:

```c#
/// <remarks>
/// Typically set alongside <see cref="AcceptsRemoteConfig"/>. Without that capability the server
/// will not send any remote configuration, so the agent will have no configuration hash to report.
/// </remarks>
```

test/OpenTelemetry.OpAmp.Client.Tests/DataGenerators/FrameBuilderTestData.cs
```diff
@@ -19,5 +20,7 @@ public FrameBuilderTestData()
         this.Add(fb => fb.AddAgentDisconnect(), m => m.AgentDisconnect);
 
         this.Add(fb => fb.AddCapabilities(), m => m.Capabilities);
+
+        this.Add(fb => fb.AddRemoteConfigStatus(new RemoteConfigStatusReport(new byte[] { 1, 2, 3 }, RemoteConfigStatusCode.Applied)), m => m.RemoteConfigStatus);
```
stevejgordon, line 24: Should we also test other statuses?

```c#
this.Add(fb => fb.AddRemoteConfigStatus(new RemoteConfigStatusReport(new byte[] { 1, 2, 3 }, RemoteConfigStatusCode.Applying)), m => m.RemoteConfigStatus);
this.Add(fb => fb.AddRemoteConfigStatus(new RemoteConfigStatusReport(new byte[] { 1, 2, 3 }, RemoteConfigStatusCode.Unset)), m => m.RemoteConfigStatus);
```

ysolomchenko: fixed in [c934492](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/c934492c2670338146ad10fd9371ea8a3617b63d)

src/OpenTelemetry.OpAmp.Client/Settings/RemoteConfigSettings.cs
ysolomchenko: fixed in [ef0c752](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/ef0c752cc0362cbf4f36f451f5f815ea975fa5b3)

test/OpenTelemetry.OpAmp.Client.Tests/OpAmpClientTests.cs
ysolomchenko: fixed in [1be2130](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/1be213029cbd8c6fb3b24ed6e1546e7e20164a8b)

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
ysolomchenko: fixed in [caa4ddd](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/caa4dddc52910fd69cefbec307b07ec768575cd4)

ysolomchenko: fixed in [a273963](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/a273963b43ace6773caf7f4c83694d7d60286607)

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
ysolomchenko: changed to `ReadOnlySpan` in [5d117cf](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4509/commits/5d117cfe493fc2057cc69edab182702390ce735d)

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
stevejgordon, line 21: We should probably add the exceptions to the XML docs since this is a public API

stevejgordon (commented):
Spotted one other small improvement

src/OpenTelemetry.OpAmp.Client/Messages/RemoteConfiguration/RemoteConfigStatusReport.cs
ysolomchenko: Added in 961f455

stevejgordon (approved):
LGTM, thanks @ysolomchenko

Kielek (approved):
LGTMRS

PR #4540, opened 2026-06-19
[Geneva] Fix TlvMetricExporter dropping instrumentation scope (Meter) tags

sandersaares:
`TlvMetricExporter` (the MessagePack/TLV serializer used by `GenevaMetricExporter`) only exported the tags present directly on each `MetricPoint` plus the prepopulated dimensions, silently dropping any tags inherited from the instrumentation scope (i.e. the `Meter`'s tags). This meant metrics emitted from a `Meter` configured with scope-level tags lost those dimensions in the Geneva TLV path, even though the OTLP-protobuf path in the same exporter already serialized them correctly as instrumentation scope attributes.

This change threads `Metric.MeterTags` through `SerializeMetricWithTLV` / `SerializeHistogramMetricWithTLV` into the dimension serialization, so instrumentation scope (Meter) tags are now emitted as metric dimensions alongside the `MetricPoint` tags and the prepopulated dimensions. The reserved `_microsoft_metrics_account` and `_microsoft_metrics_namespace` keys are honored when declared on the Meter as well (with `MetricPoint`-level values taking precedence). Since the TLV dimension model is flat (all keys, then all values, with a count), the three sources are merged in the order prepopulated → Meter → MetricPoint.

Tests were added covering scope tags being exported as dimensions (in combination with prepopulated and point tags) and reserved scope tags overriding the account/namespace without being emitted as dimensions. Existing serialization test/benchmark call sites were updated for the new parameter.

* [x] Changes in public API reviewed (if applicable) — no public API surface changed


src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Fixed `TlvMetricExporter` (used by the MessagePack/TLV `GenevaMetricExporter`)
+  dropping dimensions inherited from the instrumentation scope (`Meter` tags).
+  Meter-level tags are now exported as metric dimensions alongside the
+  `MetricPoint` tags and prepopulated dimensions.
```
martincostello: The entry should also be added to the end of the section.

```suggestion
* Fixed `TlvMetricExporter` dropping dimensions inherited from the
  instrumentation scope (`Meter` tags).
```

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
```diff
@@ -608,7 +618,40 @@ private static void SerializeHistogramBucketWithTLV(in HistogramBucket bucket, b
     }
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private void SerializeDimensionsAndGetCustomAccountNamespace(in ReadOnlyTagCollection tags, byte[] buffer, ref int bufferIndex, out string monitoringAccount, out string metricNamespace)
+    private static bool IsReservedDimensionKey(string key)
+    {
+        return string.Equals(key, GenevaMetricExporter.DimensionKeyForCustomMonitoringAccount, StringComparison.OrdinalIgnoreCase) ||
+            string.Equals(key, GenevaMetricExporter.DimensionKeyForCustomMetricsNamespace, StringComparison.OrdinalIgnoreCase);
+    }
```
martincostello:
```suggestion
    private static bool IsReservedDimensionKey(string key) =>
        string.Equals(key, GenevaMetricExporter.DimensionKeyForCustomMonitoringAccount, StringComparison.OrdinalIgnoreCase) ||
        string.Equals(key, GenevaMetricExporter.DimensionKeyForCustomMetricsNamespace, StringComparison.OrdinalIgnoreCase);
```

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
```diff
@@ -608,7 +618,40 @@ private static void SerializeHistogramBucketWithTLV(in HistogramBucket bucket, b
...
+            }
+
+            return true;
+        }
+
+        return false;
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private void SerializeDimensionsAndGetCustomAccountNamespace(in ReadOnlyTagCollection tags, IEnumerable<KeyValuePair<string, object?>>? meterTags, byte[] buffer, ref int bufferIndex, out string monitoringAccount, out string metricNamespace)
```
martincostello: This is wrapping in the GitHub web view, so the line is unreasonably long.

```suggestion
    private void SerializeDimensionsAndGetCustomAccountNamespace(
        in ReadOnlyTagCollection tags,
        IEnumerable<KeyValuePair<string, object?>>? meterTags,
        byte[] buffer,
        ref int bufferIndex,
        out string monitoringAccount,
        out string metricNamespace)
```

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
```diff
@@ -636,6 +679,30 @@ private void SerializeDimensionsAndGetCustomAccountNamespace(in ReadOnlyTagColle
...
+        var meterTagsList = meterTags as IReadOnlyList<KeyValuePair<string, object?>> ?? meterTags?.ToArray();
+
+        // Serialize instrumentation scope (Meter) Dimension keys
+        if (meterTagsList != null)
+        {
+            foreach (var tag in meterTagsList)
+            {
+                if (tag.Key.Length > GenevaMetricExporter.MaxDimensionNameSize)
+                {
+                    // TODO: Data Validation
```
martincostello, line 697: Is this something we can address here? I realise the code has just moved, but there's no tracking issue for this for someone to ever actually follow-up on.

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
```diff
@@ -664,26 +730,31 @@ private void SerializeDimensionsAndGetCustomAccountNamespace(in ReadOnlyTagColle
...
-            }
-
-            if (string.Equals(tag.Key, GenevaMetricExporter.DimensionKeyForCustomMetricsNamespace, StringComparison.OrdinalIgnoreCase) && tag.Value is string metricsNamespace)
-            {
-                if (!string.IsNullOrWhiteSpace(metricsNamespace))
+                var dimensionValue = ConvertTagValueToString(tag.Value);
+                if (dimensionValue?.Length > GenevaMetricExporter.MaxDimensionValueSize)
                 {
-                    metricNamespace = metricsNamespace;
+                    // TODO: Data Validation
```
martincostello, line 750: Same here.

martincostello (commented):
A few minor comments, but the substance of the change's approval is deferred to the Microsoft owners of this component for final approval and merge.

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
sandersaares: I think I am missing the point of this comment - are you suggesting a rewording or a moving of the entry? Happy to do either, just confused about how the comment (to move to end of section?) and the suggested change (a rewording) relate.

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
sandersaares: Yes, this also caught my eye but it is not obvious what should be done here, so I figured to keep the scope of my PR small and targeted. Any specific suggestions?

src/OpenTelemetry.Exporter.Geneva/CHANGELOG.md
martincostello: I'm suggesting that it's overly verbose _and_ that it's in the wrong place.

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
martincostello: Nothing specific - this component is "special". It lives in this repository, but it is only of use to Microsoft employees, so non-employees don't really have any knowledge of how this works end-to-end. I can only really make comments outside the specific problem domain.

@xiang17 will be able to provide more concrete guidance.

src/OpenTelemetry.Exporter.Geneva/Metrics/TlvMetricExporter.cs
martincostello, line 1: Did this use the right line endings? The error was about inconsistency, so I would have expected to just see new code added.

sandersaares: Yeah, it seems this file was an odd one out - I normalized it based on neighbors that I did not touch, as an authoritative reference. The neighbors were CRLF. Now, though, I see this file used LF instead. I will just keep it LF then - Geneva is special in more ways than are obvious, apparently. 🤷‍♂️

rajkumar-rangaraj:
This is supported through otlpprotobufencoding, no plans to get to Tlv. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Exporter.Geneva#otlpprotobufencoding

PR #4541, opened 2026-06-19
[OneCollector.Tests] Use fake endpoint for OneCollector PR tests

Kielek:
## Changes

> [!IMPORTANT]
> With this changes we lose possibility to execute integration tests before merge, but the projects seems to be stable right now. Manual execution is possible after merge

This PR removes the unsafe `pull_request_target` OneCollector integration workflow path that checked out PR code while exposing live backend secrets.

OneCollector integration tests now run by default against an in-process fake OneCollector endpoint, so they can execute safely as part of normal PR CI through `build-test-exporter-onecollector` in `ci.yml`.

The dedicated OneCollector integration workflow is now manual-only via `workflow_dispatch` and is retained for live backend validation. When `OTEL_ONECOLLECTOR_INSTRUMENTATION_KEY` is provided, the same tests use the real OneCollector backend; otherwise they use the fake endpoint.

## Why

* actions/checkout@7+ blocks possibility to checkout code by default for such scenarios, ref: https://github.com/actions/checkout/pull/2454 - not tested this TBH
* Make the zizmor analysis happy
* Make at least one other tool I have marked it manually that it is exactly as safe as merging it first, I do not remember exact name.
* Already received _security_ message that it is not safe, even if it is safe. I do not want to handle this more time.

~~* [ ] Appropriate `CHANGELOG.md` files updated for non-trivial changes~~
~~* [ ] Changes in public API reviewed (if applicable)~~

### DIFF
--- a/.github/workflows/ci-Exporter.OneCollector-Integration.yml
+++ b/.github/workflows/ci-Exporter.OneCollector-Integration.yml
@@ -1,34 +1,17 @@
-name: Integration Build OpenTelemetry.Exporter.OneCollector
+name: Live Integration Build OpenTelemetry.Exporter.OneCollector
 
 permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: false
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 on:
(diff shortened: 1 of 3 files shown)


martincostello (approved):

martincostello:
I don't even have permission to approve use of the environment this workflow uses so it's even more safe 😆

PR #4543, opened 2026-06-19
[EFCore.Tests] Drop reference to SQLite

Kielek:
Mitigate https://github.com/advisories/GHSA-2m69-gcr7-jv3q

## Changes

[EFCore.Tests] Drop reference to SQLite
Can be reverted when https://github.com/dotnet/efcore/issues/38463 is fixed

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -123,7 +123,6 @@
     <PackageVersion Include="Microsoft.AspNetCore.SignalR.Client" Version="8.0.28" />
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="8.0.28" />
     <PackageVersion Include="Microsoft.Data.SqlClient" Version="7.0.1" />
-    <PackageVersion Include="Microsoft.EntityFrameworkCore.Sqlite" Version="8.0.28" />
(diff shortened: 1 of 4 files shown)


martincostello:
Can we just comment it out rather than delete it?

Personally I think we can just suppress the NuGet audit warnings until the fix is available, especially as this is only test code (that's what I'm going to do in my own projects).

Kielek:
I was considering this, but I think the revert can be done as easy as uncommenting code.
Suppression - with the current experience, and security reports I am keen to fully remove problematic packages instead of handling later some _very important security findings_, but I will not block such solution.

martincostello:
This is what I did in one of my own projects: https://github.com/martincostello/openapi-extensions/pull/895/commits/7465d7b5d17f043874b7c0c566b788d6124c29e1

I think it's better to suppress it as:

1. We keep the test coverage we get from using SQLite for integration tests
2. It's only test code not anything we ship (and if we did, we'd only be able to wait anyway)
3. From looking at the GHSA, I don't think it's actually a problem for us as we control the queries and aren't doing anything that will hit the memory corruption
4. It's a smaller "fix" that uses a well-known mechanism for false positives

Kielek:
Replaced by https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4544

PR #4548, opened 2026-06-22
[Infra] .NET 11 preparation

martincostello:
## Changes

Cherry-pick changes out of #3867 to reduce its size:

- Fix new code analysis warnings.
- Update `HttpInListener` to handle sibling-created activities better.
- Set `SOURCE_DATE_EPOCH` during package publishing (to support NuGet's new [DeterministicTimestamp](https://github.com/dotnet/core/blob/main/release-notes/11.0/preview/preview5/nuget.md#deterministic-pack-support-is-restored) feature).

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AspNetCore/HttpContextExtensions.cs
+++ b/src/OpenTelemetry.Instrumentation.AspNetCore/HttpContextExtensions.cs
@@ -21,7 +21,6 @@ internal static class HttpContextExtensions
         var endpoint = GetOriginalEndpoint(context);
         var route = endpoint?.Metadata.GetMetadata<IRouteDiagnosticsMetadata>()?.Route;
 
-#if !NET11_0_OR_GREATER
         if (string.IsNullOrEmpty(route))
         {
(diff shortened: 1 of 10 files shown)


src/OpenTelemetry.Instrumentation.AspNetCore/HttpContextExtensions.cs
```diff
@@ -21,7 +21,6 @@ internal static class HttpContextExtensions
         var endpoint = GetOriginalEndpoint(context);
         var route = endpoint?.Metadata.GetMetadata<IRouteDiagnosticsMetadata>()?.Route;
 
-#if !NET11_0_OR_GREATER
```
Kielek, line 24: Do you know why this change is needed/was here previously?

Kielek (approved):
LGTM, one thing to double check

src/OpenTelemetry.Instrumentation.AspNetCore/HttpContextExtensions.cs
martincostello: IIRC, the previous change had this because we though this wouldn't be needed for .NET 11, but that turned out not to be the case.

Issue #1693, opened 2024-04-25
[Instrumentation.Http] - enable nullable and fix CA1508


Kielek:
Related to #894

Enable `nullable` for Instrumentation.Http and its tests.
Fix all commented out CA1508 errors.

martincostello:
Looks like this was resolved by #1941.

Kielek:
@martincostello, unfortunately no, it is still valid. It is the only class not covered in the linked PR.
See: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/f5c9b48f29b16f58f3950201b34aa2383344f47f/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs#L6

martincostello:
Ah, that wasn't clear from the commit message/PR title when I saw `#nullable` was added to the public API text files.

martincostello:
Looks like all the remaining usage is in Geneva's tests.

PR #4627, opened 2026-07-01
[Http] Enable remaining nullable

martincostello:
Contributes to #1693

## Changes

Add `#nulllable` to `HttpWebRequestActivitySource` and its tests.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
+++ b/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
@@ -64,7 +64,7 @@ internal static class HttpWebRequestActivitySource
         }
     }
 
-    internal static HttpClientTraceInstrumentationOptions TracingOptions { get; set; } = new HttpClientTraceInstrumentationOptions();
+    internal static HttpClientTraceInstrumentationOptions TracingOptions { get; set; } = new();
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static void AddRequestTagsAndInstrumentRequest(HttpWebRequest request, Activity activity)
--- a/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
+++ b/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
@@ -3,8 +3,6 @@
 
 #if NETFRAMEWORK
 
-#nullable disable
-
 using System.Collections;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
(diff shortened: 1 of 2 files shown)


src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -229,21 +221,21 @@ private static void ProcessRequest(HttpWebRequest request)
         // Eg: Parent could be the Asp.Net activity.
         InstrumentRequest(request, activityContext);
 
-        var asyncContext = writeAResultAccessor(request);
+        var asyncContext = writeAResultAccessor!(request);
```
Kielek: Do you see any possibility to reduce number of null-forgiving operator and preserve current functionalities?
I asked Codex for adivce here, proposal was to keep all reflected fields into some internal sealed Reflectionstate class, but I do not expored this option.

The number of this kind of operators you have proposed was the main reason to keep this file non-nullable friendly.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -3,8 +3,6 @@
 
 #if NETFRAMEWORK
 
-#nullable disable
```
Kielek, line 6: There is couple other places in this repos still having this note, mostly Geneva, but also Shared tests code.

martincostello: I've updated the original issue to expand the scope to the entire repo.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
martincostello: Done, but it's quite verbose IMO.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpWebRequestActivitySource.netfx.cs
```diff
@@ -86,7 +64,7 @@ static HttpWebRequestActivitySource()
         }
     }
 
-    internal static HttpClientTraceInstrumentationOptions TracingOptions { get; set; }
+    internal static HttpClientTraceInstrumentationOptions TracingOptions { get; set; } = new HttpClientTraceInstrumentationOptions();
```
martincostello:
```suggestion
    internal static HttpClientTraceInstrumentationOptions TracingOptions { get; set; } = new();
```

Kielek (approved):

PR #4629, opened 2026-07-01
[Extensions] Add consistent probability sampler

### LINKED ISSUE #3678: Implement consistent probability sampler.
# Feature Request

Implement consistent probability sampler as per spec:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md

The spec is experimental, so this should be kept as a separate non-stable package, and merged to the core SDK once stable.

martincostello:
Fixes #3678

## Changes

Add an implementation of a consistent probability sampler ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#converting-floating-point-probability-to-threshold-value)).

## Benchmarks

_Summarised by Claude 🐙_

### Analysis

All four scenarios measure the per-span cost of a single `ShouldSample` call with the sampler configured at 25%. The operations are tens-to-hundreds of nanoseconds and allocate only small, bounded amounts; the allocation figures are the most stable and meaningful signal (the net10.0 timings have wide error bars because the machine was noisier during that run).

**By scenario:**

- **`RandomTraceId` is the cheapest path on both runtimes.** No `rv` is generated or written and no other tracestate members need preserving, so the outgoing tracestate is just `ot=th:<x>`. On net462 it allocates **0 B** — the trace-id randomness is read via a span slice, the threshold is formatted without an intermediate string, and there are no lists to allocate. On net10.0 it still shows 168 B solely because `ActivityTraceId.ToHexString()` allocates a 32-char string (there is no public byte-level accessor to avoid it).
- **`RootSpan` (baseline) and `ExplicitRandomValue`** are dominated by building the outgoing `ot=th:<x>;rv:<y>` string — the `StringBuilder` plus the final serialized string, which is unavoidable since it is the return value. `ExplicitRandomValue` is slightly cheaper/leaner than the root case because it reuses the inherited `rv` instead of generating one.
- **`ExplicitRandomValueWithOtherMembers` is the most expensive** (~1.3× time and ~2× allocations vs. baseline). The two extra vendor members must be parsed into preserved strings and re-emitted. This is exactly the lazy-list code path that allocates **only when extras are present** — the common cases above pay nothing for it.

**Runtime differences:**

- net10.0 uses the more efficient `#if NET` paths (`Random.Shared.NextInt64`, `long.TryFormat` into stack buffers, `stackalloc` spans), giving lower allocations on the string-building paths than net462, which falls back to `NextBytes` + `ToString`.
- net462 is slower in absolute terms on the allocating paths but tighter/more stable in its measurements, and confirms the zero-allocation `RandomTraceId` path holds on the fallback code.

**Takeaways:**

- The struct + lazy-list + span/`TryFormat` work pays off: the hot paths allocate one small string (the required output) or nothing, and extra-member preservation is pay-for-what-you-use.
- The remaining `RootSpan`/`ExplicitRandomValue` allocations are the serialized tracestate string itself (intrinsic to the sampler's contract) plus, on net10.0, the `TraceId.ToHexString()` string on the trace-id path.

<details>

<summary>Expand to view</summary>

#### net10.0

```
BenchmarkDotNet v0.15.8, Windows 11 (10.0.26200.8655/25H2/2025Update/HudsonValley2)
13th Gen Intel Core i7-13700H 2.90GHz, 1 CPU, 20 logical and 14 physical cores
.NET SDK 10.0.301
  [Host]     : .NET 10.0.9 (10.0.9, 10.0.926.27113), X64 RyuJIT x86-64-v3
  DefaultJob : .NET 10.0.9 (10.0.9, 10.0.926.27113), X64 RyuJIT x86-64-v3
```

| Method                              | Mean      | Error    | StdDev    | Median    | Ratio | RatioSD | Allocated | Alloc Ratio |
|------------------------------------ |----------:|---------:|----------:|----------:|------:|--------:|----------:|------------:|
| RootSpan                            | 120.68 ns | 8.335 ns | 24.181 ns | 111.66 ns |  1.04 |    0.28 |     280 B |        1.00 |
| ExplicitRandomValue                 |  81.04 ns | 3.933 ns | 10.897 ns |  76.17 ns |  0.70 |    0.15 |     272 B |        0.97 |
| RandomTraceId                       |  57.12 ns | 3.308 ns |  9.598 ns |  53.74 ns |  0.49 |    0.12 |     168 B |        0.60 |
| ExplicitRandomValueWithOtherMembers | 154.03 ns | 6.455 ns | 18.520 ns | 146.02 ns |  1.32 |    0.28 |     648 B |        2.31 |

#### net462

```
BenchmarkDotNet v0.15.8, Windows 11 (10.0.26200.8655/25H2/2025Update/HudsonValley2)
13th Gen Intel Core i7-13700H 2.90GHz, 1 CPU, 20 logical and 14 physical cores
  [Host]     : .NET Framework 4.8.1 (4.8.9325.0), X64 RyuJIT VectorSize=256
  DefaultJob : .NET Framework 4.8.1 (4.8.9325.0), X64 RyuJIT VectorSize=256
```

| Method                              | Mean      | Error     | StdDev    | Median    | Ratio | RatioSD | Allocated | Alloc Ratio |
|------------------------------------ |----------:|----------:|----------:|----------:|------:|--------:|----------:|------------:|
| RootSpan                            | 229.91 ns |  3.404 ns |  2.842 ns | 229.88 ns |  1.00 |    0.02 |     401 B |        1.00 |
| ExplicitRandomValue                 | 251.62 ns |  4.835 ns |  7.944 ns | 249.63 ns |  1.09 |    0.04 |     337 B |        0.84 |
| RandomTraceId                       |  68.57 ns |  3.577 ns | 10.492 ns |  63.55 ns |  0.30 |    0.05 |         - |        0.00 |
| ExplicitRandomValueWithOtherMembers | 449.76 ns | 14.388 ns | 40.346 ns | 434.78 ns |  1.96 |    0.18 |     738 B |        1.84 |

</details>


src/OpenTelemetry.Extensions/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 * Updated OpenTelemetry core component version(s) to `1.16.0`.
   ([#4487](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4487))
 
+* Added `ConsistentProbabilitySampler`, a consistent probability based sampler
+  implementing the OpenTelemetry
+  [probability sampling](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/)
+  specification.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#4629](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4629))
```

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
martincostello, line 1: Given that the package is beta and that the new public API exposed here is effectively an implementation of a stable abstract class and otherwise only accepts a `double` value to set the probability for sampling, this is OK to not be experimental?

If the spec were to change that would just be internal implementation details and we'd just ship a new version right?

src/OpenTelemetry.Extensions/Internal/ConsistentProbability.cs
```diff
@@ -0,0 +1,271 @@
...
+    /// </summary>
+    /// <param name="threshold">The rejection threshold, in the range <c>[0, 2^56)</c>.</param>
+    /// <returns>
+    /// The adjusted count.
+    /// </returns>
+    /// <remarks>
+    /// Per the specification: <c>AdjustedCount = MaxAdjustedCount / (MaxAdjustedCount - Threshold)</c>.
+    /// </remarks>
+    public static double ThresholdToAdjustedCount(long threshold)
+        => (double)MaxAdjustedCount / (MaxAdjustedCount - threshold);
```
matt-hensley, line 220: Collector implementation uses a slightly different calculation:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b86ed1f36e28b7a50fb9a9bab97cd7ea8810b1a6/pkg/sampling/threshold.go#L139-L144

src/OpenTelemetry.Extensions/Internal/ConsistentProbability.cs
```diff
@@ -0,0 +1,271 @@
...
+            return "0";
+        }
+
+        // math.frexp(probability): probability is a positive value less than one, so the
+        // exponent is less than or equal to zero.
+        var exponent = FrexpExponent(probability);
+
+        // Raise the precision by the number of leading 'f' digits so the configured precision
+        // applies to the significant digits of the threshold for probabilities near zero.
+        precision = Math.Max(1, Math.Min(13, precision + ((-exponent) / 4)));
```
matt-hensley: This implementation matches the pseudocode in the spec, but the Collector deviates for some reason:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/89e43555904cd97c2d36605347c5d5237b1bdc8c/pkg/sampling/probability.go#L58-L60

matt-hensley (commented):
Spotted some minor differences from the Collector implementation that could cause minor interoperability issues

src/OpenTelemetry.Extensions/Internal/ConsistentProbability.cs
martincostello: I asked Claude to investigate this, and it's assessment is that this code is equivalent to the collector's implementation.

> The Collector computes `1.0 / Probability()`, i.e. `1 / ((MaxAdjustedCount - t) / MaxAdjustedCount)`, which reduces to `MaxAdjustedCount / (MaxAdjustedCount - t)` — exactly what this method computes directly. Doing it in one division rather than a division-then-reciprocal avoids an intermediate rounding, so if anything this is very slightly more accurate.
> 
> The one behavioural difference is the Collector's `NeverSampleThreshold → 0` guard. This implementation doesn't model a "never sample" threshold (probabilities are validated to `[2^-56, 1]`, so `MaxAdjustedCount - t` is always `>= 1` and never divides by zero), and this helper isn't on the sampler's hot path — it's only used to validate the spec's adjusted-count table in the tests. So I don't think a change is needed here.

src/OpenTelemetry.Extensions/Internal/ConsistentProbability.cs
martincostello: Claude's assessment on this is that there's a small discrepancy.

> You're right that the Collector deviates from the spec pseudocode here. This method is a deliberate port of the reference `probability_to_threshold_with_precision` algorithm, which the spec labels as "examples only for clarity, and not as a suggested implementation" and which explicitly points at the Collector's `pkg/sampling` package for the exact-integer method. The Collector differs in two intentional ways:
> 
> 1. It raises precision near probability `1` using `frexp(1 - fraction)`. The spec deliberately omits this ("there is not a symmetrical adjustment made for values near unit probability, as we do not believe there is a practical use for sampling very precisely near 100%").
> 2. It computes the threshold as an exact 56-bit integer (`round(prob * 2^56)` + bit-shift rounding) rather than via the IEEE-754 float mantissa, which is exact at the extremes where the float method loses precision.
> 
> So for the same input the encoded `th` can differ, e.g. (confirmed against this branch):
> 
> | input | here | Collector |
> | --- | --- | --- |
> | `0.9999` @ precision 4 | `0007` (≈0.99989) | ~7 digits (≈0.9999) |
> | `2^-56` @ precision 4 | `fffffffffffff` (≈2^-52) | `ffffffffffffff` (exact 2^-56) |
> 
> Importantly this doesn't affect sampling decisions or cross-participant consistency — those depend on propagating the `th`/`rv` values and comparing `R >= T`, not on two independent encoders agreeing on the same `th` — and the sampler always encodes at the spec's default precision of 4, so the differences are confined to probabilities extremely close to 0 or within ~1/16 of 1.
> 
> That said, digging into this I did spot a genuine nit in this line: the clamp should be `min(12, …)` to match the reference algorithm (the spec caps at 12 to keep a rounding digit within the 52-bit mantissa), whereas it currently reads `min(13, …)`, which matches neither the spec (12) nor the Collector (14).

Going up to 14 to match the collector looks like a better option than going down to 12 to match the spec for both consistency and to get an exact value for `p=2^56`.

martincostello: Updated and added additional tests ported from the collector's.

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
alanwest: Yes, since the package is not stable, we do not need to do the experimental API dance.

alanwest (approved):
I did not do a very close review of the algorithm, but this looks good to me.

A few non-blocking follow up topics for continued discussion

1. I think this is fine in OpenTelemetry.Extensions for an initial release, but we need to figure out a low friction way to expose experimental things in the actual SDK package where this ultimately will belong. We have some patterns, but maybe they're not meeting our needs?
2. In a recent SIG meeting we discussed the benefits of a cross language test that validates interoperability across SDKs. I don't think this is for our SIG alone to solve.

matt-hensley (approved):

Kielek:
I have sent some AI scan results to Martin in private channel. It needs to be reviewed before mergin.

Kielek (approved):
I pushed some feedback directly to this PR. With this LGTM.

Please double check before merge.

martincostello:
> Please double check before merge.

LGTM. I pulled through the `char` extensions/polyfills from the SDK repo and used them where relevant. When the code is ported to the SDK it will benefit automatically.

I'll look to see if there's other places in this repo that can benefit from the new polyfills separately.

PR #4633, opened 2026-07-02
[OpAMP] Add full state report

RassK:
Towards #4634

## Changes

Adds a response option after receiving `FlagsMessage` with `ReportFullState`.

### DIFF
--- a/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
@@ -145,19 +145,19 @@ internal sealed class FrameDispatcher : IDisposable
                 .AddCapabilities();
 
             // TODO: Add here features when they become available and are necessary to restore the full state in the server if requested.
-            if (report.EffectiveConfigFiles != null)
+            if (report.EffectiveConfigFiles is { } effectiveConfig)
             {
-                message.AddEffectiveConfig(report.EffectiveConfigFiles);
+                message.AddEffectiveConfig(effectiveConfig);
             }
 
-            if (report.RemoteConfigStatus != null)
+            if (report.RemoteConfigStatus is { } remoteConfigStatus)
             {
-                message.AddRemoteConfigStatus(report.RemoteConfigStatus);
+                message.AddRemoteConfigStatus(remoteConfigStatus);
             }
 
-            if (report.CustomCapabilities != null)
+            if (report.CustomCapabilities is { } customCapabilities)
             {
-                message.AddCustomCapabilities(report.CustomCapabilities);
+                message.AddCustomCapabilities(customCapabilities);
             }
 
             return message.Build();
--- a/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
@@ -6,6 +6,7 @@ using OpenTelemetry.Internal;
 using OpenTelemetry.OpAmp.Client.Internal.Services.Heartbeat;
 using OpenTelemetry.OpAmp.Client.Internal.Transport;
 using OpenTelemetry.OpAmp.Client.Messages;
+using OpenTelemetry.OpAmp.Client.Messages.Flags;
 using OpenTelemetry.OpAmp.Client.Settings;
 
 namespace OpenTelemetry.OpAmp.Client.Internal;
@@ -125,6 +126,41 @@ internal sealed class FrameDispatcher : IDisposable
         AgentToServer BuildCustomMessageMessage(FrameBuilder fb)
         {
             return fb.StartBaseMessage().AddCustomMessage(capability, type, data).Build();
+        }
+    }
+
+    public async Task DispatchFullStateReportAsync(FullStateReport report, CancellationToken token)
+    {
+        await this.DispatchFrameAsync(
+            BuildFullStateReportMessage,
+            OpAmpClientEventSource.Log.SendingFullStateReportMessage,
+            OpAmpClientEventSource.Log.SendFullStateReportMessageException,
+            token).ConfigureAwait(false);
+
+        AgentToServer BuildFullStateReportMessage(FrameBuilder fb)
+        {
+            // Start message with basic partials.
+            var message = fb.StartBaseMessage()
+                .AddAgentDescription()
+                .AddCapabilities();
+
+            // TODO: Add here features when they become available and are necessary to restore the full state in the server if requested.
+            if (report.EffectiveConfigFiles != null)
+            {
+                message.AddEffectiveConfig(report.EffectiveConfigFiles);
+            }
+
+            if (report.RemoteConfigStatus != null)
+            {
+                message.AddRemoteConfigStatus(report.RemoteConfigStatus);
+            }
+
+            if (report.CustomCapabilities != null)
+            {
+                message.AddCustomCapabilities(report.CustomCapabilities);
+            }
+
+            return message.Build();
         }
     }
 
--- a/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
+++ b/src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
@@ -145,6 +145,7 @@ internal sealed class FrameDispatcher : IDisposable
                 .AddCapabilities();
 
             // TODO: Add here features when they become available and are necessary to restore the full state in the server if requested.
+            // See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/4634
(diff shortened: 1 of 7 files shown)


src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
```diff
@@ -128,6 +129,41 @@ AgentToServer BuildCustomMessageMessage(FrameBuilder fb)
...
+            token).ConfigureAwait(false);
+
+        AgentToServer BuildFullStateReportMessage(FrameBuilder fb)
+        {
+            // Start message with basic partials.
+            var message = fb.StartBaseMessage()
+                .AddAgentDescription()
+                .AddCapabilities();
+
+            // TODO: Add here features when they become available and are necessary to restore the full state in the server if requested.
```
martincostello, line 146: Is there a tracking issue for this? If not, I would remove the `TODO` or create one and link to it.

src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
```diff
@@ -128,6 +129,41 @@ AgentToServer BuildCustomMessageMessage(FrameBuilder fb)
...
+
+            if (report.RemoteConfigStatus != null)
+            {
+                message.AddRemoteConfigStatus(report.RemoteConfigStatus);
+            }
+
+            if (report.CustomCapabilities != null)
+            {
+                message.AddCustomCapabilities(report.CustomCapabilities);
+            }
```
martincostello: Could simplify these to this style:

```csharp
if (report.Foo is { } foo)
{
    message.AddFoo(foo);
}
```

src/OpenTelemetry.OpAmp.Client/Messages/Flags/FullStateReport.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Messages.Flags;
+
+/// <summary>
+/// Contains all parts that are necessary to restore the full state in the server.
+/// </summary>
+public class FullStateReport
```
martincostello: `sealed`?

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
```diff
@@ -205,6 +206,20 @@ public Task SendCustomMessageAsync(string capability, string type, ReadOnlyMemor
         return this.dispatcher.DispatchCustomMessageAsync(capability, type, data, cancellationToken);
     }
 
+    /// <summary>
+    /// Sends full state report message to restore the lost state in the server.
```
martincostello:
```suggestion
    /// Sends a full state report message to restore the lost state in the server.
```

src/OpenTelemetry.OpAmp.Client/Internal/FrameDispatcher.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4633/commits/ae594f57152dfb3d871edc2cf9e8f76315685628

martincostello (approved):

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
```diff
@@ -205,6 +206,20 @@ public Task SendCustomMessageAsync(string capability, string type, ReadOnlyMemor
...
+    /// </summary>
+    /// <param name="report">Report that contains supported partials.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A task that represents the asynchronous send operation.</returns>
+    /// <exception cref="ObjectDisposedException">Thrown if the client has already been disposed.</exception>
+    public Task SendFullStateReportAsync(FullStateReport report, CancellationToken cancellationToken = default)
+    {
+        this.ThrowIfDisposed();
+
+        return this.dispatcher.DispatchFullStateReportAsync(report, cancellationToken);
```
lachmatt, line 230: should settings be checked before dispatching the report, in a manner similar to e.g. `SendEffectiveConfigAsync`? and `report` checked for null?

RassK: `OpAmpClientSettings` live in the client, so the client settings state is not lost.
The logic is similar to `DispatchIdentificationAsync`. No check there.

src/OpenTelemetry.OpAmp.Client/Messages/Flags/FullStateReport.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.OpAmp.Client.Messages.Flags;
```
stevejgordon: Is this the correct namespace?
```suggestion
namespace OpenTelemetry.OpAmp.Client.Messages;
```

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
```diff
@@ -205,6 +206,20 @@ public Task SendCustomMessageAsync(string capability, string type, ReadOnlyMemor
...
 
+    /// <summary>
+    /// Sends a full state report message to restore the lost state in the server.
+    /// </summary>
+    /// <param name="report">Report that contains supported partials.</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>A task that represents the asynchronous send operation.</returns>
+    /// <exception cref="ObjectDisposedException">Thrown if the client has already been disposed.</exception>
+    public Task SendFullStateReportAsync(FullStateReport report, CancellationToken cancellationToken = default)
+    {
```
stevejgordon, line 216: Should we guard against `FullStateReport` being null here?

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
```diff
@@ -205,6 +206,20 @@ public Task SendCustomMessageAsync(string capability, string type, ReadOnlyMemor
...
+        this.ThrowIfDisposed();
+
```
stevejgordon, line 229: Should we validate the `FullStateReport` only provides fields for capabilities that are enabled, in the same way as we do in the other methods? e.g.:

```c#
if (report.EffectiveConfigFiles != null && !this.settings.EffectiveConfigurationReporting.EnableReporting)
{
    throw new InvalidOperationException("Effective configuration reporting is not enabled in settings.");
}

if (report.RemoteConfigStatus != null && !this.settings.RemoteConfiguration.ReportsRemoteConfigStatus)
{
    throw new InvalidOperationException("Remote configuration status reporting is not enabled in settings.");
}
```

stevejgordon (changes requested):
A couple of consistency items.

Also, `FullStateReport` is currently in the `Messages/Flags` directory which seems wrong. Shouldn't this have a distinct directory or be in the the root `Messages` directory?

src/OpenTelemetry.OpAmp.Client/OpAmpClient.cs
stevejgordon: I raised the same point in a [comment](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4633#discussion_r3513244464) further down before spotting this thread, sorry for the near-duplicate: 

On the `DispatchIdentificationAsync` comparison; that one only ever sends `AgentDescription` + `Capabilities`, both unconditional, so there's no caller-supplied data that could disagree with what's actually enabled. `FullStateReport` is different since the caller can hand us `RemoteConfigStatus`/`EffectiveConfigFiles` regardless of whether `ReportsRemoteConfigStatus`/`EffectiveConfigurationReporting.EnableReporting` are on, so we could end up putting e.g. `remote_config_status` on the wire in a message whose capability bit says we don't support it. Feels worth guarding the same way `SendEffectiveConfigAsync`/`SendRemoteConfigStatusAsync` already do.

stevejgordon (commented):
My only remaining feedback is whether the `FullStateReport` should live in the flags directory. My view is we should bring it a level for consistency, or put it in its own subdirectory under messages.

RassK:
> My only remaining feedback is whether the `FullStateReport` should live in the flags directory. My view is we should bring it a level for consistency, or put it in its own subdirectory under messages.

It's in the Flags dir (virtual dir, correct namespace is root - as others are) due the other ServerSentFlags is connected to the response. Other messages are similar or have even weaker connection like effective config and remote config + answers.

lachmatt (approved):

stevejgordon (approved):

PR #4636, opened 2026-07-02
[ConfluentKafka] Update Semantic Conventions

### LINKED ISSUE #2567: [feature request] Kafka instrumentation should follow the Semantic Conventions

martincostello:
- Fixes #2567
- Contributes to #4064

## Changes

Update to [v1.43.0](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/messaging/kafka.md) of the semantic conventions for messaging.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
@@ -28,6 +28,9 @@
     `messaging.kafka.destination.partition`)
   * `messaging.kafka.offset` (replacing `messaging.kafka.message.offset`)
 
+  `messaging.kafka.message.key` is now emitted as a string when the key has an
+  unambiguous canonical representation; otherwise, the attribute is omitted.
+
   The metrics were renamed to:
 
   * `messaging.client.operation.duration`
--- a/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
@@ -31,10 +31,10 @@
       (replacing `messaging.receive.messages`).
   `error.type` is now reported as the language-agnostic Kafka error code and
   the exception message is recorded in the span status description.
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+  ([#4636](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4636))
 
 * Metrics and traces now include a telemetry schema URL.
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+  ([#4636](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4636))
 
 ## 0.1.0-alpha.7
 
--- a/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
@@ -15,20 +15,24 @@
   The producer `send` span (previously `publish`) and the consumer `poll`
   span (previously `receive`, and now emitted with `ActivityKind.Client`) now
   set:
-    * `messaging.operation.name` and `messaging.operation.type`
-      (replacing `messaging.operation`)
-    * `messaging.client.id` (replacing `messaging.client_id`)
-    * `messaging.consumer.group.name` (replacing
-      `messaging.kafka.consumer.group`)
-    * `messaging.destination.partition.id` (a string) (replacing
-      `messaging.kafka.destination.partition`)
-    * `messaging.kafka.offset` (replacing `messaging.kafka.message.offset`)
+
+  * `messaging.operation.name` and `messaging.operation.type`
+    (replacing `messaging.operation`)
+  * `messaging.client.id` (replacing `messaging.client_id`)
+  * `messaging.consumer.group.name` (replacing
+    `messaging.kafka.consumer.group`)
+  * `messaging.destination.partition.id` (a string) (replacing
+    `messaging.kafka.destination.partition`)
+  * `messaging.kafka.offset` (replacing `messaging.kafka.message.offset`)
+
   The metrics were renamed to:
+
     * `messaging.client.operation.duration`
       (replacing `messaging.publish.duration` and `messaging.receive.duration`)
     * `messaging.client.sent.messages` (replacing `messaging.publish.messages`)
     * `messaging.client.consumed.messages`
       (replacing `messaging.receive.messages`).
+
   `error.type` is now reported as the language-agnostic Kafka error code and
   the exception message is recorded in the span status description.
   ([#4636](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4636))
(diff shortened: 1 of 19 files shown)


src/OpenTelemetry.Instrumentation.ConfluentKafka/CHANGELOG.md
```diff
@@ -9,6 +9,33 @@
...
+      (replacing `messaging.publish.duration` and `messaging.receive.duration`)
+    * `messaging.client.sent.messages` (replacing `messaging.publish.messages`)
+    * `messaging.client.consumed.messages`
+      (replacing `messaging.receive.messages`).
+  `error.type` is now reported as the language-agnostic Kafka error code and
+  the exception message is recorded in the span status description.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
+* Metrics and traces now include a telemetry schema URL.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#4636](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4636))

* Metrics and traces now include a telemetry schema URL.
  ([#4636](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4636))
```

g7ed6e (approved):
LGTM

Kielek:
Asked Codex/GPT5.6 Sol:|

[P1] Preserve the legacy default and implement OTEL_SEMCONV_STABILITY_OPT_IN.
[ConfluentKafkaCommon.cs](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/ConfluentKafkaCommon.cs#L32-L51) unconditionally switches traces and metrics to v1.42.0. The component previously used v1.24-era conventions, but v1.42 explicitly says such instrumentations should retain their old default and support messaging and messaging/dup opt-ins, with messaging/dup taking precedence. [Messaging migration guidance](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/messaging/kafka.md#semantic-conventions-for-kafka)
Without this, existing users receive an immediate breaking change and cannot run old and new telemetry in parallel. The selection needs to cover span names/kinds, attributes, metric instruments, and schema URLs.

My comment: I am fine with upgrading if before we were in complete mess, otherwise the switch should be implemented.

[P2] Provide sampling-sensitive attributes when starting each activity.
The producer [starts the activity](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedProducer.cs#L318-L346) before setting messaging.system, destination, partition, and operation attributes. The [poll span](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs#L330-L347) and [process span](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/OpenTelemetryConsumeResultExtensions.cs#L140-L159) do the same.
Consequently, custom samplers cannot inspect these attributes. Kafka v1.42 says messaging.system should be present at creation and specifically lists group, destination, partition, and operation attributes as sampling-sensitive. [Kafka span attributes](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/messaging/kafka.md#span-attributes)
Build an initial ActivityTagsCollection and pass it to StartActivity.

[P2] Do not increment consumed.messages when no record was received.
[RecordReceive](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs#L279-L284) increments the counter for every ConsumeException, including exceptions whose ConsumerRecord is null. Those failures did not pull or deliver a message, but the v1.42 counter specifically counts messages pulled from the broker or dispatched to the application. [Consumed-messages definition](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/messaging/messaging-metrics.md#metric-messagingclientconsumedmessages)
Keep recording operation duration with error.type, but only increment the counter when the result or exception actually contains a record. This overlaps an existing Copilot thread, although that thread is too broad: failed operations containing a pulled record may still legitimately increment the counter.

[P2] Emit messaging.kafka.message.key as a string, not the generic key object.
Producer, poll, and process spans pass the generic key directly—for example [here](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/cd7de8568f26a390b8191f8132ad2b09f75784ec/src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs#L349-L352). Some tests explicitly expect a raw byte[].
V1.42 defines this attribute as a string and requires a canonical string representation for non-string keys; if none exists, the attribute must be omitted. [Kafka key requirements](https://github.com/open-telemetry/semantic-conventions/blob/v1.42.0/docs/messaging/kafka.md#span-attributes) A shared normalization helper should be used consistently across all three span types.

martincostello:
For the opt-in, I intentionally ignored it because the package is only alpha.

martincostello:
Feedback should have been addressed now.

Kielek (approved):
LGTM with some minor changes.

I think README should be updated. Typically we list all metrics produced by the package. It can be separate PR.

src/OpenTelemetry.Instrumentation.ConfluentKafka/InstrumentedConsumer.cs
```diff
@@ -293,35 +325,81 @@ private void InstrumentConsumption(DateTimeOffset startTime, DateTimeOffset endT
...
             activity.SetTag(SemanticConventions.AttributeMessagingClientId, this.Name);
-            activity.SetTag(SemanticConventions.AttributeMessagingDestinationName, topicPartitionOffset?.Topic);
-            activity.SetTag(SemanticConventions.AttributeMessagingKafkaDestinationPartition, topicPartitionOffset?.Partition.Value);
-            activity.SetTag(SemanticConventions.AttributeMessagingKafkaMessageOffset, topicPartitionOffset?.Offset.Value);
-            activity.SetTag(SemanticConventions.AttributeMessagingKafkaConsumerGroup, this.GroupId);
-            activity.SetTag(SemanticConventions.AttributeMessagingOperation, ConfluentKafkaCommon.ReceiveOperationName);
-            if (key != null)
+            activity.SetTag(SemanticConventions.AttributeMessagingKafkaOffset, topicPartitionOffset?.Offset.Value);
+
+            if (ConfluentKafkaCommon.FormatMessageKey(key) is { } messageKey)
```
martincostello, line 395: Is an empty message key valid? Asking as the check on the count was intentional.

Kielek: From: https://github.com/open-telemetry/semantic-conventions/blob/89aae438b3b3b0a8dd33003c9d70592baf7dbd0d/docs/messaging/kafka.md#L119

Message keys in Kafka are used for grouping alike messages to ensure they're processed on the same partition. They differ from messaging.message.id in that they're not unique. If the key is null, the attribute MUST NOT be set. 
: If the key type is not string, it's string representation has to be supplied for the attribute. If the key has no unambiguous, canonical string form, don't include its value.

No word about droping empyt values. I was double checking this before commit.

martincostello:
> I think README should be updated. Typically we list all metrics produced by the package. It can be separate PR.

#4674

PR #4641, opened 2026-07-02
[Hangfire] Use ActivitySource and Meter factories

martincostello:
Contributes to #4064 and #4068.

## Changes

Use `MeterFactory` and `ActivitySourceFactory`. The Hangfire instrumentation doesn't follow any of the semantic conventions, so this just changes how the `ActivitySource` and `Meter` are created.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetrics.cs
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetrics.cs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics.Metrics;
-using OpenTelemetry.Internal;
 
 namespace OpenTelemetry.Instrumentation.Hangfire.Implementation;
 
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.Hangfire/Implementation/HangfireMetrics.cs
```diff
@@ -84,5 +83,5 @@ internal static class HangfireMetrics
     /// <summary>
     /// Gets the meter instance for all Hangfire metrics.
     /// </summary>
-    private static Meter Meter => field ??= new(typeof(HangfireMetrics).Assembly.GetName().Name, typeof(HangfireMetrics).Assembly.GetPackageVersion());
+    private static Meter Meter => field ??= Metrics.MeterFactory.Create(typeof(HangfireMetrics), null); // These metrics are not in the Semantic Conventions
```
Kielek, line 86: IMO worth to switch to 
```suggestion
    private static Meter Meter => field ??= Metrics.MeterFactory.Create<HangfireMetrics>(null); // These metrics are not in the Semantic Conventions
```

martincostello: Can't do that here because the class is static,

Kielek: In fact, it can be any type from this project. Not a blocker.

Kielek (approved):

PR #4657, opened 2026-07-07
[Exporter.InfluxDB] Apply backpressure via ExportResult.Failure instead of unbounded buffering

malessainray:
## Fixes

Addresses the unbounded memory growth described in #4474 (closed as inactive).

## Changes

The InfluxDB metrics exporter writes points fire-and-forget into `WriteApi`. Its internal Rx buffer is [documented as unbounded and without backpressure in the C# client](https://github.com/influxdata/influxdb-client-csharp/blob/master/Client/WriteApi.cs), and write failures are only surfaced asynchronously via the event stream. As a result, when InfluxDB is down or cannot keep up:

- the buffer grows until the process is OOM-killed, and
- `Export()` always returns `ExportResult.Success`, so the SDK never throttles.

This PR implements the lightweight approach suggested during the review of #4474, instead of adding a custom bounded queue / background worker / new public API:

- Track the outcome of the asynchronous writes via the existing `WriteApi.EventHandler` (`WriteSuccessEvent` → healthy, `WriteErrorEvent` / `WriteRetriableErrorEvent` → unhealthy).
- While writes are failing, `Export()` stops enqueuing new points into the unbounded buffer and returns `ExportResult.Failure`.

Because the reader uses `MetricReaderTemporalityPreference.Cumulative`, returning `Failure` loses no data: the SDK keeps the accumulated values and the `PeriodicExportingMetricReader` retries on the next export cycle. This turns the reader into the natural backpressure/retry mechanism. Memory is then bounded by cardinality (and `MaxMetricPointsPerMetricStream`), not by the duration of the outage — so even a permanently unreachable InfluxDB no longer causes unbounded growth.

Trade-off: during an outage the per-interval resolution is dropped, but cumulative totals remain correct once writes resume (counters recover their full total, gauges/histograms keep their latest cumulative state).

- [x] CHANGELOG.md updated for non-trivial changes
- [x] Changes in public API reviewed (if applicable) — no public API changes

### DIFF
--- a/src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+* The metrics exporter now reports `ExportResult.Failure` when asynchronous
+  writes to InfluxDB are failing and stops enqueuing points into the unbounded
+  `WriteApi` buffer while the backend is unreachable. This prevents unbounded
+  memory growth (potential OOM) when InfluxDB is down or falling behind. Since
+  the reader uses cumulative temporality, no data is lost: the SDK retains the
+  accumulated values and the `PeriodicExportingMetricReader` retries on the next
+  export cycle.
+  ([#4474](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4474))
+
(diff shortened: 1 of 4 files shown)


martincostello:
@malessainray Please sign the CLA.

malessainray:
@martincostello I accidentally committed with the wrong account. Will redo a new PR

PR #4658, opened 2026-07-07
[Exporter.InfluxDB] Apply backpressure via ExportResult.Failure instead of unbounded buffering

Xyolyp:
## Fixes

Addresses the unbounded memory growth described in #4474 (closed as inactive).

## Changes

The InfluxDB metrics exporter writes points fire-and-forget into `WriteApi`. Its internal Rx buffer is [documented as unbounded and without backpressure in the C# client](https://github.com/influxdata/influxdb-client-csharp/blob/master/Client/WriteApi.cs), and write failures are only surfaced asynchronously via the event stream. As a result, when InfluxDB is down or cannot keep up:

- the buffer grows until the process is OOM-killed, and
- `Export()` always returns `ExportResult.Success`, so the SDK never throttles.

This PR implements the lightweight approach suggested during the review of #4474, instead of adding a custom bounded queue / background worker / new public API:

- Track the outcome of the asynchronous writes via the existing `WriteApi.EventHandler` (`WriteSuccessEvent` → healthy, `WriteErrorEvent` / `WriteRetriableErrorEvent` → unhealthy).
- While writes are failing, `Export()` stops enqueuing new points and returns `ExportResult.Failure`, but keeps **one** probe batch in flight so that a subsequent `WriteSuccessEvent` can signal recovery. This bounds the buffer to at most one outstanding batch per export cycle while still resuming automatically once InfluxDB is reachable again.

Because the reader uses `MetricReaderTemporalityPreference.Cumulative`, returning `Failure` loses no data: the SDK keeps the accumulated values and the `PeriodicExportingMetricReader` retries on the next export cycle, acting as the natural backpressure/retry mechanism. Memory is then bounded by cardinality (and `MaxMetricPointsPerMetricStream`), not by the duration of the outage — so even a permanently unreachable InfluxDB no longer causes unbounded growth.

Trade-off: during an outage the per-interval resolution is dropped, but cumulative totals remain correct once writes resume (counters recover their full total, gauges/histograms keep their latest cumulative state).

## Tests

Added following the existing `InfluxDBFakeServer` pattern (full `MeterProvider` pipeline driven via `ForceFlush`):

- `ExportReportsFailureWhenInfluxDBIsUnavailable` — the backend rejects writes; `Export()` starts reporting `Failure`.
- `ExportRecoversWhenInfluxDBBecomesAvailableAgain` — after the backend recovers, `Export()` reports `Success` again and data flows through.

`InfluxDBFakeServer` gained a configurable `ResponseStatusCode` (default `OK`, so existing tests are unaffected). All 22 tests in the project pass.

- [x] CHANGELOG.md updated for non-trivial changes
- [x] Unit tests added
- [x] Changes in public API reviewed (if applicable) — no public API changes


martincostello:
@Xyolyp Please sign the CLA.

src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+* The metrics exporter now reports `ExportResult.Failure` when asynchronous
+  writes to InfluxDB are failing and stops enqueuing points into the unbounded
+  `WriteApi` buffer while the backend is unreachable. This prevents unbounded
+  memory growth (potential OOM) when InfluxDB is down or falling behind. Since
+  the reader uses cumulative temporality, no data is lost: the SDK retains the
+  accumulated values and the `PeriodicExportingMetricReader` retries on the next
+  export cycle.
+  ([#4474](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4474))
```
martincostello:
```suggestion
  ([#4658](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4658))
```

src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+* The metrics exporter now reports `ExportResult.Failure` when asynchronous
+  writes to InfluxDB are failing and stops enqueuing points into the unbounded
+  `WriteApi` buffer while the backend is unreachable. This prevents unbounded
+  memory growth (potential OOM) when InfluxDB is down or falling behind. Since
+  the reader uses cumulative temporality, no data is lost: the SDK retains the
+  accumulated values and the `PeriodicExportingMetricReader` retries on the next
+  export cycle.
```
martincostello: This is overly verbose. Just a summary of the change like in the PR title is fine.

Xyolyp:
Thanks for the review! I've shortened the CHANGELOG entry to a title-style summary and fixed the PR reference to #4658.

Kielek:
@Havret, could you please check this one?

Havret:
@Kielek I could. Will do.

Havret (commented):
Hi @Xyolyp - first off, thank you for picking this up and putting a PR together. This is a real production issue that had been sitting inactive, and it's genuinely appreciated that you took the time to tackle it. 

I wanted to be transparent, though, because the implementation here has drifted a bit from what I'd suggested in this comment on #4474.

The core of that suggestion was to let write failures propagate directly into Export()'s return value - which, in practice, means writing synchronously so that a failed write simply surfaces as ExportResult.Failure.

This PR instead keeps the existing fire-and-forget WriteApi path and layers health-tracking on top of it (the async EventHandler, plus writeApiHealthy / writeInFlight flags). It's a reasonable interpretation, but keeping the asynchronous buffer in the loop brings back some of the tricky edge cases the synchronous approach avoids entirely.

I've gone ahead and put together an alternative that follows the original suggestion (synchronous writes via IWriteApiAsync, Failure straight through on error, plus failure/recovery tests): #4861. Given the overlap, my recommendation would be to close this PR and continue with that one - but I'd love your review and thoughts on it, and of course credit is due to you for driving this forward.

Thanks again for the contribution!

Havret:
-1

Xyolyp:
Thanks a lot, @Havret , for taking the time to look into this so thoroughly and for the kind words. I completely agree with your assessment: the synchronous approach in #4861 is the cleaner solution and avoids the edge cases that come with keeping the async buffer in the loop. It also follows what you originally outlined on #4474 more faithfully.

I'm happy to see the issue getting properly resolved and fully support going ahead with #4861. I'll leave it to the maintainers whether to close this one out. Thanks again for driving it home!

PR #4731, opened 2026-07-10
[Hangfire] Fix NullReferenceException for unresolvable jobs

reeloadead:
## Changes

The metrics instrumentation registers `HangfireMetricsStateFilter`, a global `IApplyStateFilter`. It formats a job name on every state transition, and `JobNameFormatter` dereferences `BackgroundJob.Job` with no null check.

When a job's invocation data cannot be deserialized (missing assembly, renamed type/method), `Job` is null. The filter throws `NullReferenceException`.

Affects all versions incl. `1.16.0-beta.1`.

> [!NOTE]
> This typically happens in platforms where the Hangfire scheduler (enqueues jobs) and the worker (executes jobs) are separate applications.
> The job's type/method can exist in the scheduler but be missing in the worker, so the worker can't deserialize the job and Job is null.

Observed in production (stack trimmed):

    "10 state change attempt(s) failed... moving job to the FailedState"
    System.NullReferenceException
       at ...JobNameFormatter.FormatJobName(Job job)
       at ...JobNameFormatter.FormatJobName(BackgroundJob backgroundJob)
       at ...HangfireTagBuilder.GetDefinitionName(BackgroundJob backgroundJob)

Preceded by `System.IO.FileNotFoundException: Could not resolve assembly '...'` at `InvocationData.DeserializeJob()`.

Fix: null-guard `JobNameFormatter` and return a sentinel when the job definition cannot be resolved.

* [ ] Changes in public API reviewed (if applicable) — n/a, no public API change

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 ## Unreleased
 
-* Fixed NRE in the metrics filter when a job's definition cannot be resolved.
-  ([#4731](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4731))
-
 * Assemblies are now digitally signed using cosign.
   ([#4637](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4637))
+
+* Fixed `NullReferenceException` in the metrics filter when a job's definition
+  cannot be resolved.
+  ([#4731](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4731))
 
 ## 1.16.0-beta.1
 
--- a/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed NRE in the metrics filter when a job's definition cannot be resolved.
+  ([#4731](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4731))
+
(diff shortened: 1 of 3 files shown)


src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed NRE in the metrics filter when a job's definition cannot be resolved.
+  ([#4731](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4731))
```
martincostello: Please move this to the end of the section and don't use an abbreviation.

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/JobNameFormatter.cs
```diff
@@ -9,8 +9,15 @@ namespace OpenTelemetry.Instrumentation.Hangfire.Implementation;
 
 internal static class JobNameFormatter
 {
-    internal static string FormatJobName(this Job job)
+    private const string UnknownJobName = "UNKNOWN";
```
martincostello: As this is only used in exactly one place, it can just be a literal for the return.

src/OpenTelemetry.Instrumentation.Hangfire/CHANGELOG.md
reeloadead: @martincostello, thanks, fixed

src/OpenTelemetry.Instrumentation.Hangfire/Implementation/JobNameFormatter.cs
reeloadead: fixed

martincostello (approved):

fred2u (approved):

PR #4734, opened 2026-07-11
[Runtime] Reduce RuntimeMetrics allocations

martincostello:
## Changes

I asked Copilot to do a performance analysis of the most popular instrumentations in the repo, and these changes came up as part of the results. These changes should reduce the overhead of the metric data collection.

- Reuse `Measurement` buffers for the GC observable instruments.
- Cache the result of `GC.GetGCMemoryInfo()` for a short duration (10ms) since the heap size callbacks each called it independently, doubling the cost when both are observed.
- Track `process.runtime.dotnet.assemblies.count` via an `AssemblyLoad` event handler and a counter.


martincostello:
Less compelling improvements with the fixes, so going to abandon this.

PR #4772, opened 2026-07-16
[SemanticConventions] Update to v1.43.0

martincostello:
## Changes

- Update Semantic Conventions to v1.43.0 and regenerate code.
- Update weaver to v0.24.2.


src/OpenTelemetry.SemanticConventions/Attributes/McpAttributes.cs
```diff
@@ -17,11 +17,13 @@ public static class McpAttributes
     /// <summary>
     /// The name of the request or notification method.
     /// </summary>
+    [Obsolete("Moved to the https://github.com/open-telemetry/semantic-conventions-genai repository.")]
```
martincostello: Need to decide what the weaver templates should do to deal with `<a href="foo">bar</a>` strings.

Render the HTML as-is and escape the `"`s, or strip them and replace with the bare URL?

matt-hensley: The `note` field can be any valid markdown, so may have multiple links embedded. I think  rendering the note to a doc comment would accommodate the arbitrary content without complex parsing.

Something like:
```
/// <summary>
/// The name of the request or notification method.
/// </summary>
/// <remarks>
/// Moved to the <see href="https://github.com/open-telemetry/semantic-conventions-genai">
/// OpenTelemetry GenAI semantic conventions repository</see>.
/// </remarks>
[Obsolete("Deprecated. See XML docs for migration details.")]
```

Not my favorite, but still gives the compiler warning and context.

martincostello: While definitely easier, the disadvantage to this approach is that the summary doesn't always give an alternative and the deprecation note is the only place the specifics exist. For example: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/4316c506370adf46c31063996690b0c2a1ad9d45/src/OpenTelemetry.SemanticConventions/Attributes/LinuxAttributes.cs#L17-L21

martincostello: Pushed a change that adds a macro to handle `<a>` tags and split out the text and the URL.

matt-hensley (approved):

rajkumar-rangaraj (approved):

PR #4860, opened 2026-07-21
[StackExchange.Redis] Validate flush interval

matt-hensley:
## Changes

Validates FlushInterval is set appropriately to avoid these errors:

  - Negative values cause the background flush loop to throw.
  - Zero causes a CPU-intensive busy loop.
  - Values above Int32.MaxValue milliseconds cause the background flush loop to throw.

* ~[ ] Changes in public API reviewed (if applicable)~


src/OpenTelemetry.Instrumentation.StackExchangeRedis/StackExchangeRedisInstrumentationOptions.cs
```diff
@@ -31,7 +34,20 @@ internal StackExchangeRedisInstrumentationOptions(IConfiguration configuration)
...
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// Thrown when the value is less than one millisecond or exceeds <see cref="int.MaxValue"/> milliseconds.
+    /// </exception>
+    public TimeSpan FlushInterval
+    {
+        get => this.flushInterval;
+        set
+        {
+            var milliseconds = value.TotalMilliseconds;
+            Guard.ThrowIfOutOfRange(milliseconds, nameof(value), min: 1, max: int.MaxValue);
```
martincostello: A downside of re-using this helper this way is that the caller doesn't get a TimeSpan-specific exception, but something about integers.

Maybe it would be better to just have an inline check of the timespan to `TimeSpan.FromMilliseconds(1)` instead?

martincostello (approved):

PR #4861, opened 2026-07-21
[Exporter.InfluxDB] Apply backpressure via ExportResult.Failure

### LINKED ISSUE #4473: [feature request] InfluxDB backpressure support

Havret:
Fixes #4473

## Changes

Metrics were written fire-and-forget into WriteApi's internal Rx buffer, which is unbounded and never surfaces write failures to the caller, so Export always returned Success. When InfluxDB was unavailable the buffer grew until the process was OOM-killed (#4474).

Write metrics synchronously via IWriteApiAsync and return ExportResult.Failure when a write fails. The SDK then applies backpressure and PeriodicExportingMetricReader retries on the next export cycle. Because the reader uses cumulative temporality, no data is lost: the accumulated values are re-exported once writes succeed again, so memory is bounded by cardinality rather than by outage duration. The now-unused FlushInterval option is removed.

### DIFF
--- a/src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
+++ b/src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Write metrics synchronously and report `ExportResult.Failure` when a write to
+  InfluxDB fails, so the SDK applies backpressure instead of growing an unbounded
+  in-memory buffer while InfluxDB is unavailable. The `FlushInterval` option is
+  deprecated as it no longer applies.
+  ([#4861](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4861))
+
 ## 1.0.0-alpha.10
 
 Released 2026-Jul-17
--- a/src/OpenTelemetry.Exporter.InfluxDB/IMetricsWriter.cs
+++ b/src/OpenTelemetry.Exporter.InfluxDB/IMetricsWriter.cs
@@ -1,7 +1,6 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-using InfluxDB.Client;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Resources;
 
@@ -9,5 +8,5 @@ namespace OpenTelemetry.Exporter.InfluxDB;
 
 internal interface IMetricsWriter
 {
-    void Write(Metric metric, Resource? resource, WriteApi writeApi);
+    void Write(Metric metric, Resource? resource, List<string> lineProtocol);
 }
--- a/src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
+++ b/src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using InfluxDB.Client;
-using InfluxDB.Client.Writes;
 using OpenTelemetry.Metrics;
 
 namespace OpenTelemetry.Exporter.InfluxDB;
@@ -11,36 +10,31 @@ internal sealed class InfluxDBMetricsExporter : BaseExporter<Metric>
 {
     private readonly IMetricsWriter writer;
     private readonly InfluxDBClient influxDbClient;
-    private readonly WriteApi writeApi;
+    private readonly WriteApiAsync writeApiAsync;
 
-    public InfluxDBMetricsExporter(IMetricsWriter writer, InfluxDBClient influxDbClient, WriteApi writeApi)
+    public InfluxDBMetricsExporter(IMetricsWriter writer, InfluxDBClient influxDbClient)
     {
         this.writer = writer;
         this.influxDbClient = influxDbClient;
-        this.writeApi = writeApi;
-
-        this.writeApi.EventHandler += (_, args) =>
-        {
-            switch (args)
-            {
-                case WriteErrorEvent writeErrorEvent:
-                    InfluxDBEventSource.Log.FailedToExport(writeErrorEvent.Exception.Message);
-                    break;
-                default:
-                    break;
-            }
-        };
+        this.writeApiAsync = influxDbClient.GetWriteApiAsync();
     }
 
     public override ExportResult Export(in Batch<Metric> batch)
     {
         try
         {
+            List<string> lineProtocol = [];
             foreach (var metric in batch)
             {
-                this.writer.Write(metric, this.ParentProvider?.GetResource(), this.writeApi);
+                this.writer.Write(metric, this.ParentProvider?.GetResource(), lineProtocol);
             }
 
+            if (lineProtocol.Count == 0)
+            {
+                return ExportResult.Success;
+            }
+
+            this.writeApiAsync.WriteRecordsAsync(lineProtocol).GetAwaiter().GetResult();
             return ExportResult.Success;
         }
         catch (Exception exception)
@@ -54,7 +48,6 @@ internal sealed class InfluxDBMetricsExporter : BaseExporter<Metric>
     {
         if (disposing)
         {
-            this.writeApi.Dispose();
             this.influxDbClient.Dispose();
         }
 
--- a/src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
+++ b/src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
@@ -29,12 +29,11 @@ internal sealed class InfluxDBMetricsExporter : BaseExporter<Metric>
                 this.writer.Write(metric, this.ParentProvider?.GetResource(), lineProtocol);
             }
 
-            if (lineProtocol.Count == 0)
+            if (lineProtocol.Count > 0)
             {
-                return ExportResult.Success;
+                this.writeApiAsync.WriteRecordsAsync(lineProtocol).GetAwaiter().GetResult();
             }
 
-            this.writeApiAsync.WriteRecordsAsync(lineProtocol).GetAwaiter().GetResult();
             return ExportResult.Success;
         }
         catch (Exception exception)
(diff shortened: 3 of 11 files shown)


Xyolyp:
Went through the diff — this is clearly the better approach, thanks for taking the time. Moving the line-protocol serialization out of the writers and into a plain ICollection<string> is a nice separation, and the synchronous WriteRecordsAsync + ExportResult.Failure path is exactly the backpressure model this should have. Good call also fixing the partial-read bug in InfluxDBFakeServer and adding the recovery test.

Only a single non blocking note:
Since the client-side batching from the old WriteApi is gone, a full export cycle now goes out as a single WriteRecordsAsync request. Probably fine, but for very high cardinality it might be worth chunking within an export as a follow-up.

LGTM overall, happy to see it land this way. 👍

src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -4,8 +4,6 @@ OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Bucket.get -> str
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Bucket.set -> void
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Endpoint.get -> System.Uri?
 OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.Endpoint.set -> void
-OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.FlushInterval.get -> int
-OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions.FlushInterval.set -> void
```
martincostello, line 8: Even though this package is alpha, I think it would be better as a first step to not remove and instead mark as `[Obsolete]`. Then we can remove it the release after.

src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -20,4 +18,4 @@ OpenTelemetry.Exporter.InfluxDB.MetricsSchema.None = 0 -> OpenTelemetry.Exporter
 OpenTelemetry.Exporter.InfluxDB.MetricsSchema.TelegrafPrometheusV1 = 1 -> OpenTelemetry.Exporter.InfluxDB.MetricsSchema
 OpenTelemetry.Exporter.InfluxDB.MetricsSchema.TelegrafPrometheusV2 = 2 -> OpenTelemetry.Exporter.InfluxDB.MetricsSchema
 OpenTelemetry.Metrics.InfluxDBExporterExtensions
-static OpenTelemetry.Metrics.InfluxDBExporterExtensions.AddInfluxDBMetricsExporter(this OpenTelemetry.Metrics.MeterProviderBuilder! builder, System.Action<OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions!>! configure) -> OpenTelemetry.Metrics.MeterProviderBuilder!
\ No newline at end of file
+static OpenTelemetry.Metrics.InfluxDBExporterExtensions.AddInfluxDBMetricsExporter(this OpenTelemetry.Metrics.MeterProviderBuilder! builder, System.Action<OpenTelemetry.Exporter.InfluxDB.InfluxDBMetricsExporterOptions!>! configure) -> OpenTelemetry.Metrics.MeterProviderBuilder!
```
martincostello: Line ending change?

src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* Write metrics synchronously and report `ExportResult.Failure` when a write to
+  InfluxDB fails, so the SDK applies backpressure instead of growing an unbounded
+  in-memory buffer while InfluxDB is unavailable. Because the reader uses
+  cumulative temporality, no data is lost: the accumulated values are re-exported
+  once writes succeed again. The `FlushInterval` option is removed as it no longer
+  applies.
+  ([#4658](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4658))
```
martincostello:
```suggestion
  ([#4861](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4861))
```

src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* Write metrics synchronously and report `ExportResult.Failure` when a write to
+  InfluxDB fails, so the SDK applies backpressure instead of growing an unbounded
+  in-memory buffer while InfluxDB is unavailable. Because the reader uses
+  cumulative temporality, no data is lost: the accumulated values are re-exported
+  once writes succeed again. The `FlushInterval` option is removed as it no longer
+  applies.
```
martincostello: I don't think we need the `Because...` sentence.

src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
```diff
@@ -11,36 +10,31 @@ internal sealed class InfluxDBMetricsExporter : BaseExporter<Metric>
...
-                this.writer.Write(metric, this.ParentProvider?.GetResource(), this.writeApi);
+                this.writer.Write(metric, this.ParentProvider?.GetResource(), lineProtocol);
+            }
+
+            if (lineProtocol.Count == 0)
+            {
+                return ExportResult.Success;
             }
 
+            this.writeApiAsync.WriteRecordsAsync(lineProtocol).GetAwaiter().GetResult();
```
martincostello: Is there a synchronous way to do this like `HttpClient.Send()` so we don't need to do this?

I'm not familiar with InfluxDB, so was the previous implementation being "synchronous" by just delegating an asynchronous background thread in its API to do the writing, but now the exporter is doing that explicitly using an async API, but blocking on it succeeding?

martincostello: Is this a `void`(`Task`), or is a return value being ignored?

src/OpenTelemetry.Exporter.InfluxDB/IMetricsWriter.cs
```diff
@@ -1,13 +1,12 @@
...
-using InfluxDB.Client;
 using OpenTelemetry.Metrics;
 using OpenTelemetry.Resources;
 
 namespace OpenTelemetry.Exporter.InfluxDB;
 
 internal interface IMetricsWriter
 {
-    void Write(Metric metric, Resource? resource, WriteApi writeApi);
+    void Write(Metric metric, Resource? resource, ICollection<string> lineProtocol);
```
martincostello: As this is `internal`, it would be slightly more performant to just use the concrete `List<string>` type instead of an interface.

src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
Havret: Done.

src/OpenTelemetry.Exporter.InfluxDB/.publicApi/PublicAPI.Unshipped.txt
Havret: Cleaned up.

src/OpenTelemetry.Exporter.InfluxDB/IMetricsWriter.cs
Havret: Done.

src/OpenTelemetry.Exporter.InfluxDB/CHANGELOG.md
Havret: Fixed.

src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
Havret: It is `void(Task)`.

Havret: Yes, that's exactly the case.

Havret:
@martincostello Could you please take a second look? Thanks!

martincostello (approved):

src/OpenTelemetry.Exporter.InfluxDB/InfluxDBMetricsExporter.cs
```diff
@@ -11,36 +10,31 @@ internal sealed class InfluxDBMetricsExporter : BaseExporter<Metric>
...
         try
         {
+            List<string> lineProtocol = [];
             foreach (var metric in batch)
             {
-                this.writer.Write(metric, this.ParentProvider?.GetResource(), this.writeApi);
+                this.writer.Write(metric, this.ParentProvider?.GetResource(), lineProtocol);
+            }
+
+            if (lineProtocol.Count == 0)
```
martincostello: Minor refactor here I spotted based on the missing patch coverage:

```csharp
if (lineProtocol.Count > 0)
{
    this.writeApiAsync.WriteRecordsAsync(lineProtocol).GetAwaiter().GetResult();
}

return ExportResult.Success;
```

Havret: That's neat. :) Applied.

Havret:
@martincostello If no further suggestions it seems it's ready to be merged.

martincostello (approved):

PR #4929, opened 2026-08-03
[SqlClient] SqlProcessor fixes

martincostello:
## Changes

- Handle `IN` clauses correctly when values might contain escaped identifiers, such as `)`.
- Ensure queries longer than `db.query.summary` are still redacted correctly.

Found by a Copilot/Claude scan of the repository code.

I asked Claude to investigate whether these changes have any performance impact. It's findings are summarised below.

<details>

<summary>Expand to read Claude 🐙 summary</summary>

## Performance impact of the SQL sanitization fixes

Benchmarks for the two `src/Shared/SqlProcessor.cs` sanitization fixes. Three variants were
compiled into a single process under separate namespaces so they run in one BDN session:

| Variant | Contents |
|---|---|
| `1_head` | neither fix (baseline) |
| `2_pending` | literal/comment-aware `IN (...)` close-paren scan only |
| `3_fixed` | the above **plus** removing the summary-length fast path |

### Methodology note

`GetSanitizedSql` memoises results by statement text (1000-entry cache), so the existing
`SqlProcessorBenchmarks` mostly measures a `ConcurrentDictionary` hit rather than the parser.
An additive uncached entry point (`=> SanitizeSql(sql)`) was used to measure parsing, and the
cached public API was measured separately. The existing benchmark `[Params]` corpus also
contains no statement whose summary reaches `MaxSummaryLength` (255), so the fast path was
previously unbenchmarked.

### Parser throughput (uncached)

| Query shape | `1_head` | `2_pending` | `3_fixed` | ratio | alloc ratio |
|---|---:|---:|---:|---:|---:|
| short_select | 236 ns | 211 ns | 215 ns | 0.92 | 1.00 |
| short_subquery | 542 ns | 480 ns | 492 ns | 0.92 | 1.00 |
| short_insert | 554 ns | 512 ns | 461 ns | 0.83 | 1.00 |
| short_update | 249 ns | 255 ns | 222 ns | 0.90 | 1.00 |
| short_comments | 182 ns | 183 ns | 179 ns | 0.98 | 1.00 |
| short_2tables | 388 ns | 373 ns | 381 ns | 0.98 | 1.00 |
| in_clause_200 | 279 ns | 255 ns | 257 ns | 0.92 | 1.00 |
| limit_joins12 | 3.04 µs | 2.99 µs | 3.26 µs | **1.07** | 4.51 |
| limit_tables12 | 814 ns | 800 ns | 913 ns | **1.13** | 2.17 |
| limit_longident | 548 ns | 538 ns | 707 ns | **1.30** | 17.2 |
| limit_verylong | 5.08 µs | 5.33 µs | 18.3 µs | **3.60** | 23.7 |

The first seven rows are the existing benchmark corpus: **no regression** (all ratios ≤ 1.00;
sub-1.0 values are run-to-run noise). The `IN`-clause fix is free everywhere, including a
200-value `IN` list. Cost appears only on the `limit_*` rows, i.e. statements whose summary
exceeds 255 characters — exactly where the fast path used to engage.

### Isolating CPU cost from allocation

| Control case | `1_head` | `3_fixed` | ratio | alloc ratio |
|---|---:|---:|---:|---:|
| long statement, summary stays **short** | 6.95 µs | 7.16 µs | **1.03** | 1.00 |
| past the limit, **no literals** to redact | 9.59 µs | 21.6 µs | **2.25** | **1.00** |

* The first control shows the change costs nothing for full parsing itself — a comparably long
  statement that never trips the limit is unchanged.
* The second isolates pure CPU at identical allocation: **2.25×**. That is the real cost of
  tokenizing past the limit instead of bulk-copying to the next whitespace.
* Comparing the two controls explains the allocation column, which is **not** waste. Previously
  the fast path left the text unchanged, so `SanitizeSql` took its
  `sanitizedSqlSpan.SequenceEqual(sqlSpan) ? sql : …` branch and returned the *original string
  instance* — allocating nothing, because it had leaked. Once literals are actually redacted the
  string genuinely differs and must be materialised. `limit_longident` is the clearest case:
  40 B → 688 B is the ~300-character redacted statement that previously wasn't produced at all.

### Cached path (steady state, what an application pays per execution)

| Query | `1_head` | `3_fixed` | ratio | allocated |
|---|---:|---:|---:|---:|
| joins12 | 231 ns | 244 ns | 1.06 | 0 B both |
| verylong | 1.74 µs | 1.81 µs | 1.05 | 0 B both |

Within noise, zero allocation either way. Parse cost is paid once per distinct statement text.

### Summary

* No regression for typical statements, or for long statements generally.
* Measurable cost is confined to statements whose summary exceeds `MaxSummaryLength`:
  **1.07–1.30×** for realistic shapes (12 joins, 12 comma-listed tables, long identifier) and
  **3.60×** for a deliberately pathological 5.6 KB statement.
* Amortised to **~5%** by the existing cache.
* Increased allocation reflects correct redaction now happening, not new overhead.

### Environment

```
BenchmarkDotNet v0.15.8, Windows 11 (10.0.26200.8875/25H2)
13th Gen Intel Core i7-13700H 2.90GHz, 1 CPU, 20 logical and 14 physical cores
.NET SDK 11.0.100-preview.6.26359.118
  [Host] : .NET 10.0.10 (10.0.1026.32716), X64 RyuJIT x86-64-v3
Job=ShortRun  Toolchain=InProcessEmitToolchain  IterationCount=3  LaunchCount=1  WarmupCount=3
```

ShortRun with 3 iterations gives wide error bars; differences under roughly ±10% should be
read as noise. The multi-fold figures were stable across repeated runs.

</details>

### DIFF
--- a/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+* Fixed query sanitization so that a `)` inside a value or a comment in an
+  `IN (...)` clause no longer causes the values which follow it to be left
+  unsanitized.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
+* Fixed query sanitization so that literals are still redacted after the
+  `db.query.summary` length limit is reached.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
 ## 1.17.0-beta.1
 
 Released 2026-Jul-17
--- a/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
@@ -5,11 +5,11 @@
 * Fixed query sanitization so that a `)` inside a value or a comment in an
   `IN (...)` clause no longer causes the values which follow it to be left
   unsanitized.
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
 
 * Fixed query sanitization so that literals are still redacted after the
   `db.query.summary` length limit is reached.
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
 
 ## 1.17.0-beta.1
 
--- a/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+* Fixed query sanitization so that a `)` inside a value or a comment in an
+  `IN (...)` clause no longer causes the values which follow it to be left
+  unsanitized.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
+* Fixed query sanitization so that literals are still redacted after the
+  `db.query.summary` length limit is reached.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
 ## 1.17.0
 
 Released 2026-Jul-17
--- a/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
@@ -5,11 +5,11 @@
 * Fixed query sanitization so that a `)` inside a value or a comment in an
   `IN (...)` clause no longer causes the values which follow it to be left
   unsanitized.
-  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
 
 * Fixed query sanitization so that literals are still redacted after the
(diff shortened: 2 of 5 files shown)


src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+* Fixed query sanitization so that a `)` inside a value or a comment in an
+  `IN (...)` clause no longer causes the values which follow it to be left
+  unsanitized.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
+* Fixed query sanitization so that literals are still redacted after the
+  `db.query.summary` length limit is reached.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))

* Fixed query sanitization so that literals are still redacted after the
  `db.query.summary` length limit is reached.
  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
```

src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+* Fixed query sanitization so that a `)` inside a value or a comment in an
+  `IN (...)` clause no longer causes the values which follow it to be left
+  unsanitized.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
+
+* Fixed query sanitization so that literals are still redacted after the
+  `db.query.summary` length limit is reached.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))

* Fixed query sanitization so that literals are still redacted after the
  `db.query.summary` length limit is reached.
  ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
```

rajkumar-rangaraj (approved):

PR #4930, opened 2026-08-04
[OpAMP] Refactor OpAMP communication pipe

RassK:
## What

Design discussion issue https://github.com/open-telemetry/opamp-spec/issues/366

This is about ensuring that communication is according to spec.

> When the Agent wants to send a message to the Server and the Agent has previously sent a request to the Server that is not yet responded, the Client MUST wait until the response is received before a new request can be made.
src: https://opentelemetry.io/docs/specs/opamp/#plain-http-transport

Since spec is currently not clear that messages can be accepted in sync or async manner. This PR is a basis to support both options via configuration. If it's decided that the client must block the pipe until a full response is constructed, a follow up is needed.

## Changes

- Makes sure the client is waiting for server to respond and blocks the pipe.
- Uses accumulator message to "queue" client's messages

### Breaking changes ❗

- No need for async send messages, since messages are always accumulated.

### Notes

~~❗ This PR is a preview and a discussion object how to proceed to support corner cases in the spec.~~
Since there seems to be a consensus with this PR, we can move forward

* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes < TODO until the final form is decided


src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+        lock (this.frameLock)
+        {
+            messageRequest(this.currentFrame);
+            this.hasAccumulatedData = true;
+        }
+
+        this.TryFlush();
+    }
+
+    public Task FlushAsync()
```
stevejgordon: Should this accept a `CancellationToken` passed through from `StopAsync`?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+            OpAmpClientEventSource.Log.SendMessageException(ex);
+            this.TryFlush();
+        }
+    }
+
+    private void OnServerFrameReceived(ServerToAgent message)
+    {
+        lock (this.frameLock)
+        {
+            this.isBusy = false;
```
stevejgordon, line 226: This is correct for the "happy path", but what if the the web socket connection is closed or dropped? I think we'd need a mechanism for `WsReciever` to notify through when that happens so isBusy can be reset?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+                .ConfigureAwait(false);
+        }
+
+        this.AppendMessage(MessageBuilderHelper.AppendIdentification);
+    }
+
+    public async Task StopAsync(CancellationToken token = default)
+    {
+        // Drain queued data.
+        await this.FlushAsync()
```
stevejgordon: Pass the `CancellationToken` here?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+            .ConfigureAwait(false);
+
+        this.AppendMessage(MessageBuilderHelper.AppendAgentDisconnect);
+
+        // Send disconnect.
+        await this.FlushAsync()
```
stevejgordon: +1

src/OpenTelemetry.OpAmp.Client/Internal/IFrameBuilder.cs
```diff
@@ -25,5 +24,5 @@ internal interface IFrameBuilder
 
     IFrameBuilder AddCustomMessage(string capability, string type, ReadOnlyMemory<byte> data);
 
-    AgentToServer Build();
+    IFrameBuilder Clear();
```
stevejgordon: Does this belong on this interface? It's beyond the concern of this abstraction. It seems to only be used from a test and could still existing on the `FrameBuilder` directly for that call site.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    public async Task StartAsync(CancellationToken token = default)
+    {
+        if (this.transport is WsTransport wsTransport)
+        {
+            await wsTransport.StartAsync(token)
+                .ConfigureAwait(false);
+        }
+
+        this.AppendMessage(MessageBuilderHelper.AppendIdentification);
```
stevejgordon, line 48: Should we flush here to ensure identification is sent before heartbeats (and other services) are started? Otherwise, a lost identification message will cause heartbeats to be sent to a server it had never seen.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
```diff
@@ -15,6 +15,7 @@ namespace OpenTelemetry.OpAmp.Client.Internal;
 internal sealed class FrameProcessor
 {
     private readonly ConcurrentDictionary<Type, IReadOnlyList<object>> listeners = [];
+    private readonly ConcurrentBag<Action<ServerToAgent>> internalListeners = [];
```
stevejgordon, line 18: Does this need to be `ConcurrentBag`? One one internal listener is accepted in the ctor, can we just store that?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,202 @@
...
+        this.tokenSource.Cancel();
+        this.tokenSource.Dispose();
+
+        if (this.transport is IDisposable disposableTransport)
+        {
+            disposableTransport.Dispose();
+        }
+    }
+
+    internal void TryFlush()
```
stevejgordon:
```suggestion
    private void TryFlush()
```

stevejgordon (commented):
The overall direction looks reasonable and the pipe model is the right approach to satisfy the HTTP transport spec requirement.

A few initial code comments to consider. I will do another pass soon.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: Correct, we discovered this as well in the internal overhaul to prevent hangs in the pipe. Using cancellation token here gives the control to user.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/0b456cc30fc595417a91105e61af27409f365193

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930#discussion_r3734760471

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930#discussion_r3734760471

src/OpenTelemetry.OpAmp.Client/Internal/IFrameBuilder.cs
RassK: Seems it became a leftover, removed.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: TryFlush should send it instantly since the pipe is initially free. I added Flush just in case, so nothing weird should not happen.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: Correct, retry / reconnect paths are missing currently intentionally. Seems a larger scope to focus on separately.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
RassK: The only issue today is that the pipe does not have a control over the processor (it does not construct it). Pipe itself is a user, like any internal services could be.

stevejgordon: That doesn't explain the need for `ConcurrentBag`. Looking at the code, `OpAmpPipe` is the only caller of `SubscribeToServerMessages`, and it calls it exactly once from its constructor. There's also no `Unsubscribe` method, so the concurrent-collection semantics are never exercised.

A simple `Action<ServerToAgent>?` field would be clearer and more appropriate here (as it stands) and doesn't carry the misleading implication that subscribers can be dynamically added/removed concurrently.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
```diff
@@ -81,6 +87,11 @@ private void Deserialize(ReadOnlySequence<byte> sequence)
     {
         var message = ServerToAgent.Parser.ParseFrom(sequence);
 
+        foreach (var listener in this.internalListeners)
+        {
+            listener.Invoke(message);
```
stevejgordon: Should we wrap this in try/catch to avoid exceptions bubbling out? We already do that for public listeners.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
RassK: Simple `Action<ServerToAgent>?` would meant that the owner should have set it. But because `OpAmpPipe` doesn't own it, then the control who sets it becomes vague. I do understand that there is a single use currently but such pattern leaves too many doors open.

I refactored this and removed that new functionality and reused the sub/unsub pattern already existing. This also helps in the future if you need to check multiple fields (sub-messages) without having any sync logic of waiting for multiple callbacks.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/40877b3670ead3021208a9de7ea8cb0f81d6798b

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,231 @@
...
+
+using OpAmp.Proto.V1;
+using OpenTelemetry.OpAmp.Client.Internal.Transport;
+using OpenTelemetry.OpAmp.Client.Internal.Transport.Http;
+using OpenTelemetry.OpAmp.Client.Internal.Transport.WebSocket;
+using OpenTelemetry.OpAmp.Client.Settings;
+
+namespace OpenTelemetry.OpAmp.Client.Internal;
+
+internal sealed class OpAmpPipe : IDisposable
```
stevejgordon, line 14: Genral comment for the `OpAmpPipe` implementation - For HTTP, gating on server-frame receipt looks spec-aligned: every POST gets a `ServerToAgent` response and `PlainHttpTransport` processes it before `SendAsync` returns. For WebSocket, the spec (in my re-reading) is full-duplex with no response requirement, so only clearing `isBusy` in `OnServerFrameReceived` can stall the pipe if the server doesn’t reply to every agent send (or unblock on an unrelated server message). I think we need to consider transport-specific pipe behavior: response-gated for HTTP, send-completion-gated for WebSocket.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    private async Task SendMessageAsync(AgentToServer message)
+    {
+        try
+        {
+            OpAmpClientEventSource.Log.SendingMessage();
+
+            await this.transport.SendAsync(message, this.tokenSource.Token)
```
stevejgordon: This token is only cancelled on dispose. Should we accept the tokens through from `FlushAsync`/`StopAsync` or link the the caller cancellation tokens with `tokenSource.Token` to ensure correct in-flight cancellation?

src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
```diff
@@ -112,8 +112,7 @@ private async Task HeartbeatTickAsync()
         {
             var report = this.CreateHealthReport();
 
-            await this.dispatcher.DispatchHeartbeatAsync(report, this.cts.Token)
-                .ConfigureAwait(false);
+            this.pipe.AppendMessage(MessageBuilderHelper.AppendHeartbeat(report));
```
stevejgordon, line 120: This no longer passes the cancellation token. Can this race with `Stop`? Would checking `this.cts.IsCancellationRequested` first before appending be reasonable here?

stevejgordon:
@RassK Sorry, my feedback and your new commit overlapped, so some of my comments may no longer apply. I like the direction of the refactor.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
```diff
@@ -81,6 +81,8 @@ private void Deserialize(ReadOnlySequence<byte> sequence)
     {
         var message = ServerToAgent.Parser.ParseFrom(sequence);
 
+        this.Dispatch(new ServerToAgentMessage(message));
```
stevejgordon: Worth checking `listeners.ContainsKey(typeof(ServerToAgentMessage))` before allocating and dispatching a message no one is subscriber to?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: seems a good catch, https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/3ae9596c2c0771bc380a8758b5837ff193b2d420

src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
RassK: yep, seems `IsCancellationRequested` check is appropriate here. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/d288329aeebbcf5cee322d67dd001be9b6791cfb

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
RassK: that might be useful for other types as well.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: I went currently with a short change, https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/cdc5ac77c2a116a00254c176947904b6dfc114c2

Definitely a good point that it could block for too long.

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
RassK: That should do it https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/be8b15bc2df12cd373b3cdd13d79e83b269e70fd

src/OpenTelemetry.OpAmp.Client/Internal/FrameProcessor.cs
RassK: outdated

RassK:
@stevejgordon could you take a look again ⬆️

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -0,0 +1,264 @@
...
+            return;
+        }
+
+        lock (this.frameLock)
+        {
+            messageRequest(this.currentFrame);
+            this.hasAccumulatedData = true;
+        }
+
+        this.TryFlush(this.tokenSource.Token);
```
stevejgordon, line 112: This still does not propagate the caller cancellation token to an already in-flight send. `AppendMessage` starts the operation with the pipe lifetime token here; if `FlushAsync`/`StopAsync` is then called while `isBusy` is true, its token only cancels `WaitForFlushAsync` and the transport operation continues, leaving the pipe busy until the send eventually completes or the pipe is disposed. The controlled transport also ignores its token, so the new cancellation tests do not cover this. Can the active send be made cancellation-aware for the caller?

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
```diff
@@ -173,8 +145,6 @@ IFrameBuilder IFrameBuilder.AddCustomCapabilities(IEnumerable<string> capabiliti
 
     IFrameBuilder IFrameBuilder.AddCustomMessage(string capability, string type, ReadOnlyMemory<byte> data)
     {
-        this.EnsureInitialized();
-
         this.currentMessage.CustomMessage = new CustomMessage
```
stevejgordon, line 148: Multiple custom messages can be silently lost while another send is active. The first queued custom message remains in `currentFrame`, but every subsequent `SendCustomMessage` overwrites this singular property before `Build()` runs, whereas the previous dispatcher serialized and sent every call. Custom messages are not coalescible state; they need separate queued frames (or equivalent) so each payload is delivered.

src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
```diff
@@ -108,12 +108,16 @@ private void CreateOrUpdateTimer(TimeSpan interval)
...
+            return;
+        }
+
         try
         {
             var report = this.CreateHealthReport();
 
-            await this.dispatcher.DispatchHeartbeatAsync(report, this.cts.Token)
-                .ConfigureAwait(false);
+            this.pipe.AppendMessage(MessageBuilderHelper.AppendHeartbeat(report));
```
stevejgordon, line 120: The cancellation pre-check still leaves a shutdown race. A timer callback can pass the check, then `Stop()` can cancel/disable the timer and `OpAmpClient.StopAsync` can begin draining/queue the disconnect, after which this callback appends a heartbeat. Because disabling a `Timer` does not wait for an executing callback, that heartbeat can be sent after the disconnect frame. Please synchronize stop with active callbacks or make this append atomically reject work after service cancellation.

stevejgordon (changes requested):
Thanks for addressing the earlier feedback. The transport-specific WebSocket gating, identification flush, listener changes, dispatch allocation check, and interface cleanup now look resolved. I still see three correctness issues around cancellation, non-coalescible custom messages, and heartbeat shutdown ordering.

stevejgordon:
@RassK Sorry for the delay. I've tweaked my GMail filters and this didn't go where I thought it would! I had a look and didn't spot much directly in review, but I had AI do a pass and push some feedback on my behalf which we may need to consider.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: There might be some other flaws with cancellation. Should make sure that this should not cause any half states and wondering about loss of accumulated messages.

src/OpenTelemetry.OpAmp.Client/Internal/Services/Heartbeat/HeartbeatService.cs
RassK: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/750827961844878810453d08b78ff7c55d7a8e27
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930/commits/48a08f8aaf6b76b16b55a683483462b7868c52ef

This should now avoid races when the pipe enters to the stopped state. All new messages are dropped.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
RassK: Codex about the comment:

> Yes—the hypothesis is wrong as a defect claim.
> The observed behavior is real, but intentional: an in-flight send started by AppendMessage belongs to the pipe lifetime, not to a later FlushAsync or StopAsync caller. Their cancellation tokens should cancel their waits, not abort that autonomous send.
> When FlushAsync itself starts a send, its token is already passed to the transport.

src/OpenTelemetry.OpAmp.Client/Internal/FrameBuilder.cs
RassK: Separated the issue > https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/5142

stevejgordon (approved):
I think this looks good now with the follow up issue in place to handle the possibility of dropped custom messages.

Kielek (approved):
LGTMRS

PR #4944, opened 2026-08-06
[SqlClient] Improve tag addition efficiency

martincostello:
## Changes

Turn an `O(m * n)` scan for tags into `O(n)`.

Identified by a scan of the code by Claude Code.

A throwaway benchmark gives the following numbers for the change. TL;DR: ~30-40% improvement.

<details>

<summary>Click to expand</summary>

| Method | TagCount | Duplicates | Mean | Ratio | Allocated |
|---|---|---|---:|---:|---:|
| Before (main) | 5 | 0 | 44.26 ns | 1.00 | – |
| AfterOriginal | 5 | 0 | 23.67 ns | 0.53 | – |
| **AfterDeduped (branch)** | 5 | 0 | 25.09 ns | **0.57** | – |
| Before (main) | 9 | 0 | 57.35 ns | 1.00 | – |
| AfterOriginal | 9 | 0 | 34.14 ns | 0.60 | – |
| **AfterDeduped (branch)** | 9 | 0 | 36.77 ns | **0.64** | – |
| Before (main) | 10 | 0 | 61.15 ns | 1.00 | – |
| AfterOriginal | 10 | 0 | 39.30 ns | 0.64 | – |
| **AfterDeduped (branch)** | 10 | 0 | 43.85 ns | **0.72** | – |
| Before (main) | 5 | 2 | 50.96 ns | 1.00 | – |
| AfterOriginal | 5 | 2 | 31.91 ns | 0.63 | – |
| **AfterDeduped (branch)** | 5 | 2 | 27.94 ns | **0.55** | – |
| Before (main) | 9 | 2 | 60.58 ns | 1.00 | – |
| AfterOriginal | 9 | 2 | 57.72 ns | 0.95 | **280 B** |
| **AfterDeduped (branch)** | 9 | 2 | 40.11 ns | **0.66** | – |
| Before (main) | 10 | 2 | 68.65 ns | 1.00 | – |
| AfterOriginal | 10 | 2 | 65.18 ns | 0.95 | **280 B** |
| **AfterDeduped (branch)** | 10 | 2 | 45.56 ns | **0.66** | – |

</details>

### DIFF
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs
@@ -141,7 +141,7 @@ internal sealed class SqlTelemetryHelper
 
     private static Dictionary<string, int> CreateSharedTagNameIndexes()
     {
-        Debug.Assert(SharedTagNames.Length <= 32, "There are too many shared tag names to track with a 32-bit mask.");
+        Debug.Assert(SharedTagNames.Length < 32, "There are too many shared tag names to track with a 32-bit mask.");
 
         var indexes = new Dictionary<string, int>(SharedTagNames.Length, StringComparer.Ordinal);
 
--- a/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs
@@ -44,22 +44,41 @@ internal sealed class SqlTelemetryHelper
     ];
 
 #if NET
-    private static readonly FrozenSet<string> SharedTagNameSet = SharedTagNames.ToFrozenSet(StringComparer.Ordinal);
+    private static readonly FrozenDictionary<string, int> SharedTagNameIndexes = CreateSharedTagNameIndexes().ToFrozenDictionary(StringComparer.Ordinal);
 #else
-    private static readonly HashSet<string> SharedTagNameSet = new(SharedTagNames, StringComparer.Ordinal);
+    private static readonly Dictionary<string, int> SharedTagNameIndexes = CreateSharedTagNameIndexes();
 #endif
+
+    private static readonly int AllSharedTagsFoundMask = (1 << SharedTagNames.Length) - 1;
 
     public static void AddSharedTags(Activity activity, ref TagList tags)
     {
         var enumerator = activity.EnumerateTagObjects();
+        var found = 0;
 
         while (enumerator.MoveNext())
         {
             ref readonly var tag = ref enumerator.Current;
 
-            if (tag.Value != null && SharedTagNameSet.Contains(tag.Key))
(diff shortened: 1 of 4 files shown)


src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs
```diff
@@ -95,4 +138,18 @@ public static TagList GetTagListFromConnectionInfo(string? dataSource, string? d
 
     internal static double CalculateDurationFromTimestamp(long begin)
         => Stopwatch.GetElapsedTime(begin).TotalSeconds;
+
+    private static Dictionary<string, int> CreateSharedTagNameIndexes()
+    {
+        Debug.Assert(SharedTagNames.Length <= 32, "There are too many shared tag names to track with a 32-bit mask.");
```
stevejgordon, line 144:
```suggestion
        Debug.Assert(SharedTagNames.Length < 32, "There are too many shared tag names to track with a 32-bit mask.");
```

Should this be less than 32?

martincostello: 32-bit and a flag per bit, so `<=32` is right and it would be `<32` if it was an index not a length/count?

stevejgordon: I just checked with AI!

A 32-bit mask can hold 32 flags, but (1 << SharedTagNames.Length) - 1 breaks at length 32 because C# wraps that shift to 1 << 0 (For int/uint, C# only uses the low 5 bits of the shift count (count & 0x1F)). So for this code, the assert should stay < 32 unless that mask formula is fixed.

martincostello: I love it when AI finds issues that the AI that wrote the code in the first place shouldn't have done in the first place... 😮‍💨

Kielek:
From Codex review - it is complaining that some wrongly created tags (`null`) could potentially broke changes.

More details:
[P2] Guard null tag keys before dictionary lookup — [SqlTelemetryHelper.cs:63](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/67bf5d9a95e9ebb716215a11eb7f02181d8ec243/src/OpenTelemetry.Instrumentation.SqlClient/Implementation/SqlTelemetryHelper.cs#L63)
Activity.AddTag(null!, value) stores a null key at runtime ([source](https://github.com/dotnet/runtime/blob/1e4f4bc96e9a45eeba82a5d358d1779795bfbcb1/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L489-L496)). The previous GetTagItem scan safely ignored it, but the new Dictionary/FrozenDictionary.TryGetValue call throws ArgumentNullException. The listener catches that exception, but metric recording aborts, dropping that SQL duration measurement. Add tag.Key is null to the guard and a regression test using activity.AddTag(null!, "value").

Kielek (approved):

Issue #4742, opened 2026-07-13
Dynamic Control (META)


stevejgordon:
### Component

This is a new component

### Is your feature request related to a problem?

Currently, the SDK and instrumentations are configured once at startup and configuration cannot be adjusted at runtime without deploying an updated configuration and restarting the instrumented application. In some situations it can be useful to dynamically control the configuration at runtime, without requiring a restart. In large enterprises it can be complex and time consuming to deploy configuration changes to a many distributed services. This can include changing logging levels, sampling rates and potentially disabling exporting of signals. This may be for general operational needs, or to prevent a flood of signal data due to unexpected load or external misuse.

### What is the expected behavior?

My colleague has been working on an initial version of dynamic control (via [telemetry policies](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4738-telemetry-policy.md)) for the Java SDK ([contrib](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/dynamic-control)).

I plan to create a .NET version of this same concept to support remote, dynamic configuration of the SDK and its components. The core design will follow that of the Java version and the OTEP for telemetry policies with .NET idioms and alignment with the work I'm doing in the SDK for declarative config.

Initially, I propose that I work in one new project for this POC with a longer term plan to potentially refactor into several projects. The reasoning is that telemetry policies can be applied from files, HTTP, OpAMP and custom providers. Some consumers may want the file-based telemetry policies, and likely won't want OpAMP dependencies etc. Eventually I could see the need for a core/abstractions project, with implementations for each provider in their own projects. Until a working POC is in place and a design is settled, it's probably easier to maintain this as a single project. This is the same plan that exists for the Java implementation.

I'm introducing this issue to capture discussions around the initial design and implementation. This issue is the meta issue so that the overall outline and progress are clearer. Note this not a generic reconfiguring of arbitrary agent features/config/capability, this is very specific to control maybe a dozen capabilities in some dynamically adjustable way

## Task breakdown

- [x] Phase 1: Package, tests, solution, CI, and packaging infrastructure (#4744)
- [x] Phase 2: Internal policy model and validation (#4950)
- [x] Phase 3: Source identity, immutable snapshots, and atomic policy store (3a #5011 / 3b #5017)
- [x] Phase 4: Ordered and coalesced policy-store subscriptions (#5135)
[...]

PR #4950, opened 2026-08-07
[Dynamic Control] Phase 2: Add ITelemetryPolicy and TraceSamplingRatePolicy

stevejgordon:
Contributes to: #4742

## Changes

Introduces the internal policy model layer: ITelemetryPolicy as the base contract and TraceSamplingRatePolicy as the first concrete implementation for the `trace-sampling` policy type from the Telemetry Policy OTEP.

This is phase 2 of the dynamic control work — the policy model layer that downstream phases will build on (policy source ingestion, sampler integration).


src/OpenTelemetry.DynamicControl/Internal/Policies/ITelemetryPolicy.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Represents a resolved telemetry policy as defined by the Telemetry Policy OTEP.
+/// A policy carries an identity (<see cref="Id"/>, <see cref="Name"/>) and a
+/// <see cref="PolicyType"/> discriminator that determines which concrete
+/// implementation handles it (e.g. trace-sampling).
+/// </summary>
+internal interface ITelemetryPolicy
```
martincostello, line 12: Do other properties from the [protobuf schema](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4738-telemetry-policy.md#typed-schema) need to be fed through here?

src/OpenTelemetry.DynamicControl/Internal/Policies/ITelemetryPolicy.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Represents a resolved telemetry policy as defined by the Telemetry Policy OTEP.
+/// A policy carries an identity (<see cref="Id"/>, <see cref="Name"/>) and a
+/// <see cref="PolicyType"/> discriminator that determines which concrete
+/// implementation handles it (e.g. trace-sampling).
```
martincostello:
```suggestion
/// implementation handles it (e.g. <c>trace-sampling</c>).
```

src/OpenTelemetry.DynamicControl/Internal/Policies/ITelemetryPolicy.cs
stevejgordon: Eventually, yes, but the exact shape is subject to change. For now, I've implemented the minimal parts to conform to "Every policy MUST have an ID and name. Each policy MAY specify associated labels and metadata about its creation." In Java, they are using a proposed short-form for the source messages which this is sufficient to replicate as discussions occur on the exact details. My plan for this PR and all future ones is to follow that path for now and extend as and when needed, ideally when a more complete spec is in place. Everything will be internal until it needs to be exposed, and public (but experimental/alpha) until the spec itself is stable and more implementations have guided that design.

martincostello (approved):

PR #4954, opened 2026-08-10
[Http] Redact only named query string parameters

### LINKED ISSUE #1791: [Http Instrumentation] Make it easier to scrub sensitive URI details from output
# Feature Request

Before opening a feature request against this repo, consider whether the feature
should/could be implemented in the [other OpenTelemetry client
libraries](https://github.com/open-telemetry/). If so, please [open an issue on
opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/issues/new)
first.

**Is your feature request related to a problem?**

If so, provide a concise description of the problem.

> The HTTP instrumentation package currently writes out the full request path for most requests, the exception being requests that contain Uri.UserInfo. This is not sufficient for many resource types. For example, the Azure Storage SDK's requests show up in logs with a SAS token appended. 
>
> For example:
> `https://mystorage.blob.core.windows.net/container/myfile.txt?sv=2019-07-07&sr=c&sig=placeholder&se=2021-09-08T22%3A03%3A23Z&sp=r&api-version=2019-07-07&`
>
> I can also imagine other scenarios where the URI might otherwise contain sensitive information that cannot be logged.

**Describe the solution you'd like:**

What do you want to happen instead? What is the expected behavior?

> It would be helpful if OpenTelemetry had a configurable high-performance URL scrubber for anywhere URLs could be logged,
> to make it as easy as possible for developers to sanitize their log output. This would support (minimally) blocking out any specified query string parameters.

**Describe alternatives you've considered.**

Which alternative solutions or features have you considered?

> In our project today, we configure an enricher to run a Regex replace on all URLs we log, but it feels like there is room for improvement on the default behavior.

## Additional Context

Add any other context about the feature request here.

> This is definitely in the "nice to have" category. There are ways to solve the problem with the current shipped libraries.

PetarJerinic:
Fixes #1791

## Changes

`OTEL_DOTNET_EXPERIMENTAL_HTTPCLIENT_SENSITIVE_QUERY_PARAMETERS` takes a comma-separated list of query parameter names.  Those values are replaced with `REDACTED` in `url.full`; every other value survives. 
Names are matched case-sensitively, per semconv. Applies to .NET Framework and up to .NET 8.

There are no default query parameters giving the user full control.

### Open questions
- `RedactionHelper.SensitiveRedactedText` - should it be a separate variable or should existing `RedactionHelper.RedactedText` be used? A new variable was added since semconv states that REDACTED should be uppercase. I am happy to use the existing one to maintain consistent behavior, or update all usages of the existing one to the new one.
- ~In `HttpHandlerDiagnosticListener` (lines ~150-159, .NET 9+ runtimes) -  Runtime writes `url.full` itself and redacts the query to ?*. When this variable is set, the instrumentation overwrites that one attribute. This goes against the README's "does not override native instrumentation" guarantee (line 47) for opted-in users. Should this change be removed or should the README be amended?~
Resolved: per suggestion below, .NET 9+ behavior remains the same


martincostello:
> In HttpHandlerDiagnosticListener (lines ~150-159, .NET 9+ runtimes) - Runtime writes url.full itself and redacts the query to ?*. When this variable is set, the instrumentation overwrites that one attribute. This goes against the README's "does not override native instrumentation" guarantee (line 47) for opted-in users. Should this change be removed or should the README be amended?

My initial thought is that we should stick with the README and not overwrite it, which effectively means implementing this request is two pieces of work, because the .NET runtime would also need to implement it. However they almost certainly won't implement it while it's non-stable.

Here's some further links and context here around how `url.query` redaction and configurability is handled in various parts of .NET (or not): 
- https://github.com/dotnet/runtime/issues/110018
- https://github.com/dotnet/aspnetcore/issues/64850
- https://github.com/dotnet/aspnetcore/issues/65873#issuecomment-4811267649

src/OpenTelemetry.Instrumentation.Http/HttpClientTraceInstrumentationOptions.cs
```diff
@@ -156,6 +170,17 @@ internal HttpClientTraceInstrumentationOptions(IConfiguration configuration)
...
     internal bool DisableUrlQueryRedaction { get; set; }
 
+    /// <summary>
+    /// Gets or sets the query parameter names whose values are redacted.
+    /// </summary>
+    /// <remarks>
+    /// When <see langword="null" /> every query parameter value is redacted. When
+    /// set, only the values of the named parameters are redacted and every other
+    /// query parameter value is preserved. Names are matched case-sensitively, as
+    /// semantic conventions require.
```
martincostello:
```suggestion
    /// query parameter value is preserved.
```

PetarJerinic:
@martincostello I updated the SensitiveQueryParameters description and removed the change in `HttpHandlerDiagnosticListener`. 

I also updated the description of PR, please re-review when time permits.

With your comment in mind for `url.query` redaction and configurability, there will be some wait time until this is supported in newer versions of .NET.

martincostello:
@rajkumar-rangaraj Thoughts on the path forward for this PR after our discussion in the SIG call this Tuesday?

rajkumar-rangaraj (changes requested):
Thanks @PetarJerinic for the work here, the implementation itself is solid.

After the SIG discussion, I'd rather not take this one. The durable fix belongs in the runtime: `System.Net` owns `url.full` on .NET 9+, and configurable query redaction should be solved there via [[dotnet/runtime#110018]](https://github.com/dotnet/runtime/issues/110018). Shipping a contrib-only knob would mean the same configuration behaves differently depending on target framework - active on .NET Framework and .NET 8, silently doing nothing on .NET 9+. That's confusing to explain and support, and it's a surface we'd carry and eventually deprecate once declarative configuration lands.


With .NET 8 reaching end of support this November, the practical window for it is also quite narrow.


Proposal: let's keep #1791 open and track [[dotnet/runtime#110018]](https://github.com/dotnet/runtime/issues/110018) as the primary path. 
@PetarJerinic, if you're up for it, pushing on the runtime issue would have a much broader impact than landing this here, happy to support that.

PetarJerinic:
@rajkumar-rangaraj Thanks for the feedback! 

I fully agree with the reasoning and am closing PR this in favour of https://github.com/dotnet/runtime/issues/110018. I'll analyse the runtime repo and post a proposal (with semconv in mind) in the linked issue. If you have a pov of what should be included in the proposal there, I am open to hearing it.

PR #4981, opened 2026-08-15
[perf] Avoid boxing port values

martincostello:
## Changes

Use a cache of common pre-boxed HTTP(S) port numbers to avoid boxing every time an `server.port` or `network.peer.port` attribute is added to telemetry.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.Http/Implementation/TelemetryHelper.cs
+++ b/src/OpenTelemetry.Instrumentation.Http/Implementation/TelemetryHelper.cs
@@ -10,6 +10,12 @@ internal static class TelemetryHelper
 {
     public static readonly (object, string)[] BoxedStatusCodes = InitializeBoxedStatusCodes();
 
+    private static readonly object Port80 = 80;
+    private static readonly object Port443 = 443;
+    private static readonly object Port8080 = 8080;
+
+    private static object? lastBoxedPort;
+
     public static object GetBoxedStatusCode(HttpStatusCode statusCode)
     {
         var intStatusCode = (int)statusCode;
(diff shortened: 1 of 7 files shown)


src/Shared/GrpcTagHelper.cs
```diff
@@ -29,6 +29,39 @@ internal static class GrpcTagHelper
     public const string GrpcStatusCodeTagName = "grpc.status_code";
     public const string GrpcTargetTagName = "grpc.target";
 
+    private static readonly object Port80 = 80;
+    private static readonly object Port443 = 443;
+    private static readonly object Port8080 = 8080;
+
+    private static object? lastBoxedPort;
+
+    public static object GetBoxedPort(int port)
```
stevejgordon, line 38: This looks like a duplication of the one in `TelemetryHelper`. Can we share this code?

src/OpenTelemetry.Instrumentation.Http/Implementation/TelemetryHelper.cs
```diff
@@ -10,6 +10,12 @@ internal static class TelemetryHelper
 {
     public static readonly (object, string)[] BoxedStatusCodes = InitializeBoxedStatusCodes();
 
+    private static readonly object Port80 = 80;
+    private static readonly object Port443 = 443;
+    private static readonly object Port8080 = 8080;
+
+    private static object? lastBoxedPort;
```
stevejgordon, line 17: Should this be marked `volatile`?

martincostello: I don't _think_ we need that for this as in most cases it'll have an unchanging value fairly soon after startup.

src/Shared/GrpcTagHelper.cs
martincostello: `TelemetryHelper` isn't used by the gRPC instrumentation and doesn't need any of the extra functionality in it, which would be dead code and a start-up hit from building the unused static table of boxed HTTP status codes.

I think in the case the minimal duplication is fine, rather than extracting these parts into a new shared helper and then re-using that everywhere.

rajkumar-rangaraj (approved):

PR #4986, opened 2026-08-17
[EFCore] Fix MySQL/MariaDB literal sanitization

martincostello:
## Changes

Sanitize SQL for MySQL and MariaDB when dollar-quoted literals, double-quoted literals and backslash-escaped single quotes are present.


src/Shared/SqlProcessor.cs
```diff
@@ -766,6 +790,21 @@ private static bool SanitizeStringLiteral(ReadOnlySpan<char> sql, Span<char> buf
                 }
 
                 searchPos += quoteIndex;
+
+                // Skip a backslash-escaped quote (\'). MySQL/MariaDB (with the default
+                // NO_BACKSLASH_ESCAPES disabled) treat a backslash as a string escape
+                // character, so a quote preceded by an odd number of backslashes does
+                // not terminate the literal. Without this a value such as 'a\'secret'
+                // would be mis-parsed and the trailing "secret" copied into the sanitized
```
martincostello:
```suggestion
                // would be incorrectly parsed and the trailing "secret" copied into the sanitized
```

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@
   `db.query.summary` length limit is reached.
   ([#4929](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4929))
 
+* Fixed query sanitization so that backslash-escaped quotes (`'a\'b'`) in
+  MySQL/MariaDB string literals and PostgreSQL dollar-quoted strings
+  (`$$...$$`) are correctly redacted.
+  ([#4985](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4985))
```
stevejgordon:
```suggestion
  ([#4986](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4986))
```

src/Shared/SqlProcessor.cs
```diff
@@ -793,6 +832,69 @@ private static bool SanitizeStringLiteral(ReadOnlySpan<char> sql, Span<char> buf
...
+            return false;
+        }
+
+        var delimiter = sql.Slice(start, tagEnd - start + 1);
+        var bodyStart = tagEnd + 1;
+
+        var closeOffset = sql.Slice(bodyStart).IndexOf(delimiter);
+        if (closeOffset < 0)
+        {
+            return false;
```
stevejgordon: Returning false here leaves `state.ParsePosition` unchanged, so the `$` character falls through to `ParseNextToken` and is emitted verbatim. The body of the unterminated literal then flows through subsequent iterations as plain identifier/keyword tokens and is not redacted.

`SanitizeStringLiteral` handles its equivalent unterminated case by advancing `state.ParsePosition` to `sql.Length` and emitting `?`, which keeps the security guarantee consistent. Worth doing the same here I think.

src/Shared/SqlProcessor.cs
```diff
@@ -143,35 +147,52 @@ private enum SqlKeyword
...
+    /// <param name="useBackslashEscapes">
+    /// <see langword="true"/> if the source database treats a backslash as a string-literal escape
+    /// character (MySQL and MariaDB with the default <c>NO_BACKSLASH_ESCAPES</c> mode disabled);
+    /// otherwise <see langword="false"/>.
+    /// </param>
+    /// <returns>The sanitized SQL and query summary.</returns>
+    public static SqlStatementInfo GetSanitizedSql(string? sql, bool useBackslashEscapes = false) =>
+        sql != null
+        ? useBackslashEscapes
+        ? GetSanitizedSql(sql, BackslashEscapeCache, ref approxBackslashEscapeCacheCount, useBackslashEscapes: true)
```
stevejgordon, line 167: If an application uses more tha one DB technology, one using backslash escapes and one not, I think we could end up with two caches up to 1000 entries each. Arguably, that may be okay, but we might need to document that, or maintain a shared count across both caches?

src/Shared/SqlProcessor.cs
```diff
@@ -143,35 +147,52 @@ private enum SqlKeyword
...
+    /// <param name="sql">The SQL statement to sanitize.</param>
+    /// <param name="useBackslashEscapes">
+    /// <see langword="true"/> if the source database treats a backslash as a string-literal escape
+    /// character (MySQL and MariaDB with the default <c>NO_BACKSLASH_ESCAPES</c> mode disabled);
+    /// otherwise <see langword="false"/>.
+    /// </param>
+    /// <returns>The sanitized SQL and query summary.</returns>
+    public static SqlStatementInfo GetSanitizedSql(string? sql, bool useBackslashEscapes = false) =>
+        sql != null
+        ? useBackslashEscapes
```
stevejgordon: This code is a little hard to read now with nested ternaries. Would an if/else be clearer? This may depend on the decision on cache count sharing.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
```diff
@@ -369,6 +375,16 @@ DbSystemNames.Sqlite or
         };
     }
 
+    internal static bool IsBackslashEscapeProvider(string? providerOrCommandName)
+    {
+        // MySQL and MariaDB (which use the MySQL providers) treat a backslash as a
+        // string-literal escape character unless the NO_BACKSLASH_ESCAPES SQL mode
+        // is enabled. The other supported engines follow the SQL standard where
+        // only a doubled quote ('') escapes a quote.
+        (_, var dbSystemName) = GetDbSystemNames(providerOrCommandName);
```
stevejgordon, line 391: Does anything actually detect if `NO_BACKSLASH_ESCAPES` is enabled or not and adjust the logic accordingly?

src/Shared/SqlProcessor.cs
martincostello: It ended up like this because the code I started with had "use expression bodied member", then once that was applied it had "simplify if statement", then that had "use ternary". I just kept accepting until the IDE stopped suggesting refactorings. I figured that was easier than having to suppress it to keep the simple `if`.

src/OpenTelemetry.Instrumentation.EntityFrameworkCore/Implementation/EntityFrameworkDiagnosticListener.cs
martincostello: No.

src/Shared/SqlProcessor.cs
martincostello: I'm not sure that's needed - if you have a multi-engine application and do enough SQL with it for there to be an internal cache with 2000 entries, that's probably neither here nor there in the grander scheme?

src/Shared/SqlProcessor.cs
martincostello:
```csharp
public static SqlStatementInfo GetSanitizedSql(string? sql, bool useBackslashEscapes = false)
{
    if (sql == null)
    {
        return default;
    }

    return useBackslashEscapes
        ? GetSanitizedSql(sql, BackslashEscapeCache, ref approxBackslashEscapeCacheCount, useBackslashEscapes: true)
        : GetSanitizedSql(sql, Cache, ref approxCacheCount, useBackslashEscapes: false);
}
```

image

then:

```csharp
public static SqlStatementInfo GetSanitizedSql(string? sql, bool useBackslashEscapes = false)
{
    return sql == null
        ? default
        : useBackslashEscapes
        ? GetSanitizedSql(sql, BackslashEscapeCache, ref approxBackslashEscapeCacheCount, useBackslashEscapes: true)
        : GetSanitizedSql(sql, Cache, ref approxCacheCount, useBackslashEscapes: false);
}
```

image

amirani8137:
Thanks for this — the double-quote fix in particular closes a real gap. One follow-up on the NO_BACKSLASH_ESCAPES point above: since both that and the ANSI_QUOTES assumption for double-quoted literals are documented rather than detected, would it be worth a brief note (comment or CHANGELOG) on which direction a wrong assumption fails in? I'd guess it only ever leads to safe over-redaction rather than a literal slipping through unredacted, but since this is security-sensitive, it seemed worth confirming explicitly rather than leaving it implicit.

matt-hensley (approved):

rajkumar-rangaraj (approved):

PR #5011, opened 2026-08-19
[Dynamic Control] Phase 3a: Add PolicyKey, SourceRegistrationId, and PolicySourceMetadata

stevejgordon:
Contributes to: #4742

## Changes

Adds `PolicyKey`, `SourceRegistrationId`, and `PolicySourceMetadata` identity types

Introduces internal identity types for keying policies and configured policy sources. This is intentially kept small for easier review before building on these types to introduce the `PolicyStore` in 3b.

### DIFF
--- a/src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
+++ b/src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
@@ -1,0 +1,116 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Identifies the slot that a telemetry policy occupies, across all configured sources.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Cross-source policy identity is the composite of the policy type and the
+/// provider-assigned policy identifier, rather than the identifier alone.
+/// </para>
+/// </remarks>
+internal readonly struct PolicyKey : IEquatable<PolicyKey>
+{
+    private readonly string? policyType;
+    private readonly string? policyId;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PolicyKey"/> struct.
+    /// </summary>
+    /// <param name="policyType">The policy type discriminator. Must not be null or whitespace.</param>
+    /// <param name="policyId">The provider-assigned policy identifier. Must not be null or whitespace.</param>
+    /// <exception cref="ArgumentException">Thrown when either component is null, empty, or whitespace.</exception>
+    public PolicyKey(string policyType, string policyId)
+    {
+        Guard.ThrowIfNullOrWhitespace(policyType, nameof(policyType));
+        Guard.ThrowIfNullOrWhitespace(policyId, nameof(policyId));
+
+        this.policyType = policyType;
+        this.policyId = policyId;
+    }
+
+    /// <summary>
+    /// Gets the policy type discriminator (e.g. <c>trace-sampling</c>).
+    /// </summary>
+    /// <remarks>
+    /// The constructor rejects a blank type. <see langword="default"/> instances bypass
+    /// every constructor, so they report an empty value rather than <see langword="null"/>.
+    /// </remarks>
+    public string PolicyType => this.policyType ?? string.Empty;
+
+    /// <summary>
+    /// Gets the opaque, provider-assigned policy identifier.
+    /// </summary>
+    /// <remarks>
+    /// The constructor rejects a blank identifier. <see langword="default"/> instances
+    /// bypass every constructor, so they report an empty value rather than
+    /// <see langword="null"/>.
+    /// </remarks>
+    public string PolicyId => this.policyId ?? string.Empty;
+
+    /// <summary>
+    /// Determines whether two <see cref="PolicyKey"/> instances identify the same policy.
+    /// </summary>
+    /// <param name="left">The first key to compare.</param>
+    /// <param name="right">The second key to compare.</param>
+    /// <returns><see langword="true"/> if the keys are equal; otherwise, <see langword="false"/>.</returns>
+    public static bool operator ==(PolicyKey left, PolicyKey right) => left.Equals(right);
+
+    /// <summary>
+    /// Determines whether two <see cref="PolicyKey"/> instances identify different policies.
+    /// </summary>
+    /// <param name="left">The first key to compare.</param>
+    /// <param name="right">The second key to compare.</param>
+    /// <returns><see langword="true"/> if the keys are not equal; otherwise, <see langword="false"/>.</returns>
+    public static bool operator !=(PolicyKey left, PolicyKey right) => !left.Equals(right);
+
+    /// <summary>
+    /// Creates the key identifying the slot that <paramref name="policy"/> occupies.
+    /// </summary>
+    /// <param name="policy">The policy to derive a key from.</param>
+    /// <returns>The key composed of the policy's type and identifier.</returns>
+    /// <remarks>
+    /// The returned key depends only on identity, never on policy content, so an updated
+    /// value for the same policy resolves to the same slot.
+    /// </remarks>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="policy"/> is null.</exception>
+    /// <exception cref="ArgumentException">
+    /// Thrown when the policy reports a blank type or identifier. Validated policy models
+    /// cannot, so this indicates a faulty <see cref="ITelemetryPolicy"/> implementation.
+    /// </exception>
+    public static PolicyKey FromPolicy(ITelemetryPolicy policy)
+    {
+        Guard.ThrowIfNull(policy);
+
+        return new PolicyKey(policy.PolicyType, policy.Id);
+    }
+
+    /// <inheritdoc/>
+    public bool Equals(PolicyKey other)
+        => string.Equals(this.policyType ?? string.Empty, other.policyType ?? string.Empty, StringComparison.Ordinal)
+            && string.Equals(this.policyId ?? string.Empty, other.policyId ?? string.Empty, StringComparison.Ordinal);
+
+    /// <inheritdoc/>
+    public override bool Equals(object? obj) => obj is PolicyKey other && this.Equals(other);
+
+    /// <inheritdoc/>
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            var hash = (17 * 31) + StringComparer.Ordinal.GetHashCode(this.PolicyType);
+            return (hash * 31) + StringComparer.Ordinal.GetHashCode(this.PolicyId);
+        }
+    }
+
+    /// <summary>
+    /// Returns a diagnostic representation of the key.
+    /// </summary>
+    /// <returns>The policy type and identifier, separated by a forward slash.</returns>
+    public override string ToString() => this.PolicyType + "/" + this.PolicyId;
+}
--- a/src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
+++ b/src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
@@ -101,11 +101,17 @@ internal readonly struct PolicyKey : IEquatable<PolicyKey>
     /// <inheritdoc/>
     public override int GetHashCode()
     {
+#if NET || NETSTANDARD2_1_OR_GREATER
+        return HashCode.Combine(
+            StringComparer.Ordinal.GetHashCode(this.PolicyType),
+            StringComparer.Ordinal.GetHashCode(this.PolicyId));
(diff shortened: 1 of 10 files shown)


src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
```diff
@@ -0,0 +1,116 @@
...
+        this.policyId = policyId;
+    }
+
+    /// <summary>
+    /// Gets the policy type discriminator (e.g. <c>trace-sampling</c>).
+    /// </summary>
+    /// <remarks>
+    /// The constructor rejects a blank type. <see langword="default"/> instances bypass
+    /// every constructor, so they report an empty value rather than <see langword="null"/>.
+    /// </remarks>
```
martincostello: I think we can remove the two comments about how `default` works.

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// <inheritdoc/>
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            var hash = (17 * 31) + StringComparer.Ordinal.GetHashCode(this.PolicyType);
+            return (hash * 31) + StringComparer.Ordinal.GetHashCode(this.PolicyId);
+        }
+    }
```
martincostello, line 112: Use `HashCode` or `HashCodeCombine()` for TFMs that support it.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceKind.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace OpenTelemetry.DynamicControl.Internal.Sources;
+
+/// <summary>
+/// Describes the transport or backing store that a configured policy source reads from.
+/// </summary>
+/// <remarks>
+/// The kind is diagnostic and grouping information only; it never identifies a source,
+/// because several sources of the same kind may be configured.
+/// </remarks>
+internal enum PolicySourceKind
```
martincostello, line 13: Something to consider later: when this becomes extensible and a user could write their own weird and wacky custom implementation, how would we model that with an enum that users can't extend?

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+/// during aggregation.
+/// </summary>
+internal readonly struct PolicySourceMetadata : IEquatable<PolicySourceMetadata>
+{
+    /// <summary>
+    /// The precedence applied to a source that does not specify one.
+    /// </summary>
+    /// <remarks>
+    /// Because a lower value wins, the default is the lowest possible precedence
+    /// (<see cref="int.MaxValue"/>) rather than zero. A source that does not opt into
```
martincostello: I would reword this slightly as it almost sounds like `int.MaxValue` is the lowest possible value, which it isn't in general terms.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// applies when a constructor call omits the priority argument; a
+    /// <see langword="default"/> instance bypasses the constructor entirely and reports a
+    /// <see cref="Priority"/> of zero instead. See <see cref="Priority"/> for details.
+    /// </remarks>
+    public const int DefaultPriority = int.MaxValue;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PolicySourceMetadata"/> struct.
+    /// </summary>
+    /// <param name="registrationId">The identity of the configured source. Must not be <see langword="default"/>.</param>
```
martincostello: Maybe create a well-known property that is equivalent to empty (like `Guid.Empty`) to make the code more self-descriptive when saying what it shouldn't be?

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+        int priority = DefaultPriority)
+    {
+        if (registrationId == default)
+        {
+            throw new ArgumentException("The source registration ID is required.", nameof(registrationId));
+        }
+
+        if (kind == PolicySourceKind.Unknown)
+        {
+            throw new ArgumentException("The source kind is required.", nameof(kind));
```
martincostello: Use `ArgumentOutOfRange` and include the invalid value. Also needs to guard against values outside the defined set.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// </summary>
+    /// <remarks>
+    /// A lower value takes precedence over a higher one, matching the provider-priority
+    /// convention from the Telemetry Policy OTEP (e.g. OpAmp=1, Http=2, File=3). Aggregation
+    /// resolves equal priorities deterministically rather than by update order; those rules
+    /// are defined with aggregation itself, not here. A constructor call that omits the
+    /// priority argument reports <see cref="DefaultPriority"/>; a <see langword="default"/>
+    /// instance bypasses every constructor, so it reports zero instead, the same as any
+    /// other field on a zero-initialized struct.
+    /// </remarks>
```
martincostello: Same here about documenting how `default` works for structs.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// <inheritdoc/>
+    public override int GetHashCode()
+    {
+        unchecked
+        {
+            var hash = (17 * 31) + this.RegistrationId.GetHashCode();
+            hash = (hash * 31) + (int)this.Kind;
+            return (hash * 31) + this.Priority;
+        }
```
martincostello, line 115: `HashCode` usage.

src/OpenTelemetry.DynamicControl/Internal/Sources/SourceRegistrationId.cs
```diff
@@ -0,0 +1,71 @@
...
+        this.value = value;
+    }
+
+    /// <summary>
+    /// Gets the opaque registration value.
+    /// </summary>
+    /// <remarks>
+    /// The constructor rejects a blank value. <see langword="default"/> instances bypass
+    /// every constructor, so they report an empty value rather than <see langword="null"/>.
+    /// </remarks>
```
martincostello: Same here.

test/OpenTelemetry.DynamicControl.Tests/PolicyKeyTests.cs
```diff
@@ -0,0 +1,218 @@
...
+    [Fact]
+    public void Equals_WithSameComponents_ReturnsTrue()
+    {
+        var left = new PolicyKey("trace-sampling", "policy-id");
+        var right = new PolicyKey("trace-sampling", "policy-id");
+
+        Assert.True(left.Equals(right), "Typed Equals should be true");
+        Assert.True(left.Equals((object)right), "Object Equals should be true");
+        Assert.True(left == right, "Equals (==) operator should be true");
+        Assert.False(left != right, "Not equals (!=) operator should be false");
```
martincostello: For symmetry, should assert the same with left and right swapped? (i.e. `1+2=2+1=3`)

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
stevejgordon: I considered this but didn't find a precedent for pre-processor directives for this scenario. I think a couple of other structs use this approach too. I don't mind doing it, just wonder if we do it as a wider consistency pass elsewhere too in a follow up?

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceKind.cs
stevejgordon: Good point! I forgot that the spec allows "custom" for sources. Let me think on this.

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
martincostello: I had a conversation with someone in the SDK repo about this yesterday, which resulted in https://github.com/open-telemetry/opentelemetry-dotnet/pull/7644.

If you copy that approach for the new ones you're adding, and any others can be done separately in a follow-up.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceKind.cs
stevejgordon: Actually, the set is bounded, so I just need to add the missing "Http" and "Custom" options. This mirrors how Java defines this using an enum also.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in b6b7c728ca0f5e262819ec99f6681e4ce000335e

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in b6b7c728ca0f5e262819ec99f6681e4ce000335e

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceKind.cs
stevejgordon: Addressed in b6b7c728ca0f5e262819ec99f6681e4ce000335e

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
stevejgordon: Addressed in b6b7c728ca0f5e262819ec99f6681e4ce000335e

src/OpenTelemetry.DynamicControl/Internal/Sources/SourceRegistrationId.cs
stevejgordon: Addressed in b6b7c728ca0f5e262819ec99f6681e4ce000335e

test/OpenTelemetry.DynamicControl.Tests/PolicyKeyTests.cs
stevejgordon: Addressed in e2cc2ec9ecb25304004637432aa602300148edbd

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in e2cc2ec9ecb25304004637432aa602300148edbd

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in e2cc2ec9ecb25304004637432aa602300148edbd

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in e2cc2ec9ecb25304004637432aa602300148edbd

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
stevejgordon: Addressed in e2cc2ec9ecb25304004637432aa602300148edbd

martincostello (approved):

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKey.cs
martincostello: Looking across the code base, it seems the only candidate to use it is this one:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/9f928922543089e38fa7c577424f7684650c5d99/src/OpenTelemetry.Exporter.Geneva/External/LinuxTracepoints-Net/Provider/EventHeaderDynamicProvider.cs#L269-L272

All the other overrides already use `HashCode`, or delegate to a single value's hash code.

PR #5017, opened 2026-08-20
[Dynamic Control] Phase 3b: Add PolicyStore

stevejgordon:
Contributes to: #4742

## Changes

Continues the Dynamic Control component implementation with the policy store layer.

- Replaces the `ITelemetryPolicy` interface with an abstract `TelemetryPolicy` base class, enabling shared behaviour and a cleaner extension point for concrete immutable policy types.
- Reorders `PolicySourceKind` to match the provider-priority order defined in the Telemetry Policy OTEP.
- Adds `PolicySourceVersion` for sequence-based versioning of per-source policy data.
- Adds an internal copy-on-write `PolicyStore` built on immutable snapshots with:
  - Stale-write rejection via sequence numbers
  - Version-based suppression to skip no-op updates
  - Metadata pinning (source registration metadata captured at write time)
  - Deterministic policy and source ordering via `PolicyKeyComparer`
  - Atomic snapshot replacement for lock-free reads


stevejgordon:
Regarding the codecov. The missing line is technically unreachable but required to satisfy CS8509. Three options:

1. Ignore it (but codecov % can slip over time if we have more similar scenarios that we add.
2. Use ExcludeFromCodeCoverage on the method, but any future arms could be missed from testing without codecov detection.
3. Use a reflection-based test which is a slight test hack but would cover the actual code. Only issue is it's tied to the method name etc.

Do you have a preference @martincostello?

martincostello:
It's fine to not have 100% coverage, particularly for stuff that can't be reasonably covered or is just to hit a number.

It's just for new code, particularly when you're using an agent, I figured aiming for higher coverage is reasonable. In the previous PR I think there was a number of non-trivial paths that didn't have coverage even though it was still 90%+.

Also even though it's an ❌, it's just to draw attention to it - it's not a hard requirement we enforce.

src/OpenTelemetry.DynamicControl/CHANGELOG.md
```diff
@@ -11,4 +11,14 @@
   configured sources.
   ([#5011](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5011))
 
+* Replaced internal `ITelemetryPolicy` with an abstract `TelemetryPolicy`
+  base class. Reordered `PolicySourceKind` to match the Telemetry Policy OTEP
+  provider-priority order.
+  ([#5017](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5017))
```
martincostello: The internal-refactoring part probably isn't worth including in the change notes for external users. I think just documenting the incremental features being added is fine.

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKeyComparer.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Compares <see cref="PolicyKey"/> instances by type then ID.
+/// </summary>
+internal sealed class PolicyKeyComparer : IComparer<PolicyKey>
```
martincostello: I wonder if it's worth making this an `IEqualityComparer<PolicyKey>` too, and moving the implementation out of the type itself to here. Then `IEquatable<T>` can delegate to the singleton. Could then similarly add `IComparable<PolicyKey>` to the type.

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKeyComparer.cs
```diff
@@ -0,0 +1,30 @@
...
+    // "a deterministic order for processing", not "one policy key is less than another".
+
+    private PolicyKeyComparer()
+    {
+    }
+
+    /// <summary>
+    /// Gets the singleton instance of <see cref="PolicyKeyComparer"/>.
+    /// </summary>
+    public static PolicyKeyComparer Instance { get; } = new();
```
martincostello: Maybe rename to `Default` for symmetry with `{Equality}Comparer<T>.Default`?

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -38,20 +32,23 @@ namespace OpenTelemetry.DynamicControl.Internal.Sources;
...
-            priority,
-            "The priority must be non-negative. Per the Telemetry Policy OTEP, lower values represent higher precedence (e.g. OpAmp=1, Http=2, File=3).",
-            nameof(priority));
+        if (priority.HasValue)
+        {
+            Guard.ThrowIfNegative(
+                priority.Value,
+                "The priority must be non-negative. Per the Telemetry Policy OTEP, lower values represent higher precedence (e.g. OpAmp=1, Http=2, File=3).",
+                nameof(priority));
+        }
```
martincostello:
```suggestion
        if (priority is { } priorityValue)
        {
            Guard.ThrowIfNegative(
                priorityValue,
                "The priority must be non-negative. Per the Telemetry Policy OTEP, lower values represent higher precedence (e.g. OpAmp=1, Http=2, File=3).",
                nameof(priority));
        }
```

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
```diff
@@ -122,4 +119,13 @@ public override int GetHashCode()
...
     public override string ToString()
         => this.RegistrationId.Value + "/" + this.Kind + "/" + this.Priority;
+
+    private static int KindDefaultPriority(PolicySourceKind kind) => kind switch
+    {
+        PolicySourceKind.OpAmp => 1,
+        PolicySourceKind.Http => 2,
+        PolicySourceKind.File => 3,
+        PolicySourceKind.Custom => 1000,
+        _ => throw new ArgumentOutOfRangeException(nameof(kind), kind, null),
```
martincostello: For `#if NET` could consider using `UnreachableException`.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStoreUpdateStatus.cs
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.DynamicControl.Internal.Sources;
+
+namespace OpenTelemetry.DynamicControl.Internal.Store;
+
+/// <summary>
+/// Describes the outcome of a single submission to the policy store.
+/// </summary>
+internal enum PolicyStoreUpdateStatus
```
martincostello, line 11: Add a `None`/`Unknown` value for 0?

src/OpenTelemetry.DynamicControl/OpenTelemetry.DynamicControl.csproj
```diff
@@ -20,7 +20,12 @@
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)\src\Shared\Guard.cs" Link="Includes\Guard.cs" />
+    <Compile Include="$(RepoRoot)\src\Shared\Lock.cs" Link="Includes\Lock.cs" />
     <Compile Include="$(RepoRoot)\src\Shared\NullableAttributes.cs" Link="Includes\NullableAttributes.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.Collections.Immutable" />
```
martincostello, line 28: Is this needed for all TFMs?

test/OpenTelemetry.DynamicControl.Tests/TelemetryPolicyImmutabilityTests.cs
```diff
@@ -0,0 +1,107 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Reflection;
+using OpenTelemetry.DynamicControl.Internal.Policies;
+
+namespace OpenTelemetry.DynamicControl.Tests;
+
+public class TelemetryPolicyImmutabilityTests
```
martincostello, line 9: Maybe this is more of "Specification" tests (which happens to require that things are immutable, amongst other things (like being `sealed`))?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.DynamicControl.Internal.Store;
+
+/// <summary>
+/// One policy source's complete, validated, immutable policy set at a single point in time.
+/// </summary>
+internal sealed class PolicySourceSnapshot
+{
+    private readonly Dictionary<PolicyKey, TelemetryPolicy> lookup;
```
martincostello, line 22: Does this need to be safe for concurrent reads while something else adds new entries?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+        this.lookup = lookup;
+    }
+
+    /// <summary>
+    /// Gets the metadata that fully describes the source that produced this snapshot.
+    /// </summary>
+    public PolicySourceMetadata Metadata { get; }
+
+    /// <summary>
+    /// Gets the registration identity of the source. Equivalent to <see cref="Metadata"/>.RegistrationId.
```
martincostello: It's a bit weird to `.Prop` a `<see>` like this IMHO.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// lifecycle event).
+    /// </remarks>
+    public bool IsEmpty => this.Policies.IsEmpty;
+
+    /// <summary>
+    /// Attempts to create a new <see cref="PolicySourceSnapshot"/> from the supplied inputs.
+    /// </summary>
+    /// <param name="metadata">
+    /// The metadata of the source. Must not be <see langword="default"/>; a default value
+    /// would bypass all constructor validation and could assign this snapshot top precedence.
```
martincostello:
```suggestion
    /// The metadata of the source. Must not be <see langword="default"/>.
```

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// <param name="error">
+    /// When this method returns <see langword="false"/>, a description of the validation
+    /// failure suitable for reporting back to the transport or coordinator. When
+    /// <see langword="true"/>, <see langword="null"/>.
+    /// </param>
+    /// <returns>
+    /// <see langword="true"/> if the snapshot was created successfully; <see langword="false"/>
+    /// if validation failed.
+    /// </returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="policies"/> is null.</exception>
```
martincostello:
```suggestion
    /// <exception cref="ArgumentNullException"><paramref name="policies"/> is null.</exception>
```

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+        [NotNullWhen(true)] out PolicySourceSnapshot? snapshot,
+        [NotNullWhen(false)] out string? error)
+    {
+        Guard.ThrowIfNull(policies);
+
+        if (metadata.Equals(default))
+        {
+            snapshot = null;
+            error = "The metadata must not be a default PolicySourceMetadata instance. " +
+                    "A default value bypasses constructor validation and would assign this snapshot top precedence.";
```
martincostello: Not sure we need this part.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            snapshot = null;
+            error = $"The sequence must be greater than or equal to 1 (was {sequence}).";
+            return false;
+        }
+
+        var count = policies.Count;
+        var keys = new PolicyKey[count];
+        var policyItems = new TelemetryPolicy[count];
+        var lookupDict = new Dictionary<PolicyKey, TelemetryPolicy>(count);
```
martincostello: `policyMap`?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            var key = new PolicyKey(policy.PolicyType, policy.Id);
+            if (lookupDict.ContainsKey(key))
+            {
+                snapshot = null;
+                error = $"Duplicate policy key: {key}. A policy snapshot must not contain two policies with the same PolicyType and Id.";
+                return false;
+            }
+
+            lookupDict[key] = policy;
```
martincostello: Use `!TryAdd()` to condense into one operation?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// by a concurrent update and produce incorrect results.
+    /// </remarks>
+    public PolicyStoreSnapshot Current => Volatile.Read(ref this.current);
+
+    /// <summary>
+    /// Replaces the policy set for the source described by <paramref name="snapshot"/>.
+    /// The submission is evaluated against metadata consistency, sequence staleness, and
+    /// version suppression gates before being applied.
+    /// </summary>
+    /// <param name="snapshot">The new snapshot to commit. Must not be null.</param>
```
martincostello: I think we avoid stating the requirement given we're using nullable reference types (plus we have the `<exception>` docs for completeness.

```suggestion
    /// <param name="snapshot">The new snapshot to commit.</param>
```

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
```diff
@@ -0,0 +1,143 @@
...
+                return new PolicyStoreUpdateResult(PolicyStoreUpdateStatus.RejectedStaleSequence, this.current);
+            }
+
+            // Gate 3: version suppression. Advance the maximum sequence even on
+            // suppression to prevent a lower-sequence later submission from winning.
+            if (!snapshot.Version.IsEmpty
+                && existing != null
+                && snapshot.Version.Equals(existing.Version))
+            {
+                this.maxSequence[id] = snapshot.Sequence;
```
martincostello, line 89: Is this actually _advancing_ something? It just looks like it's assigning whatever it got from the caller, which doesn't necessarily mean the number increased?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This is a configuration lifecycle operation. It must not be called in response to
+    /// a transport disconnect; a disconnect is not a retraction, and calling
+    /// <see cref="RemoveSource"/> in response to one would clear the effective policies
+    /// and leave downstream samplers without configuration until the next connection.
+    /// </para>
+    /// <para>
+    /// After removal, no tombstone is retained. Re-adding the same
```
martincostello: I don't think we've used the term "tombstone" anywhere else so far?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
```diff
@@ -0,0 +1,143 @@
...
+        Guard.ThrowIfDefault(registrationId);
+
+        lock (this.updateLock)
+        {
+            if (!this.sources.ContainsKey(registrationId))
+            {
+                return new PolicyStoreUpdateResult(PolicyStoreUpdateStatus.SourceNotFound, this.current);
+            }
+
+            this.sources.Remove(registrationId);
```
martincostello:
```suggestion
            if (!this.sources.Remove(registrationId))
            {
                return new PolicyStoreUpdateResult(PolicyStoreUpdateStatus.SourceNotFound, this.current);
            }

```

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStoreSnapshot.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </remarks>
+    public long Revision { get; }
+
+    /// <summary>
+    /// Gets the current source snapshots, sorted by ordinal <see cref="SourceRegistrationId.Value"/>.
+    /// </summary>
+    /// <remarks>
+    /// Ordering is by identity, not by <see cref="PolicySourceMetadata.Priority"/>. Priority
+    /// ordering is an aggregation decision reserved for Phase 6; identity ordering is
+    /// neutral and stable, giving Phase 6 a deterministic starting order to sort from.
```
martincostello: I don't think we need the "Phase 6" stuff in the `///` comments.

src/OpenTelemetry.DynamicControl/CHANGELOG.md
stevejgordon: It was only to correct the prior change showing "Added internal `ITelemetryPolicy`". Right now, everything is mostly internal, so arguably we could forgoe any changelogs until we make stuff public, but I did wonder if a short history is useful for following the main changes during development. I'll refine this change and refer to the new internal bits for now, but happy to discuss the best direction while this is pre-release.

martincostello: I think noting the incremental change is fine, I just think it doesn't need to be too detailed, as ultimately on the first release users will just see what it does as a snapshot, so don't necessarily need to see all the changes along the way (like type renaming).

test/OpenTelemetry.DynamicControl.Tests/TelemetryPolicyImmutabilityTests.cs
stevejgordon: Currently all test constraints exist to enforce and protect the immutability guarantee for lock-free sharing across snapshots. Personally I'd stick with a clearer name unless we expanded to other specification elements beyond immutability.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: The copy-on-write model means each `PolicySourceSnapshot` is fully constructed before being published via `Volatile.Write`. The lookup dictionary is populated entirely inside `TryCreate` before the constructor runs and is never mutated after. Concurrent reads of a published snapshot are safe by construction. I'll add a note to the XML docs.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: I thought it worthwhile to describe a non-obvious security rule.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStoreSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStore.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/OpenTelemetry.DynamicControl.csproj
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyStoreUpdateStatus.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicySourceMetadata.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKeyComparer.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/Internal/Policies/PolicyKeyComparer.cs
stevejgordon: Addressed in 3b62d27

src/OpenTelemetry.DynamicControl/CHANGELOG.md
stevejgordon: Addressed in 3b62d27

test/OpenTelemetry.DynamicControl.Tests/TelemetryPolicyImmutabilityTests.cs
martincostello: Sealed doesn't seem like immutability to me, but we can leave it as is.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
martincostello: Would accepting an `IReadOnlyDictionary<K, V>` make it more obvious? The question mainly came about because a dictionary is being passed in from somewhere else, so there's no guarantee from reading this code that the caller won't have a reference to it and mutate it.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
```diff
@@ -13,6 +13,11 @@ namespace OpenTelemetry.DynamicControl.Internal.Store;
 /// <summary>
 /// One policy source's complete, validated, immutable policy set at a single point in time.
 /// </summary>
+/// <remarks>
+/// Once published by <see cref="PolicyStore"/>, this instance is read-only.
+/// The <c>lookup</c> dictionary is populated once during construction and never modified.
```
martincostello: If this type gets made public later it doesn't seem appropriate to refer to private fields. Maybe make it more generic, or move the comment inside the type as a `//` comment?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
martincostello: That's not a _security_ boundary though, is it? I would have thought it enough to tell the caller they're "holding it wrong" and not have to explain why (the internal design reasons).

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Good point! Actually, it might be even safer to make a copy in the ctor.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Fair point, but it's a required property for the immutability guarantee to hold across the type hierarchy. Happy to rename to TelemetryPolicyContractTests perhaps?

martincostello: Sure, that works for me.

stevejgordon:
> That's not a security boundary though, is it?

It's a correctness boundary, `default` has no registration identity and its default priority value 0 would place the snapshot at the top of the priority ordering, silently corrupting the store. I'll look to rephrase

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in a10b164

src/OpenTelemetry.DynamicControl/Internal/Store/PolicySourceSnapshot.cs
stevejgordon: Addressed in a10b164

martincostello (approved):

PR #5108, opened 2026-08-21
[Sampler.BottomFloor] Add fixed-memory bottom-k log sampler

Yun-Ting:
Design discussion issue #5119

## Changes

Adds `OpenTelemetry.Sampler.BottomFloor`, a fixed-memory, self-calibrating
bottom-k sampler for log records.

Chatty services have no good options today: raising the log level loses the rare
error buried among a flood of routine records, fixed-rate head sampling loses it
too and reports nothing about what was dropped, and per-logger rate limiting
silently truncates whatever happens to be loudest. This component bounds export
volume while keeping rare callsites representable, and stamps every retained
record with the arithmetic needed to reconstruct the original counts.

### How it works

Each arrival is assigned an exponential priority scaled by its callsite's
weight, and the reservoir retains the `k + 1` smallest. When the window closes,
the `k` best form the sample and the remaining entry becomes the inclusion
threshold. Per-callsite estimates then set the next window's weights, so
frequent callsites are pushed down and rare or newly seen ones keep a high
chance of inclusion. The only required setting is the per-window budget `k`.

Retained records carry `otel.logs.adjusted_count` (the reciprocal of the
record's inclusion probability) and `otel.logs.cv2` (an adequacy signal for the
estimate), so downstream aggregation switches from `COUNT(*)` to
`SUM(otel.logs.adjusted_count)` and stays unbiased.

### What is included

The public API is deliberately small - 14 entries in total:

* `BottomFloorLogExporter` - decorates a `BaseExporter<LogRecord>` and samples
  each export batch.
* `BottomFloorLogRecordProcessor` - the ready-to-register form, combining
  batching and sampling. One export batch is one sampling window.
* `BottomFloorLogSamplerOptions` - `Budget` plus the two attribute names.

The sampling algorithm itself is internal. It is a general bottom-k stream
sampler with no OpenTelemetry dependency, so exposing it publicly is worth doing
eventually, but it is not needed to sample logs and it would roughly quadruple
the public surface under review here.

### Deliberately deferred

Two features were removed from this PR so that everything shipped is covered by
tests. Both are tracked as later increments in #5119:

* Per-span log coverage (`MaxLogsPerSpanPerWindow`), which bounds how many logs
  a single chatty span contributes.
* Public access to `BottomFloorSampler<TCallsite>` for non-log streams.

### Notes for reviewers

* **Compression ratio under load is `Budget / maxExportBatchSize`.** A window no
  larger than the budget keeps everything it holds, so the constructor rejects
  that combination rather than silently no-op sampling. Quiet windows pass
  through whole, which is the "bottom floor" behaviour.
* **Two dependencies on SDK internals**, both isolated and both degrading safely:
  * `LogRecord.Copy()` for retention. Records are pooled and reclaimed as the
    batch enumerator advances, so anything held past that visit must be copied.
    If it cannot be bound, the exporter forwards every batch **unsampled**
    rather than emitting recycled records.
  * `BaseExporter<T>.ParentProvider` forwarding, so a decorated exporter still
    resolves its `Resource`. Uses `UnsafeAccessor` on .NET 9+ and reflection
    earlier. If it cannot be bound the inner exporter is left without a
    provider, which surfaces at export time rather than silently.

  `BaseExporter<T>` exposes no supported seam for the second one: the setter is
  internal, there is no virtual hook, and no constructor overload. The one
  supported alternative - wrapping a `BaseProcessor<T>` and subclassing
  `CompositeProcessor<T>`, which does forward to children - was rejected because
  it would give up the batching and flush timer inherited from
  `BatchExportProcessor` and would not remove the `LogRecord.Copy()` dependency
  anyway. Rationale is recorded in `DESIGN.md`; an upstream API request is
  planned.
* **39 tests run on `net462`, `net8.0`, `net9.0` and `net10.0`** - the
  `net462` leg is what exercises the downlevel reflection paths. The few
  uncovered lines are all the "SDK internals could not be bound" degradation
  paths, which by construction only execute if a future SDK removes the members
  they probe for; each is guarded by a test that fails if the member becomes
  unreachable.
* **README snippets are compiled and executed as tests** (`ReadmeExampleTests`),
  and a test asserts the README still matches the compiled source character for
  character. Editing a snippet therefore requires editing both files.
* Sampling accuracy is exercised end to end by `examples/bottom-floor`, which
  recovers the total arrival count to 0.22% at 10x compression, and every
  individual callsite to within a few percent.


martincostello:
If we were to add this, I think it needs more test coverage.

Yun-Ting:
> If we were to add this, I think it needs more test coverage.

Good call, @martincostello! 
I've added a bunch more test coverage. 
I've also scoping down the PR and opened #5119 for the design discussion, where we can settle how this package got introduced before it lands.

Thank you.

Yun-Ting:
cc. @rajkumar-rangaraj

martincostello:
Moved to draft until #5119 is agreed.

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        var summary = this.sampler.CloseWindow();
+
+        var kept = new LogRecord[summary.KeptItems.Count];
+        var count = 0;
+        foreach (var item in summary.KeptItems)
+        {
+            if (!this.buffer.TryGetValue(item.Token, out var record))
+            {
+                continue;
```
jmacd, line 123: What would cause this `continue`?

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        return this.innerExporter.Export(new Batch<LogRecord>(kept, count));
+    }
+
+    /// <summary>
+    /// Computes the stable callsite identity used to key the sampler. It is a
+    /// 64-bit FNV-1a hash over the log category, the numeric event id, and the
+    /// event name, which together identify the emitting statement independently
+    /// of the formatted message.
```
jmacd, line 150: (I'm new to C# types of question)

Is there a dotnet standard hash function, to avoid an inline FNV hash?
I believe there is an underlying identifier that the dotnet runtime has for identifying log statements--is the OpenTelemetry SDK obscuring that IOW forcing us to recompute an identifier? (e.g., like in Rust the Callsite [Identifier](https://docs.rs/tracing-core/0.1.36/tracing_core/callsite/struct.Identifier.html) directly supports hashing.

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogRecordProcessor.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <param name="exporter">
+    /// The exporter that receives the kept, adjusted-count-stamped records. The
+    /// processor takes ownership of it and disposes it.
+    /// </param>
+    /// <param name="options">The sampler options, including the budget.</param>
+    /// <param name="maxExportBatchSize">
+    /// The export size: the maximum number of records a single window holds
+    /// before it closes and is sampled. Must be greater than
+    /// <see cref="BottomFloorLogSamplerOptions.Budget"/>, since a window no larger
+    /// than the budget keeps everything it holds. The default is 2048.
```
jmacd, line 44: I was confused by this at first. I read forward and think I understand; this is the maximum size of the "population" being sampled in one round, it determines the worst-case sampling rate. When arrivals reach 2048/5000ms that is we begin flushing (I think blocking) so that records will begin dropping indiscriminately under sustained load above this many per interval.
```suggestion
    /// The export size: the maximum number of records a single window holds
    /// before it closes and is sampled. Must be greater than
    /// <see cref="BottomFloorLogSamplerOptions.Budget"/>, since a window no larger
    /// than the budget keeps everything it holds. Budget divided by the export size sets
    /// the worst-case sampling probability. The default is 2048.
```

jmacd: Maybe consider calling this population size?

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+    {
+    }
+
+    internal BottomFloorLogExporter(BaseExporter<LogRecord> innerExporter, BottomFloorLogSamplerOptions options, Random? random)
+    {
+        this.innerExporter = innerExporter ?? throw new ArgumentNullException(nameof(innerExporter));
+
+        Guard.ThrowIfNull(options);
+
+        if (options.Budget < 1)
```
jmacd, line 62: nit: I'm not sure the algorithm is a good choice when Budget == 1. I would say < 10 sounds like a bad idea, even. If you wanted Budget == 1, you'd choose an unweighted sampling algorithm.

examples/bottom-floor/README.md
```diff
@@ -0,0 +1,24 @@
+# Bottom-Floor sampler example
```
jmacd, line 1: Cool, for the record I see

```
Windows=300  logs=300000  budget=100/window
Forwarded=30000  compression=10.0x
Recovery (sum of otel.logs.adjusted_count vs actual arrivals)
  callsite         p(theory)   share   actual    estimate   rel.err
  App.Callsite01    0.3550   0.355    106392    102401.1  -3.75 %
  App.Callsite02    0.1656   0.166     49677     49003.4  -1.36 %
  App.Callsite03    0.1060   0.106     31761     31064.4  -2.19 %
  App.Callsite04    0.0773   0.077     22998     22706.2  -1.27 %
  App.Callsite05    0.0604   0.061     18222     18088.7  -0.73 %
  App.Callsite06    0.0495   0.050     14948     14991.6   0.29 %
  App.Callsite07    0.0417   0.042     12559     12652.9   0.75 %
  App.Callsite08    0.0360   0.036     10652     10867.5   2.02 %
  App.Callsite09    0.0317   0.032      9719      9720.9   0.02 %
  App.Callsite10    0.0282   0.028      8459      8717.5   3.06 %
  App.Callsite11    0.0254   0.025      7607      7402.9  -2.68 %
  App.Callsite12    0.0231   0.023      7006      7003.8  -0.03 %
  TOTAL             1.0000   1.000    300000    294621.2  -1.79 %
```

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+        // seeded positive, so the priority key below never divides by zero.
+        Debug.Assert(weight > 0.0, "callsite weight must be strictly positive");
+
+        // Non-cryptographic randomness is intentional: this is statistical sampling,
+        // not a security context, and the fast path must stay cheap per arrival.
+#pragma warning disable CA5394 // Do not use insecure randomness
+        var u = 1.0 - this.random.NextDouble();
+#pragma warning restore CA5394
+
+        var key = -Math.Log(u) / weight;
```
jmacd, line 118:
```suggestion
        // Exponential-weight Bottom-k math
        var key = -Math.Log(u) / weight;
```

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        if (this.count < this.capacity)
```
jmacd, line 120:
```suggestion
        // Fill the initial reservoir
        if (this.count < this.capacity)
```

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+    /// </param>
+    public BottomFloorSampler(int budget, Random? random = null)
+    {
+        if (budget < 1)
+        {
+            throw new ArgumentOutOfRangeException(nameof(budget), budget, "Budget must be at least one.");
+        }
+
+        this.budget = budget;
+        this.capacity = budget + 1;
```
jmacd, line 74:
```suggestion
        // Bottom-k uses a heap of size (budget+1)
        this.capacity = budget + 1;
```

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+        var key = -Math.Log(u) / weight;
+
+        if (this.count < this.capacity)
+        {
+            var token = this.nextToken++;
+            this.HeapPush(new Entry(key, callsite, token));
+            return OfferOutcome.Admit(token);
+        }
+
+        if (key < this.heap[0].Key)
```
jmacd, line 127:
```suggestion
        // Replace the max-key and re-heapify
        if (key < this.heap[0].Key)
```

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+            // its count is exact: an infinite threshold gives an inclusion
+            // probability of one and leaves the Horvitz-Thompson count uncorrected.
+            threshold = double.PositiveInfinity;
+            sampleSize = this.count;
+        }
+        else
+        {
+            // The reservoir holds k+1 keys; the largest is the boundary and the
+            // remaining k form the sample.
+            var boundary = this.HeapPopMax();
```
jmacd, line 162: minor: you can avoid the final heapify call here, meaning just set threshold to `this.heap[0].Key` and use `for (var i = 1; ...` below.

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+            this.Callsite = callsite;
+            this.Token = token;
+        }
+
+        public double Key { get; }
+
+        public TCallsite Callsite { get; }
+
+        public long Token { get; }
+    }
```
jmacd, line 294: 😂 no standard heap library? (Same is true in Rust, I'm thinking of the [Golang Heap](https://pkg.go.dev/container/heap).)

src/OpenTelemetry.Sampler.BottomFloor/CallsiteEstimate.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// <summary>
+    /// Gets the number of the callsite's arrivals retained in the sample, <c>m</c>.
+    /// </summary>
+    public long KeptCount { get; }
+
+    /// <summary>
+    /// Gets the sampling weight applied to the callsite this window, <c>w</c>.
+    /// </summary>
+    public double Weight { get; }
```
jmacd, line 41: I think Weight is not useful to consumers?

src/OpenTelemetry.Sampler.BottomFloor/CallsiteEstimate.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// <summary>
+    /// Gets the per-arrival inclusion probability <c>1 - exp(-tau w)</c> the count
+    /// is corrected by. It is one when the window's reservoir never filled.
+    /// </summary>
+    public double InclusionProbability { get; }
+
+    /// <summary>
+    /// Gets the unbiased Horvitz-Thompson count estimate <c>nhat = m / inclusion</c>,
```
jmacd, line 50:
```suggestion
    /// Gets the unbiased Horvitz-Thompson count estimate <c>nhat = KeptCount / inclusion</c>,
```

src/OpenTelemetry.Sampler.BottomFloor/CallsiteEstimate.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// <summary>
+    /// Gets the sampling weight applied to the callsite this window, <c>w</c>.
+    /// </summary>
+    public double Weight { get; }
+
+    /// <summary>
+    /// Gets the per-arrival inclusion probability <c>1 - exp(-tau w)</c> the count
+    /// is corrected by. It is one when the window's reservoir never filled.
+    /// </summary>
+    public double InclusionProbability { get; }
```
jmacd, line 47: Afaict, this value is not used directly, but we often use its inverse. I wonder if we should store the inverse, I would call AdjustedCount, which is the per-event adjustement. Then EstimatedCount = KeptCount * AdjustedCount and each record gets otel.logs.adjusted_count = AdjustedCount. I believe this is fewer divisions, more multiplications. Not sure you need to store both InclusionProbability and AdjustedCount.

src/OpenTelemetry.Sampler.BottomFloor/DESIGN.md
```diff
@@ -0,0 +1,171 @@
...
+internals access at all.
+
+That restructuring was not taken because it trades one coupling for a larger
+redesign: the component would give up the batching and flush timer it inherits
+from `BatchExportProcessor` and reimplement windowing on top of `OnEnd`, and the
+exporter-decorator form would have to be dropped. Crucially it would not make the
+component internals-free, because retention still depends on the internal
+`LogRecord.Copy()` and no public equivalent exists. The coupling is therefore
+kept, but confined to one class, degraded to a no-op on failure, and covered by
+a test that fails if the member can no longer be reached.
```
jmacd, line 171:
```suggestion
a test that fails if the member can no longer be reached.

### References

[Edith Cohen and Haim Kaplan. 2007. Bottom-k sketches: better and more efficient estimation of aggregates. SIGMETRICS Perform. Eval. Rev. 35, 1 (June 2007), 353–354. https://dl.acm.org/doi/10.1145/1269899.1254926](https://dl.acm.org/doi/10.1145/1269899.1254926), 

[Daniel Ting. 2022. Adaptive Threshold Sampling. In Proceedings of the 2022 International Conference on Management of Data (SIGMOD '22). Association for Computing Machinery, New York, NY, USA, 1612–1625. https://doi.org/10.1145/3514221.3526122](https://doi.org/10.1145/3514221.3526122)

```

src/OpenTelemetry.Sampler.BottomFloor/ParentProviderPropagation.cs
```diff
@@ -0,0 +1,130 @@
+// Copyright The OpenTelemetry Authors
```
jmacd, line 1: I have the feeling this is copied from somewhere else?

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+/// export, so exporters that resolve their <c>Resource</c> from the provider work
+/// when wrapped.
+/// </summary>
+public sealed class BottomFloorLogExporter : BaseExporter<LogRecord>
+{
+    private readonly BaseExporter<LogRecord> innerExporter;
+    private readonly BottomFloorSampler<long> sampler;
+    private readonly string adjustedCountAttribute;
+    private readonly string squaredCvAttribute;
+    private readonly Dictionary<long, LogRecord> buffer = new();
```
jmacd, line 40: I was going to suggest naming this `reservoir`. I now understand that this is keyed by a Token (long) which is the insertion sequence number for the window. This Token (long) is not the same as TCallsite (long) .

This could be a suggestion for a future PR--or it could be me thinking about .NET with a beginner's mindset: I think about combining "buffer" into "sampler", to remove the concept of Token entirely, just to let the sampler's heap store LogRecord values directly, then there will be no `Remove()` call needed, the eviction removes the item.

(If I understand this correctly,) I think if the heap would store the pair `(LogRecord, TCallsite)` then you could eliminate the `KeptItems` array entirely.

Apologies if I'm mis-understanding how .NET works here!

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    private void Stamp(LogRecord record, in RecordStamp stamp)
+    {
+        var existing = record.Attributes;
+        var attributes = new List<KeyValuePair<string, object?>>((existing?.Count ?? 0) + 2);
+        if (existing != null)
+        {
+            attributes.AddRange(existing);
+        }
```
jmacd, line 234: Suggest not to allocate this unless `stamp.StreamAdjustedCount > 1.0`. If == 1.0, return without modifying `record.Attributes`.

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
```diff
@@ -0,0 +1,256 @@
...
+
+        // An adjusted count of one means the record was fully included, so it
+        // carries no information and is omitted. The variance companion is only
+        // meaningful when the record was actually subsampled.
+        if (stamp.StreamAdjustedCount != 1.0)
+        {
+            attributes.Add(new KeyValuePair<string, object?>(this.adjustedCountAttribute, stamp.StreamAdjustedCount));
+            if (stamp.StreamAdjustedCount > 1.0)
```
jmacd, line 242: Is StreamAdjustedCount ever less than 1.0? I suspect not, so != 1.0 is the same as > 1.0?

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
```diff
@@ -0,0 +1,295 @@
...
+
+            var theta = threshold * weight;
+            var inclusion = Math.Max(InclusionProbability(theta), MinInclusionProbability);
+            var estimatedCount = m / inclusion;
+            var cv2 = (1.0 - inclusion) / m;
+            estimates[callsite] = new CallsiteEstimate<TCallsite>(callsite, m, weight, inclusion, estimatedCount, cv2);
+
+            var nextWeight = 1.0 / estimatedCount;
+            nextWeights[callsite] = nextWeight;
+            if (nextWeight > nextUnseen)
```
jmacd, line 193:
```suggestion
            // nextUnseen is the maximum nextWeight
            if (nextWeight > nextUnseen)
```

jmacd:
@Yun-Ting Thank you--I made a very thorough review, please forgive me!

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
Yun-Ting: Ah, learned something new today, I didn't know there's no heap in Rust!

There is heap data structure in .NET.
But the problem here is that PriorityQueue<TElement, TPriority> starts to be supported net6.0 onward.
And the TFMs for libraries has some older versions: netstandard2.0 and (net462)https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/3e219d836ebc5ac7041ef185aa915071aea42690/build/Common.props#L8.

So, if we were to support those TFMs, we'll have to add multi-targeting logic, and I'm not sure whether it is worth it.

Lemme do some benchmarking.

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorLogExporter.cs
Yun-Ting: good catch! There is, in c# runtime, the EventId.Id is a hash of from the EventName (method name if folks use compile-time source gen.)
Let me see how to access it here.
I saw this one before, but forgot close these split ends...

Yun-Ting: Ah, and in runtime, their using the FNV implementation for hashing, too!

https://github.com/dotnet/runtime/blob/d9a6b9e82ecac933f7dcfb6cc80ca0319838925a/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Parser.cs#L234-L235

https://github.com/dotnet/runtime/blob/d9a6b9e82ecac933f7dcfb6cc80ca0319838925a/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Parser.cs#L1249-L1259

src/OpenTelemetry.Sampler.BottomFloor/BottomFloorSampler.cs
Yun-Ting: @martincostello, @rajkumar-rangaraj, I'm curious to learn - 
How strictly should a newly added project follow the version requirement in Common.props?
Just wanna make sure I'm not missing any context here.

src/OpenTelemetry.Sampler.BottomFloor/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
+# Changelog - OpenTelemetry.Sampler.BottomFloor
+
+## Unreleased
+
+Initial release of `OpenTelemetry.Sampler.BottomFloor`.
```
Yun-Ting, line 5: remove this line, as it's unreleased

PR #5115, opened 2026-08-23
[Extensions] Add W3CTraceState

### LINKED ISSUE #3091: [feature request] Support best-practice tracestate mutation
### Package

OpenTelemetry

Hi great people, 👋

The W3C [recommend](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field) several practices when mutating the `tracestate` header:
* "Vendors SHOULD NOT delete keys that were not generated by them."
* "Modified keys SHOULD be moved to the beginning (left) of the list."
* "the order of unmodified key/value pairs MUST be preserved"
* "Spaces and horizontal tabs surrounding list-members are ignored"
* "Empty list members are allowed"

It is tricky and time-consuming to write `tracestate` mutation code that follows these practices, so I wonder if we might consider providing an API in this package. Perhaps a `TraceState` struct that wraps the string value and provides "Update" methods, or a static helper class.

I am happy to author a PR if the idea meets approval.

Thanks

An API for performing `tracestate` mutation that adheres to the W3C's recommended practices

-

### Additional context

_No response_

PetarJerinic:
Fixes #3091 

## Changes

Adds `OpenTelemetry.Trace.W3CTraceState`. It is a W3C tracestate header exposing the four operations the [Trace State](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#tracestate) spec defines.
.NET exposes tracestate only as the raw ActivityContext.TraceState string. A custom sampler would have to re-implement the W3C specification itself.

`OtelTraceState` is left as-is.

### Notes:
- Name of the class: `W3C` prefix was added even though naming it `TraceState` would compile as well. The spec's required get/add/update/delete on TraceState is still unimplemented in .NET, if core adds TraceState to `OpenTelemetry.Trace`, every consumer of both packages will get CS0104. Happy to change the naming, if needed.


src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
```diff
@@ -25,3 +30,4 @@ static Microsoft.Extensions.Logging.OpenTelemetryLoggingExtensions.AttachLogsToA
 static OpenTelemetry.Trace.BaggageActivityProcessor.AllowAllBaggageKeys.get -> System.Predicate<string!>!
 static OpenTelemetry.Trace.TracerProviderBuilderExtensions.AddAutoFlushActivityProcessor(this OpenTelemetry.Trace.TracerProviderBuilder! builder, System.Func<System.Diagnostics.Activity!, bool>! predicate, int timeoutMilliseconds = 10000) -> OpenTelemetry.Trace.TracerProviderBuilder!
 static OpenTelemetry.Trace.TracerProviderBuilderExtensions.AddBaggageActivityProcessor(this OpenTelemetry.Trace.TracerProviderBuilder! builder, System.Predicate<string!>! baggageKeyPredicate) -> OpenTelemetry.Trace.TracerProviderBuilder!
+static OpenTelemetry.Trace.W3CTraceState.Parse(string? tracestate) -> OpenTelemetry.Trace.W3CTraceState!
```
martincostello, line 33: For a public API designed for parsing, I think it would make sense to follow the `Parse`/`TryParse` pattern in the runtime for types like `int` so the caller has an explicit signal for invalid trace state values.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,340 @@
...
+/// Members this instance did not write are handed back exactly as they arrived, malformed ones
+/// included, so that a long chain of edits never erodes another vendor's entries.
+/// <para/>
+/// Nothing here mutates and nothing here throws: an edit returns a new instance, and an edit naming
+/// an invalid key or value returns one carrying the receiver's contents unchanged.
+/// <para/>
+/// At most 32 members are kept, which is all the header grammar allows; anything past that is
+/// dropped from the right as it arrives.
+/// </remarks>
+public sealed class W3CTraceState
```
martincostello, line 28: Is there a specific design reason to implement this as a reference type rather than a value type? If not, would it maybe be more performant to do that?

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,340 @@
...
+
+    // Never longer than MemberLimit: Parse and Set both stop growing it there, and Remove only ever
+    // shrinks it. Bounding it here rather than at serialization keeps a wire-supplied header from
+    // being retained in full.
+    private readonly List<Member> members;
+
+    private W3CTraceState(List<Member> members)
+    {
+        this.members = members;
+    }
```
martincostello: If this type is meant to be immutable, then there's more efficient types that can be used to hold the members.

PetarJerinic: Agreed, will update it to be an array.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
PetarJerinic: The main reason for this was the null guard, where we could protect the members array from being null (by doing `default(W3CTraceState)` if it was a struct) and thus avoiding null checks (both now and in the future) whenever used within the class. 
The class, since it serves as a wrapper around members, adds one 24-byte object header. This trade-off seemed sensible to me, but we can go with a struct if that is the preferred option.

src/OpenTelemetry.Extensions/.publicApi/PublicAPI.Unshipped.txt
PetarJerinic: The reason I initially omitted it is that `tracestate` should keep even malformed members, having Parse actually "succeed" in that case, along with a couple of other "error" cases. 

It does make sense to add it to signal the caller. `TryParse` could be modified to return false if one of the members is malformed, or if there is another error, such as passing more members than the limit states, while still populating the `W3CTraceState` as per specification. Does that work in your opinion?

martincostello: I guess there's a grey area between completely valid and completely invalid where _some_ of the members are "bad". For `TryParse()` I'm just thinking for returning `false` in the case that the header is completely invalid/garbage.

PetarJerinic: Pushed. `TryParse` returns false in the following cases:
- `"malformed", "VENDORA=1", "=novalue"`
- `"<32 malformed members>,vendora=1"` - since the valid pair is past 32-member limit

src/OpenTelemetry.Extensions/README.md
```diff
@@ -161,3 +161,55 @@ builder.Services.AddOpenTelemetry()
...
+header and for a header mixing valid pairs with text kept verbatim. The state is
+populated the same way whichever value is returned, so `false` never yields less
+than `true` would.
+
+The value reports on the members retained, not on the header as it arrived. At
+most 32 members are kept either way, so a header of 32 unusable members followed
+by a well-formed pair reports `false`, because the pair sits past the limit and
+was never taken on. Discarding well-formed members to stay inside that limit is
+not itself a failure: a header of 40 valid pairs keeps the first 32 and reports
+`true`.
```
martincostello: This feels a bit overly descriptive to me.

src/OpenTelemetry.Extensions/README.md
```diff
@@ -161,3 +161,55 @@ builder.Services.AddOpenTelemetry()
...
+}
+```
+
+The modified key is placed first and every other member keeps its relative
+position, as the specification requires. At most 32 members are kept, which is
+all the header grammar allows, and the right-most ones are dropped as they
+arrive rather than at serialization, so a wire-supplied header is never retained
+in full. No length limit is imposed: the 512 characters vendors should be able
+to propagate is a floor on capability rather than a ceiling on output, so
+truncating to fit a transport limit stays the caller's decision.
```
martincostello: Same here - I think it's focusing more on how it's implemented, rather than what people should use it for and how.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+/// At most 32 members are kept, which is all the header grammar allows; anything past that is
+/// dropped from the right as it arrives.
+/// </remarks>
+public sealed class W3CTraceState
+{
+    /// <summary>
+    /// The maximum number of members in a W3C <c>tracestate</c> value, and so the most this type
+    /// ever retains.
+    /// </summary>
+    private const int MemberLimit = 32;
```
martincostello: `MaxMembers` or something like that?

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    /// <summary>
+    /// The maximum length of a key, per the <c>key = ( lcalpha / DIGIT ) 0*255 ( keychar )</c>
+    /// production.
+    /// </summary>
+    private const int KeyLengthLimit = 256;
+
+    /// <summary>
+    /// The maximum length of a value, per the <c>value = 0*255(chr) nblk-chr</c> production.
+    /// </summary>
+    private const int ValueLengthLimit = 256;
```
martincostello: Similarly, `Max{X}Length` for consistency with other code in the codebase and .NET that use `Max` and `Min` prefixes (like `int.MaxValue`).

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+
+    private static readonly W3CTraceState Empty = new([]);
+
+    // Never longer than MemberLimit: Parse and Set both stop filling it there, and Remove only ever
+    // shrinks it. Bounding it here rather than at serialization keeps a wire-supplied header from
+    // being retained in full. It is sized exactly at every allocation, because nothing is ever
+    // appended to an instance once it exists.
```
martincostello: Needed?

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    private readonly Member[] members;
+
+    private W3CTraceState(Member[] members)
+    {
+        this.members = members;
+    }
+
+    /// <summary>Reads a W3C <c>tracestate</c> value into a state that can be queried and edited.</summary>
```
martincostello:
```suggestion
    /// <summary>
    /// Parses a W3C <c>tracestate</c> value.
    /// </summary>
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    /// with it keys it never generated.
+    /// <para/>
+    /// The first 32 members are kept and the rest of the header is discarded, matching the limit
+    /// the grammar puts on a <c>tracestate</c> list.
+    /// </remarks>
+    public static W3CTraceState Parse(string? tracestate) => ParseCore(tracestate, out _);
+
+    /// <summary>
+    /// Reads a W3C <c>tracestate</c> value into a state that can be queried and edited, and reports
+    /// whether the header carried anything this type could make sense of.
```
martincostello: Similar to above.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    /// </summary>
+    /// <param name="tracestate">
+    /// The <c>tracestate</c> value, which may be <see langword="null"/> or empty.
```
martincostello: Implied by the type of the parameter.

```suggestion
    /// The <c>tracestate</c> value.
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    /// <para/>
+    /// An incoming header may legitimately carry the same key more than once. Every such member is
+    /// collapsed into the single one written here, so that the add does not leave the key present
+    /// multiple times; the key is by then one this instance generated. Members kept verbatim are
+    /// opaque and so are never collapsed, which means a member such as <c>vendora=</c>, carrying no
+    /// value and therefore not a valid pair, survives alongside a <c>vendora</c> pair added later.
+    /// <para/>
+    /// The new member counts towards the 32 an instance keeps, so setting a key on a state that is
+    /// already full drops the right-most member.
+    /// </remarks>
```
martincostello, line 157: I think this could be more simply explained. For example, the last paragraph could just be something like:

_If setting the value would exceed the maximum of 32 members, the right-most value is dropped_.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    public W3CTraceState Set(string key, string value)
+    {
+        // A null key or value is simply invalid: this is validated but never thrown for.
```
martincostello:
```suggestion
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+            if (!string.Equals(member.Key, key, StringComparison.Ordinal))
+            {
+                members[written++] = member;
+            }
+        }
+
+        return new W3CTraceState(members);
+    }
+
+    /// <summary>Deletes the pair a key carries, when there is one.</summary>
```
martincostello:
```suggestion
    /// <summary>
    /// Deletes the pair a key carries, if found.
    /// </summary>
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+            if (!string.Equals(member.Key, key, StringComparison.Ordinal))
+            {
+                members[written++] = member;
+            }
+        }
+
+        return new W3CTraceState(members);
+    }
+
+    /// <summary>Writes the state back out as a W3C <c>tracestate</c> value.</summary>
```
martincostello:
```suggestion
    /// <summary>
    /// Returns a string containing the W3C <c>tracestate</c> representation of the current instance.
    /// </summary>
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+        }
+
+        return builder.ToString();
+    }
+
+    // The single parsing path, so that Parse and TryParse can never come to disagree about what a
+    // header holds. The header is walked twice, once to size the store and once to fill it: an
+    // instance never grows after construction, so paying for the count is what keeps the store from
+    // carrying growth slack it will never use.
+    private static W3CTraceState ParseCore(string? tracestate, out bool understood)
```
martincostello: `understood` is an unusual name for this IMHO.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+            builder.Append(member.Value);
+        }
+
+        return builder.ToString();
+    }
+
+    // The single parsing path, so that Parse and TryParse can never come to disagree about what a
+    // header holds. The header is walked twice, once to size the store and once to fill it: an
+    // instance never grows after construction, so paying for the count is what keeps the store from
+    // carrying growth slack it will never use.
```
martincostello: I think this mostly isn't needed as the code is pretty self-descriptive.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+        if (count == 0)
+        {
+            return Empty;
+        }
+
+        var members = new Member[count];
+        var written = 0;
+        understood = false;
+
+        var remaining = tracestate.AsSpan();
```
martincostello: This could be moved up to above `count`'s declaration, and then passed into `CountMembers()`.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+        while (!remaining.IsEmpty && written < count)
+        {
+            var comma = remaining.IndexOf(',');
+            var member = (comma < 0 ? remaining : remaining.Slice(0, comma)).Trim();
+            remaining = comma < 0 ? default : remaining.Slice(comma + 1);
```
martincostello: I think this would be more readable as an `if (command >= 0)` block containing the two mutations.

```csharp
var member = remaining;

if (comma >=0 )
{
    member = member.Slice(0, comma);
    remaining = remaining.Slice(comma + 1);
}

member = member.Trim();
```

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+
+            members[written++] = created;
+        }
+
+        return new W3CTraceState(members);
+    }
+
+    // Counts the members a header yields, stopping at MemberLimit so that the right-most ones are
+    // never taken on. Empty members are dropped rather than counted.
+    private static int CountMembers(ReadOnlySpan<char> tracestate)
```
martincostello: This could be a local method to the one above as it's only used in one place.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+    {
+        var count = 0;
+
+        var remaining = tracestate;
+        while (!remaining.IsEmpty && count < MemberLimit)
+        {
+            var comma = remaining.IndexOf(',');
+            var member = (comma < 0 ? remaining : remaining.Slice(0, comma)).Trim();
+            remaining = comma < 0 ? default : remaining.Slice(comma + 1);
```
martincostello: Similar comment to above.

src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs
```diff
@@ -0,0 +1,449 @@
...
+
+        // The value ends in nblk-chr, so an interior space is allowed but a trailing one is not.
+        return value[value.Length - 1] != ' ';
+    }
+
+    /// <summary>
+    /// One member of a <c>tracestate</c> list: either a key and its value, or a run of text kept
+    /// verbatim because it did not match the grammar.
+    /// </summary>
+    private readonly struct Member
```
martincostello: Simplify with primary constructor?

test/OpenTelemetry.Extensions.Tests/Trace/W3CTraceStateTests.cs
```diff
@@ -0,0 +1,454 @@
...
+{
+    // "There can be a maximum of 32 list-members in a list."
+    // https://www.w3.org/TR/2021/REC-trace-context-1-20211123/#tracestate-header
+    private const int MemberLimit = 32;
+
+    // key = ( lcalpha / DIGIT ) 0*255 ( keychar ), so 256 characters is the maximum.
+    private const int KeyLengthLimit = 256;
+
+    // value = 0*255(chr) nblk-chr, so 256 characters is the maximum.
+    private const int ValueLengthLimit = 256;
```
martincostello: `Max` not `Limit`.

martincostello (approved):

Kielek:
Sorry for late review. Finally I was able to look into this. This code is intended to work directly with potentially unsafe inputs. The question is how it should be handled. Consider following feedback, assisted by chatgpt, to improve user experience.

## P1: Do not emit malformed `TraceState` members

`src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs`, lines 384–385.

Malformed members are retained and emitted unchanged. This allows the public
`W3CTraceState` type to produce an invalid `tracestate`, even though it is
documented as implementing the OpenTelemetry `TraceState` operations.

Examples:

```csharp
W3CTraceState.Parse("malformed").ToString()
// Returns: "malformed"
// Expected: an empty/valid TraceState, or a parse failure.

W3CTraceState.Parse("vendor=1,malformed").ToString()
// Returns: "vendor=1,malformed"
// This is not a valid W3C tracestate list.

W3CTraceState.TryParse("vendor=1,malformed", out var state)
// Returns: true
// state still contains the malformed member.
```

The [OpenTelemetry `TraceState` requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#tracestate)
require `TraceState` to remain valid at all times and prohibit returning a
state containing invalid data. W3C permits invalid entries to be discarded.

Please discard invalid/duplicate entries, or keep them only in a separate
raw-forwarding representation. The typed `TraceState` must remain valid.

## P2: Trim only W3C optional whitespace

`src/OpenTelemetry.Extensions/Trace/W3CTraceState.cs`, line 320.

`ReadOnlySpan.Trim()` applies .NET Unicode-whitespace rules, while W3C
`tracestate` only ignores spaces and horizontal tabs around members. Inputs
containing characters such as NBSP or CR/LF are therefore silently normalized
and accepted. For example, `vendor=value\u00A0` becomes `vendor=value`.

Use trimming limited to space and horizontal tab so other characters are
rejected or preserved according to the selected parse policy. See the
[W3C Trace Context grammar](https://www.w3.org/TR/trace-context-2/).

PetarJerinic:
@Kielek Thanks, pushed code changes for both parts of the feedback.

@martincostello for vis, I updated `TryParse` accordingly, it now returns false whenever an invalid or duplicate member is dropped, with valid parts present in `state`.

PR #5117, opened 2026-08-24
[StackExchange.Redis] Fix race during `ProfilingSession` instantiation

matt-hensley:
## Changes

Fixes race condition that could orphan `ProfilingSession` when instantiated for the same trace concurrently.

On cache misses, a new session was always created, but the cache insertion could fail if another session with the same cache key was added concurrently. This could result in Redis commands being attached to an orphaned session.

`GetOrAdd` always returns the "winning" cache entry, even if the factory is invoked. The extra session is does not collect commands and is discarded.


src/OpenTelemetry.Instrumentation.StackExchangeRedis/StackExchangeRedisConnectionInstrumentation.cs
```diff
@@ -120,13 +120,7 @@ public StackExchangeRedisConnectionInstrumentation(
...
         // Try to reuse a session for all activities created under the same TraceId+SpanId.
         var cacheKey = (parent.TraceId, parent.SpanId);
-        if (!this.Cache.TryGetValue(cacheKey, out var session))
-        {
-            // This runs on the thread that issued the command, the only point at which
-            // the caller's Baggage is still reachable. Draining happens elsewhere.
-            session = (parent, new ProfilingSession(), Baggage.Current);
-            this.Cache.TryAdd(cacheKey, session);
-        }
+        var session = this.Cache.GetOrAdd(cacheKey, _ => (parent, new(), Baggage.Current));
```
martincostello, line 127: For `NET` I think there's an overload you can use to pass `parent` through as state to avoid the need to allocate the closure for the factory so it can be static. Something like:

```csharp
var session = this.Cache.GetOrAdd(cacheKey, static (_, parent) => (parent, new(), Baggage.Current), parent);
```

martincostello (approved):

PR #5122, opened 2026-08-25
[Dynamic Control] Phase 5a: Add policy readers

stevejgordon:
Contributes to: #4742

## Changes

Continues the Dynamic Control component implementation with the policy reader layer, which turns a single value carried in a policy payload into a validated policy model. How a policy type is decoded is now defined once by that policy type, rather than by each source that carries it.

- Adds `PolicyType` and `PolicyId` identity value types and reworks `PolicyKey` to be composed of them.
- Adds an abstract `PolicyReader` contract that reads a JSON payload value into a `PolicyReadResult`. This is either a validated `TelemetryPolicy` or a rejection with a `PolicyRejectionReason` and error detail.
- Adds `TraceSamplingRatePolicyReader` and `TraceSamplingRateParser`, moving sampling-rate validation out of the policy model.
- Adds a second policy type: `LogLevelPolicy`, with `DiagnosticLogLevel` and `DiagnosticLogLevelParser`. This isn't yet a settled policy in the OTEP, but is a good candicate to validate this work with our new self diagnostics feature in OTel SDK core.

The JSON key-value payload the readers decode is an interim format: the Telemetry Policy OTEP makes no requirement on how policies are encoded, and its trace target is still unsettled.


src/OpenTelemetry.DynamicControl/Internal/Policies/DiagnosticLogLevel.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// The severity threshold at or above which the SDK emits its own diagnostic logs.
```
martincostello: Idle thought: is it an SDK here? Typically by "SDK" we mean the OpenTelemetry library.

src/OpenTelemetry.DynamicControl/Internal/Policies/LogLevelPolicy.cs
```diff
@@ -0,0 +1,92 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Represents a validated policy setting the severity threshold for the SDK's own
```
martincostello: Just a note that if we don't call this an "SDK", then any comments that refer to it as such should all be updated.

src/OpenTelemetry.DynamicControl/Internal/Policies/LogLevelPolicy.cs
```diff
@@ -0,0 +1,92 @@
...
+        return true;
+    }
+
+    private static bool IsSupported(DiagnosticLogLevel level)
+        => level is DiagnosticLogLevel.Trace
+            or DiagnosticLogLevel.Debug
+            or DiagnosticLogLevel.Information
+            or DiagnosticLogLevel.Warning
+            or DiagnosticLogLevel.Error
+            or DiagnosticLogLevel.None;
```
martincostello, line 91: _Maybe_ Worth a comment in the enum after the last value saying to update this if a new member is added?

src/OpenTelemetry.DynamicControl/Internal/Policies/TelemetryPolicy.cs
```diff
@@ -47,5 +47,5 @@ protected TelemetryPolicy(string id, string name)
     /// Gets the policy type discriminator (e.g. <c>trace-sampling</c>).
```
martincostello, line 47: Maybe this needs updating as it's no longer a string literal?

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace OpenTelemetry.DynamicControl.Internal.Policies;
+
+/// <summary>
+/// Parses the textual form of a trace sampling probability.
+/// </summary>
+/// <remarks>
+/// Parsing uses the invariant culture, permits surrounding white space and exponents, and
+/// rejects negative values, group separators, and values that underflow to zero.
+/// </remarks>
```
martincostello, line 13: Do we need this?

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
```diff
@@ -0,0 +1,92 @@
...
+    public static bool TryParse(string? text, out double probability)
+    {
+        probability = default;
+
+        if (text is null)
+        {
+            return false;
+        }
+
+        if (!double.TryParse(text, NumberStyles.Float, CultureInfo.InvariantCulture, out probability))
```
martincostello: Use `NumberStyles` to disallow (some?) of the values being explicitly checked for below?

e.g. `AllowLeadingWhite | AllowTrailingWhite | AllowDecimalPoint | AllowExponent` but not `AllowLeadingSign`.

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRatePolicy.cs
```diff
@@ -51,20 +51,20 @@ private TraceSamplingRatePolicy(string id, string name, double samplingProbabili
...
-        if (id is not { Length: > 0 } || string.IsNullOrWhiteSpace(id))
+        if (id.IsEmpty)
         {
             policy = null;
             error = "The policy ID is required.";
             return false;
         }
 
-        if (name is not { Length: > 0 } || string.IsNullOrWhiteSpace(name))
+        if (string.IsNullOrWhiteSpace(name))
```
martincostello, line 67: I don't immediately see what's changed that won't cause this removal to re-trigger a nullability warning in #3867.

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyReadResult.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// <summary>
+    /// Gets the rejection message, or <see langword="null"/> on success.
+    /// </summary>
+    public string? Error { get; }
+
+    /// <summary>
+    /// Creates a successful result.
+    /// </summary>
+    /// <param name="policy">The validated policy. Must not be null.</param>
```
martincostello:
```suggestion
    /// <param name="policy">The validated policy.</param>
```

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyReadResult.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <returns>A rejected result.</returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="reason"/> is <see cref="PolicyRejectionReason.None"/>, or
+    /// when <paramref name="error"/> is null, empty, or whitespace.
+    /// </exception>
+    public static PolicyReadResult Reject(PolicyRejectionReason reason, string error)
+    {
+        if (reason == PolicyRejectionReason.None)
+        {
+            throw new ArgumentException("A rejection must state a reason.", nameof(reason));
```
martincostello: Nit: _technically_ it did, it was just "None".

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyReader.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// <summary>
+    /// Reads one payload value.
+    /// </summary>
+    /// <param name="value">The value the payload associates with <see cref="PayloadKey"/>.</param>
+    /// <returns>
+    /// A result containing either the validated policy or its rejection.
+    /// </returns>
+    /// <remarks>
+    /// Implementations report unusable values as rejections instead of throwing.
```
martincostello:
```suggestion
    /// Implementations should report unusable values as rejections instead of throwing.
```

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyRejectionReason.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// <summary>
+    /// The value was not rejected.
+    /// </summary>
+    None = 0,
+
+    /// <summary>
+    /// The entry was not in a shape the payload format accepts, such as an unusable value
+    /// kind or a missing required member.
+    /// </summary>
+    InvalidPayloadShape = 1,
```
martincostello: `InvalidPayloadFormat`?

martincostello (commented):
Reviewed everything except the tests.

src/OpenTelemetry.DynamicControl/Internal/Policies/DiagnosticLogLevel.cs
stevejgordon: In this case, yes as this would later control the self diagnostics log level from the core SDK. Note that right now, there's no precedent for configuring the SDK from telemetry policies as such but it's a direction I think will be likely as reference implementations show the value and we can define specs. I plan to prototype that scenario once self-diagnostics lands and I understand Java may also be trialing that soon too.

src/OpenTelemetry.DynamicControl/Internal/Policies/LogLevelPolicy.cs
stevejgordon: As above.

src/OpenTelemetry.DynamicControl/Internal/Policies/LogLevelPolicy.cs
stevejgordon: Went with tests driven off the enum rather than a comment.

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
stevejgordon: I thought it useful to document but it could be a bit more concise perhaps.

src/OpenTelemetry.DynamicControl/Internal/Policies/DiagnosticLogLevel.cs
stevejgordon: I'll update the comment to make it more specific

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyRejectionReason.cs
stevejgordon: Yeah this was a hard one to name. I worried format implied not JSON, rather than an incorrect structure/shape. I considered `InvalidPayloadStructure` too.

martincostello: `Schema` maybe?

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRatePolicy.cs
stevejgordon: The param is non-null now so I think this is ok?

src/OpenTelemetry.DynamicControl/Internal/Sources/PolicyRejectionReason.cs
stevejgordon: `SchemaMismatch` could work

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRatePolicy.cs
martincostello: Ah, that must be what I missed.

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
```diff
@@ -9,11 +9,17 @@ namespace OpenTelemetry.DynamicControl.Internal.Policies;
...
+/// Parsing uses the invariant culture and rejects values that underflow to zero.
 /// </remarks>
 internal static class TraceSamplingRateParser
 {
+    // NumberStyles.Float without AllowLeadingSign.
+    private const NumberStyles ProbabilityStyles =
+        NumberStyles.AllowLeadingWhite
+        | NumberStyles.AllowTrailingWhite
+        | NumberStyles.AllowDecimalPoint
+        | NumberStyles.AllowExponent;
```
martincostello: I think that can be simplified as `NumberStyles.Float & ~NumberStyles.AllowLeadingSign` to turn it off with a mask?

test/OpenTelemetry.DynamicControl.Tests/EnumTestHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+    // The generic Enum.GetValues<TEnum>() overload that CA2263 prefers is not available
+    // on .NET Framework, so the choice is made here rather than at every call site.
+    public static TEnum[] Values<TEnum>()
+        where TEnum : struct, Enum
+#if NET
+        => Enum.GetValues<TEnum>();
+#else
+        => (TEnum[])Enum.GetValues(typeof(TEnum));
+#endif
+}
```
martincostello, line 17: Could this be an `extension` instead?

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
stevejgordon: Nice!

test/OpenTelemetry.DynamicControl.Tests/EnumTestHelper.cs
stevejgordon: I don't think so in this test project.

test/OpenTelemetry.DynamicControl.Tests/PolicyValueRandomizedTests.cs
martincostello, line 1: Implement as explicit fuzz tests using FsCheck?

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
stevejgordon: Revisited this after comparing with the Java implementation. Java accepts  -0.0  and normalizes it to  +0.0 , as the policy model already does. I’ve therefore restored `NumberStyles.Float` and reject values numerically less than zero instead. Negative underflow remains rejected. Leading  `+`  in numeric strings is also accepted, matching Java parsing. OTEP doesn't specify anything on this yet but an eventual spec may. Matching the implementations feels like the best choice for now for these edge cases.

test/OpenTelemetry.DynamicControl.Tests/PolicyValueRandomizedTests.cs
stevejgordon: See 05d41f2e4b55ee552b2ea6f148dd853b8144d4b4

src/OpenTelemetry.DynamicControl/Internal/Policies/TraceSamplingRateParser.cs
stevejgordon: See 05d41f2e4b55ee552b2ea6f148dd853b8144d4b4

test/OpenTelemetry.DynamicControl.FuzzTests/OpenTelemetry.DynamicControl.FuzzTests.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- OmniSharp/VS Code requires TargetFrameworks to be in descending order for IntelliSense and analysis. -->
```
martincostello: I've been deleting these as files are touched as almost all of them don't contain any literal versions anymore, so the comment is redundant as the sorting is with the property defintion.

```suggestion
```

test/OpenTelemetry.DynamicControl.FuzzTests/FuzzInput.cs
```diff
@@ -0,0 +1,64 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.DynamicControl.Internal.Policies;
+
+namespace OpenTelemetry.DynamicControl.FuzzTests;
+
+internal static class FuzzInput
```
martincostello, line 8: I would investigate whether some of the fuzz testing code can be made more idiomatic for FsCheck by using `Gen<T>` and `Arbitrary<T>`. See [`Generators`](https://github.com/open-telemetry/opentelemetry-dotnet/blob/c059ef68ccbdfe0df02c4b60c2cef482562dcee3/test/OpenTelemetry.Api.FuzzTests/Context/Propagation/Generators.cs) for an example.

This could reduce the amount of code in the tests themselves to generate input from primitive types that FsCheck supports out-of-the-box and would make them more focused on Act and Assert and need less code to Arrange.

stevejgordon: I'll take a look. I mostly let Claude Code handle this as I've not used fuzz testing myself.

martincostello: To be fair, I got Claude (or maybe Copilot) to generate a lot of that code anyway, so it should be able to action it reasonably well.

stevejgordon: Hopefully 1b3f25444f1663e49eb6d5473c9f9af694d3756e improves things!

test/OpenTelemetry.DynamicControl.FuzzTests/PolicyReaderTests.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Globalization;
+using System.Text;
+using System.Text.Json;
+using FsCheck.Xunit;
+using OpenTelemetry.DynamicControl.Internal.Policies;
+using OpenTelemetry.DynamicControl.Internal.Sources;
+
+namespace OpenTelemetry.DynamicControl.FuzzTests;
+
+public static class PolicyReaderTests
```
martincostello, line 12: Can these tests be refactored to benefit from `Generators` too? `byte` seems to be being used as a proxy for generating JSON schema shapes. Even if it's just lifted and shifted and passed into a proxy type like the probabilities are, I think it would be a bit easier to follow to the casual observer.

stevejgordon: Let's try 20a2aa15ff5bb23ba43789ad8ac60446fe4958bf

martincostello (approved):

PR #5124, opened 2026-08-25
[WCF] Avoid creating activity

martincostello:
## Changes

Avoid creating an activity and associated tags if there are no trace listeners configured for the activity source.


matt-hensley (approved):

Kielek:
@martincostello, could you please describe why we need this?
What is the use case? The instrumentation should be added by `.AddWcfInstrumentation()` otherwise this code shouldn't be called at all.

martincostello:
This was found by a scan of the codebase with Claude Code for correctness to avoid redundant work. Taking a quick look at the code, it could technically be reached through manual use of the public `TelemetryEndpointBehavior` class.

Kielek:
TBH, I would not merge this, only case when it can be reached is purely wrong set-up.

PR #5135, opened 2026-08-27
[Dynamic Control] Phase 4: Add ordered, coalesced PolicyStore subscriptions

stevejgordon:
Contributes to: #4742

## Changes

Adds PolicyChangeNotifier/PolicyChangeSubscription so that a subscriber can register with PolicyStore.Subscribe and receive an atomic replay of the current snapshot followed by every subsequent commit, coalesced to the latest revision and never delivered out of order or overlapping. Subscriber callback failures are isolated per-subscription and reported via a new DynamicControlEventSource rather than propagating to the committing thread.

PolicyStore is now IDisposable to release its subscriptions.

### DIFF
--- a/src/OpenTelemetry.DynamicControl/Internal/Diagnostics/DynamicControlEventSource.cs
+++ b/src/OpenTelemetry.DynamicControl/Internal/Diagnostics/DynamicControlEventSource.cs
@@ -1,0 +1,39 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using System.Diagnostics.Tracing;
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.DynamicControl.Internal.Diagnostics;
+
+/// <summary>
+/// The <see cref="EventSource"/> for this component's internal diagnostics.
+/// </summary>
+[EventSource(Name = "OpenTelemetry-DynamicControl")]
+internal sealed class DynamicControlEventSource : EventSource
+{
+    public static readonly DynamicControlEventSource Log = new();
+
+    private const int EventIdPolicyChangeSubscriberFailure = 1;
+
+    /// <summary>
+    /// Records that a policy change subscriber's callback threw an exception. The
+    /// exception is isolated to the subscription that raised it; it does not affect the
+    /// store, other subscribers, or the commit that triggered the notification.
+    /// </summary>
+    /// <param name="ex">The exception thrown by the subscriber callback.</param>
+    [NonEvent]
+    public void PolicyChangeSubscriberException(Exception ex)
+    {
+        if (this.IsEnabled(EventLevel.Warning, EventKeywords.All))
+        {
+            this.PolicyChangeSubscriberFailure(ex.ToInvariantString());
+        }
+    }
+
+    [Event(EventIdPolicyChangeSubscriberFailure, Message = "Policy change subscriber callback failed: {0}", Level = EventLevel.Warning)]
+    public void PolicyChangeSubscriberFailure(string exception)
+    {
+        this.WriteEvent(EventIdPolicyChangeSubscriberFailure, exception);
+    }
+}
--- a/src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
+++ b/src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
@@ -69,7 +69,7 @@ internal sealed class PolicyChangeNotifier : IDisposable
             var current = this.subscribers;
             var updated = new PolicyChangeSubscription[current.Length + 1];
             Array.Copy(current, updated, current.Length);
-            updated[current.Length] = subscription;
+            updated[updated.Length - 1] = subscription;
             Volatile.Write(ref this.subscribers, updated);
 
             return subscription;
@@ -77,7 +77,7 @@ internal sealed class PolicyChangeNotifier : IDisposable
     }
 
     /// <summary>
-    /// Removes a subscription. Called by <see cref="PolicyChangeSubscription.Dispose"/>.
+    /// Removes a subscription.
     /// </summary>
     /// <param name="subscription">The subscription to remove.</param>
     internal void Remove(PolicyChangeSubscription subscription)
--- a/src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
+++ b/src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
@@ -1,0 +1,119 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+using OpenTelemetry.Internal;
+
+namespace OpenTelemetry.DynamicControl.Internal.Store;
+
+/// <summary>
+/// Owns the set of subscriptions registered against a <see cref="PolicyStore"/> and
+/// dispatches published snapshots to them.
+/// </summary>
+internal sealed class PolicyChangeNotifier : IDisposable
+{
+    private static readonly PolicyChangeSubscription[] EmptySubscribers = [];
+
+    private readonly Lock registrationLock = new();
+
+    // The subscriber list is a copy-on-write array.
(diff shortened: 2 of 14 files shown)


src/OpenTelemetry.DynamicControl/Internal/Diagnostics/DynamicControlEventSource.cs
martincostello, line 1: Please add a test for this [like this](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/test/OpenTelemetry.Extensions.Tests/EventSourceTests.cs).

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        lock (this.registrationLock)
+        {
+            this.ThrowIfDisposed();
+
+            var subscription = new PolicyChangeSubscription(onChanged, this);
+            var current = this.subscribers;
+            var updated = new PolicyChangeSubscription[current.Length + 1];
+            Array.Copy(current, updated, current.Length);
+            updated[current.Length] = subscription;
```
martincostello: I think this would be _slightly_ more readable as `updated[updated.Length - 1]`, as otherwise you need to read more than one line to check this is correctly adding to the end of the array.

It's probably also a stronger hint to the JIT that the write is in bounds.

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
```diff
@@ -0,0 +1,119 @@
...
+            Volatile.Write(ref this.subscribers, updated);
+
+            return subscription;
+        }
+    }
+
+    /// <summary>
+    /// Removes a subscription. Called by <see cref="PolicyChangeSubscription.Dispose"/>.
```
martincostello: One less thing to update if that ever changes.

```suggestion
    /// Removes a subscription.
```

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (current.Length == 1)
+            {
+                Volatile.Write(ref this.subscribers, EmptySubscribers);
+                return;
+            }
+
+            var updated = new PolicyChangeSubscription[current.Length - 1];
+            Array.Copy(current, updated, index);
+            Array.Copy(current, index + 1, updated, index, current.Length - index - 1);
+            Volatile.Write(ref this.subscribers, updated);
```
martincostello, line 103: As I've spent a minute trying to trace the bounds arithmetic in my head here, if not already present let's make sure there's test coverage for these boundary cases:

- 2 & 3 element array, remove the first item
- 2 & 3 element array, remove the last item
- 3 element array, remove the middle item

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeSubscription.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            this.dispatching = true;
+        }
+
+        ThreadPool.UnsafeQueueUserWorkItem(static state => ((PolicyChangeSubscription)state!).Drain(), this);
+    }
+
+    private void Drain()
+    {
+        while (true)
```
martincostello, line 94: 1. Should we have a `CancellationTokenSource` to signal from `Dispose` to stop the loop here (I'm always extra wary of `while (true)`).
2. Should we have any backpressure on how often this publishes a change?

test/OpenTelemetry.DynamicControl.Tests/PolicyStoreConcurrencyTests.cs
```diff
@@ -0,0 +1,165 @@
...
+        while (!condition())
+        {
+            if (sw.ElapsedMilliseconds > timeoutMilliseconds)
+            {
+                Assert.Fail("Timed out waiting for condition.");
+            }
+
+            await Task.Delay(10);
+        }
+    }
```
martincostello: Move to a helper as there's ~~at least two~~ three copies of the same pattern here?

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeSubscription.cs
stevejgordon: On the CTS: I don't think we need one here. The exit condition (`disposed || pending is null`) is checked under the same lock that `Dispose()` sets `disposed` under, so the loop is guaranteed to observe disposal and exit on the next itteration.

On backpressure: I had AI reassess against the telemetry-policy OTEP and it requires policy updates to be async and fail-open, and explicitly says a stale/out-of-date policy must never be lethal to the system. Blocking the committer to apply backpressure would violate that, so coalesce-to-latest (current behavior) is spec-aligned, not a gap. I'll add a comment citing this so it reads as intentional rather than an oversight.

test/OpenTelemetry.DynamicControl.Tests/PolicyStoreConcurrencyTests.cs
stevejgordon: Addressed in 2c875ac5eeb2edf29e77d935986cce92ffe9434a

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
stevejgordon: Addressed in 2c875ac5eeb2edf29e77d935986cce92ffe9434a

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
stevejgordon: Addressed in 2c875ac5eeb2edf29e77d935986cce92ffe9434a

src/OpenTelemetry.DynamicControl/Internal/Store/PolicyChangeNotifier.cs
stevejgordon: Addressed in 2c875ac5eeb2edf29e77d935986cce92ffe9434a

src/OpenTelemetry.DynamicControl/Internal/Diagnostics/DynamicControlEventSource.cs
stevejgordon: Addressed in 2c875ac5eeb2edf29e77d935986cce92ffe9434a

test/OpenTelemetry.DynamicControl.Tests/OpenTelemetry.DynamicControl.Tests.csproj
```diff
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <Compile Include="$(RepoRoot)\test\Shared\InMemoryEventListener.cs" Link="Includes\InMemoryEventListener.cs" />
+    <Compile Include="$(RepoRoot)\test\Shared\EventSourceTestHelper.cs" Link="Includes\EventSourceTestHelper.cs" />
```
martincostello, line 14: Nit: sorting.

test/OpenTelemetry.DynamicControl.Tests/PolicyStoreConcurrencyTests.cs
martincostello: Looks like this didn't get committed.

stevejgordon:
> Looks like this didn't get committed.

I'm confused? It's called from the three places?

stevejgordon:
> I'm confused? It's called from the three places?

Oh wait. I see. I fixed a slightly different issue!

martincostello:
> I'm confused? It's called from the three places?

- `PolicyChangeSubscriptionTests.WaitUntil()`
- `PolicyStoreConcurrencyTests.WaitUntil()`
- `PolicyStoreSubscriptionTests.WaitUntil()`

These three methods all seem identical except for having a different default timeout and delay.

martincostello (approved):

PR #5147, opened 2026-09-01
[Geneva] Fix OTLP protobuf lengths at and above 16 KiB

maksmara:
## Changes

Fix the Geneva metric exporter's fixed-width protobuf length encoding when a
nested message requires three significant varint bytes.

This affects Geneva metric export when OTLP protobuf encoding is enabled.
Legacy Geneva metric encoding does not use this serializer and is unaffected.

Previously, the encoder set the continuation bit on the final reserved byte for
lengths at or above 16,384 bytes. This caused protobuf decoders to consume the
first payload byte as part of the length and reject the complete
`ExportMetricsServiceRequest`.

The updated encoder:

* Always writes exactly three length bytes, with the third byte terminating the
  varint.
* Rejects lengths outside the supported three-byte range.
* Guards against insufficient buffer space when backpatching the fixed-width
  prefix.
* Reports invalid lengths through the existing metric buffer-overflow path.

The regression coverage:

* Tests the custom encoder around the `16,383`/`16,384` transition and at its
  supported limits.
* Verifies that a large explicit histogram exercises a three-byte nested-message
  length.
* Parses payloads with and without transport framing.
* Verifies that all explicit bounds and bucket counts survive serialization.

* [x] Changes in public API reviewed (not applicable; no public API changes)


rajkumar-rangaraj (approved):
LGTM

PR #5161, opened 2026-09-02
[OpAMP] Preserve custom messages queued during active sends

### LINKED ISSUE #5142: [bug] [OpenTelemetry.OpAmp.Client] Fast appends corrupt accumulated custom messages.
### Component

OpenTelemetry.OpAmp.Client

### Package Version

| OpenTelemetry.OpAmp.Client       | 0.6.0-alpha.1  |

### Runtime Version

net462,net8,net10

### Description

> Multiple custom messages can be silently lost while another send is active. The first queued custom message remains in currentFrame, but every subsequent SendCustomMessage overwrites this singular property before Build() runs, whereas the previous dispatcher serialized and sent every call. Custom messages are not coalescible state; they need separate queued frames (or equivalent) so each payload is delivered.

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4930#discussion_r3869609454

### Steps to Reproduce

Append multiple custom messages either concurrently or in series fast before the flush happens.

### Expected Result

Accumulated data does not get overridden (no data loss)

### Actual Result

Accumulated data gets overridden (data loss)

### Additional Context

https://github.com/open-telemetry/opamp-spec/pull/387#discussion_r3872244598

### Tip

ysolomchenko:
Fixes #5142

## Changes

`AgentToServer.custom_message` is a singular field. When several custom messages
were appended while another send was in progress, later calls overwrote the
previously accumulated message before its frame was built, causing silent data
loss.

This change:

* Snapshots each custom message into a separate `AgentToServer` frame and drains
  the frames in FIFO order.
* Bounds the pending custom-message queue at 2,048 messages and 64 MiB of
  aggregate payload data. `SendCustomMessage()` throws
  `InvalidOperationException` when either limit would be exceeded.
* Includes pending custom-message frames when determining whether flush and
  graceful shutdown operations have completed.
* Dispatches sends asynchronously so synchronously completing transports cannot
  recursively drain the queue and overflow the call stack.
* Documents the new backpressure behavior.
* Adds tests covering message preservation, FIFO ordering, both queue limits,
  capacity reclamation, and synchronously completing transports.

> [!NOTE]
> The OpAMP specification does not prescribe limits for a client-side
> custom-message queue. The 2,048-message limit matches the standard
> `OTEL_BSP_MAX_QUEUE_SIZE` and `OTEL_BLRP_MAX_QUEUE_SIZE` defaults used by
> OpenTelemetry batch processors, making it a familiar OpenTelemetry default.
>
> The 64 MiB aggregate payload limit is an internal safeguard chosen by analogy
> with the 64 MiB default recommended by the
> [OpAMP specification](https://github.com/open-telemetry/opamp-spec/blob/v0.20.0/specification.md#websocket-message-size-limits)
> for a single complete WebSocket message. It is not an aggregate queue limit
> prescribed by OpAMP.
>
> Both limits remain internal to keep this bug fix focused. A follow-up PR will
> make them configurable through `OpAmpClientSettings`, allowing users to adjust
> them for their workload and memory constraints.


src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -13,12 +13,15 @@ namespace OpenTelemetry.OpAmp.Client.Internal;
 
 internal sealed class OpAmpPipe : IDisposable
 {
+    internal const int MaxPendingCustomMessages = 2048;
```
stevejgordon, line 16: Worth adding this to `OpAmpClientSettings` to make it configurable? We could make that a follow up also to not block this PR.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -170,6 +164,39 @@ private static async Task WaitForFlushAsync(Task flushTask, CancellationToken to
...
+        lock (this.frameLock)
+        {
+            if (this.isStopped ||
+                this.isDisposed ||
+                this.tokenSource.IsCancellationRequested)
+            {
+                return; // Discard any new messages
+            }
+
+            if (queueFrame && this.pendingFrames.Count >= MaxPendingCustomMessages)
```
stevejgordon: While we bound the queue size to 2048, we have no bounding for the actual data being held in memory. If, for some reason the custom message payloads are large, holding over two thousand of them could be problematic. I'm wondering if we should consider a `MaxPendingCustomMessageBytes` option too. In `AppendCustomMessage` we have the payload byte count and could track the actual queued bytes, as well as message counts. This would add an extra layer of protection to prevent a runaway growth in the queue.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -205,11 +232,12 @@ private void TryFlush(CancellationToken token)
     }
 
     private bool IsFlushCompleteLocked()
-        => this.isDisposed || (!this.hasAccumulatedData && !this.isBusy);
+        => this.isDisposed ||
+            (!this.hasAccumulatedData && this.pendingFrames.Count == 0 && !this.isBusy);
```
lachmatt: Nit: it might be worth to convert to positive predicate:
```csharp
private bool HasDataToSendLocked() => this.hasAccumulatedData || this.pendingFrames.Count > 0;
```
and then use it as:
```csharp
private bool IsFlushCompleteLocked() => this.isDisposed || (!this.HasDataToSendLocked() && !this.isBusy);
```
Similar adjustment to be made in `TryStartFlushLocked`

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -221,10 +249,23 @@ private bool IsFlushCompleteLocked()
...
+        {
+            message = this.currentFrame.Build();
+            this.hasAccumulatedData = false;
+        }
 
-        this.flushTask = this.SendMessageAsync(message, token);
+        // Always dispatch asynchronously so a synchronously completing transport cannot
+        // recursively drain the queue and eventually overflow the call stack. The send task
+        // must start even when token is canceled so SendMessageAsync can release the busy state.
+        this.flushTask = Task.Run(
```
stevejgordon, line 284: This introduces a little overhead (`Task`, closure, thread pool hop, etc.) per message compared to the existing code.  For realistic, low volume messages I think this is okay, but if the message volumes are large (as our max queue size potentially permits) then this adds up. For HTTP where we need the response, I'm not sure we can improve it much. But for WebSocket, perhaps a solution could be message pump that drains queued messages in order. I wouldn't block this PR on that though as it needs some thought. But potentially worth a follow up issue to test/benchmark it at some point?

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
ysolomchenko: Added in [1d1ca11](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5161/commits/1d1ca118210a66cf951fe90a480730aca215f073)

stevejgordon (commented):
A couple of thoughts. The only one I could argue as blocking would be unbounded byte growth in the queue. The other two might be worth considering as follow ups.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
ysolomchenko: Yes! This definitely should be configurable. I was planning to address this in a separate PR. Please see the `Note` in the PR description.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
ysolomchenko: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/5168

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
ysolomchenko: Added in [65b4cc4](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5161/commits/65b4cc4ff5e9fe8ace08c017d4dd9cc5f0a6c10d)

I will add `MaxPendingCustomMessageBytes` to `OpAmpClientSettings` to make it configurable in a separate PR.

src/OpenTelemetry.OpAmp.Client/Internal/OpAmpPipe.cs
```diff
@@ -170,6 +167,49 @@ private static async Task WaitForFlushAsync(Task flushTask, CancellationToken to
         await flushTask.ConfigureAwait(false);
     }
 
+    private void AppendMessage(Action<IFrameBuilder> messageRequest, int? customMessagePayloadBytes)
```
lachmatt: Nit: I'd remove the nullable payload-length parameter from the generic append path and use `AppendMessage(Action<IFrameBuilder>)` exclusively for coalescible state, and implement `AppendCustomMessage` as a separate path.

stevejgordon: I was thinking something similar, or if not, why not make this non-nullable, and treat -1 as the default and then simplify the checks against this being >= 0?

ysolomchenko: Fixed in [214366d](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5161/commits/214366dbb093de39922b9586fa4142fe465f131f)

stevejgordon (approved):

lachmatt (approved):

Kielek (approved):
LGTMRS

PR #5138, opened 2026-08-27
[Resources.Azure] Add App Service account and resource group attributes

Lewis-E:
## Changes

Update the Azure App Service resource detector to emit:
* `cloud.account.id` - from `WEBSITE_OWNER_NAME` [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-account-id)]
* `azure.resource_group.name` - from `WEBSITE_RESOURCE_GROUP` [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/azure/#azure-resource-group-name)]

Also include these values as params for `GetAzureResourceURI` instead of parsing them inside the uri function. 

Goal: Align the .NET detector with the Azure App Service mapping tracked in [collector-contrib #49616](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49616) and implemented for the Collector in [collector-contrib #50287](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/50287).

**AI Disclosure:** Pi Coding Agent was used to draft this PR, but I wrote this PR description and can explain all changes.

## Testing

* [x] Unit Tests
* [x] Deployed a test app & verified the attributes were visible.

Screenshot 2026-08-28 at 2 19 09 PM

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [x] Changes in public API reviewed (if applicable)

PR #5165, opened 2026-09-02
[Resources.Azure] Add Azure Functions resource detector

Lewis-E:
## Changes

Adds an Azure Functions resource detector alongside the previous Azure App Services resource detector. https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5138

This new resource detector emits:
* `service.name` from `WEBSITE_SITE_NAME` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-name)]
* `cloud.provider=azure` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-provider)]
* `cloud.platform=azure.functions` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-platform)]
* `cloud.region` from `REGION_NAME` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-region)]
* `cloud.account.id` from the subscription portion of `WEBSITE_OWNER_NAME` before `+` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-account-id)], [[discussion](https://github.com/open-telemetry/semantic-conventions/issues/3708)]
* `azure.resource_group.name` from `WEBSITE_RESOURCE_GROUP` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/azure/#azure-resource-group-name)], [[discussion](https://github.com/open-telemetry/semantic-conventions/issues/3697)]
* `faas.instance` from `WEBSITE_INSTANCE_ID`, falling back to `WEBSITE_POD_NAME` and then `CONTAINER_NAME` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/#faas-instance)]
* `deployment.environment.name` from `WEBSITE_SLOT_NAME` - [[sem-conv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/deployment/#deployment-environment-name)]

For `faas.instance`, note that `WEBSITE_INSTANCE_ID` is absent for Flex Consumption Plans & Azure Functions on Azure Container Apps. `WEBSITE_POD_NAME` and `CONTAINER_NAME` are our next best fallbacks. See also https://github.com/Azure/azure-functions-host/blob/cd7d51ee7ca03b37dfeeb66f709fab1dbec081b8/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L543. Further discussion of choices at https://github.com/open-telemetry/opentelemetry-go-contrib/pull/9290

We previously tried to add `cloud.resource.id`, but the [faas specification](https://opentelemetry.io/docs/specs/semconv/resource/faas/) actually indicates that is a span attribute that we cannot add here. We need the function name to create the full resource_id.

Changes also include readme updates & unit tests.

**AI Disclosure:** Pi Coding Agent was used to draft this PR, but I wrote this PR description and can explain all changes.

## Testing
- [x] Unit Tests
- [x] Test App

Screenshot 2026-09-02 at 9 39 11 PM


src/OpenTelemetry.Resources.Azure/CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
   * `azure_vm` to `azure.vm`
   ([#5143](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5143))
 
+* Detect Azure Functions environments before Azure App Service and emit Azure
+  Functions resource attributes.
```
stevejgordon, line 17: This needs the link to this pull request adding.

stevejgordon, line 17: I think it may be worth calling out that this is breaking even though this is beta. Any existing user who has `FUNCTIONS_WORKER_RUNTIME` set will loose some attributes.

src/OpenTelemetry.Resources.Azure/README.md
```diff
@@ -62,6 +63,25 @@ using var loggerFactory = LoggerFactory.Create(builder =>
...
+| cloud.provider              | The cloud service provider. In this context, it is always "azure".                                                                                             |
+| cloud.region                | The Azure region from `REGION_NAME`.                                                                                                                           |
+| cloud.resource_id           | The Azure Resource Manager URI identifying the function app. Emitted when the site name, resource group, and subscription ID are available.                    |
+| deployment.environment.name | The deployment slot from `WEBSITE_SLOT_NAME`.                                                                                                                  |
+| faas.instance               | The platform instance ID from `WEBSITE_INSTANCE_ID`, falling back to `WEBSITE_POD_NAME` and then `CONTAINER_NAME` for Linux and Flex Consumption environments. |
+| service.name                | The function app name from `WEBSITE_SITE_NAME`.                                                                                                                |
+
```
stevejgordon, line 84: Is it worth calling out here that `FUNCTIONS_WORKER_RUNTIME` is set by the Azure Functions Core Tools in local development even when there is no Azure subscription involved meaning that the detector emits cloud.provider=azure and cloud.platform=azure.functions with no region, account, or resource ID etc.?

src/OpenTelemetry.Resources.Azure/ResourceAttributeConstants.cs
```diff
@@ -34,6 +39,15 @@ internal sealed class ResourceAttributeConstants
     // Azure resource attributes constant values
     internal const string AzureAppServicePlatformValue = "azure.app_service";
     internal const string AzureCloudProviderValue = "azure";
+    internal const string AzureFunctionsPlatformValue = "azure.functions";
     internal const string AzureVmCloudPlatformValue = "azure.vm";
     internal const string AzureContainerAppsPlatformValue = "azure.container_apps";
+
+    // Match the Azure Functions host's instance ID precedence across hosting plans.
+    internal static readonly string[] AzureFunctionsInstanceIdEnvVars =
```
stevejgordon, line 47: Worth moving up to the `// Azure Functions environment variables` section above as these all relate?

stevejgordon, line 47:
```suggestion
    internal static readonly ImmutableArray<string> AzureFunctionsInstanceIdEnvVars =
```

stevejgordon: Will require a package reference to `System.Collections.Immutable` in the project also.

src/OpenTelemetry.Resources.Azure/AzureFunctionsResourceDetector.cs
```diff
@@ -0,0 +1,92 @@
...
+using OpenTelemetry.Trace;
+
+namespace OpenTelemetry.Resources.Azure;
+
+/// <summary>
+/// Resource detector for Azure Functions environments.
+/// </summary>
+internal sealed class AzureFunctionsResourceDetector : IResourceDetector
+{
+    internal static readonly IReadOnlyDictionary<string, string> AzureFunctionsResourceAttributes = new Dictionary<string, string>
```
stevejgordon, line 13: We could use `FrozenDictionary` on an `#if NET` path here. This is a convention we try to follow here for marginal performance improvements.

src/OpenTelemetry.Resources.Azure/AzureResourcesEventSource.cs
stevejgordon, line 1: I think we could leverage `InMemoryEventListener` in some of the tests (or add new ones) to improve coverage of the events themselves, ensuring they fire when expected.

test/OpenTelemetry.Resources.Azure.Tests/AzureResourceDetectorTests.cs
```diff
@@ -117,6 +118,93 @@ public void AppServiceResourceDetectorReturnsResourceWithAttributes()
...
+            Assert.Equal(2, resource.Attributes.Count());
+            Assert.DoesNotContain(resource.Attributes, attribute => attribute.Key == ResourceSemanticConventions.AttributeFaasVersion);
+        }
+    }
+
+    [Theory]
+    [InlineData("website-instance", "pod-instance", "container-instance", "website-instance")]
+    [InlineData(null, "pod-instance", "container-instance", "pod-instance")]
+    [InlineData(null, null, "container-instance", "container-instance")]
+    public void AzureFunctionsResourceDetectorSelectsAuthoritativeInstanceId(
```
stevejgordon, line 186: Can/should this also test the scenario of all three instance IDs being null?

src/OpenTelemetry.Resources.Azure/AppServiceResourceDetector.cs
```diff
@@ -24,6 +24,11 @@ public Resource Detect()
     {
         try
         {
+            if (Environment.GetEnvironmentVariable(ResourceAttributeConstants.AzureFunctionsWorkerRuntimeEnvVar) != null)
+            {
+                return new AzureFunctionsResourceDetector().Detect();
```
martincostello, line 29: I think this is a bit weird - it would be better to just register the new detector like we do with the existing ones, and have this one and the other each skip based on the environment, rather than have one delegate to the other.

src/OpenTelemetry.Resources.Azure/AzureResourcesEventSource.cs
Lewis-E: Added it to `AzureResourcesEventSourceLogsDetectionFailures` -- I didn't see any obvious existing tests for this, but if there's a good structure to test the failure without calling it directly, let me know and I'll switch to that.

src/OpenTelemetry.Resources.Azure/README.md
Lewis-E: Added to the readme in the upper section ✅

src/OpenTelemetry.Resources.Azure/CHANGELOG.md
Lewis-E: I've called it out as breaking here, but if you want any kind of change in the ordering logic or related logging, let me know.

src/OpenTelemetry.Resources.Azure/ResourceAttributeConstants.cs
Lewis-E: done and referenced

src/OpenTelemetry.Resources.Azure/CHANGELOG.md
Lewis-E: breaking change noted ✔️ 
let me know if you want any other breaking change handling

src/OpenTelemetry.Resources.Azure/AppServiceResourceDetector.cs
Lewis-E: made some changes, let me know if this is better? No more delegation.

Lewis-E:
@martincostello @stevejgordon apologies for the delay (oncall shifts are real), please let me know if the revised PR is decent

src/OpenTelemetry.Resources.Azure/CHANGELOG.md
```diff
@@ -20,6 +26,13 @@ Released 2026-Sep-10
   * `azure_vm` to `azure.vm`
   ([#5143](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5143))
 
+* **Breaking change**: Detect Azure Functions environments before Azure App
+  Service and emit Azure Functions resource attributes. When
+  `FUNCTIONS_WORKER_RUNTIME` is set, the detector no longer emits the App
+  Service-specific `azure.app.service.stamp`, `host.id`, or
+  `service.instance.id` attributes.
+  ([#5165](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5165))
```
martincostello: This is in the wrong section.

Lewis-E: Fixed, and set it back to the one expected change note.

src/OpenTelemetry.Resources.Azure/AzureFunctionsResourceDetector.cs
```diff
@@ -0,0 +1,103 @@
...
+        catch (Exception ex)
+        {
+            AzureResourcesEventSource.Log.FailedToDetectAzureFunctionsResources(ex);
+            return Resource.Empty;
+        }
+    }
+
+    private static Dictionary<string, string> CreateAzureFunctionsResourceAttributes()
+    {
+        return new Dictionary<string, string>
```
martincostello, line 84:
```suggestion
        return new Dictionary<string, string>(2)
```

src/OpenTelemetry.Resources.Azure/AzureResourceBuilderExtensions.cs
```diff
@@ -14,14 +14,16 @@ public static class AzureResourceBuilderExtensions
     internal static readonly Version SemanticConventionsVersion = new(1, 44, 0);
 
     /// <summary>
-    /// Enables Azure App Service resource detector.
+    /// Enables Azure App Service and Azure Functions resource detection.
```
martincostello, line 17: Is it correct that functions is being treated as part of App Service rather than its own separate opt-in extension method (e.g. `AddAzureFunctionsDetector()`)?

src/OpenTelemetry.Resources.Azure/OpenTelemetry.Resources.Azure.csproj
```diff
@@ -25,6 +25,10 @@
     <PackageReference Include="System.Text.Json" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'net462'">
+    <PackageReference Include="System.Collections.Immutable" />
+  </ItemGroup>
+
```
martincostello, line 31: I would avoid adding this dependency.

rajkumar-rangaraj:
I'm involving the Azure Functions team and would like approval from an Azure Functions engineer before this is merged.

RohitRanjanMS:
Hi [[@Lewis-E](https://github.com/Lewis-E)], we already maintain a `FunctionsResourceDetector` for the .NET isolated worker and register it through [`[UseFunctionsWorkerDefaults()](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.OpenTelemetry/ConfigureFunctionsOpenTelemetry.cs)`](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.OpenTelemetry/ConfigureFunctionsOpenTelemetry.cs). It could use some updates, and I’m happy to make those. You are welcome to open a PR we well.


Functions resource detection needs to account for several hosting environments. For .NET isolated-worker apps, I’d recommend using the detector we maintain and addressing any gaps there rather than introducing a second, independent implementation.


I’m also concerned about implicitly enabling Functions detection through `AddAzureAppServiceDetector()`, since App Service is only one of the environments where Functions can run. Although the skip logic prevents the two detectors in this PR from overwriting each other, combining the new detector with our existing worker detector can still produce conflicting attributes, such as `cloud.platform`, or override an explicitly configured `service.name`. I’d prefer separate opt-in registration and alignment with our existing detector’s behavior.

Lewis-E:
@RohitRanjanMS Thank you -- I will open a PR over there.

PR #5166, opened 2026-09-02
[Instrumentation.ServiceFabricRemoting] Support custom exception convertors

sablancoleis:
## Changes

Service Fabric lets applications register `IExceptionConvertor` implementations
so that custom exception types survive a remoting call. The provider attributes
in this package build the listener and the client factory internally and call
the Service Fabric overloads that take no convertors, so applications using them
have no way to opt in.

This became a problem in **Service Fabric SDK 8 (runtime 11)**, which stopped
enabling the `BinaryFormatter` fallback for remoting exception serialization by
default; **SDK 9 (runtime 12)** removes it altogether. See the
[Remoting V1 deprecation strategy](https://github.com/microsoft/service-fabric/blob/master/release_notes/Deprecated/RemotingV1.md).
Up to SDK 7.1 the defaults were `BinaryFormatter` on the listener and `Fallback`
on the client, which round-tripped any `[Serializable]` exception with no
configuration. From SDK 8 an exception type is only preserved if a convertor is
registered for it; anything else reaches the client as a `ServiceException`
wrapped in an `AggregateException`, so `catch (MyCustomException)` stops
matching.

* Unseal `TraceContextEnrichedServiceRemotingProviderAttribute` and
  `TraceContextEnrichedActorRemotingProviderAttribute` and add
  `GetServiceExceptionConvertors()` / `GetClientExceptionConvertors()` hooks,
  passing the results to the Service Fabric listener and client factory
  overloads that accept exception convertors. This extends the existing Service
  Fabric provider-attribute model rather than introducing a new abstraction, and
  keeps real convertor instances so they can take dependencies.
* Add a `RemotingExceptionDepth` property to control how many levels of inner
  exceptions are serialized.
* Document the manual composition path in the README. Both adapters are already
  public, so applications needing full control over listener or client factory
  construction can wrap their own objects directly. This was previously
  undocumented.

Behaviour is unchanged by default: the hooks return `null`, and
`RemotingExceptionDepth` only overrides the Service Fabric default when set.


src/OpenTelemetry.Instrumentation.ServiceFabricRemoting/CHANGELOG.md
```diff
@@ -2,6 +2,22 @@
...
+  registered convertor reaches the client as a `ServiceException`.
+  `TraceContextEnrichedServiceRemotingProviderAttribute` and
+  `TraceContextEnrichedActorRemotingProviderAttribute` are no longer sealed and
+  expose `GetServiceExceptionConvertors()` / `GetClientExceptionConvertors()`
+  for this purpose. A `RemotingExceptionDepth` property was also added to
+  control how many levels of inner exceptions are serialized.
+
+* Documented how to compose the instrumentation adapters manually, for
+  applications that need to configure the Service Fabric listener or client
+  factory directly instead of using the provider attributes.
```
martincostello: We don't typically put documentation changes into the CHANGELOGs.

src/OpenTelemetry.Instrumentation.ServiceFabricRemoting/README.md
```diff
@@ -189,6 +190,116 @@ the exception will be recorded at the server as an `ActivityEvent` or not
...
+        {
+            return [new MyExceptionConvertorClient()];
+        }
+    }
+```
+
+Then apply the derived attribute instead of the built-in one:
+
+```csharp
+   [assembly: MyRemotingProvider()]
```
martincostello:
```suggestion
   [assembly: MyRemotingProvider]
```

martincostello (approved):

src/OpenTelemetry.Instrumentation.ServiceFabricRemoting/CHANGELOG.md
sablancoleis: removed

src/OpenTelemetry.Instrumentation.ServiceFabricRemoting/README.md
sablancoleis: fixed

martincostello (approved):

PR #5172, opened 2026-09-05
[Resources.Host] Add host.ip and host.mac resource attributes

PetarJerinic:
Design discussion issue #1516.

First of the three proposed points there.

## Changes

`host.ip` and `host.mac` are emitted when `OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_NETWORK_ADDRESSES` is set to true. Both are opt_in in semconv, so they are off by default, meaning existing users are not affected.

Implementation follows the OpenTelemetry Collector's [system detector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.160.0/internal/metadataproviders/system/metadata.go#L274-L322) (interfaces that are Up and not loopback, skipping loopback addresses). There are 2 differences:
- An interface with an empty physical address contributes no MAC
- Duplicate values are removed from both attributes, first occurrence is kept

### DIFF
--- a/src/OpenTelemetry.Resources.Host/HostDetector.cs
+++ b/src/OpenTelemetry.Resources.Host/HostDetector.cs
@@ -6,6 +6,9 @@ using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Text;
 #endif
+using System.Net;
+using System.Net.NetworkInformation;
+using System.Net.Sockets;
 using Microsoft.Win32;
 using OpenTelemetry.Internal;
 
@@ -16,6 +19,8 @@ namespace OpenTelemetry.Resources.Host;
 /// </summary>
 internal sealed class HostDetector : IResourceDetector
 {
+    internal const string EnableNetworkAddressesEnvVarName = "OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_NETWORK_ADDRESSES";
+
 #if !NETFRAMEWORK
     private const string ETCMACHINEID = "/etc/machine-id";
     private const string ETCVARDBUSMACHINEID = "/var/lib/dbus/machine-id";
@@ -113,7 +118,9 @@ internal sealed class HostDetector : IResourceDetector
     {
         try
         {
-            var attributes = new List<KeyValuePair<string, object>>(3)
+            var networkAddressesEnabled = IsNetworkAddressesEnabled();
+
+            var attributes = new List<KeyValuePair<string, object>>(networkAddressesEnabled ? 5 : 3)
             {
                 new(HostSemanticConventions.AttributeHostName, Environment.MachineName),
             };
@@ -136,6 +143,11 @@ internal sealed class HostDetector : IResourceDetector
 #error Architecture is available in .NET Framework 4.7.1+, enable it when we move to that as minimum supported version
 #endif
 
+            if (networkAddressesEnabled)
+            {
+                AddNetworkAddresses(attributes);
+            }
+
             return new Resource(attributes, SchemaUrls.Get(SemanticConventionsVersion));
         }
         catch (InvalidOperationException ex)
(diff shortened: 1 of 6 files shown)


src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -10,7 +10,7 @@
 [![codecov.io](https://codecov.io/gh/open-telemetry/opentelemetry-dotnet-contrib/branch/main/graphs/badge.svg?flag=unittests-Resources.Host)](https://app.codecov.io/gh/open-telemetry/opentelemetry-dotnet-contrib?flags[0]=unittests-Resources.Host)
 
 > [!IMPORTANT]
-> Resources detected by this packages are defined by [experimental semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/v1.24.0/docs/resource/host.md).
+> Resources detected by this packages are defined by [experimental semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/resource/host.md).
```
martincostello, line 13: `SemanticConventionsVersion` in `HostDetector` needs updating to 1.44.0. You'll also need to check whether changing that version from 1.43.0 requires updating any other behaviour.

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -57,8 +57,30 @@ your application is running:
...
+[opt-in](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/resource/host.md)
+attributes and identify the machine, so they are not emitted by default. Set
+`OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_NETWORK_ADDRESSES` to `true` to
+emit them.
+
+Both attributes are read from network interfaces that are up, skipping
+loopback interfaces. `host.ip` also leaves out link-local addresses, and
+`host.mac` only includes interfaces that have a physical address. The values
+are captured once, when the resource is built, so addresses assigned after
+startup are not picked up.
```
martincostello: This isn't quite accurate - new addresses would be picked-up if resources were re-built by an application.

src/OpenTelemetry.Resources.Host/README.md
PetarJerinic: Updated to 1.44.0. It [renames](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/schemas/1.44.0#L6-L12) container/k8s metrics, nothing else needs changing.

martincostello (approved):

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -147,6 +159,35 @@ public Resource Detect()
         return Resource.Empty;
     }
 
+    internal static bool ShouldIncludeNetworkInterface(OperationalStatus operationalStatus, NetworkInterfaceType networkInterfaceType) =>
+        operationalStatus == OperationalStatus.Up && networkInterfaceType != NetworkInterfaceType.Loopback;
+
+    internal static bool ShouldIncludeIpAddress(IPAddress address)
+    {
+        if (IPAddress.IsLoopback(address) || address.IsIPv6LinkLocal)
```
lachmatt: why did you decide to filter out link local addresses, even though semantic conventions have one in [their example](https://github.com/open-telemetry/semantic-conventions/blob/24493fdd14534d5c3fc6bba2c3e8402af070fdb4/docs/resource/host.md?plain=1#L29)?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -184,6 +225,55 @@ private static IEnumerable<string> GetFilePaths()
     }
 #endif
 
+    private static bool IsNetworkAddressesEnabled() =>
+        bool.TryParse(Environment.GetEnvironmentVariable(EnableNetworkAddressesEnvVarName), out var enabled) && enabled;
+
+    private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attributes)
+    {
+        var ipAddresses = new List<string>();
+        var macAddresses = new List<string>();
```
lachmatt, line 226: I think deduplication might be needed

src/OpenTelemetry.Resources.Host/HostDetector.cs
PetarJerinic: Agreed to follow semconv.
It was to cut the noise. During testing on my mac, including them resulted in 10 addresses instead of 3. .NET formats them with a zone suffix (`fe80::1%14`), which is now stripped.

src/OpenTelemetry.Resources.Host/HostDetector.cs
PetarJerinic: Added for both ip and mac while keeping the first occurrence.

lachmatt (approved):

Issue #4417, opened 2026-05-20
[Exporter.Geneva] Remove AllowUnsafeBlocks, MemoryMarshal and Unsafe


martincostello:
Context: https://github.com/open-telemetry/opentelemetry-dotnet/issues/7324

The Geneva ~~InfluxDB and OneCollector~~ exporter uses `unsafe` for performance reasons:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/bfa4e8f7cef497066d8c0d15fa5cba6b36a5d24a/src/OpenTelemetry.Exporter.Geneva/OpenTelemetry.Exporter.Geneva.csproj#L6

The Geneva exporter also uses `MemoryMarshal` ([code search](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-dotnet-contrib+MemoryMarshal+language%3Acsharp&type=code)) and `Unsafe`:

https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/ac5a82448c9f606f7b8e8b1339d0fcd41d5ad591/src/OpenTelemetry.Exporter.Geneva/External/LinuxTracepoints-Net/Provider/EventHeaderValue128.cs#L47

C# ~~15~~ 16 in the .NET ~~11~~ 12 timeframe is looking to make changes to unsafe code to reduce its usage for security reasons, see [_Unsafe Evolution_](https://github.com/dotnet/csharplang/blob/main/proposals/unsafe-evolution.md) for further context.

To improve the security posture of the exporter, we should remove the remaining usages of unsafe code even in the face of potential performance degradation.

This may take a while to achieve as if the degradation is large, we may have to make other changes elsewhere to compensate.

The `MemoryMarshal` and `Unsafe` usages _may_ get surfaced in later previews of .NET 11 in #3867.

martincostello:
[Improving C# Memory Safety](https://devblogs.microsoft.com/dotnet/improving-csharp-memory-safety/)

martincostello:
Marking as help wanted, but it's likely someone from Microsoft will need to work on this as they're the only consumers and have access to a real endpoint end-to-end for validation.

martincostello:
Testing with .NET 11 rc.1 and the following project settings, there are 166 compiler errors generated:

```xml
<PropertyGroup>
  <Features>$(Features);updated-memory-safety-rules</Features>
  <LangVersion>preview</LangVersion>
</PropertyGroup>
```

PR #5173, opened 2026-09-06
[Resources.Host] Add host.cpu.* resource attributes

PetarJerinic:
Design discussion issue #1516

Second of the three proposed points there.

## Changes

The six `host.cpu.*` attributes are emitted when `OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO` is set to true. All six are opt_in in the semconv, so they are off by default, which means nothing changes for an existing user.

Sources are per OS, explanation is present in README.md


PetarJerinic:
> This PR was marked stale due to lack of activity. It will be closed in 7 days.

Waiting for #5172 to be completed.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,115 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+            _ => 0,
+        };
+
+        if (multiplier == 0)
+        {
+            return null;
+        }
+
+        size *= multiplier;
+        return size <= int.MaxValue ? (int?)size : null;
```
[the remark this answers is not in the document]
PetarJerinic: Addressed.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -266,6 +418,254 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
...
+            };
+
+            using var process = Process.Start(startInfo);
+            if (process != null)
+            {
+                if (!process.WaitForExit(timeoutMilliseconds))
+                {
+                    HostResourceEventSource.Log.ProcessTimeout("Process did not exit within the given timeout");
+                    return null;
+                }
```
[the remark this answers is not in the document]
PetarJerinic: Addressed.

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -81,6 +82,24 @@ includes interfaces that have a physical address. The values are read when the
...
+### CPU information
+
+`host.cpu.vendor.id`, `host.cpu.family`, `host.cpu.model.id`,
+`host.cpu.model.name`, `host.cpu.stepping` and `host.cpu.cache.l2.size` are
+[opt-in](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/resource/host.md)
+attributes, so they are not emitted by default. Set
+`OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO` to `true` to emit them.
+
+Which attributes are available depends on the CPU architecture rather than
+the operating system, and an attribute whose source is missing is left out.
```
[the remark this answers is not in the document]
PetarJerinic: Addressed.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -19,15 +23,39 @@ namespace OpenTelemetry.Resources.Host;
...
+    [
+        new("machdep.cpu.vendor", HostSemanticConventions.AttributeHostCpuVendorId),
+        new("machdep.cpu.family", HostSemanticConventions.AttributeHostCpuFamily),
+        new("machdep.cpu.model", HostSemanticConventions.AttributeHostCpuModelId),
+        new("machdep.cpu.brand_string", HostSemanticConventions.AttributeHostCpuModelName),
+        new("machdep.cpu.stepping", HostSemanticConventions.AttributeHostCpuStepping),
+    ];
+
+    private static readonly string MacOsCpuSysctlArguments =
+        string.Join(" ", MacOsCpuAttributes.Select(attribute => attribute.Key)) + " " + MacOsCacheSizeKey;
```
martincostello: Where available, [`ArgumentList`](https://learn.microsoft.com/dotnet/api/system.diagnostics.processstartinfo.argumentlist) is preferably to building the commands into a string like this.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -118,8 +146,10 @@ public Resource Detect()
         try
         {
             var networkAddressesEnabled = IsNetworkAddressesEnabled();
+            var cpuInfoEnabled = IsCpuInfoEnabled();
+            var capacity = GetAttributeCapacity(networkAddressesEnabled, cpuInfoEnabled);
```
martincostello: Not sure this needs its own method.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
     internal static string[] RemoveDuplicates(IEnumerable<string> values) =>
         values.Distinct().ToArray();
 
+    internal static string? NormalizeCpuValue(string? value)
```
martincostello: This doesn't appear to be CPU-specific. It's just a trim that treats empty as `null`.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+        }
+
+        var trimmed = value.Trim();
+        return trimmed.Length == 0 ? null : trimmed;
+    }
+
+    // The processor blocks after the first repeat these fields for the other cores, so the
+    // first match is the one to report.
+    internal static string? ParseCpuInfoField(string? cpuInfo, string fieldName) =>
+        ParseFieldValue(cpuInfo, fieldName);
```
martincostello: What's the benefit to this method other than it having a different name?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+
+    internal static string? ParseCpuIdentifierToken(string? identifier, string keyword)
```
martincostello: Similarly this seems CPU independent - it's just splitting and trimming a sequence of values separated by spaces.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+            if (string.Equals(tokens[i], keyword, StringComparison.Ordinal))
+            {
+                return NormalizeCpuValue(tokens[i + 1]);
+            }
+        }
+
+        return null;
+    }
+
+    internal static string? ParseSysctlField(string? output, string key) =>
```
martincostello: Again, why do we need this method with a different name?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+    {
+        if (getRegistryValue != null)
+        {
+            try
+            {
+                AddCpuAttribute(attributes, HostSemanticConventions.AttributeHostCpuVendorId, NormalizeCpuValue(getRegistryValue("VendorIdentifier")));
+                AddCpuAttribute(attributes, HostSemanticConventions.AttributeHostCpuModelName, NormalizeCpuValue(getRegistryValue("ProcessorNameString")));
+
+                // Family, model and stepping have no registry value of their own; they are only
+                // available inside "AMD64 Family 25 Model 1 Stepping 1", whose leading token is the
```
martincostello, line 332: Is the value documented as being stable with regards to this format?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -180,6 +215,120 @@ internal static string[] FormatIpAddresses(IEnumerable<IPAddress> addresses) =>
...
+                AddCpuAttribute(attributes, HostSemanticConventions.AttributeHostCpuStepping, ParseCpuIdentifierToken(identifier, "Stepping"));
+            }
+            catch (Exception ex)
+            {
+                HostResourceEventSource.Log.ResourceAttributesExtractException(nameof(HostDetector), ex);
+            }
+        }
+
+        // The cache size comes from kernel32, so neither a missing registry key nor a failed
+        // read may skip it.
```
martincostello:
```suggestion
```

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -266,6 +423,265 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
...
+    }
+
+    private static void AddCpuInfoWindows(List<KeyValuePair<string, object>> attributes)
+    {
+#if NET
+        if (!OperatingSystem.IsWindows())
+        {
+            return;
+        }
+#endif
```
martincostello, line 459: If you guard the methods with `[SupportedOSPlatform("windows")]` (or `linux`, `macos`, etc.) then these OS guards should be able to be removed because they're already guarded?

Then you wouldn't need the dedicated `CreateRegistryValueReader` method?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -266,6 +423,265 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
...
+        {
+            uint length = 0;
+            _ = NativeMethods.GetLogicalProcessorInformation(IntPtr.Zero, ref length);
+            if (length == 0)
+            {
+                return null;
+            }
+
+            buffer = Marshal.AllocHGlobal((int)length);
+            if (!NativeMethods.GetLogicalProcessorInformation(buffer, ref length))
```
martincostello, line 493: Is the length guaranteed not to be able to change since the previous call?

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -19,15 +23,39 @@ namespace OpenTelemetry.Resources.Host;
...
     internal const string EnableNetworkAddressesEnvVarName = "OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_NETWORK_ADDRESSES";
+    internal const string EnableCpuInfoEnvVarName = "OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO";
+
+    private const string WindowsCpuRegistryKey = @"HARDWARE\DESCRIPTION\System\CentralProcessor\0";
 
 #if !NETFRAMEWORK
     private const string ETCMACHINEID = "/etc/machine-id";
     private const string ETCVARDBUSMACHINEID = "/var/lib/dbus/machine-id";
+    private const string ProcCpuInfo = "/proc/cpuinfo";
+    private const string SysfsCpuCache = "/sys/devices/system/cpu/cpu0/cache";
```
martincostello: As this only seems to be used in one place, I'd make the constant local to the method.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -266,6 +423,265 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+            };
+
+            using var process = Process.Start(startInfo);
+            if (process != null)
+            {
+                if (!process.WaitForExit(timeoutMilliseconds))
+                {
+                    HostResourceEventSource.Log.ProcessTimeout("Process did not exit within the given timeout");
```
martincostello: Tell the user what the timeout was.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -266,6 +423,265 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
...
+        catch (Exception ex)
+        {
+            HostResourceEventSource.Log.ResourceAttributesExtractException(nameof(HostDetector), ex);
+        }
+
+        return null;
+    }
+#endif
+
+    private static void AddCpuAttribute(List<KeyValuePair<string, object>> attributes, string key, string? value)
```
martincostello: These methods don't appear to be CPU-specific - they'd just guarded helpers.

src/OpenTelemetry.Resources.Host/HostDetector.cs
```diff
@@ -373,4 +789,47 @@ private static void AddNetworkAddresses(List<KeyValuePair<string, object>> attri
         return null;
     }
 #endif
+
+    internal static class NativeMethods
```
martincostello: Could we use the P/Invoke source generator here for the TFMs that support it without over-complicating the code?

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -81,6 +82,24 @@ includes interfaces that have a physical address. The values are read when the
 resource is built, so addresses assigned afterwards are only picked up if the
 resource is built again.
 
+### CPU information
+
+`host.cpu.vendor.id`, `host.cpu.family`, `host.cpu.model.id`,
+`host.cpu.model.name`, `host.cpu.stepping` and `host.cpu.cache.l2.size` are
+[opt-in](https://github.com/open-telemetry/semantic-conventions/blob/v1.44.0/docs/resource/host.md)
+attributes, so they are not emitted by default. Set
+`OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO` to `true` to emit them.
```
martincostello: This would be more readable as the summary, then the an `<ul>` list of attributes.

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -3,15 +3,55 @@
...
+        "machdep.cpu.brand_string: Apple M5\n" +
+        "hw.l2cachesize: 6291456\n";
+
+    private const string IntelSysctlOutput =
+        "machdep.cpu.vendor: GenuineIntel\n" +
+        "machdep.cpu.family: 6\n" +
+        "machdep.cpu.model: 6\n" +
+        "machdep.cpu.stepping: 1\n" +
+        "machdep.cpu.brand_string: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz\n" +
+        "hw.l2cachesize: 12288000\n";
```
martincostello, line 54: IMHO these would be more readable as fields built with a `string.Join("\n", ...)`.

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -316,4 +357,214 @@ public void TestFormatPhysicalAddress(byte[] address, string expected) =>
     [Fact]
     public void TestFormatPhysicalAddressWithEmptyAddress() =>
         Assert.Null(HostDetector.FormatPhysicalAddress(PhysicalAddress.None));
+
+    [Theory]
+    [InlineData("true")]
+    [InlineData("True")]
+    [InlineData("TRUE")]
+    public void TestHostCpuInfoEnabled(string value)
```
martincostello: This would probably be less complicated if there was a different test per OS, which is skipped at the start of the current OS doesn't match the test.

src/OpenTelemetry.Resources.Host/HostDetector.cs
PetarJerinic: No, it is not. If the format changes the keyword isn't found, and information is left out in that case. I couldn't find any documented registry source, WMI's `Win32_Processor.Family` is the SMBIOS code, not the CPUID family
If preferred, I can drop these on Windows due to the lack of a stable source.

src/OpenTelemetry.Resources.Host/HostDetector.cs
PetarJerinic: No, but it can't overrun. If the size grows, the second call would fail with `ERROR_INSUFFICIENT_BUFFER` in which case null would be returned. [GetLogicalProcessorInformation page](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformation) has a pattern with a retry loop, we could implement something similar, if preferred.

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
PetarJerinic: They're consts atm because they're used in `[InlineData]`, which needs constants. With `string.Join("\n", ...)` affected theories would have to use `MemberData`. Is that fine?

PetarJerinic: Done.

src/OpenTelemetry.Resources.Host/HostDetector.cs
PetarJerinic: Done, removed the dedicated `CreateRegistryValueReader` method in favor of OS attribute guards.

src/OpenTelemetry.Resources.Host/HostDetector.cs
martincostello: I think it's fine to make an assumption as long as the code is robust to if the format changes.

src/OpenTelemetry.Resources.Host/HostDetector.cs
martincostello: Same is true for `ProcCpuInfo`.

src/OpenTelemetry.Resources.Host/OpenTelemetry.Resources.Host.csproj
```diff
@@ -9,6 +9,7 @@
 
   <PropertyGroup Condition="'$([MSBuild]::GetTargetFrameworkIdentifier(`$(TargetFramework)`))' == '.NETCoreApp'">
     <IsAotCompatible>true</IsAotCompatible>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
martincostello, line 12: Is this required by the source generator? If so, then we should revert it. We could revisit in future when/if the P/Invoke source generator doesn't require unsafe code with the future changes to memory safety rules (for example the JSON source generator is making such changes).

See https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/4417.

src/OpenTelemetry.Resources.Host/README.md
```diff
@@ -84,11 +84,17 @@ resource is built again.
...
-`OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO` to `true` to emit them.
+attributes are not emitted by default. Set
+`OTEL_DOTNET_EXPERIMENTAL_HOST_RESOURCE_ENABLE_CPU_INFO` to `true` to emit them:
+
+- `host.cpu.vendor.id`,
+- `host.cpu.family`,
+- `host.cpu.model.id`,
+- `host.cpu.model.name`,
+- `host.cpu.stepping`,
+- `host.cpu.cache.l2.size`.
```
martincostello, line 97: Nit: sorting.

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -316,4 +393,246 @@ public void TestFormatPhysicalAddress(byte[] address, string expected) =>
...
+
+#if NET
+    [SkippableTheory]
+    [InlineData("true")]
+    [InlineData("True")]
+    [InlineData("TRUE")]
+    public void TestHostCpuInfoEnabledLinux(string value)
+    {
+        Skip.IfNot(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+            && RuntimeInformation.ProcessArchitecture is not (Architecture.Arm or Architecture.Arm64));
```
martincostello, line 432:
```suggestion
        Skip.IfNot(
            RuntimeInformation.IsOSPlatform(OSPlatform.Linux) &&
            RuntimeInformation.ProcessArchitecture is not (Architecture.Arm or Architecture.Arm64));
```

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -316,4 +393,246 @@ public void TestFormatPhysicalAddress(byte[] address, string expected) =>
...
+    }
+
+    [SkippableTheory]
+    [InlineData("true")]
+    [InlineData("True")]
+    [InlineData("TRUE")]
+    public void TestHostCpuInfoEnabledArmLinux(string value)
+    {
+        Skip.IfNot(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
+            && RuntimeInformation.ProcessArchitecture is Architecture.Arm or Architecture.Arm64);
```
martincostello, line 455:
```suggestion
        Skip.IfNot(
            RuntimeInformation.IsOSPlatform(OSPlatform.Linux) &&
            RuntimeInformation.ProcessArchitecture is Architecture.Arm or Architecture.Arm64);
```

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -316,4 +393,246 @@ public void TestFormatPhysicalAddress(byte[] address, string expected) =>
...
+    }
+
+#if NET
+    [SkippableTheory]
+    [InlineData("true")]
+    [InlineData("True")]
+    [InlineData("TRUE")]
+    public void TestHostCpuInfoEnabledLinux(string value)
+    {
+        Skip.IfNot(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)
```
martincostello, line 431: As these tests are in a `#if NET` block you can use the `OperatingSystem.Is*()` methods instead.

test/OpenTelemetry.Resources.Host.Tests/HostDetectorTests.cs
```diff
@@ -316,4 +393,246 @@ public void TestFormatPhysicalAddress(byte[] address, string expected) =>
...
+    [InlineData(null, null)]
+    [InlineData("9223372036854775807 K", null)]
+    [InlineData("18014398509481985 K", null)]
+    [InlineData("2097151 K", 2147482624)]
+    [InlineData("2097152 K", null)]
+    public void TestParseCacheSize(string? value, int? expected) =>
+        Assert.Equal(expected, HostDetector.ParseCacheSize(value));
+
+    [Fact]
+    public void TestSystemLogicalProcessorInformationCacheLayout()
```
martincostello, line 555: Do we need this test? Seems a bit low-level given that it needs to do P/Invoke. Can a similar test outcome be achieved just through the host detector itself rather than needing to know the P/Invoke implementation details?

PR #5177, opened 2026-09-07
[Instrumentation.Http] Invoke EnrichWithException for canceled requests on pre-.NET 10 targets

### LINKED ISSUE #1793: Allow listening to TaskCanceled exceptions on HTTP client instrumentation
# Feature Request

**Is your feature request related to a problem?**

I would like to add custom tags when timeouts happen on clients using the auto instrumentation, but currently when a HTTP Request is canceled and the request fails neither `OnStopActivity` nor `OnException` are called using the `options.Enrich` hook.

It seems this happens because when a Task is canceled the onStopActivity is called (and it handles tasks canceled):
https://github.com/open-telemetry/opentelemetry-dotnet/blob/884b224508351a65ab260c404fad5e25b341db97/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L160-L178

And since there is no response the `onStopActivity` will not be called:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/884b224508351a65ab260c404fad5e25b341db97/src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs#L180-L198

**Describe the solution you'd like:**

Maybe OnStopActivity or OnException could be called signaling the timeout?

shoumikchakravarty-dev:
Fixes #1793 

## Changes
Canceling an HttpClient request in versions before .NET 10 never raised the System.Net.Http.Exception diagnostic event, so OnException and options.EnrichWithException were not called for cancellations or timeouts. OnStopActivity did set the error status and error.type, but callers could not run their own enrichment logic for these cases.

This change adds an EnrichWithException call to the existing cancellation path in OnStopActivity for non‑NET10 targets, using a TaskCanceledException and following the same try/catch pattern used elsewhere. .NET 10 behavior is unchanged because its diagnostics already emit the exception event.

A new test, HttpCancellationInvokesEnrichWithException, is included for NET && !NET10_0_OR_GREATER and verifies that EnrichWithException receives a TaskCanceledException on a canceled request. The full HttpClientTests suite passes, including the existing HttpCancellationLogsError test.

## AI Disclosure

AI helped identify the root cause (traced the code to confirm `OnException` never fires for cancellation on pre-.NET 10 targets) . I reviewed the change, ran `dotnet build` and the full `HttpClientTests` suite locally before opening this PR, and confirmed no regressions.


src/OpenTelemetry.Instrumentation.Http/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Fix `EnrichWithException` not being invoked for canceled HTTP client
+  requests on target frameworks prior to .NET 10.
+  ([#1793](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1793))
```
stevejgordon: This should be the link to this PR.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -194,6 +194,21 @@ public void OnStopActivity(Activity activity, object? payload)
                             activity.SetStatus(ActivityStatusCode.Error, "Task Canceled");
                             activity.SetTag(SemanticConventions.AttributeErrorType, typeof(TaskCanceledException).FullName);
                         }
+
+                        // Task cancellation won't trigger OnException, so EnrichWithException is
```
stevejgordon: This opener repeats the comment above. Should we move that up higher as the justification for special casing various details and let the code document the rest?

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
```diff
@@ -194,6 +194,21 @@ public void OnStopActivity(Activity activity, object? payload)
...
+                        // never invoked for canceled requests on this code path. Invoke it here so
+                        // that callers get the same enrichment opportunity they get for other errors.
+                        if (this.options.EnrichWithException is { } enrichOnCancellation)
```
stevejgordon: Nit: Everywhere else uses just `enrich` for the variable name. Worth doing for consistency?

test/OpenTelemetry.Instrumentation.Http.Tests/HttpClientTests.cs
```diff
@@ -202,6 +202,52 @@ public async Task HttpCancellationLogsError()
...
 #endif
 
+#if NET && !NET10_0_OR_GREATER
+    [Fact]
+    public async Task HttpCancellationInvokesEnrichWithException()
+    {
+        // Prior to .NET 10, task cancellation does not raise the diagnostic
+        // "Exception" event, so OnException (and therefore EnrichWithException)
+        // is never invoked for canceled requests unless the cancellation branch
+        // in OnStopActivity invokes it directly. See issue #1793.
```
stevejgordon: Please add the full URL for the issue to the comment.

stevejgordon (commented):
I've added some code comments, but as a whole, I'm not sure this is the right solution to the original issue. Since all this does it invoke the callback with a synthetic new `TaskCanceledException`, there's nothing the delegate can really do that provides much value here. I think a consumer could achieve the same today with a processor handling `OnEnd` when `error.type` is set with `System.Threading.Tasks.TaskCanceledException`.

src/OpenTelemetry.Instrumentation.Http/Implementation/HttpHandlerDiagnosticListener.cs
shoumikchakravarty-dev: makes sense.

shoumikchakravarty-dev:
> I've added some code comments, but as a whole, I'm not sure this is the right solution to the original issue. Since all this does it invoke the callback with a synthetic new `TaskCanceledException`, there's nothing the delegate can really do that provides much value here. I think a consumer could achieve the same today with a processor handling `OnEnd` when `error.type` is set with `System.Threading.Tasks.TaskCanceledException`.

The goal here was to keep the enrichment callback firing for cancellations on older runtimes. It’s not about the exception itself being useful but to keep the behavior consistent. I can add the comment if that is useful. Please advise.

shoumikchakravarty-dev:
Hello @stevejgordon - Could you please review further for any other changes required here. Thank you.

shoumikchakravarty-dev:
This PR is relevant. The review comments were addressed . waiting for the reviewer @stevejgordon to approve and merge or if there is any further comments to it

martincostello:
We discussed this PR in this week's SIG call.

We don't feel it's worth emulating the different behaviour here to account for a difference in how .NET itself behaves. .NET 8 and 9 also go out of support in less than two months.

Issue #5178, opened 2026-09-07
[AWSLambda] Support the datasource faas.trigger type for S3 events


ashishsinghnr:
### Component

OpenTelemetry.Instrumentation.AWSLambda

### Is your feature request related to a problem?

`GetFaasTrigger` classifies HTTP triggers as `http` and, SQS/SNS triggers as `pubsub`. Every other event source still reports`faas.trigger = other`, including S3.

A function triggered by an S3 object event is responding to a data source operation, which the conventions describe as the `datasource` trigger type. The conventions name S3 explicitly in the attribute definitions:

* [`faas.document.collection`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/):
  "The name of the source on which the triggering operation was performed. For example, in Cloud Storage or **S3 corresponds to the bucket name**"
* [`faas.document.name`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/faas/):
  "The document name/table subjected to the operation. For example, in Cloud Storage or **S3 is the name of the file**"

So S3 is not an inferred fit for `datasource` — it is one of the two examples the specification was written around, and it currently reports `other`.

### What is the expected behavior?

Report `faas.trigger = datasource` for S3 events, together with the `faas.document.*` attributes the conventions require.

Only `datasource` is not satisfied by setting `faas.trigger` alone. The [Datasource section](https://opentelemetry.io/docs/specs/semconv/faas/faas-spans/#datasource)
requires:

| Attribute | Requirement level |
| --------- | ----------------- |
| `faas.document.collection` | **Required** |
| `faas.document.operation`  | **Required** |
| `faas.document.name`       | Recommended |
| `faas.document.time`       | Recommended |

Setting `faas.trigger = datasource` without the two required attributes would be less conformant than the current `other`, so the attributes have to be part of the same change.

**S3 maps completely.** Every attribute has a real property on [S3Event.S3EventNotificationRecord](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/event/S3EventNotification.S3EventNotificationRecord.html) :

| Attribute | Source | Type |
| --------- | ------ | ---- |
| `faas.document.collection` | `S3.Bucket.Name` | `string` |
| `faas.document.name`       | `S3.Object.Key` | `string` |
| `faas.document.operation`  | `EventName`, mapped below | `string` |
[...]

PR #5146, opened 2026-08-28
[AWSLambda] Set faas.trigger to pubsub for SQS and SNS triggered functions

ashishsinghnr:
Fixes #5145

## Changes

GetFaasTrigger in AWSLambdaUtils only distinguishes HTTP request types, so every non-HTTP event source falls through to "other". Functions triggered by SQS or SNS are message-driven, and both the general FaaS conventions and the AWS Lambda conventions require faas.trigger to be pubsub for these sources:

[FaaS spans — PubSub](https://opentelemetry.io/docs/specs/semconv/faas/faas-spans/#pubsub)
[AWS Lambda — SQS event](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-event):
"faas.trigger MUST be set to pubsub."
[AWS Lambda — SQS message](https://opentelemetry.io/docs/specs/semconv/faas/aws-lambda/#sqs-message):
"faas.trigger MUST be set to pubsub."

This adds an IsPubSubRequest check, mirroring the existing IsHttpRequest, so faas.trigger is pubsub for SQSEvent, SQSEvent.SQSMessage, SNSEvent and SNSEvent.SNSRecord.

The record-level types are included because ExtractParentContext already treats them as message-driven sources in order to extract parent context from them, while GetFaasTrigger did not classify them. Both are called with the same input in OnFunctionStart, so passing an individual message previously produced correct trace linkage but a trigger type of other.

Behaviour is unchanged for all other inputs: HTTP request types still report http, and plain payloads from a direct invocation still report other.

## Merge requirement checklist

* [x] [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/CONTRIBUTING.md) guidelines followed (license requirements, nullable enabled, static analysis, etc.)
* [x] Unit tests added/updated
* [x] Appropriate `CHANGELOG.md` files updated for non-trivial changes
* [ ] ~~Changes in public API reviewed (if applicable)~~

PR #5180, opened 2026-09-09
[AWSLambda] Add EnrichWithInput to describe unrecognised triggers

ashishsinghnr:
Fixes #
Design discussion issue #5178 

## Changes

`GetFaasTrigger` classifies API Gateway and Application Load Balancer requests as `http` and, since #5146, SQS and SNS events as `pubsub`. Every other event source reports `other`.

#5178 describe about supporting `datasource` for S3, and the answer there was to avoid new dependencies and instead make this achievable by the user for an arbitrary trigger. This adds that:

```csharp
public Action<Activity, object?, ILambdaContext>? EnrichWithInput { get; set; }
```
The action is invoked with the invocation Activity ,the function input and the ILambdaContext, so a function can describe its own trigger. This follows the EnrichWith* pattern already used by the AspNetCore, Http, GrpcNetClient and ElasticsearchClient instrumentations.

For the S3 case from #5178, the user references `Amazon.Lambda.S3Events` in their own function, where they already have it and this package takes no new dependency:

```
.AddAWSLambdaConfigurations(options => options.EnrichWithInput = (activity, input, context) =>
{
    if (input is S3Event { Records.Count: 1 } s3Event)
    {
        var record = s3Event.Records[0];
        activity.SetTag("faas.trigger", "datasource");
        activity.SetTag("faas.document.collection", record.S3.Bucket.Name);
        activity.SetTag("faas.document.operation", "insert");
        activity.SetTag("faas.document.name", WebUtility.UrlDecode(record.S3.Object.Key));
    }

    activity.SetTag("aws.lambda.log_stream", context.LogStreamName);
})
```
This also covers two cases a built-in classification could not:

- Event families with no well-known `faas.document.operation` value. S3 publishes eleven families; only `ObjectCreated` and `ObjectRemoved` map to insert/delete, and the conventions permit a custom value for the rest.
- Sources with no .NET event package. There is no `Amazon.Lambda.DocumentDBEvents`, so DocumentDB change streams are deserialised into user-defined types that no type-based detection in this library could classify.

Named EnrichWithInput to match the `EnrichWith* ` convention  used by the other instrumentations, and the `input `parameter name already in this package's public API. 
Happy to rename if you would prefer something else — it is unshipped.

### Verification
faas.trigger values, the trigger-override path, the URL-decoded object key,the ILambdaContext argument, the IsAllDataRequested gate, the async path, and the throwing-callback path are covered by unit tests. Each was checked by mutation — removing the invocation fails 4 tests,
removing the IsAllDataRequested guard fails 1, removing the try/catch fails 1.

The example above was also run as a real consumer: packed from this branch into a local NuGet feed and executed against the [S3 notification payload from the AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html), deserialised by DefaultLambdaJsonSerializer. The URL-encoded key S3 actually sends (`exports/2026-09-08+daily%3D1.csv`) is decoded to `exports/2026-09-08 daily=1.csv`, and faas.trigger carries a single datasource tag.

### DIFF
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 using System.Diagnostics;
+using Amazon.Lambda.Core;
 using OpenTelemetry.AWS;
 
 namespace OpenTelemetry.Instrumentation.AWSLambda;
@@ -26,15 +27,16 @@ public class AWSLambdaInstrumentationOptions
     public bool SetParentFromBatch { get; set; }
 
     /// <summary>
-    /// Gets or sets an action to enrich the invocation Activity from the function input.
-    /// Use it to describe a trigger the instrumentation does not recognise, for example to set
-    /// <c>faas.trigger</c> and the <c>faas.document.*</c> attributes for an Amazon S3 event.
+    /// Gets or sets an action to enrich the invocation Activity from the function input and
+    /// the Lambda context. Use it to describe a trigger the instrumentation does not recognise,
+    /// for example to set <c>faas.trigger</c> and the <c>faas.document.*</c> attributes for an
+    /// Amazon S3 event.
     /// </summary>
     /// <remarks>
     /// The action runs after the activity is created, so values it sets are not seen by samplers,
     /// and it applies process-wide rather than per-provider.
     /// </remarks>
-    public Action<Activity, object?>? EnrichWithInput { get; set; }
+    public Action<Activity, object?, ILambdaContext>? EnrichWithInput { get; set; }
 
     /// <inheritdoc cref="AWSLambda.SemanticConventionVersion"/>
     public SemanticConventionVersion SemanticConventionVersion { get; set; } = AWSSemanticConventions.DefaultSemanticConventionVersion;
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+using System.Diagnostics;
 using OpenTelemetry.AWS;
 
 namespace OpenTelemetry.Instrumentation.AWSLambda;
@@ -23,6 +24,17 @@ public class AWSLambdaInstrumentationOptions
     /// Currently, the only event type to which this applies is SQS.
     /// </remarks>
     public bool SetParentFromBatch { get; set; }
+
+    /// <summary>
+    /// Gets or sets an action to enrich the invocation Activity from the function input.
+    /// Use it to describe a trigger the instrumentation does not recognise, for example to set
+    /// <c>faas.trigger</c> and the <c>faas.document.*</c> attributes for an Amazon S3 event.
+    /// </summary>
+    /// <remarks>
+    /// The action runs after the activity is created, so values it sets are not seen by samplers,
+    /// and it applies process-wide rather than per-provider.
+    /// </remarks>
+    public Action<Activity, object?>? EnrichWithInput { get; set; }
 
     /// <inheritdoc cref="AWSLambda.SemanticConventionVersion"/>
     public SemanticConventionVersion SemanticConventionVersion { get; set; } = AWSSemanticConventions.DefaultSemanticConventionVersion;
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -2,16 +2,16 @@
 
 ## Unreleased
 
+* The `faas.trigger` span attribute is now set to `pubsub` for functions
+  triggered by SQS or SNS events, instead of `other`.
+  ([#5146](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5146))
+
 * Added `AWSLambdaInstrumentationOptions.EnrichWithInput`, an action invoked with
   the invocation `Activity` and the function input. This allows a function to
   describe a trigger the instrumentation does not classify itself, such as setting
   `faas.trigger` and the `faas.document.*` attributes for an Amazon S3 or Amazon
   DynamoDB event.
   ([#5180](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5180))
-
-* The `faas.trigger` span attribute is now set to `pubsub` for functions
-  triggered by SQS or SNS events, instead of `other`.
-  ([#5146](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5146))
 
 ## 1.18.0
 
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -6,8 +6,7 @@
   the invocation `Activity` and the function input. This allows a function to
   describe a trigger the instrumentation does not classify itself, such as setting
   `faas.trigger` and the `faas.document.*` attributes for an Amazon S3 or Amazon
-  DynamoDB event, without this package taking a dependency on those event
-  packages.
+  DynamoDB event.
   ([#5180](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5180))
 
 * The `faas.trigger` span attribute is now set to `pubsub` for functions
--- a/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
+++ b/src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
@@ -7,10 +7,10 @@
   ([#5146](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5146))
 
 * Added `AWSLambdaInstrumentationOptions.EnrichWithInput`, an action invoked with
-  the invocation `Activity` and the function input. This allows a function to
-  describe a trigger the instrumentation does not classify itself, such as setting
-  `faas.trigger` and the `faas.document.*` attributes for an Amazon S3 or Amazon
-  DynamoDB event.
+  the invocation `Activity`, the function input and the `ILambdaContext`. This
+  allows a function to describe a trigger the instrumentation does not classify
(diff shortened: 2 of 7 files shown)


src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* Added `AWSLambdaInstrumentationOptions.EnrichWithInput`, an action invoked with
+  the invocation `Activity` and the function input. This allows a function to
+  describe a trigger the instrumentation does not classify itself, such as setting
+  `faas.trigger` and the `faas.document.*` attributes for an Amazon S3 or Amazon
+  DynamoDB event, without this package taking a dependency on those event
+  packages.
```
martincostello:
```suggestion
* Added `AWSLambdaInstrumentationOptions.EnrichWithInput`, an action invoked with
  the invocation `Activity` and the function input. This allows a function to
  describe a trigger the instrumentation does not classify itself, such as setting
  `faas.trigger` and the `faas.document.*` attributes for an Amazon S3 or Amazon
  DynamoDB event.
```

src/OpenTelemetry.Instrumentation.AWSLambda/CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+* Added `AWSLambdaInstrumentationOptions.EnrichWithInput`, an action invoked with
```
martincostello, line 9: This should be at the end of the section.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -35,6 +35,48 @@ AWS lambda instrumentation:
...
 * [`DisableAwsXRayContextExtraction`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
 * [`SetParentFromBatch`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+* [`EnrichWithInput`](/src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs)
+
+### Enriching the invocation span
+
+`faas.trigger` is set to `http` for API Gateway and Application Load Balancer
+requests and to `pubsub` for SQS and SNS events. Other event sources report
+`other`, because classifying them would require this package to depend on their
+event packages.
```
martincostello: I don't think we need the design justification here in the user-facing documentation.

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
```diff
@@ -35,6 +35,48 @@ AWS lambda instrumentation:
...
+        }
+    })
+    .Build();
+```
+
+> [!NOTE]
+> The action runs after the activity is created, so attributes it sets are not
+> visible to samplers. Exceptions thrown by the action are caught so that they
+> cannot fail the invocation, and are not reported; keep the action defensive, or
+> handle errors within it if you need them logged.
```
martincostello: Not sure we need the "don't throw exceptions in your code" note here.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
```diff
@@ -406,6 +407,168 @@ public void TraceSyncSnsRecordSetsPubSubTrigger()
...
+        }
+
+        var item = Assert.Single(exportedItems);
+        Assert.Equal("value", item.GetTagValue("custom.attribute"));
+    }
+
+    [Fact]
+    public void EnrichWithInputCanOverrideFaasTrigger()
+    {
+        // This is what the datasource trigger types need.
```
martincostello:
```suggestion
```

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
```diff
@@ -406,6 +407,168 @@ public void TraceSyncSnsRecordSetsPubSubTrigger()
...
+    public void EnrichWithInputExceptionDoesNotFailTheInvocation()
+    {
+        var exportedItems = new List<Activity>();
+        var handlerRan = false;
+
+        using (var tracerProvider = Sdk.CreateTracerProviderBuilder()
+                   .AddAWSLambdaConfigurations(opt =>
+                   {
+                       opt.SemanticConventionVersion = SemanticConventionVersion.Latest;
+                       opt.EnrichWithInput = (_, _) => throw new InvalidOperationException("enrichment failure");
```
martincostello: There's no assertion that the enricher actually ran.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
```diff
@@ -406,6 +407,168 @@ public void TraceSyncSnsRecordSetsPubSubTrigger()
...
+        }
+
+        var item = Assert.Single(exportedItems);
+        Assert.Equal("async", item.GetTagValue("custom.attribute"));
+    }
+
+    [Fact]
+    public void EnrichWithInputIsNotInvokedWhenTheActivityIsNotRecorded()
+    {
+        // Attributes on a non-recording activity are discarded, so the action is skipped.
```
martincostello: Comment is effectively the same as the test method's name.

```suggestion
```

src/OpenTelemetry.Instrumentation.AWSLambda/README.md
ashishsinghnr: ok, let's remove this.

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
ashishsinghnr: fixed. Thanks!

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
```diff
@@ -406,6 +407,172 @@ public void TraceSyncSnsRecordSetsPubSubTrigger()
...
+        {
+            AWSLambdaWrapper.Trace(
+                tracerProvider,
+                (string _, ILambdaContext _) => handlerRan = true,
+                "TestStream",
+                this.sampleLambdaContext);
+        }
+
+        Assert.True(enricherRan);
+        Assert.True(handlerRan);
```
martincostello, line 515:
```suggestion
        Assert.True(enricherRan, "The enricher was not invoked.");
        Assert.True(handlerRan, "The handler was not invoked.");
```

test/OpenTelemetry.Instrumentation.AWSLambda.Tests/AWSLambdaWrapperTests.cs
```diff
@@ -406,6 +407,172 @@ public void TraceSyncSnsRecordSetsPubSubTrigger()
...
+                       opt.SemanticConventionVersion = SemanticConventionVersion.Latest;
+                       opt.EnrichWithInput = (_, _) => invoked = true;
+                   })
+                   .SetSampler(new AlwaysOffSampler())
+                   .Build()!)
+        {
+            AWSLambdaWrapper.Trace(tracerProvider, this.sampleHandlers.SampleHandlerSyncInputAndNoReturn, "TestStream", this.sampleLambdaContext);
+        }
+
+        Assert.False(invoked);
```
martincostello, line 558:
```suggestion
        Assert.False(invoked, "The enricher should not have been invoked.");
```

martincostello (approved):
Looks OK to me.

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
```diff
@@ -24,6 +25,17 @@ public class AWSLambdaInstrumentationOptions
...
+    /// <summary>
+    /// Gets or sets an action to enrich the invocation Activity from the function input.
+    /// Use it to describe a trigger the instrumentation does not recognise, for example to set
+    /// <c>faas.trigger</c> and the <c>faas.document.*</c> attributes for an Amazon S3 event.
+    /// </summary>
+    /// <remarks>
+    /// The action runs after the activity is created, so values it sets are not seen by samplers,
+    /// and it applies process-wide rather than per-provider.
+    /// </remarks>
+    public Action<Activity, object?>? EnrichWithInput { get; set; }
```
normj: Just wondering if we should make this feature a bit more generic and also include the `ILambdaContext` into the callback for enriching. I want to avoid a later feature request where users want to enrich the activity based on data in the `ILambdaContext` and then we have to add yet another callback.

ashishsinghnr: good point. Done, it is now `Action<Activity, object?, ILambdaContext>`

src/OpenTelemetry.Instrumentation.AWSLambda/AWSLambdaInstrumentationOptions.cs
```diff
@@ -26,15 +27,16 @@ public class AWSLambdaInstrumentationOptions
...
+    /// the Lambda context. Use it to describe a trigger the instrumentation does not recognise,
+    /// for example to set <c>faas.trigger</c> and the <c>faas.document.*</c> attributes for an
+    /// Amazon S3 event.
     /// </summary>
     /// <remarks>
     /// The action runs after the activity is created, so values it sets are not seen by samplers,
     /// and it applies process-wide rather than per-provider.
     /// </remarks>
-    public Action<Activity, object?>? EnrichWithInput { get; set; }
+    public Action<Activity, object?, ILambdaContext>? EnrichWithInput { get; set; }
```
martincostello, line 39: Just an observation at this stage: I can't decide whether it would be better or not to have the untyped parameter last.

ashishsinghnr: The current order similar to the Lambda handler signature — handlers are `(TInput input, ILambdaContext context)`, so the callback reads as that with the Activity prepended.
It also matches `AWSLambdaWrapper`, which uses `(input, context)` throughout (TraceRequest, OnFunctionStart, etc.), and the EnrichWith* convention elsewhere in the repo of Activity first, then the primary subject, with supplementary parameters after — e.g. EnrichAtClientFromResponse is `(Activity, IServiceRemotingResponseMessage?, Exception?)`. So here,  i am matching the handler shape.

Happy to flip it to `(Activity, ILambdaContext, object?)` if you'd rather have the untyped parameter last. Please suggest

normj (approved):

PR #5211, opened 2026-09-15
[Resources.Process] Fix process creation time fallback

Kielek:
Fixes internal Splunk/Cisco Codex scan
Design discussion issue #

## Changes

Fix process creation time fallback, for some edge cases it could return DateTime.Min instead of null


src/OpenTelemetry.Resources.Process/ProcessDetector.cs
```diff
@@ -44,19 +44,19 @@ static void GetProcessAttributes(
             using var process = CurrentProcess.GetCurrentProcess();
             processId = process.Id;
 
-            creationTime = SafeGet(process, (p) => p.StartTime);
```
martincostello, line 47: Would `SafeGet<DateTime?>(process, (p) => p.StartTime);` be an equivalent smaller change with the same overall effect? Then we keep `SafeGet` for other usage in future (IIRC, it exists because previous revisions used the method on other properties of other types).

martincostello (approved):

lachmatt (approved):

PR #5213, opened 2026-09-15
[Dynamic Control] Phase 5b: Add policy parsers

stevejgordon:
Contributes to: #4742

## Changes

Adds an internal JSON parser that decodes a complete JSON policy payload into validated policy models.


.typos.toml
```diff
@@ -16,7 +16,7 @@ vertica = "vertica"
 WRONLY = "WRONLY"
 # PowerShell -Force abbreviation
 fo = "fo"
-# Part of "café" (caf\u00e9)
+# Part of "café" (café)
```
martincostello: Was this changed intentionally?

src/OpenTelemetry.DynamicControl/CHANGELOG.md
```diff
@@ -27,4 +27,11 @@
   new `PolicyChangeNotifier`/`PolicyChangeSubscription` pair.
   ([#5135](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5135))
 
+* Added an internal `JsonKeyValuePolicyParser` that decodes a complete UTF-8
+  JSON key-value policy payload into validated policy models, distinguishing a
+  payload that could not be decoded from one that decoded into an empty or
+  partially usable set via `PolicyPayloadParseResult`,
+  `PolicyPayloadRejection`, and `PayloadEntryLocation`.
+  ([#TODO](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/TODO))
```
martincostello:
```suggestion
  ([#5213](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5213))
```

src/OpenTelemetry.DynamicControl/CHANGELOG.md
```diff
@@ -27,4 +27,11 @@
   new `PolicyChangeNotifier`/`PolicyChangeSubscription` pair.
   ([#5135](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/5135))
 
+* Added an internal `JsonKeyValuePolicyParser` that decodes a complete UTF-8
+  JSON key-value policy payload into validated policy models, distinguishing a
+  payload that could not be decoded from one that decoded into an empty or
+  partially usable set via `PolicyPayloadParseResult`,
+  `PolicyPayloadRejection`, and `PayloadEntryLocation`.
```
martincostello: Maybe a little overly detailed?

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+    // The depth cap is stated rather than inherited so that raising it is a deliberate act.
+    // It is set at the System.Text.Json default.
+    private static readonly JsonDocumentOptions PayloadOptions = new()
+    {
+        AllowTrailingCommas = false,
+        CommentHandling = JsonCommentHandling.Disallow,
+        MaxDepth = 64,
+    };
+
+    private static ReadOnlySpan<byte> Utf8ByteOrderMark => [0xEF, 0xBB, 0xBF];
```
martincostello, line 49: You could use [`Encoding.UTF8.Preamble`](https://learn.microsoft.com/dotnet/api/system.text.encoding.preamble?view=net-10.0#remarks) instead.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            document = JsonDocument.Parse(payload, PayloadOptions);
+        }
+        catch (JsonException ex)
+        {
+            return PolicyPayloadParseResult.Malformed("The payload could not be decoded as JSON. " + ex.Message);
+        }
+
+        // Every policy is built before the document is disposed, so nothing that outlives
+        // this scope holds a JsonElement into the document's pooled buffers.
```
martincostello:
```suggestion
```

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+        using (document)
+        {
+            var root = document.RootElement;
+
+            return root.ValueKind switch
+            {
+                JsonValueKind.Array => BuildResult(CollectArrayEntries(root)),
+                JsonValueKind.Object => BuildResult(CollectObjectEntries(root)),
+                _ => PolicyPayloadParseResult.Malformed("The payload root must be a JSON object or array."),
+            };
```
martincostello: I think .NET 11 will complain about the incomplete switch (which is a lot of cases), so it might be easiest to just do `if (array) { ... } else if (object) { ... } else { throw }`.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            entries.Add(CreateEntry(property, default));
+        }
+
+        return entries;
+    }
+
+    private static List<PayloadEntry> CollectArrayEntries(in JsonElement root)
+    {
+        List<PayloadEntry> entries = [];
```
martincostello: I didn't dig into the implementation too much to check, but it _may_ be able to use `root.GetArrayLength()` to pre-size the collection if it doesn't duplicate any work `EnumerateArray()` would do.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+            ? new(key, property.Value)
+            : new(new PolicyPayloadRejection(
+                fallbackLocation,
+                PolicyRejectionReason.SchemaMismatch,
+                "The payload declares a key that cannot be read as text."));
+
+    private static bool TryGetKey(in JsonProperty property, [NotNullWhen(true)] out string? key)
+    {
+        // NameEquals compares against the encoded name, which avoids an allocation for the
+        // common case. A name with no string equivalent cannot match a readable payload key.
```
martincostello: I would put this comment above line 156 rather than here.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    private static PolicyPayloadParseResult BuildResult(List<PayloadEntry> entries)
+    {
+        var duplicateKeys = FindDuplicateKeys(entries);
+
+        List<TelemetryPolicy> policies = [];
+        List<PolicyPayloadRejection> rejections = [];
+        List<string> ignoredKeys = [];
+        HashSet<string> seenIgnoredKeys = new(StringComparer.Ordinal);
```
martincostello, line 202: I'm not quite following why we need both collections when we could just have the HashSet? It looks like both are case sensitive, so `seenIgnoredKeys` and `ignoredKeys` would have the same contents at the end?

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
```diff
@@ -0,0 +1,316 @@
...
+        return PolicyPayloadParseResult.Decoded(
+            [.. policies],
+            [.. rejections],
+            [.. ignoredKeys]);
+    }
+
+    // A recognized key maps to exactly one policy type, and identity is derived from that
+    // type, so a key repeated anywhere in the payload would otherwise yield two policies
+    // occupying one PolicyKey. Rejecting every occurrence is what prevents that; resolving
+    // to the last would make the committed set depend on the order the payload was written.
```
martincostello: I would put this inside the method as it isn't a `///` comment.

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
```diff
@@ -0,0 +1,160 @@
...
+    public static readonly PayloadEntryLocation Empty;
+
+    private readonly string? key;
+
+    private readonly int index;
+
+    // Held separately rather than encoded into the index, so that a default instance
+    // identifies no location while every int, including int.MaxValue, remains a usable
+    // position. An earlier revision stored the index offset by one, which silently
+    // discarded the location of the last representable position.
```
martincostello: I don't think we need the comment about what an older version did.

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
```diff
@@ -0,0 +1,160 @@
...
+    /// <returns><see langword="true"/> if the locations are not equal; otherwise, <see langword="false"/>.</returns>
+    public static bool operator !=(PayloadEntryLocation left, PayloadEntryLocation right) => !left.Equals(right);
+
+    /// <summary>
+    /// Creates a location identifying the entry that declared a given key.
+    /// </summary>
+    /// <param name="key">The key the entry declared. Must not be null or whitespace.</param>
+    /// <returns>A location identified by key.</returns>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="key"/> is null, empty, or whitespace.
```
martincostello: Throw is implied by `<exception>`.

```suggestion
    /// <paramref name="key"/> is null, empty, or whitespace.
```

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
```diff
@@ -0,0 +1,160 @@
...
+        return new(key, 0, false);
+    }
+
+    /// <summary>
+    /// Creates a location identifying an entry by its position in the payload.
+    /// </summary>
+    /// <param name="index">The zero-based position of the entry. Must be non-negative.</param>
+    /// <returns>A location identified by position.</returns>
+    /// <exception cref="ArgumentOutOfRangeException">
+    /// Thrown when <paramref name="index"/> is negative.
```
martincostello:
```suggestion
    /// <paramref name="index"/> is negative.
```

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadParseResult.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// </summary>
+    /// <remarks>
+    /// When <see langword="true"/>, nothing about the payload's intended content is known,
+    /// so <see cref="Policies"/> must not be treated as a replacement set.
+    /// </remarks>
+    public bool IsMalformed { get; }
+
+    /// <summary>
+    /// Gets a description of why the payload could not be decoded, or <see langword="null"/>
+    /// when it was decoded.
```
martincostello:
```suggestion
    /// when it was decoded successfully.
```

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
martincostello: Ah, is this for ordering purposes? If so, add a comment.

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadParseResult.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        return new(true, error, [], [], []);
+    }
+
+    /// <summary>
+    /// Creates a result carrying the complete policy set a payload declared.
+    /// </summary>
+    /// <param name="policies">The policies the payload declared. May be empty.</param>
+    /// <param name="rejections">The entries that could not be used. May be empty.</param>
+    /// <param name="ignoredKeys">The distinct unrecognized keys the payload carried. May be empty.</param>
```
martincostello:
```suggestion
    /// <param name="policies">The policies the payload declared, if any.</param>
    /// <param name="rejections">The entries that could not be used, if any.</param>
    /// <param name="ignoredKeys">The distinct unrecognized keys the payload carried, if any.</param>
```

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadRejection.cs
```diff
@@ -0,0 +1,49 @@
...
+internal sealed class PolicyPayloadRejection
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="PolicyPayloadRejection"/> class.
+    /// </summary>
+    /// <param name="location">Where in the payload the rejected entry appeared.</param>
+    /// <param name="reason">The category of failure.</param>
+    /// <param name="message">A description of the failure. Must not be null or whitespace.</param>
+    /// <exception cref="ArgumentException">
+    /// Thrown when <paramref name="message"/> is null, empty, or whitespace.
```
martincostello:
```suggestion
    /// <paramref name="message"/> is null, empty, or whitespace.
```

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadRejection.cs
```diff
@@ -0,0 +1,49 @@
...
+        this.Message = message;
+    }
+
+    /// <summary>
+    /// Gets the location in the payload of the rejected entry.
+    /// </summary>
+    public PayloadEntryLocation Location { get; }
+
+    /// <summary>
+    /// Gets the category of failure.
```
martincostello:
```suggestion
    /// Gets the reason for failure, if any.
```

test/OpenTelemetry.DynamicControl.Tests/JsonKeyValuePolicyParserTests.cs
```diff
@@ -0,0 +1,607 @@
...
+        Assert.Empty(result.Rejections);
+        Assert.Equal(["LOG_LEVEL"], (IEnumerable<string>)result.IgnoredKeys);
+    }
+
+    private static string CreateNestedArrays(int depth)
+        => new string('[', depth) + "0.5" + new string(']', depth);
+
+    // The parser reads UTF-8 because that is the form every transport delivers a payload in.
+    // Tests state their payloads as source text and encode them here, so that each case
+    // remains readable and the encoding step is not repeated in every one.
```
martincostello:
```suggestion
```

test/OpenTelemetry.DynamicControl.Tests/JsonKeyValuePolicyParserTests.cs
```diff
@@ -0,0 +1,607 @@
...
+        JsonKeyValuePolicyParser.Parse(Encoding.UTF8.GetBytes(payload));
+
+    // Encoding a string cannot produce invalid UTF-8, so the bytes are assembled directly.
+    // 0xC3 opens a two-byte sequence and 0x28 is not a continuation byte, which System.Text.Json
+    // accepts while parsing and only fails on when the text it encodes is read.
+    private static ReadOnlyMemory<byte> CreatePayloadWithInvalidUtf8(string before, string after)
+    {
+        byte[] payload = [.. Encoding.UTF8.GetBytes(before), 0xC3, 0x28, .. Encoding.UTF8.GetBytes(after)];
+        return payload;
+    }
```
martincostello:
```suggestion
    private static ReadOnlyMemory<byte> CreatePayloadWithInvalidUtf8(string before, string after)
    {
        // Encoding a string cannot produce invalid UTF-8, so the bytes are assembled directly.
        // 0xC3 opens a two-byte sequence and 0x28 is not a continuation byte, which System.Text.Json
        // accepts while parsing and only fails on when the text it encodes is read.
        byte[] payload = [.. Encoding.UTF8.GetBytes(before), 0xC3, 0x28, .. Encoding.UTF8.GetBytes(after)];
        return payload;
    }
```

.typos.toml
stevejgordon: No! Not sure how the slipped through!

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: The explicit bytes avoid a heap allocation on our older targets (netstandard2.0, net462) where Encoding.UTF8.Preamble returns byte[]. The span literal is stored as static data. Marginal performance optimisation.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: This comment captures a subtle lifetime consideration: `JsonDocument` pools its internal buffers, and any `JsonElement` that outlives the using would be reading freed memory. I included this to avoid that potential mistake.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: I thought the default arm should be considered exhaustive. Potentially one to leave and test on the .NET 11 branch?

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
martincostello: Just feels a bit OTT to duplicate it into the static data for every TFM, when on newer runtimes the JIT is probably smart enough to leverage the equivalent value directly from the runtime from [here](https://github.com/dotnet/runtime/blob/d1a3ac4365f4819773ed81cba1956faa847226b0/src/libraries/System.Private.CoreLib/src/System/Text/UTF8Encoding.cs#L64).

If you want to keep it I'd make it TFM-aware with an `#if NET` and avoid the extra copy of the data.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
martincostello: I figured it was implied by the `using` itself as that's the scope of it.

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
martincostello: I'm 99% certain it will complain - I've had to make several changes to code in my .NET 11 branch to account for it. It's less that it's exhaustive, but the analyser will complain that the code isn't being _explicit_.

It's better to fix here, as otherwise I'll just have to fix it in #3867, and then possibly end up cherry-picking it into main to land ahead of time anyway.

test/OpenTelemetry.DynamicControl.Tests/JsonKeyValuePolicyParserTests.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

test/OpenTelemetry.DynamicControl.Tests/JsonKeyValuePolicyParserTests.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadRejection.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadRejection.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadParseResult.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PolicyPayloadParseResult.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/PayloadEntryLocation.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/CHANGELOG.md
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/CHANGELOG.md
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

.typos.toml
stevejgordon: Addressed in a1d336633195ac2c00bf2a49be2efd3af1c9e906

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in aab6f785ec5a8b55e1e484fc9599eba65d37fa89

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in aab6f785ec5a8b55e1e484fc9599eba65d37fa89

src/OpenTelemetry.DynamicControl/Internal/Providers/JsonKeyValuePolicyParser.cs
stevejgordon: Addressed in aab6f785ec5a8b55e1e484fc9599eba65d37fa89

.typos.toml
```diff
@@ -27,3 +27,4 @@ DELET = "DELET"
 INSER = "INSER"
 SELEC = "SELEC"
 UPDAT = "UPDAT"
+probabilit = "probabilit"
```
martincostello, line 30: Why do we need this intentional typo?

martincostello (approved):

.typos.toml
stevejgordon: It's used in `JsonKeyValuePolicyParserTests` for some of the malformed and escaped tests.

martincostello: Cool - I couldn't immediately see it in a Ctrl+F other than here.

PR #5237, opened 2026-09-18
[SqlClient/EFCore] Keep delimited identifiers whole when a closing bracket is escaped

mdmosfikurrahman:
Fixes # (none open; found while probing the sanitizer with T-SQL constructs not covered by the existing cases)
Design discussion issue #

## Changes

In T-SQL a doubled closing bracket inside a delimited identifier is an escaped bracket, so `[Tab]]le]` names the table `Tab]le`. `IsValidTokenCharacter` recognised the pair but did not consume both characters: it accepted the first bracket because another one followed, then judged the second on its own, found an ordinary character after it, and treated it as the terminator.

The identifier therefore ended halfway through the escape, and the captured target reached `db.query.summary` truncated:

| Query | `db.query.summary` before | after |
|---|---|---|
| `SELECT * FROM [Tab]]le]` | `SELECT [Tab]]` | `SELECT [Tab]]le]` |
| `SELECT * FROM [dbo].[Order]]s] WHERE Id = 1` | `SELECT [dbo].[Order]]` | `SELECT [dbo].[Order]]s]` |
| `SELECT * FROM [A]]]]B]` | `SELECT [A]]]]` | `SELECT [A]]]]B]` |

`db.query.summary` is specified as a low-cardinality grouping key, so two tables whose names differ only after an escaped bracket collapse onto the same series, and a dashboard grouped by it attributes one table's queries to the other.

A closing bracket is now read in the context of the brackets immediately preceding it within the token. An even run means it may open a pair, so the following character decides as before; an odd run means it completes a pair and the identifier continues past it. The scan stays local to the bracket run, so there is no change to the cost of ordinary identifiers.

`db.query.text` was already correct for these inputs and is unchanged; only the summary was affected.

## Testing

Three cases added to `SqlProcessorAdditionalTestCases.json` covering a plain escaped bracket, a schema-qualified identifier, and consecutive escapes. All three fail on `main` with exactly the truncated values above.

Verified locally on net462, net8.0, net9.0 and net10.0:

- `OpenTelemetry.Contrib.Shared.Tests` — 421 passed per framework
- `OpenTelemetry.Contrib.Shared.FuzzTests` — 30 properties passed per framework

* [x] Changes in public API reviewed (if applicable) — no public API change


src/OpenTelemetry.Instrumentation.SqlClient/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Fixed `db.query.summary` so that a delimited identifier containing an escaped
```
martincostello, line 5: Nit: please move to the end of the section.

src/Shared/SqlProcessor.cs
```diff
@@ -201,6 +201,23 @@ private static bool IsValidTokenCharacter(ReadOnlySpan<char> sql, int currentPos
...
+                // brackets immediately precede it within the token. An even run means this
+                // bracket is the first of a potential pair, so the next character decides. An
+                // odd run means it completes a pair and the identifier continues past it,
+                // which is what makes [Tab]]le] a single identifier named Tab]le.
+                var tokenStart = currentPosition - indexInToken;
+                var precedingBrackets = 0;
+                for (var i = currentPosition - 1; i >= tokenStart && sql[i] == CloseSquareBracketChar; i--)
+                {
+                    precedingBrackets++;
+                }
```
martincostello, line 214: This scales `O(n^2)` when scanning for `]` characters.

Could you refactor this to track the count incrementally in `ParseState` instead? That should make it `O(1)`.

PR #5326, opened 2026-09-19
[Exporter.Geneva] Fix number data point prepopulated attributes copy length

### LINKED ISSUE #5019: [Geneva] WriteNumberDataPoint copies prepopulated attributes using the histogram length
In `OtlpProtobufSerializer.WriteNumberDataPoint`, the prepopulated *number* data point attributes are copied using the prepopulated *histogram* attribute length, but the cursor is then advanced by the *number* attribute length:

```csharp
if (this.prepopulatedNumberDataPointAttributes != null)
{
    Array.Copy(this.prepopulatedNumberDataPointAttributes, 0, buffer, cursor, this.prepopulatedHistogramDataPointAttributesLength);
    cursor += this.prepopulatedNumberDataPointAttributesLength;
}
```

Compare with the histogram and exponential histogram paths, which consistently use their own length for both the copy and the cursor advance.

### Impact

Benign today, latent. The two buffers hold the same dimensions serialized under different field numbers — `NumberDataPoint_attributes` (7) and `HistogramDataPoint_attributes` (9). Both are below 16, so both encode to a single-byte tag and the two lengths are always equal. The bug therefore cannot be observed with the current field numbers.

It becomes a real bug the moment those two lengths can diverge — for example if a field number were to change to 16 or above (2-byte tag), which would make the copy either truncate the attributes or read past the end of the source array.

### Suggested fix

Use `this.prepopulatedNumberDataPointAttributesLength` for the `Array.Copy` length.

### Notes

Found while working on a separate change to the same method. Deliberately not fixed there to keep that change scoped.

amirani8137:
Use prepopulatedNumberDataPointAttributesLength instead of prepopulatedHistogramDataPointAttributesLength when copying prepopulated number data point attributes in WriteNumberDataPoint.

Fixes #5019

## Changes

Use prepopulatedNumberDataPointAttributesLength instead of prepopulatedHistogramDataPointAttributesLength when copying prepopulated number data point attributes in WriteNumberDataPoint.


rajkumar-rangaraj:
@amirani8137 How did you identify the issue, and did you complete E2E testing?

PR #5327, opened 2026-09-20
[gRPC] Inject context regardless of SuppressDownstreamInstrumentation

### LINKED ISSUE #1780: Adding gRPC instrumentation but not HttpClient causes issues with parenting
Both the gRPC .NET and HttpClient libraries generate their own Activity for outgoing calls.

gRPC .NET uses HttpClient behind the scenes, so when you configure OpenTelemetry with both gRPC and HttpClient instrumentation you'll see that the Activity generated by gRPC is the parent of the Activity generated by HttpClient.

However, if you only include gRPC instrumentation or if you only include HttpClient instrumentation the chain of parenting can become broken since the library still generates their respective Activity. Another way this can occur is if you add both gRPC and HttpClient instrumentation, but you configure `SuppressDownstreamInstrumentation` for the gRPC instrumentation thereby excluding the child HttpClient Activity from the trace.

This issue is somewhat related to open-telemetry/opentelemetry-dotnet-contrib#1779 in that the general issue is that libraries generate Activities regardless of whether OpenTelemetry has been configured to be interested in them. I'm not sure there are good answers here, but it would be nice if there were a way to inform a library of whether or not to generate Activities based on how OpenTelemetry is configured for an app.

PetarJerinic:
Fixes #1780
Design discussion issue https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1780#issuecomment-5679208872

## Changes

The gRPC client instrumentation now injects context on every call, not only when `SuppressDownstreamInstrumentation` is `true`.

Notable changes, presented as breaking changes in the CHANGELOG:

gRPC only
- The downstream span is parented to the gRPC client span, and baggage is now sent.

gRPC + HttpClient 
- With the default propagator is unaffected. 
- With only TraceContextPropagator: the downstream parent moves from the HttpClient span to the gRPC client span, because HttpClient instrumentation does not inject for that propagator.


src/OpenTelemetry.Instrumentation.GrpcNetClient/README.md
```diff
@@ -108,6 +106,15 @@ using var tracerProvider = Sdk.CreateTracerProviderBuilder()
     .Build();
 ```
 
+### Context propagation
+
+Grpc.Net.Client instrumentation propagates context (ActivityContext and
+Baggage) to downstream services using the configured propagator, so the
```
martincostello:
```suggestion
Grpc.Net.Client instrumentation propagates context (`ActivityContext` and
`Baggage`) to downstream services using the configured propagator, so the
```

test/OpenTelemetry.Instrumentation.GrpcNetClient.Tests/GrpcTests.client.cs
```diff
@@ -259,6 +256,52 @@ public void GrpcClientCallsAreNotCollectedWhenFilterThrows()
...
+            {
+                HttpClient = httpClient,
+            });
+            var client = new Greeter.GreeterClient(channel);
+            _ = client.SayHello(new HelloRequest());
+        }
+
+        Assert.NotNull(clientActivity);
+        var expectedSampledFlag = filterMatches ? "01" : "00";
+        Assert.Equal($"00-{clientActivity.TraceId}-{clientActivity.SpanId}-{expectedSampledFlag}", traceparent);
```
martincostello: This _might_ fail in .NET 11 where the `02` flag if a random trace ID could be set (in which cause it would be `03` from `1|2`).

martincostello (approved):
Code changes look OK to me, but would be good to get a review from an SME in the gRPC area and from the original issue.

rajkumar-rangaraj (approved):

